PR #3, opened 2016-03-30
GUAC-1511: Add low-level support for audio input to libguac.

mike-jumper:
This change adds support for the "audio" instruction when received from a connected user. Prior to this, audio streams were only provided in the outbound direction, with inbound handling of the "audio" instruction restricted to the JavaScript `Guacamole.Client` class.

### DIFF
--- a/src/libguac/guacamole/user-fntypes.h
+++ b/src/libguac/guacamole/user-fntypes.h
@@ -116,6 +116,29 @@ typedef int guac_user_mouse_handler(guac_user* user, int x, int y,
 typedef int guac_user_key_handler(guac_user* user, int keysym, int pressed);
 
 /**
+ * Handler for Guacamole audio streams received from a user. Each such audio
+ * stream begins when the user sends an "audio" instruction. To handle received
+ * data along this stream, implementations of this handler must assign blob and
+ * end handlers to the given stream object.
+ *
+ * @param user
+ *     The user that opened the audio stream.
+ *
+ * @param stream
+ *     The stream object allocated by libguac to represent the audio stream
+ *     opened by the user.
+ *
+ * @param mimetype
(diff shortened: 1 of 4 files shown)


src/libguac/guacamole/user-fntypes.h
```diff
@@ -116,6 +116,29 @@ typedef int guac_user_mouse_handler(guac_user* user, int x, int y,
 typedef int guac_user_key_handler(guac_user* user, int keysym, int pressed);
 
 /**
+ * Handler for Guacamole audio streams received from a user. Each such audio
+ * stream begins when the user sends a "audio" instruction. To handle received
```
jmuehlner: An "audio" instruction? :)

mike-jumper: Yes: http://guac-dev.org/doc/gug/protocol-reference.html#audio-stream-instruction

The "audio" instruction was previously only used for server -> client. These changes allow it in either direction.

jmuehlner: I meant should be "an" not "a".

An audio stream.

src/libguac/user-handlers.c
```diff
@@ -267,6 +268,29 @@ static guac_stream* __init_input_stream(guac_user* user, int stream_index) {
...
+    if (user->audio_handler)
+        return user->audio_handler(
+            user,
+            stream,
+            argv[1] /* mimetype */
+        );
+
+    /* Otherwise, abort */
+    guac_protocol_send_ack(user->socket, stream,
+            "Audip input unsupported", GUAC_PROTOCOL_STATUS_UNSUPPORTED);
```
mike-jumper: Ughhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhh...

"Audip"

PR #11, opened 2016-05-27
GUACAMOLE-39: Do not abort VNC/RDP connection if SFTP fails

tylermauthe:
Please review and let me know if you'd like to suggest any changes.

### DIFF
--- a/src/common-ssh/guac_ssh.c
+++ b/src/common-ssh/guac_ssh.c
@@ -275,11 +275,43 @@ static void guac_common_ssh_kbd_callback(const char *name, int name_len,
 }
 
 /**
+ * A handler for SSH client errors which logs the error and can optionally
+ * abort the given guac_client with the given guac_protocol_status.
+ *
+ * @param client
+ *     Guacamole client instance to use when logging or aborting.
+ *
+ * @param status
+ *      Protocol status to return if aborting the connection. This is only
+ *      used if abort_on_error is set to true.
+ *
+ * @param abort_on_error
+ *     Whether to abort the connection as well as log the error.
+ *     
+ * @param format
+ *     Format string for log messages. This is also used when aborting 
+ *     the connection.
+ *
+ */
+static void guac_common_ssh_client_error(guac_client* client,
+        guac_protocol_status status, int abort_on_error, const char* format, ...) {
+    va_list args;
+    va_start(args, format);
+    if(abort_on_error == 1) {
+        vguac_client_abort(client, status, format, args);
+    }
+    else {
+        vguac_client_log(client, GUAC_LOG_ERROR, format, args);
+    }
+    va_end(args);
+}
+
+/**
  * Authenticates the user associated with the given session over SSH. All
  * required credentials must already be present within the user object
  * associated with the given session.
  *
- * @param session
+ * @param common_session
  *     The session associated with the user to be authenticated.
  *
  * @return
@@ -299,7 +331,7 @@ static int guac_common_ssh_authenticate(guac_common_ssh_session* common_session)
 
     /* Validate username provided */
     if (username == NULL) {
-        guac_client_abort(client, GUAC_PROTOCOL_STATUS_CLIENT_UNAUTHORIZED,
+        guac_client_log(client, GUAC_LOG_ERROR,
                 "SSH authentication requires a username.");
         return 1;
     }
@@ -315,7 +347,7 @@ static int guac_common_ssh_authenticate(guac_common_ssh_session* common_session)
 
         /* Check if public key auth is supported on the server */
         if (strstr(user_authlist, "publickey") == NULL) {
-            guac_client_abort(client, GUAC_PROTOCOL_STATUS_CLIENT_UNAUTHORIZED,
+            guac_client_log(client, GUAC_LOG_ERROR,
                     "Public key authentication is not supported by "
                     "the SSH server");
             return 1;
@@ -329,7 +361,7 @@ static int guac_common_ssh_authenticate(guac_common_ssh_session* common_session)
             /* Abort on failure */
             char* error_message;
             libssh2_session_last_error(session, &error_message, NULL, 0);
-            guac_client_abort(client, GUAC_PROTOCOL_STATUS_CLIENT_UNAUTHORIZED,
+            guac_client_log(client, GUAC_LOG_ERROR,
                     "Public key authentication failed: %s", error_message);
 
             return 1;
@@ -353,8 +385,7 @@ static int guac_common_ssh_authenticate(guac_common_ssh_session* common_session)
                 /* Abort on failure */
                 char* error_message;
                 libssh2_session_last_error(session, &error_message, NULL, 0);
-                guac_client_abort(client,
-                        GUAC_PROTOCOL_STATUS_CLIENT_UNAUTHORIZED,
+                guac_client_log(client, GUAC_LOG_ERROR,
                         "Password authentication failed: %s", error_message);
 
                 return 1;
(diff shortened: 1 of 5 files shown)


src/common-ssh/guac_ssh.c
```diff
@@ -274,19 +275,36 @@ static void guac_common_ssh_kbd_callback(const char *name, int name_len,
 
 }
 
+void ssh_client_error(bool abortOnError, guac_client* client, guac_protocol_status status,
```
mike-jumper: The are a few issues here:

1. The other functions defined in this part of guacamole-server all have the namespace/prefix `guac_common_ssh_`. Should be the same here.
2. If this is not used externally (and not declared in a header file), it should be `static`.
3. Documentation of the function and its parameters is missing.
4. The variable `abortOnError` does not follow the naming style of the rest of this file or guacamole-server. Please user lowercase, separating words by underscores.

As a minor nitpick, it's strange to me to a flag variable as the first argument in a function. Normally such things are toward the end of the parameter list. As this is essentially a selective wrapper for `guac_client_abort()` and `guac_client_log()`, I would recommend following the same ordering to avoid confusion when people try to develop against this in the future.

I also notice that the `status` parameter is ignored in the case that the abort flag is set. I don't immediately see a way to avoid that with this approach, but perhaps that is a sign that a different approach would be better.

src/common-ssh/guac_ssh.c
```diff
@@ -274,19 +275,36 @@ static void guac_common_ssh_kbd_callback(const char *name, int name_len,
...
+        const char* format, ...) {
+    va_list args;
```
mike-jumper: The `stdarg.h` header required for `va_list` and related functions is missing.

src/common-ssh/guac_ssh.h
```diff
@@ -101,6 +101,33 @@ guac_common_ssh_session* guac_common_ssh_create_session(guac_client* client,
...
+ *     The port to connect to on the given hostname.
+ *
+ * @param user
+ *     The user to authenticate as, once connected.
+ *
+ * @return
+ *     A new SSH session if the connection and authentication succeed, or NULL
+ *     if the connection or authentication were not successful.
+ */
+guac_common_ssh_session* guac_common_ssh_create_secondary_session(guac_client* client,
```
mike-jumper: Why define a whole new function for this?

src/common-ssh/guac_ssh.h
```diff
@@ -101,6 +101,33 @@ guac_common_ssh_session* guac_common_ssh_create_session(guac_client* client,
         const char* hostname, const char* port, guac_common_ssh_user* user);
 
 /**
+ * Used within VNC and RDP connections to create an SFTP connection using the
```
mike-jumper: Dictating exactly where this function is used within the documentation of the function seems questionable to me. Can this not instead be described at a high level, strictly from the perspective of the common SSH client implemented here?

src/common-ssh/guac_ssh.c
```diff
@@ -274,19 +275,36 @@ static void guac_common_ssh_kbd_callback(const char *name, int name_len,
...
+ *
+ * @param common_session
  *     The session associated with the user to be authenticated.
  *
  * @return
  *     Zero if authentication succeeds, or non-zero if authentication has
  *     failed.
  */
-static int guac_common_ssh_authenticate(guac_common_ssh_session* common_session) {
+static int guac_common_ssh_authenticate(bool abortOnError, guac_common_ssh_session* common_session) {
```
mike-jumper: Since the function is already defined as returning zero / non-zero depending on whether the operation succeeds, why not leverage that and let the caller dictate whether to abort?

It seems like we shouldn't have had `guac_common_ssh_authenticate()` automatically handle the full blown abort in the first place - way out of scope for what it's suppose to do / be aware of.

mike-jumper:
@flangelo / @jmuehlner - anything I'm missing?

jmuehlner:
@mike-jumper Not that I can see.

flangelo:
@mike-jumper, Think you got it all.

tylermauthe:
Thanks for the feedback. I've incorporated your changes, sorry for the delay. I've also squashed my branch into a single commit.

Let me know if there's anything else.

tylermauthe:
Hello, just curious if there is anything else I can do to help out with getting this merged? Happy to assist in any way.

mike-jumper:
Thanks for the ping, @tylermauthe - I had completely forgotten about this.

Will reread the PR and try to get up to speed on where we left off.

mike-jumper:
OK - in general this looks much better. There are still a few things I think are worth fixing, but I like what you've done.

src/common-ssh/guac_ssh.c
```diff
@@ -275,11 +275,43 @@ static void guac_common_ssh_kbd_callback(const char *name, int name_len,
...
+ * @param format
+ *     Format string for log messages. This is also used when aborting 
+ *     the connection.
+ *
+ */
+static void guac_common_ssh_client_error(guac_client* client,
+        guac_protocol_status status, int abort_on_error, const char* format, ...) {
+    va_list args;
+    va_start(args, format);
+    if(abort_on_error == 1) {
```
mike-jumper: Two things here:

1. Direct comparisons against `1` are a little bit funky in C. Normally, you use non-zero to represent true, as this is what C uses internally for `if` tests, etc. Things will work more as expected with:

    ```
    if (abort_on_error) {
        ...
    }
    ```

2. `if` is not a function. There should be a space between the `if` and the opening paren.

Beyond that, I think this is OK. The function seems a bit crowded to me due to the lack of blank lines / "code paragraphs", but that's not a deal-breaker. I'd like you to consider omitting the braces where unnecessary (as that's our preference - see the [style guidelines](http://guacamole.incubator.apache.org/guac-style/)), but as we say in that doc, these aren't hard rules - use your own best judgement.

src/common-ssh/guac_ssh.h
```diff
@@ -76,10 +76,10 @@ void guac_common_ssh_uninit();
 /**
  * Connects to the SSH server running at the given hostname and port, and
  * authenticates as the given user. If an error occurs while connecting or
- * authenticating, the Guacamole client will automatically and fatally abort.
- * The user object provided must eventually be explicitly destroyed, but should
- * not be destroyed until this session is destroyed, assuming the session is
- * successfully created.
+ * authenticating, the Guacamole client will only automatically and fatally
+ * abort if the abort_on_error parameter is set to 1 and errors will always
```
mike-jumper: As mentioned earlier, requiring `1` is odd for C. Better for this to simply be non-zero. Other than that, still looking good.

src/protocols/rdp/rdp.c
```diff
@@ -1003,36 +1003,27 @@ void* guac_rdp_client_thread(void* data) {
...
+            guac_client_for_owner(client,
+                    guac_common_ssh_expose_sftp_filesystem,
+                    rdp_client->sftp_filesystem);
+
+            /* Verify that SFTP connection succeeded */
+            if (rdp_client->sftp_filesystem != NULL) {
+                guac_client_log(client, GUAC_LOG_DEBUG,
+                    "SFTP connection succeeded.");
+            }
         }
```
mike-jumper: I think you've accidentally killed the required calls to `guac_common_ssh_destroy_user()`, as well as the other cleanup in the case that the filesystem object creation fails, etc.

If the session fails to start, the user object will be leaked. If the filesystem fails, the user and session objects will be leaked.

src/protocols/vnc/vnc.c
```diff
@@ -283,43 +283,32 @@ void* guac_vnc_client_thread(void* data) {
...
+            if (vnc_client->sftp_filesystem != NULL) {
+                /* If specified, configure destination for basic uploads */
+                if (settings->sftp_directory != NULL)
+                    guac_common_ssh_sftp_set_upload_path(
+                        vnc_client->sftp_filesystem,
+                        settings->sftp_directory);
+
+                guac_client_log(client, GUAC_LOG_DEBUG,
+                    "SFTP connection succeeded.");
+            }
```
mike-jumper: Same deal here. It looks like the user and filesystem object cleanup was deleted without being re-added.

necouchman:
Ping.

@tylermauthe: Tyler, any progress here in making the tweaks that Mike and James suggested?

tylermauthe:
I've changed jobs and, as a result, won't be finishing this PR up. Sorry about that!

If anyone wants to take it over, please feel free to do so.

PR #13, opened 2016-06-03
GUACAMOLE-25: Avoid unnecessary allocation of temporary string for guac_rdp_ptr_to_string().

mike-jumper:
The current code leaks the storage for the stringified pointer allocated by `guac_rdp_ptr_to_string()`, but it's really not necessary allocate the string at all. Better that `guac_rdp_ptr_to_string()` be provided a sufficiently-large buffer.


mike-jumper:
Oops. GUACAMOLE-25, not GUACAMOLE-44.

PR #15, opened 2016-06-16
GUACAMOLE-25: Refactor the RDPDR printer to synchronous operation.

mike-jumper:
This change replaces the old printing thread with a handler for the "ack" responses sent by the client. Rather than continuously reading and sending blobs, ignoring client responses, the handler will only read and send additional print data when the client acknowledges that it is ready.


mike-jumper:
Cripes ... why do I always tag the wrong JIRA.

This should be GUACAMOLE-44.

PR #16, opened 2016-06-16
GUACAMOLE-44: Refactor the RDPDR printer to synchronous operation.

mike-jumper:
This change replaces the old printing thread with a handler for the "ack" responses sent by the client. Rather than continuously reading and sending blobs, ignoring client responses, the handler will only read and send additional print data when the client acknowledges that it is ready.

**NOTE:** This was originally opened as #15, but with the wrong JIRA issue tag.

### DIFF
--- a/src/protocols/rdp/guac_rdpdr/rdpdr_printer.h
+++ b/src/protocols/rdp/guac_rdpdr/rdpdr_printer.h
@@ -25,15 +25,11 @@
 
 #include "rdpdr_service.h"
 
-#include <guacamole/stream.h>
-
 #ifdef ENABLE_WINPR
 #include <winpr/stream.h>
 #else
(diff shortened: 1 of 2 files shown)


src/protocols/rdp/guac_rdpdr/rdpdr_printer.h
```diff
@@ -25,44 +25,32 @@
 
 #include "rdpdr_service.h"
 
-#include <guacamole/stream.h>
+#include <guacamole/user.h>
```
jmuehlner: What is this new import used for?

mike-jumper: Nothing. It's a remnant of a part of the change which ended up not being necessary. Will remove.

PR #17, opened 2016-07-25
GUACAMOLE-5: Add per-user "read-only" parameter to all protocols.

mike-jumper:
Previously, the only protocol providing a "read-only" parameter was VNC, as VNC servers inherently support screen sharing. Now that Guacamole supports screen sharing regardless of underlying protocol, such parameters should be provided across the board.

This change adds a "read-only" parameter to the RDP, SSH, and telnet protocols, and extends the existing "read-only" parameter of VNC to be handled on a per-user (rather than per-connection) basis.

### DIFF
--- a/src/protocols/rdp/user.c
+++ b/src/protocols/rdp/user.c
@@ -44,19 +44,25 @@ int guac_rdp_user_join_handler(guac_user* user, int argc, char** argv) {
 
     guac_rdp_client* rdp_client = (guac_rdp_client*) user->client->data;
 
+    /* Parse provided arguments */
+    guac_rdp_settings* settings = guac_rdp_parse_args(user,
+            argc, (const char**) argv);
+
+    /* Fail if settings cannot be parsed */
+    if (settings == NULL) {
+        guac_user_log(user, GUAC_LOG_INFO,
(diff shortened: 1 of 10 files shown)


src/protocols/rdp/user.c
```diff
@@ -88,12 +94,24 @@ int guac_rdp_user_join_handler(guac_user* user, int argc, char** argv) {
...
     }
 
-    user->file_handler = guac_rdp_user_file_handler;
-    user->mouse_handler = guac_rdp_user_mouse_handler;
-    user->key_handler = guac_rdp_user_key_handler;
-    user->size_handler = guac_rdp_user_size_handler;
-    user->pipe_handler = guac_rdp_svc_pipe_handler;
-    user->clipboard_handler = guac_rdp_clipboard_handler;
+    /* Only handle events if not read-only */
+    if (settings->read_only == 0) {
```
jmuehlner: This is sort of a weird way to check for false.

mike-jumper: Ugh... oops... and this just after I commented on exactly that sort of thing for #11.

Will fix.

PR #18, opened 2016-08-15
GUACAMOLE-51: Separate RDP keyboard handling into dedicated files / structures.

mike-jumper:
Finishing the RDP portion of [GUACAMOLE-51](https://issues.apache.org/jira/browse/GUACAMOLE-51) is getting a bit hairy since the keymap structures are interwoven with the rest of the RDP client code. Really shouldn't have done it that way.

This change moves the main keymap-driven keyboard translation code into its own files and structures, centered around the `guac_rdp_keyboard` abstraction.

### DIFF
--- a/src/protocols/rdp/keyboard.c
+++ b/src/protocols/rdp/keyboard.c
@@ -32,6 +32,88 @@
 #include <stdlib.h>
 
 /**
+ * Immediately sends an RDP key event having the given scancode and flags.
+ *
+ * @param rdp_client
+ *     The RDP client instance associated with the RDP session along which the
+ *     key event should be sent.
+ *
+ * @param scancode
+ *     The scancode of the key to press or release via the RDP key event.
+ *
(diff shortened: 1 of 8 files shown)


src/protocols/rdp/keyboard.c
```diff
@@ -0,0 +1,283 @@
...
+    rdp_inst->input->KeyboardEvent(rdp_inst->input,
+            flags | pressed_flags, scancode);
+
+    pthread_mutex_unlock(&(rdp_client->rdp_lock));
+
+}
+
+/**
+ * Immediately sends an RDP Unicode event having the given Unicode codepoint.
+ * Unlike key events, RDP Unicode events do not a pressed or released state.
```
jmuehlner: This sentence no verb.

mike-jumper: It has a verb, but it actually two.

PR #19, opened 2016-08-17
GUACAMOLE-51: Synchronize implied num lock state with RDP server.

mike-jumper:
If a user attempts to use the number pad to type numbers over RDP, they will only succeed if the RDP server has num lock active. This change automatically activates num lock if necessary.


jmuehlner:
Looks good to me.

PR #20, opened 2016-08-26
GUACAMOLE-87: Bump guacamole-server version numbers to 0.9.10-incubating.

jmuehlner:
The Dockerfile is untouched because it needs more work before the release; simply bumping the version numbers in it would be insufficient.


mike-jumper:
The libtool version info for libguac needs bumping, too.

PR #21, opened 2016-08-26
GUACAMOLE-93: Build guacd-docker using parent guacamole-server source.

mike-jumper:


mike-jumper:
As with apache/incubator-guacamole-client#67, I've also moved the `Dockerfile` to the root (see 4f84be1).

PR #24, opened 2016-10-27
GUACAMOLE-107 Remove references to old website and Apachify project name

jmuehlner:
It looks like only the README had to be touched here.

### DIFF
--- a/README
+++ b/README
@@ -6,7 +6,11 @@
 This README is intended to provide quick and to-the-point documentation for
 technical users intending to compile parts of Apache Guacamole themselves.
 
-A full manual is provided on the Apache Guacamole web site:
+Source archives are available from the downloads section of the project website:
+ 
+    http://guacamole.incubator.apache.org/
+
+A full manual is available as well:
 
     http://guacamole.incubator.apache.org/doc/gug/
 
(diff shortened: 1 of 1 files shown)


README
```diff
@@ -4,15 +4,11 @@
 ------------------------------------------------------------
 
 This README is intended to provide quick and to-the-point documentation for
-technical users intending to compile parts of Guacamole themselves.
+technical users intending to compile parts of Apache Guacamole themselves.
 
-Source archives are available from the files section of the main project page:
```
mike-jumper: This should probably still be here, but instead refer to the "downloads" section. It is still valid, though definitely should not point to SourceForge.

jmuehlner: Ok

README
```diff
@@ -6,11 +6,16 @@
...
+Source archives are available from the downloads section of the project website:
+ 
+    http://guacamole.incubator.apache.org/
+
+A full manual is available as well:
 
     http://guacamole.incubator.apache.org/doc/gug/
 
 
+
```
mike-jumper: Why did you add this blank line?

jmuehlner: :face_with_head_bandage:

PR #35, opened 2017-01-11
GUACAMOLE-157: Initialize SSH SSL lock array to NULL and test for NUL…

flangelo:
…L-ness before freeing.


mike-jumper:
Nice find. Probably worth expanding scope of 0.9.11 to include this, IMHO... what do you think?

flangelo:
I agree.

PR #36, opened 2017-01-16
GUACAMOLE-86: Remove terminal emulator's STDOUT pipe.

mike-jumper:
This change removes use of the STDOUT pipe entirely, replacing the blocking/signaling functionality with a `pthread_cond_t` and corresponding mutex. Writes which previously would go through the pipe now go through direct calls to `guac_terminal_write()`.


jmuehlner:
@flangelo Any thoughts on this change? You're probably more familiar with a lot of this code.

flangelo:
It looks good. Great work @mike-jumper .

jmuehlner:
@flangelo Go for it - merge away! :shipit:

mike-jumper:
Also beware: since the 0.9.11-incubating release branch exists, and this is a 0.9.11-incubating change, it needs to be merged to `staging/0.9.11-incubating` first, followed by `staging/0.9.11-incubating` to `master`.

PR #39, opened 2017-01-24
GUACAMOLE-171: Move broadcast socket to public API

mike-jumper:
This change refactors the previously-private broadcast socket, bringing it into the public portion of the libguac API. Downstream users can now create `guac_socket` instances with broadcast semantics using the `guac_socket_broadcast()` function.


src/libguac/client.c
```diff
@@ -544,12 +273,12 @@ int guac_client_add_user(guac_client* client, guac_user* user, int argc, char**
 
     int retval = 0;
 
-    pthread_rwlock_wrlock(&(client->__users_lock));
-
     /* Call handler, if defined */
     if (client->join_handler)
         retval = client->join_handler(user, argc, argv);
 
+    pthread_rwlock_wrlock(&(client->__users_lock));
```
jmuehlner: Why are these moving around?

mike-jumper: So that implementations of the `join_handler` and the `leave_handler` need not be aware of the internals of `guac_client` (that the `guac_client` will internally acquire a lock on the user list, preventing use of any broadcast socket).

Prior to this, writing to a broadcast socket within join/leave lead to undefined results (most likely a segfault). Now, writing to the broadcast socket within join/leave sends data to all users except the one joining/leaving.

jmuehlner: Ok

PR #40, opened 2017-01-24
GUACAMOLE-172: Improve/fix processing lag calculations

mike-jumper:
This change modifies the lag calculations provided by libguac and used within the VNC and RDP protocols such that:

* Server processing time is excluded
* Bad client timestamps are ignored


jmuehlner:
Looks reasonable to me - any concerns @flangelo?

flangelo:
Looks good to me.

PR #41, opened 2017-01-24
GUACAMOLE-175: Move core of guacd into utility library (libguacd)

mike-jumper:
This change moves much of the reusable innards of guacd into a utility library, "libguacd", which other components of guacamole-server can use to provide guacd-like Guacamole protocol implementations.


jmuehlner:
Cool; everything looks good to me.

PR #42, opened 2017-01-25
GUACAMOLE-176: Support layer location and opacity for guac_common_surface

mike-jumper:


src/common/common/surface.h
```diff
@@ -317,10 +363,77 @@ void guac_common_surface_clip(guac_common_surface* surface, int x, int y, int w,
...
+ */
+void guac_common_surface_stack(guac_common_surface* surface, int z);
+
+/**
+ * Changes the parent layer of ths given surface. By default, layers will be
+ * children of the default layer. If the surface does not represent a
+ * non-default visible layer, then this function has no effect.
+ *
+ * @param surface
+ *     The surface to move relative to its parent layer.
```
jmuehlner: Looks like maybe some bad copypasta here? I don't think this parameter description works here, or in guac_common_surface_set_opacity.

mike-jumper: Oops.

src/common/common/surface.h
```diff
@@ -317,10 +363,77 @@ void guac_common_surface_clip(guac_common_surface* surface, int x, int y, int w,
...
+ */
+void guac_common_surface_set_parent(guac_common_surface* surface,
+        const guac_layer* parent);
+
+/**
+ * Sets the opacity of the surface. If the surface does not represent a
+ * non-default visible layer, then this function has no effect.
+ *
+ * @param surface
+ *     The surface to move relative to its parent layer.
```
jmuehlner: To move? Doesn't this function change opacity?

mike-jumper: Sigh ... yet more copypasta ... 🍝

PR #44, opened 2017-01-25
GUACAMOLE-179: Move PulseAudio support into utility library.

mike-jumper:
This change moves the PulseAudio support provided by the VNC client into a common utility library which can be shared by other protocol implementations, including the WIP X.Org driver.

Part of this change involved modifying libguac's audio stream implementation such that it does not rely on a connection "owner" existing. Not all guacd-like Guacamole protocol implementations will have a connected user at all times.


jmuehlner:
Looks good to me!

PR #45, opened 2017-01-25
GUACAMOLE-180: Ensure common surface and display objects are threadsafe.

mike-jumper:


jmuehlner:
@flangelo You might also want to have a look at this one.

flangelo:
I think this is good.

PR #47, opened 2017-01-26
GUACAMOLE-185: Fit source rectangle to bounds for copy/transfer.

mike-jumper:


jmuehlner:
Ok, I think everything looks good here.

PR #56, opened 2017-01-29
GUACAMOLE-148: Fix handling of terminal scroll.

mike-jumper:
Terminal emulators maintain the scrolling region and the bounds of the terminal separately. Console codes which allow the scrolling region to be redefined facilitate utilities like top, tmux, or screen which must maintain control of what scrolls and what doesn't.

Guacamole handles scrolling improperly, automatically scrolling whenever the cursor exceeds the bounds of the scroll region. This is only correct behavior when the cursor is within the scrolling region. The terminal should not be scrolled *at all* when the cursor is outside the scrolling region. If outside the scrolling region, the cursor should simply be kept within terminal bounds.

### DIFF
--- a/src/terminal/terminal.h
+++ b/src/terminal/terminal.h
@@ -259,12 +259,18 @@ struct guac_terminal {
     int scroll_end;
 
     /**
-     * The current row location of the cursor.
+     * The current row location of the cursor. Note that while most terminal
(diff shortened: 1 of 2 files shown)


jmuehlner:
Ok, this makes sense to me.

PR #57, opened 2017-01-31
GUACAMOLE-193: Take character attributes into account when testing for visibility.

mike-jumper:


src/terminal/terminal.c
```diff
@@ -847,6 +847,45 @@ int guac_terminal_clear_range(guac_terminal* term,
...
+        return false;
+
+    /* Characters with glyphs are ALWAYS visible */
+    if (guac_terminal_has_glyph(c->value))
+        return true;
+
+    int background;
+
+    /* Determine actual background color of character */
+    if (c->attributes.reverse != c->attributes.cursor)
```
jmuehlner: I'm not sure I understand what this check is about - comparing the reverse attribute to the cursor attribute?

mike-jumper:
Both the reverse attribute and the cursor attribute swap the foreground and background, so the attributes only have a visible effect when only one is set.

PR #58, opened 2017-02-01
GUACAMOLE-149: Ignore zero-width characters.

mike-jumper:


jmuehlner:
Are there no valid zero-width characters that we might need to keep? We can just throw them all out?

mike-jumper:
For the sake of rendering, we can, yes. I have yet to find a terminal emulator which hangs on to such characters within the buffer just so that they come back when copied, and it makes sense that they wouldn't: doing so would require that each row of buffer data could be arbitrarily long, no longer confined to at most N characters.

PR #59, opened 2017-02-16
GUACAMOLE-200: Do not allow read() to block input handling

mike-jumper:
From [GUACAMOLE-200](https://issues.apache.org/jira/browse/GUACAMOLE-200):

>
> Recent changes moved the RDP printer within Guacamole from asynchronous transfer (independent of received "ack" messages) to synchronous transfer. This is fine in itself, but the `read()` call pulling more data from GhostScript in reponse to an "ack" actually occurs within the handler for that "ack". If GhostScript is busy processing things and that `read()` blocks, handling of user input as a whole gets blocked.
>
> The actual reads need to be moved into a printer-specific thread, with receipt of "ack" signalling that more data should be read, allowing the ack handler to finish regardless of GhostScript's state.
>

This change refactors the printer implementation such that printer reads occur within a separate thread.

### DIFF
--- a/src/protocols/rdp/guac_rdpdr/rdpdr_printer.c
+++ b/src/protocols/rdp/guac_rdpdr/rdpdr_printer.c
@@ -117,7 +117,7 @@ void guac_rdpdr_process_print_job_close(guac_rdpdr_device* device,
     /* End print job */
     if (job != NULL) {
         guac_rdp_print_job_free(job);
-        device->data = NULL;
+        rdp_client->active_job = NULL;
     }
 
     wStream* output_stream = guac_rdpdr_new_io_completion(device,
--- a/src/protocols/rdp/guac_rdpdr/rdpdr_printer.c
+++ b/src/protocols/rdp/guac_rdpdr/rdpdr_printer.c
@@ -21,6 +21,7 @@
 
 #include "rdpdr_messages.h"
 #include "rdpdr_printer.h"
+#include "rdpdr_print_job.h"
 #include "rdpdr_service.h"
 #include "rdp_status.h"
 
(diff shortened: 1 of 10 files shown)


src/protocols/rdp/rdp_print_job.h
```diff
@@ -0,0 +1,232 @@
...
+ *     The PostScript print data to write to the given print job.
+ *
+ * @param length
+ *     The number of bytes of PostScript print data to write.
+ *
+ * @return
+ *     The number of bytes written, or -1 if an error occurs which prevents
+ *     further writes.
+ */
+int guac_rdp_print_job_write(guac_rdp_print_job* job,
```
jmuehlner: Missing javadoc for the actual job.

mike-jumper: 😓

src/protocols/rdp/guac_rdpdr/rdpdr_printer.c
```diff
@@ -373,21 +110,25 @@ void guac_rdpdr_process_print_job_write(guac_rdpdr_device* device,
...
-    guac_rdpdr_printer_data* printer_data =
-        (guac_rdpdr_printer_data*) device->data;
+    guac_client* client = device->rdpdr->client;
+    guac_rdp_client* rdp_client = (guac_rdp_client*) client->data;
+    guac_rdp_print_job* job = (guac_rdp_print_job*) rdp_client->active_job;
+
+    /* End print job */
+    if (job != NULL) {
+        guac_rdp_print_job_free(job);
+        device->data = NULL;
```
jmuehlner: Is this what you mean to be clearing here?

mike-jumper: Erg. Definitely not.

jmuehlner:
Looks good to me!

PR #60, opened 2017-02-16
GUACAMOLE-206: Do not drop keys while Alt is held.

mike-jumper:


jmuehlner:
Oops - well it's good to fix this.

PR #62, opened 2017-02-16
GUACAMOLE-208: Define new status codes for network/connection state. Leverage within protocol support.

mike-jumper:


jmuehlner:
Why only leverage these within RDP? These look pretty universal.

mike-jumper:
Fair enough. I'll quickly update the status code usage elsewhere.

mike-jumper:
The branch name is somewhat silly now, but the other protocols should be OK. It turns out they were already pretty alright with the existing codes - only `UPSTREAM_NOT_FOUND` was relevant (for network connect failures).

jmuehlner:
All right, looks good to me now!

flangelo:
+1

PR #64, opened 2017-02-26
GUACAMOLE-188: Use Porter-Duff "over" operator when drawing with ARGB surfaces.

mike-jumper:


jmuehlner:
Is this the complete fix? I noticed this comment on the JIRA issue:

"This isn't quite done - the alpha channel is being ignored in all but the fully-opaque and fully-transparent cases."

mike-jumper:
Yep. That is exactly what this PR addresses - the not-quite-doneness of the previously-merged code.

jmuehlner:
👍

PR #69, opened 2017-02-27
GUACAMOLE-169: Move common-ssh and terminal headers to namespaced directories.

mike-jumper:


mike-jumper:
**NOTE:** Merge destination is `staging/0.9.12-incubating`. See [GUACAMOLE-169](https://issues.apache.org/jira/browse/GUACAMOLE-169) and #37.

I suspect this will conflict when `staging/0.9.12-incubating` is merged back to `master`, so if/when merge occurs, send me a share link if you want a double check on the merge result.

PR #73, opened 2017-03-01
GUACAMOLE-226: Properly initialize new layers as fully opaque.

mike-jumper:
All layers are fully opaque by default, and `guac_common_surface` should accurately represent this. Doing otherwise means that newly-joining users of a shared connection will receive the wrong layer opacity, effectively hiding all but the default layer.


mike-jumper:
**NOTE:** Intended for `staging/0.9.12-incubating`. This is a regression due to changes included in 0.9.12-incubating ([GUACAMOLE-176](https://issues.apache.org/jira/browse/GUACAMOLE-176)).

PR #74, opened 2017-03-10
GUACAMOLE-239: When available, use CLOCK_MONOTONIC to prevent backwards running time

gamlerhart:
Fix disconnects due to back ward running time. Fix by using CLOCK_MONOTONIC.

Fix: https://issues.apache.org/jira/browse/GUACAMOLE-239

### DIFF
--- a/src/libguac/timestamp.c
+++ b/src/libguac/timestamp.c
@@ -33,9 +33,13 @@ guac_timestamp guac_timestamp_current() {
 
     struct timespec current;
 
-    /* Get current time */
+    /* Get current time, monotonically increasing */ 
+#ifdef CLOCK_MONOTONIC
+    clock_gettime(CLOCK_MONOTONIC, &current);
+#else
     clock_gettime(CLOCK_REALTIME, &current);
-    
+#endif    
+
     /* Calculate milliseconds */
     return (guac_timestamp) current.tv_sec * 1000 + current.tv_nsec / 1000000;
 


mike-jumper:
Is `CLOCK_MONOTONIC` always available?

Reading [the manpage for `clock_gettime()`](https://linux.die.net/man/3/clock_gettime), it looks like `CLOCK_REALTIME` is the only value guaranteed in all cases. If so, that's OK, but there would need to be a configure test to check whether `CLOCK_MONOTONIC` is present before using it, relying on `CLOCK_REALTIME` if necessary.

gamlerhart:
I don't know what platform are missing CLOCK_MONOTONIC.
My assumption is, that 'server/desktop' linux will have CLOCK_MONOTONIC.

mike-jumper:
According to POSIX, that isn't a safe assumption. [The standard's definition for `clock_gettime()`](http://pubs.opengroup.org/onlinepubs/9699919799/functions/clock_getres.html) describes support for `CLOCK_MONOTONIC` as optional.

Thankfully, [the standard's definition for `time.h`](http://pubs.opengroup.org/onlinepubs/9699919799/basedefs/time.h.html) requires that `CLOCK_MONOTONIC` be defined as a macro, so the issues with portability here are addressable with a simple `#ifdef CLOCK_MONOTONIC` (no need to write a configure test).

gamlerhart:
Added ifdef check around the `CLOCK_MONOTINIC` usage.

src/libguac/timestamp.c
```diff
@@ -33,9 +33,21 @@ guac_timestamp guac_timestamp_current() {
 
     struct timespec current;
 
-    /* Get current time */
+    /* Get current time, monotinically increasing.
```
mike-jumper: monotonically*

src/libguac/timestamp.c
```diff
@@ -33,9 +33,21 @@ guac_timestamp guac_timestamp_current() {
...
+       guacd expects that a client sync timestamp <= last_sent_timestamp. 
+       When time moves backwards, when last_sent_timestamp can move back, 
+       violating the assumption of forward moving time.
+       Ex. Obsorved a Azure Linux instance,
```
mike-jumper: observed*

src/libguac/timestamp.c
```diff
@@ -33,9 +33,21 @@ guac_timestamp guac_timestamp_current() {
...
 
-    /* Get current time */
+    /* Get current time, monotinically increasing.
+       guacd expects that a client sync timestamp <= last_sent_timestamp. 
+       When time moves backwards, when last_sent_timestamp can move back, 
+       violating the assumption of forward moving time.
+       Ex. Obsorved a Azure Linux instance, 
+       where CLOCK_REALTIME jumped back often, causing disconnects.
+     
+    Posix does not require CLOCK_MONOTONIC.
```
mike-jumper: POSIX*

Why is the indentation of this part of the block comment different from the rest?

src/libguac/timestamp.c
```diff
@@ -33,9 +33,21 @@ guac_timestamp guac_timestamp_current() {
 
     struct timespec current;
 
-    /* Get current time */
+    /* Get current time, monotinically increasing.
+       guacd expects that a client sync timestamp <= last_sent_timestamp.
```
mike-jumper: Perhaps leaving this whole thing simply as:

    /* Get current time, monotonically increasing */

would be better?

While that is indeed enforced, and violation of that rule results in the connection being summarily closed, the definition of `guac_timestamp_current()` already states:

> The difference between return values of any two calls is equal to the amount of time in milliseconds between those calls.

which implicitly requires that the time be monotonically increasing. This function isn't supposed to care about the internals of the Guacamole protocol, guacd, libguac, etc.; only timestamps. It's good to document this within the JIRA issue related to the change (which I believe you've already done), but documenting external behavior here violates separation of concerns IMHO, and the function definition itself is already enough to require this change.

ie: The issue is not that guacd expects timestamps will be monotonically increasing, but rather that it (and any other user of this function) is supposed to be able to rely on that fact.

mike-jumper:
Beyond the points mentioned above, I just noticed: is there any reason you're copying the JIRA issue summary into each commit?

By convention, git commits within the Guacamole codebase consist of the JIRA issue tag, a colon, and then the high-level message describing the change. You more-or-less have that, but it's redundant (and inconsistent with the rest of the git history) to copy the summary.

When you get a chance ... mind amending your commits to correct that?

Other than that - looks good to me.

gamlerhart:
Did:
- Reduce the comment, without all the 'background' information.
- Squashed and amended commit message.

src/libguac/timestamp.c
```diff
@@ -29,13 +31,21 @@
...
+#undef CLOCK_MONOTONIC
 #ifdef HAVE_CLOCK_GETTIME
 
     struct timespec current;
 
-    /* Get current time */
+    /* Get current time, monotonically increasing. 
+     Using monotonic clock when available, to prevent backward running time */
+#ifdef CLOCK_MONOTONIC   
+    printf("MONOTONIC USED");
```
mike-jumper: Please remove the debugging statements. Looks like these `printf()`, the `#include <stdio.h>`, and the `#undef CLOCK_MONOTONIC` must have snuck in here during testing. ;)

gamlerhart: I'm so sorry. Yes...this was part of the testing.

Fixing.

gamlerhart: Fixed...Thanks for your patience to clean up my commits.

src/libguac/timestamp.c
```diff
@@ -33,9 +33,14 @@ guac_timestamp guac_timestamp_current() {
 
     struct timespec current;
 
-    /* Get current time */
+    /* Get current time, monotonically increasing. 
+     Using monotonic clock when available, to prevent backward running time */
```
mike-jumper: The only other issue I see here with respect to wording is:

> Using monotonic clock when available, to prevent backward running time

seems to simply restate that the time is monotonically increasing, which is already stated above... essentially, the whole comment reads to me as: "Get current time, monotonically increasing. Use monotonically increasing time, to ensure time is monotonically increasing."

If you disagree here and feel it adds something, I'm willing to back off and let it be, but it should at least be indented to align with the content of the comment line above it (with leading `*`). I think I may have mentioned this before in prior review, but if not - my apologies. Currently, it's at the same level as a normal line of code, which hurts readability.

Searching through the rest of the guacamole-server codebase, it's rare that we have any comments which span multiple lines, so finding precedent to draw from is tricky, but there are a few examples of how things look in established code, such as:

    /* Iterate over all the heat map cells for the area
     * and calculate the average framerate */
    for (y = min_y; y < max_y; y++) {

(see https://github.com/apache/incubator-guacamole-server/blob/516c4a0593fd78604774bf52c9b547aad5e22619/src/common/surface.c#L395-L396)

Other than that, unless your next revision adds more `printf()` statements, I think we're good to go. ;)

src/libguac/timestamp.c
mike-jumper: No problem. Thanks for your patience in cleaning up your commits.

src/libguac/timestamp.c
gamlerhart: Dropped the second, repeating comment line =)

anycodes:
Hello, I am a Chinese, my English is not very good. When I use this soft(guacamole) in Ubuntu-16 ,my browser always tells me : Connection closed, But I find all my hard disk, I did not find this file which named timestamp.c,  I only found a file named timestamp.h, but it is block! Could you help me?

mike-jumper:
Hi @anycodes,

This is a closed pull request for changes which have already been merged. If you have questions regarding Guacamole, please instead pose them in one of the mailing lists:

http://guacamole.incubator.apache.org/support/#mailing-lists

anycodes:
Thank You

PR #75, opened 2017-03-10
GUACAMOLE-231: Notify terminal when display is modified due to mouse cursor movement.

mike-jumper:


mike-jumper:
**NOTE:** Intended for `staging/0.9.12-incubating` (regression due to [GUACAMOLE-117](https://issues.apache.org/jira/browse/GUACAMOLE-177)).

PR #76, opened 2017-03-11
GUACAMOLE-231: Report server-side mouse position using "mouse" instruction.

mike-jumper:
This change adds a new "mouse" instruction which reports the current location of the mouse cursor to all users not actively moving the mouse, allowing the client to handle rendering of the shared cursor (rather than relying on the server to render the cursor itself).

### DIFF
--- a/src/common/cursor.c
+++ b/src/common/cursor.c
@@ -26,7 +26,6 @@
 
 #include <cairo/cairo.h>
 #include <guacamole/client.h>
-#include <guacamole/layer.h>
 #include <guacamole/protocol.h>
 #include <guacamole/socket.h>
 #include <guacamole/user.h>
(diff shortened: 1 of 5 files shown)


mike-jumper:
**NOTE:** Intended for `staging/0.9.12-incubating` (regression due to [GUACAMOLE-117](https://issues.apache.org/jira/browse/GUACAMOLE-177)).

Client-side implementation of this is on apache/incubator-guacamole-client#128.

mike-jumper:
Errr ... one moment - need to fetch (and remove) the now-irrelevant changes from #75.

mike-jumper:
There. Should be all good.

src/common/cursor.c
```diff
@@ -295,25 +209,19 @@ void guac_common_cursor_set_argb(guac_common_cursor* cursor, int hx, int hy,
...
-            cursor->y - hy,
-            INT_MAX);
-
     /* Broadcast new cursor image to all users */
-    guac_protocol_send_size(cursor->client->socket, cursor->layer,
+    guac_protocol_send_size(cursor->client->socket, cursor->buffer,
             width, height);
 
-    guac_client_foreach_user(cursor->client, __send_user_cursor_image, cursor);
+    guac_client_stream_png(cursor->client, cursor->client->socket,
```
jmuehlner: Why is guac_client_foreach_user no longer required here?

mike-jumper: It was actually never required.

I think this was written prior to implementation of `guac_client_stream_png()` and client-scoped streams. The old `guac_client_foreach_user()` call here just invoked `guac_user_stream_png()` for each connected user. That's equivalent to `guac_client_stream_png()` in that each user receives the image, but bad because the encoding process occurs once per user.

With `guac_client_stream_png()`, a stream is allocated at the client scope and the image (encoded only the once) is streamed over the broadcast socket.

jmuehlner: Ok, makes sense.

PR #77, opened 2017-03-16
GUACAMOLE-240: Always pass guac_pa_stream to PulseAudio callbacks.

mike-jumper:
To avoid this sort of problem going forward, this change also adds long-missing documentation for the PulseAudio callbacks being used.

### DIFF
--- a/src/pulse/pulse.c
+++ b/src/pulse/pulse.c
@@ -58,6 +58,20 @@ static int guac_pa_is_silence(const void* buffer, size_t length) {
 
 }
 
+/**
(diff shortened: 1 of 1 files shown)


mike-jumper:
**NOTE:** Intended for `staging/0.9.12-incubating` (regression due to [GUACAMOLE-179](https://issues.apache.org/jira/browse/GUACAMOLE-179)).

jmuehlner:
Yikes! Well it certainly looks better now. Good catch. And I also appreciate the new documentation for existing functions.

PR #80, opened 2017-03-31
GUACAMOLE-259: Add TRACE log level.

mike-jumper:
We've avoided adding log levels below DEBUG, as syslog does not support such levels, but as a consequence we are forced to avoid logging which may affect performance, useful though it may be.

This change adds a new TRACE log level to libguac's existing logging facilities, allowing logging of low-level events (like those required by [GUACAMOLE-259](https://issues.apache.org/jira/browse/GUACAMOLE-259)) which would otherwise pollute use of the DEBUG level.


jmuehlner:
Looks reasonable to me!

PR #83, opened 2017-04-03
GUACAMOLE-262: set VNC trueColour flag to 1 instead of 255

dborth:


src/protocols/vnc/display.c
```diff
@@ -153,6 +153,7 @@ void guac_vnc_copyrect(rfbClient* client, int src_x, int src_y, int w, int h, in
 }
 
 void guac_vnc_set_pixel_format(rfbClient* client, int color_depth) {
+    client->format.trueColour = 1;
```
mike-jumper: So this applies to all pixel formats regardless of desired depth?

dborth: Yes, that's correct. The default in rfbGetClient() in libvncserver which you're calling is already setting:
client->format.trueColour = TRUE;

This is just overriding that value so that it's 1 instead of -1.

mike-jumper:
Well ... It's unfortunate that this has to be done, but I guess it is what it is. This is definitely not the only workaround for libvncclient quirks present in the guacamole-server codebase. We should probably look into migrating away from libvncclient when possible.

PR #86, opened 2017-04-23
GUACAMOLE-278: Add support for xterm's 256-color control codes

mike-jumper:
This change adds support for the codes used by xterm to set colors beyond the normal 16 color terminal palette. Specifically:

* Set foreground to RGB color: `CSI 38 ; 2 ; R ; G ; B`
* Set foreground to 256-color palette index N: `CSI 38 ; 5 ; N`
* Set background to RGB color: `CSI 48 ; 2 ; R ; G ; B`
* Set background to 256-color palette index N: `CSI 48 ; 5 ; N`

See: http://invisible-island.net/xterm/ctlseqs/ctlseqs.html.


mike-jumper:
Sorry - I just now made corrective changes after having already opened the PR, so things have been updated here.

Handling of OSC should indeed stop after an invalid character is read (that aspect of these changes has now been restored to the old behavior). Since invalid OSC characters result the remainder of the sequence being echoed to the terminal, I had to add stub handling for the 256-color palette redefinition OSC, which programs that depend on 256-color codes will often use. That sequence will need to be implemented at a later time.

mike-jumper:
For reference, before:

guac-terminal-256-before

After:

guac-terminal-256-after

src/terminal/display.c
```diff
@@ -127,24 +104,27 @@ int __guac_terminal_hash_codepoint(int codepoint) {
...
+        foreground = &attributes->foreground;
+        background = &attributes->background;
     }
 
     /* Handle bold */
-    if (attributes->bold && foreground <= 7)
-        foreground += 8;
+    if (attributes->bold && foreground->palette_index >= 0
+            && foreground->palette_index <= 7) {
+        foreground = &guac_terminal_palette[foreground->palette_index + 8];
```
jmuehlner: Constants that better explain what this 7 and 8 are about would make me happy.

mike-jumper: Alrighty. I will do this.

src/terminal/terminal_handlers.c
```diff
@@ -779,13 +919,30 @@ int guac_terminal_csi(guac_terminal* term, unsigned char c) {
 
                     /* Foreground */
                     else if (value >= 30 && value <= 37)
```
jmuehlner: Some constants for these numbers here would also make me very happy.

mike-jumper: If you're referring to 30 and 37 here, these are the values from the standard defining the console codes used by the Guacamole terminal emulator.

Might not be a bad idea to do go through the terminal emulator code and add constants for all of these, but I'd say we should do that as a separate cleanup issue, rather than having these be the only two such constants.

jmuehlner: Sounds fair to me.

jmuehlner:
Ok, everything looks good to me. If @flangelo or anyone else sees any issues with this, we can always come back to it later. For now, I'm going to merge this so we can all enjoy the :rainbow:

PR #88, opened 2017-04-27
GUACAMOLE-278: Add support for xterm's bright SGR colors (90-97 and 100-107).

mike-jumper:


src/terminal/terminal_handlers.c
```diff
@@ -967,6 +967,18 @@ int guac_terminal_csi(guac_terminal* term, unsigned char c) {
                         term->current_attributes.background =
                             term->default_char.attributes.background;
 
+                    /* Intense foreground */
+                    else if (value >= 90 && value <= 97)
```
jmuehlner: You probably know what I'm going to say here :stuck_out_tongue:  :memo:

jmuehlner: But yeah, some constants would be nice in the future...

mike-jumper: :sob:

PR #89, opened 2017-04-27
GUACAMOLE-265: Set connection name when terminal window title is changed.

mike-jumper:


src/terminal/terminal_handlers.c
```diff
@@ -1196,6 +1226,10 @@ int guac_terminal_osc(guac_terminal* term, unsigned char c) {
         else if (operation == 482203)
             term->char_handler = guac_terminal_close_pipe_stream;
 
+        /* Set window title OSC */
+        else if (operation == 0 || operation == 2)
```
jmuehlner: Same as on the other PR - it would be lovely if these were constants, but that can wait for a cleanup change later I suppose.

mike-jumper: I hereby nominate you to open an impassioned JIRA issue to track this cleanup.

PR #92, opened 2017-05-21
GUACAMOLE-280: Add support for faint (low) intensity SGR.

mike-jumper:


src/terminal/display.c
```diff
@@ -118,7 +118,7 @@ int __guac_terminal_set_colors(guac_terminal_display* display,
     }
 
     /* Handle bold */
-    if (attributes->bold
+    if (attributes->bold && !attributes->half_bright
```
jmuehlner: I guess if both bold and half_bright are set, they cancel each other out. That makes sense.

PR #93, opened 2017-05-31
GUACAMOLE-203: Support SSH Keep Alive

necouchman:
This is my stab at implementing the equivalent of ServerAliveInterval in guacd via the libssh2_keepalive_config() and libssh2_keepalive_send() calls.  It definitely needs some more testing and someone else to look and make sure I implemented it correctly, but from the documentation I was able to find on libssh2 and those two methods, I think it should do the trick.

### DIFF
--- a/src/common-ssh/common-ssh/ssh.h
+++ b/src/common-ssh/common-ssh/ssh.h
@@ -98,7 +98,7 @@ void guac_common_ssh_uninit();
  *     if the connection or authentication were not successful.
  */
 guac_common_ssh_session* guac_common_ssh_create_session(guac_client* client,
-        const char* hostname, const char* port, guac_common_ssh_user* user, const int keepalive);
+        const char* hostname, const char* port, guac_common_ssh_user* user, int keepalive);
 
 /**
  * Disconnects and destroys the given SSH session, freeing all associated
--- a/src/common-ssh/common-ssh/ssh.h
+++ b/src/common-ssh/common-ssh/ssh.h
@@ -98,7 +98,7 @@ void guac_common_ssh_uninit();
  *     if the connection or authentication were not successful.
  */
 guac_common_ssh_session* guac_common_ssh_create_session(guac_client* client,
-        const char* hostname, const char* port, guac_common_ssh_user* user);
+        const char* hostname, const char* port, guac_common_ssh_user* user, const int keepalive);
 
 /**
  * Disconnects and destroys the given SSH session, freeing all associated
--- a/src/common-ssh/ssh.c
+++ b/src/common-ssh/ssh.c
@@ -536,6 +536,12 @@ guac_common_ssh_session* guac_common_ssh_create_session(guac_client* client,
     if (keepalive > 0)
         libssh2_keepalive_config(common_session->session, 1, keepalive);
 
+    /* Warn if keepalive below minimum value */
+    if (keepalive == 1) {
+        guac_user_log(user, GUAC_LOG_WARNING, "keepalive interval will "
+            "be rounded up to minimum value of 2.");
+    }
+
     /* Return created session */
     return common_session;
 
--- a/src/common-ssh/ssh.c
+++ b/src/common-ssh/ssh.c
@@ -414,7 +414,7 @@ static int guac_common_ssh_authenticate(guac_common_ssh_session* common_session)
 }
 
 guac_common_ssh_session* guac_common_ssh_create_session(guac_client* client,
-        const char* hostname, const char* port, guac_common_ssh_user* user) {
+        const char* hostname, const char* port, guac_common_ssh_user* user, const int keepalive) {
 
     int retval;
 
@@ -531,6 +531,10 @@ guac_common_ssh_session* guac_common_ssh_create_session(guac_client* client,
         close(fd);
         return NULL;
     }
+
+    /* Configure session keepalive */
+    if (keepalive > 0)
+        libssh2_keepalive_config(common_session->session, 1, keepalive);
 
     /* Return created session */
     return common_session;
--- a/src/common-ssh/ssh.c
+++ b/src/common-ssh/ssh.c
@@ -414,7 +414,7 @@ static int guac_common_ssh_authenticate(guac_common_ssh_session* common_session)
 }
 
 guac_common_ssh_session* guac_common_ssh_create_session(guac_client* client,
-        const char* hostname, const char* port, guac_common_ssh_user* user, const int keepalive) {
+        const char* hostname, const char* port, guac_common_ssh_user* user, int keepalive) {
 
     int retval;
 
--- a/src/common-ssh/ssh.c
+++ b/src/common-ssh/ssh.c
@@ -538,7 +538,7 @@ guac_common_ssh_session* guac_common_ssh_create_session(guac_client* client,
         guac_client_log(client, GUAC_LOG_WARNING, "negative keepalive intervals "
             "are converted to 0, disabling keepalive.");
     }
-    else if(keepalive == 1) {
+    else if (keepalive == 1) {
         guac_client_log(client, GUAC_LOG_WARNING, "keepalive interval will "
             "be rounded up to minimum value of 2.");
     }
--- a/src/common-ssh/ssh.c
+++ b/src/common-ssh/ssh.c
@@ -538,7 +538,7 @@ guac_common_ssh_session* guac_common_ssh_create_session(guac_client* client,
 
     /* Warn if keepalive below minimum value */
     if (keepalive == 1) {
-        guac_user_log(user, GUAC_LOG_WARNING, "keepalive interval will "
+        guac_client_log(client, GUAC_LOG_WARNING, "keepalive interval will "
             "be rounded up to minimum value of 2.");
     }
 
--- a/src/common-ssh/ssh.c
+++ b/src/common-ssh/ssh.c
@@ -532,15 +532,19 @@ guac_common_ssh_session* guac_common_ssh_create_session(guac_client* client,
         return NULL;
     }
 
-    /* Configure session keepalive */
-    if (keepalive > 0)
-        libssh2_keepalive_config(common_session->session, 1, keepalive);
-
     /* Warn if keepalive below minimum value */
-    if (keepalive == 1) {
+    if (keepalive < 0) {
+        keepalive = 0;
+        guac_client_log(client, GUAC_LOG_WARNING, "negative keepalive intervals "
+            "are converted to 0, disabling keepalive.");
+    }
+    else if(keepalive == 1) {
         guac_client_log(client, GUAC_LOG_WARNING, "keepalive interval will "
             "be rounded up to minimum value of 2.");
     }
+
+    /* Configure session keepalive */
+    libssh2_keepalive_config(common_session->session, 1, keepalive);
 
     /* Return created session */
     return common_session;
--- a/src/protocols/rdp/rdp_settings.c
+++ b/src/protocols/rdp/rdp_settings.c
@@ -790,7 +790,7 @@ guac_rdp_settings* guac_rdp_parse_args(guac_user* user,
                 IDX_SFTP_SERVER_ALIVE_INTERVAL, 0);
     if (settings->sftp_server_alive_interval == 1)
         guac_user_log(user, GUAC_LOG_WARNING, "The minimum allowed "
-                "value for keepalives is 2 seconds.");
+                "value for keepalives is 2 seconds, this will be rounded up.");
 #endif
 
     /* Read recording path */
--- a/src/protocols/rdp/rdp_settings.c
+++ b/src/protocols/rdp/rdp_settings.c
@@ -84,7 +84,7 @@ const char* GUAC_RDP_CLIENT_ARGS[] = {
     "sftp-private-key",
     "sftp-passphrase",
     "sftp-directory",
-    "sftp-keepalive",
+    "sftp-server-alive-interval",
 #endif
 
     "recording-path",
@@ -373,7 +373,7 @@ enum RDP_ARGS_IDX {
      * of 1 is automatically increased to 2 by libssh2 to avoid busy loop corner
      * cases.
      */
-    IDX_SFTP_KEEPALIVE,
+    IDX_SFTP_SERVER_ALIVE_INTERVAL,
 #endif
 
     /**
@@ -787,7 +787,7 @@ guac_rdp_settings* guac_rdp_parse_args(guac_user* user,
     /* Default keepalive value */
     settings->sftp_keepalive =
         guac_user_parse_args_int(user, GUAC_RDP_CLIENT_ARGS, argv,
-                IDX_SFTP_KEEPALIVE, 0);
+                IDX_SFTP_SERVER_ALIVE_INTERVAL, 0);
     if (settings->sftp_keepalive == 1)
         guac_user_log(user, GUAC_LOG_WARNING, "The minimum allowed "
                 "value for keepalives by libssh2 is 2 seconds.");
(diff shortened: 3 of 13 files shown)


src/protocols/ssh/ssh.c
```diff
@@ -318,11 +323,18 @@ void* ssh_client_thread(void* data) {
 
     /* While data available, write to terminal */
     int bytes_read = 0;
+    int timeout = 0;
```
mike-jumper: The scope of `timeout` can be tightened - there's no need to declare it outside the only block which uses its value (the `if` block within the loop).

src/protocols/ssh/ssh.c
```diff
@@ -318,11 +323,18 @@ void* ssh_client_thread(void* data) {
...
     for (;;) {
 
         /* Track total amount of data read */
         int total_read = 0;
 
+        /* Set up return value for keepalives */
+        int alive = 0;
```
mike-jumper: I recommend cleaning this up a bit:

It's unnecessarily wide scope (it's only used within that `if` block), unnecessary initialization (this initial zero value will always be overwritten by the call to `libssh2_keepalive_send()`), and an unnecessary variable (you can just as easily simply call `libssh2_keepalive_send()` within the `if` that checks the return value).

src/protocols/ssh/ssh.c
```diff
@@ -318,11 +323,18 @@ void* ssh_client_thread(void* data) {
...
+
+        /* Timer for keepalives */
+        int sleep = 0;
```
mike-jumper: Unnecessary initialization. This value is always overwritten.

src/protocols/ssh/ssh.c
```diff
@@ -331,6 +343,18 @@ void* ssh_client_thread(void* data) {
...
+        if (settings->server_alive_interval > 0) {
+            alive = libssh2_keepalive_send(ssh_client->session->session, &timeout);
+            /* Sending the keepalive failed, so we break out */
+            if (alive > 0)
+                break;
+            sleep = timeout * 1000;
+        }
+        /* If keepalive is not configured, sleep for the default of 1 second */
+        else
+            sleep = 1000;
```
mike-jumper: We should probably take this opportunity to define a constant for that `1000`.

src/protocols/ssh/settings.c
```diff
@@ -165,6 +166,12 @@ enum SSH_ARGS_IDX {
      */
     IDX_READ_ONLY,
 
+    /**
+     * Number of seconds between sending alive packets.  A default of 0
```
mike-jumper: The [documentation for `libssh2_keepalive_config()`](https://www.libssh2.org/libssh2_keepalive_config.html) defines some behavior which will make this not exactly correct:

> To avoid some busy-loop corner-cases, if you specify an interval of 1 it will be treated as 2.

That oddity will need to be documented. Perhaps we should just define 2 as the minimum value allowed?

src/protocols/ssh/ssh.c
mike-jumper: `sleep` is a bit of a misnomer here, since we aren't going to actually explicitly sleep for that period of time, but rather wait for additional data for no more than this amount of time.

mike-jumper:
Do similar things need to be done for SFTP?

src/protocols/ssh/ssh.c
necouchman: Okay, I think I have it placed correctly, now.

src/protocols/ssh/ssh.c
necouchman: Removed alive and called the libssh2_keepalive_send() method inside the if() statement as you suggested.

src/protocols/ssh/ssh.c
necouchman: Changed the name to "timer" - hopefully that makes more sense?  If not I'll take suggestions...

src/protocols/ssh/ssh.c
necouchman: Defined GUAC_SSH_DEFAULT_POLL_TIMER in settings.h and used that here, instead.  Is that placement/name okay?

src/protocols/ssh/settings.c
necouchman: I beefed up that comment to document that - I'll also verify what I put in the manual PR to make sure that correctly documents it.  I also changed some of the checks within the code to > 1 instead of > 0.  However, I'm not sure if it makes sense to actually change those and do any additional checks, or just let libssh2 handle it since libssh2 already handles it - as long as it's documented??  After changing those checks to > 1 it seems like maybe that would be more troublesome to someone trying to enable keepalives than just checking for > 0 and making sure documentations specifies that the minimum will never be < 2 seconds.  Or is there another check/warning/error I should implement here to throw something into the log file that warns of this configuration issue?

necouchman:
Yeah, probably need to look at the SFTP code and see about that.  I'm guessing we want separate options for that...

necouchman:
Okay, did some refactoring to support the keepalive for both SSH and SFTP across all protocols.  I moved the libssh2_keepalive_config() call into the guac_common_ssh_create_session() function to unify this across all of the protocols that support sftp.  I also added in the settings entries for VNC and RDP.

src/protocols/ssh/settings.c
necouchman: I did end up adding some logging to the settings to check when the user enters 1 for the keepalive interval and spit out a warning in the log.  Hopefully this, plus documentation, addresses this concern?

necouchman:
Ping.

@mike-jumper: Any further comments on this?  Other things that need to be tweaked?

src/protocols/rdp/rdp_settings.c
```diff
@@ -775,6 +783,14 @@ guac_rdp_settings* guac_rdp_parse_args(guac_user* user,
     settings->sftp_directory =
         guac_user_parse_args_string(user, GUAC_RDP_CLIENT_ARGS, argv,
                 IDX_SFTP_DIRECTORY, NULL);
+
+    /* Default keepalive value */
+    settings->sftp_keepalive =
+        guac_user_parse_args_int(user, GUAC_RDP_CLIENT_ARGS, argv,
+                IDX_SFTP_KEEPALIVE, 0);
+    if (settings->sftp_keepalive == 1)
```
mike-jumper: Shouldn't this be `<= 1` ?

src/protocols/rdp/rdp_settings.c
```diff
@@ -775,6 +783,14 @@ guac_rdp_settings* guac_rdp_parse_args(guac_user* user,
...
+        guac_user_log(user, GUAC_LOG_WARNING, "The minimum allowed "
+                "value for keepalives by libssh2 is 2 seconds.");
```
mike-jumper: It's important that the reasoning for behavior be documented internally, but citing such internals in the logs will only lead to confusion. To the administrator, they're looking through the logs to debug Guacamole, not its dependencies, and the fact that it's libssh2 that imposes this restriction has no practical implications. All that matters is that the specified keep-alive interval has been automatically rounded up to the minimum value of 2.

src/protocols/ssh/settings.c
```diff
@@ -279,6 +287,14 @@ guac_ssh_settings* guac_ssh_parse_args(guac_user* user,
         guac_user_parse_args_boolean(user, GUAC_SSH_CLIENT_ARGS, argv,
                 IDX_CREATE_RECORDING_PATH, false);
 
+    /* Parse server alive interval */
+    settings->server_alive_interval =
+        guac_user_parse_args_int(user, GUAC_SSH_CLIENT_ARGS, argv,
+                IDX_SERVER_ALIVE_INTERVAL, 0);
+    if (settings->server_alive_interval == 1)
+        guac_user_log(user, GUAC_LOG_WARNING, "Minimum keepalive interval "
```
mike-jumper: Perhaps this should instead be logged within the shared SSH client code, rather than repeated for all users of that shared code?

src/protocols/ssh/settings.h
```diff
@@ -54,6 +54,11 @@
 #define GUAC_SSH_DEFAULT_RECORDING_NAME "recording"
 
 /**
+ * The default polling timer for SSH activity in milliseconds.
```
mike-jumper: I don't get the use of "timer" in referring to the polling timeout. A timer is an object, not a length of time. There must be a better way to word this which matches standard terminology.

My suggestion would be "timeout", but I'm open to anything that makes reasonable sense.

src/protocols/ssh/ssh.c
```diff
@@ -331,6 +334,17 @@ void* ssh_client_thread(void* data) {
             break;
         }
 
+        /* Send keepalive at configured interval */
+        if (settings->server_alive_interval > 0) {
+            int timeout = 0;
+            if(libssh2_keepalive_send(ssh_client->session->session, &timeout) > 0)
```
mike-jumper: `if` is not a function.

src/protocols/vnc/settings.c
```diff
@@ -66,6 +66,7 @@ const char* GUAC_VNC_CLIENT_ARGS[] = {
     "sftp-private-key",
     "sftp-passphrase",
     "sftp-directory",
+    "sftp-keepalive",
```
mike-jumper: I suggest finding consistent wording between the `sftp-` parameters and those used by the SSH client. Here, things are referred to as `keepalive`, while the SSH client uses `server-alive-interval`. One or the other should be altered such that things are predictable.

mike-jumper:
As an aside, I'm not sure the best way to deal with this, but the way the documentation is referring to the internals of dependencies when describing the minimum values allowed makes me itchy. It's an important thing to document internally for sure, but from a user's perspective it's unnecessary. Such details should be encapsulated.

If we are limited to values >= 2 for the keep-alive interval, I feel like we should just own that restriction, document it as the minimum allowed value, ensure that any other non-zero values are rounded up to the known-good value of 2, and leave it at that.

src/protocols/rdp/rdp_settings.c
necouchman: Well, 0 is also a valid value for disabling keepalives, so <= 1 wouldn't work, I don't think.  However, it is probably also good to check for values < 0 and avoid those.

src/protocols/ssh/ssh.c
necouchman: Fixed.

src/protocols/vnc/settings.c
necouchman: Tweaked this, and several of the other instances of keepalive to server-alive-interval.

necouchman:
> As an aside, I'm not sure the best way to deal with this, but the way the documentation is referring to the internals of dependencies when describing the minimum values allowed makes me itchy. It's an important thing to document internally for sure, but from a user's perspective it's unnecessary. Such details should be encapsulated.

I'm fine removing references in error logs, etc., to libssh2's limitation on the keepalive.

> If we are limited to values >= 2 for the keep-alive interval, I feel like we should just own that restriction, document it as the minimum allowed value, ensure that any other non-zero values are rounded up to the known-good value of 2, and leave it at that.

Fine by me.

As a side note I changed the type of this parameter from int to unsigned, as this matches what libssh2 uses internally, since anything < 0 is invalid.  This really puts 1 as the only value that has to be dealt with - 0 is valid for disabling keepalives, and >= 2 is fine.

src/protocols/rdp/rdp_settings.c
necouchman: So, I made these unsigned instead of plain int variables, which should help eliminate the negative values.  However, I'm wondering if something needs to be done client-side to prevent negative values??

src/protocols/rdp/rdp_settings.c
necouchman: Tweaked the error message.

src/protocols/ssh/settings.c
necouchman: Makes sense...done.

src/protocols/ssh/settings.h
necouchman: Changed both the comment and the constant to timeout instead of timer.

necouchman:
Okay, tweaked several things, hopefully it's in better shape.  Anything else that needs attention?

necouchman:
Ping...anything else that needs attention, here?

src/protocols/rdp/rdp_settings.h
```diff
@@ -359,6 +359,14 @@ typedef struct guac_rdp_settings {
...
      */
     char* sftp_directory;
+
+    /**
+     * The interval at which SSH keepalive messages are sent to the server for
+     * SFTP connections.  The default is 0 (disabling keepalives), and a value
+     * of 1 is automatically increased to 2 by libssh2 to avoid busy loop corner
+     * cases.
+     */
+    unsigned sftp_server_alive_interval;
```
mike-jumper: Using an `unsigned int` for the result of `guac_user_parse_args_int()` is dangerous. Because `guac_user_parse_args_int()` accepts negative values, such values will be assigned directly without any warning and interpreted as enormous positive values.

If you want to disallow negative values, it would be better to explicitly check for those values and round up to a legal value as necessary.

src/protocols/ssh/ssh.c
```diff
@@ -323,6 +323,9 @@ void* ssh_client_thread(void* data) {
         /* Track total amount of data read */
         int total_read = 0;
 
+        /* Timer for polling socket activity */
+        int timer;
```
mike-jumper: Missed a few spots for 8c24c77's changes ("GUACAMOLE-203: Change timer to timeout").

mike-jumper:
> Ping...anything else that needs attention, here?

Sorry for the delay.

At this point, I think things look generally fine. There's just a few spots where you didn't update the wording for timer vs. timeout, and the dangerous `unsigned int`.

src/protocols/rdp/rdp_settings.h
necouchman: How does it look, now?  It isn't the cleanest, because 0 is actually valid (to disable keepalive).  So, I'm setting negative values to 0 (since negative values don't make sense), and then just warning that 1 gets rounded up to 2 and allowing libssh2 to do that internally.

src/protocols/ssh/ssh.c
necouchman: Okay, I think I've converted everything, now.

src/common-ssh/ssh.c
```diff
@@ -532,16 +532,20 @@ guac_common_ssh_session* guac_common_ssh_create_session(guac_client* client,
...
-        libssh2_keepalive_config(common_session->session, 1, keepalive);
-
     /* Warn if keepalive below minimum value */
-    if (keepalive == 1) {
+    if (keepalive < 0) {
+        keepalive = 0;
+        guac_client_log(client, GUAC_LOG_WARNING, "negative keepalive intervals "
+            "are converted to 0, disabling keepalive.");
+    }
+    else if(keepalive == 1) {
```
mike-jumper: `if` is not a function.

necouchman: Fixed.

src/common-ssh/common-ssh/ssh.h
```diff
@@ -98,7 +98,7 @@ void guac_common_ssh_uninit();
  *     if the connection or authentication were not successful.
  */
 guac_common_ssh_session* guac_common_ssh_create_session(guac_client* client,
-        const char* hostname, const char* port, guac_common_ssh_user* user);
+        const char* hostname, const char* port, guac_common_ssh_user* user, const int keepalive);
```
mike-jumper: As of 711cdd6, the prototype for `guac_common_ssh_create_session()` no longer matches the function definition (`const int keepalive` vs. `int keepalive`).

necouchman: Fixed.

PR #95, opened 2017-06-13
GUACAMOLE-39: Do not abort VNC/RDP connection if SFTP fails

cfroese:
Re-implementation in the current code base of the changes in the previous pull request for GUACAMOLE-39.  For VNC/RDP connections, this PR now logs the failure rather than aborting the connection.  This allows the primary connection to succeed regardless of the secondary SFTP connection.

### DIFF
--- a/src/common-ssh/common-ssh/ssh.h
+++ b/src/common-ssh/common-ssh/ssh.h
@@ -76,10 +76,11 @@ void guac_common_ssh_uninit();
 /**
  * Connects to the SSH server running at the given hostname and port, and
  * authenticates as the given user. If an error occurs while connecting or
- * authenticating, the Guacamole client will automatically and fatally abort.
- * The user object provided must eventually be explicitly destroyed, but should
- * not be destroyed until this session is destroyed, assuming the session is
- * successfully created.
+ * authenticating, the Guacamole client will only automatically and fatally 
+ * abort if the abort_on_error parameter is non-zero. Errors will always be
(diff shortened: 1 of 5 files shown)


src/common-ssh/common-ssh/ssh.h
```diff
@@ -93,12 +94,17 @@ void guac_common_ssh_uninit();
  * @param user
  *     The user to authenticate as, once connected.
  *
+ * @param abort_on_error
+ *     Whether to automatically and fatally abort if an error occurs whil
```
jmuehlner: Nitpick: missing an "e" in while. :)

src/common-ssh/ssh.c
```diff
@@ -285,6 +285,38 @@ static void guac_common_ssh_kbd_callback(const char *name, int name_len,
 }
 
 /**
+ *  * A handler for SSH client errors which logs the error and can optionally
```
jmuehlner: Woah, what's happened to this comment? It's all diagonal.

cfroese: I wish I could explain that.  :)  It's fixed now.

cfroese:
In theory it was helpful for a feature in our product but said feature turned out not so well.  This change also exposes a problem similar to https://issues.apache.org/jira/browse/GUACAMOLE-257.  It may let you in without problems but do not try to use drag and drop, it will kick you out and there's a segfault in the logs.  At this point we do not have the resources to make this work as well as it should. So we're abandoning this change.

PR #97, opened 2017-06-29
GUACAMOLE-303: Allow root directory of SFTP filesystem to be configured.

mike-jumper:
This change adds a new "sftp-root-directory" parameter to VNC, RDP, and SSH which defines the root directory exposed via Guacamole's file browser. Some care is taken to ensure that requests outside this root directory cannot be made via the file browse, in that both the root directory and the requested filename are normalized prior to being concatenated.

### DIFF
--- a/src/common-ssh/common-ssh/sftp.h
+++ b/src/common-ssh/common-ssh/sftp.h
@@ -34,6 +34,11 @@
 #define GUAC_COMMON_SSH_SFTP_MAX_PATH 2048
 
 /**
+ * Maximum number of path components per path.
+ */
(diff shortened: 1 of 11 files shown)


necouchman:
Looks good to me.  @jmuehlner anything you see before it gets merged to master?

jmuehlner:
Looks reasonable to me!

swplotner:
Testing this with guacamole-0.9.14 the feature sftp-root-directory does not appear to work with VNC. I have not tested the other protocol paths.

necouchman:
@swplotner If you're having issues with this please open a thread on the mailing list - this is a closed/merged PR.

PR #98, opened 2017-07-02
GUACAMOLE-337: Move libguacd functionality into public libguac API.

mike-jumper:
This change removes the libguacd utility library, moving its functionality instead into the public libguac library. With these changes, libguac now provides an SSL/TLS implementation of `guac_socket` (if guacamole-server was built with SSL support) and a new `guac_user_handle_connection()` function which deals with the entire Guacamole protocol handshake and subsequent I/O in a blocking manner.

The new SSL/TLS `guac_socket` is defined within the `guacamole/socket-ssl.h` header file (rather than within the main `guacamole/socket.h` header), which will only be installed if SSL support has been included in the guacamole-server build.

guacd has been refactored to take these changes into account.


necouchman:
I'm good with this, but this is complicated enough that it probably warrants review by someone else.  @jmuehlner??

I'm happy to do the merge work after a review.

src/libguac/guacamole/user.h
```diff
@@ -493,6 +493,30 @@ guac_user* guac_user_alloc();
 void guac_user_free(guac_user* user);
 
 /**
+ * Handles the portion of a user's Guacamole protocol handshake following the
+ * "select" instruction all subsequent I/O, automatically populating the
```
jmuehlner: Should there be some more punctuation here or something? This first sentence is really hard to parse.

mike-jumper: Nooooo ... my beautiful beautiful sentence ...

I will ... uh ... try to make this make sense.

jmuehlner:
Ok, I think this all looks good now.

PR #100, opened 2017-07-06
GUACAMOLE-325: Restore Windows-compatibility at libguac level.

mike-jumper:
This change brings Windows compatibility back to the libguac portion of the guacamole-server build, allowing it to be built with the [MinGW](http://www.mingw.org/) toolchain, so long as libguac's normal dependencies are also available, as well as Windows ports of:

* libdl
* libpthread

This change also adds a Windows Socket API (WSA, also known as "winsock") implementation of `guac_socket`, allowing Windows applications to be implemented which leverage libguac to serve Guacamole protocol directly.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -89,9 +89,6 @@ AC_CHECK_DECL([uuid_make],,
 # cunit
 AC_CHECK_LIB([cunit], [CU_run_test], [CUNIT_LIBS=-lcunit])
 
-# WinSock
-AC_CHECK_LIB([wsock32], [main])
-
 AC_SUBST(DL_LIBS)
 AC_SUBST(MATH_LIBS)
 AC_SUBST(PNG_LIBS)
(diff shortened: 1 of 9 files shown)


necouchman:
Would love to be the one to review this, but definitely not confident enough in my C skills to evaluate this.  Again, happy to do the actual merge work...

necouchman:
@jmuehlner Any chance you could review this one?  I'll be happy to do the legwork to merge it, but not sure I'm familiar enough with the server code or confident enough in my C skills to catch what might pop up here...

jmuehlner:
@necouchman I'm also not super fresh on my C coding skills, but I had a look through and everything looks reasonable as far as I can tell.

necouchman:
Okay, from what I can tell it looks okay to me, so I'll go ahead and merge.  Thanks.

PR #101, opened 2017-07-11
GUACAMOLE-279: Migrate to mutable terminal color palette.

mike-jumper:


mike-jumper:
This change modifies the color handling of Guacamole's internal terminal emulator such that it uses a mutable palette, defined at the display level, rather than a global immutable palette. The global palette still exists, but is now used to provide the initial values for the mutable palette.

necouchman:
Looks fine to me, will merge to master, here, shortly.

PR #102, opened 2017-07-11
GUACAMOLE-296: Fix linking issue that causes audio issues with FreeRDP 1.1

necouchman:
This is a very minor tweak to the configure.ac file that adds the winpr-utils library to the list of libraries linked when the winpr/stream.h file is found.  This works correctly for FreeRDP 1.0 (winpr/stream.h is not present) and FreeRDP 1.1 (winpr/stream.h is present), but probably not FreeRDP 2.0, as winpr libraries got significantly collapsed there.  But, I think there may be other issues with FreeRDP 2.0 and the Guacamole server.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -506,6 +506,7 @@ fi
 
 have_freerdp=disabled
 RDP_LIBS=
+WINPR_LIBS=
 AC_ARG_WITH([rdp],
             [AS_HELP_STRING([--with-rdp],
                             [support RDP @<:@default=check@:>@])],
@@ -695,7 +696,7 @@ fi
 # Check for stream support via WinPR
 if test "x${have_freerdp}" = "xyes"
 then
-    AC_CHECK_HEADER(winpr/stream.h,[RDP_LIBS="$RDP_LIBS -lwinpr-utils"],
+    AC_CHECK_HEADER(winpr/stream.h,[WINPR_LIBS="$WINPR_LIBS -lwinpr-utils"],
                     [have_winpr=no,
                      AC_CHECK_DECL([stream_write_uint8],,
                                   [AC_MSG_WARN([
@@ -1015,6 +1016,7 @@ AM_CONDITIONAL([ENABLE_WINPR], [test "x${have_winpr}"   = "xyes"])
 AM_CONDITIONAL([ENABLE_RDP],   [test "x${have_freerdp}" = "xyes"])
 
 AC_SUBST(RDP_LIBS)
+AC_SUBST(WINPR_LIBS)
 
 #
 # libssh2
--- a/configure.ac
+++ b/configure.ac
@@ -695,7 +695,7 @@ fi
 # Check for stream support via WinPR
 if test "x${have_freerdp}" = "xyes"
 then
-    AC_CHECK_HEADER(winpr/stream.h,,
+    AC_CHECK_HEADER(winpr/stream.h,[RDP_LIBS="$RDP_LIBS -lwinpr-utils"],
                     [have_winpr=no,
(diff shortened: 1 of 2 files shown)


configure.ac
```diff
@@ -662,7 +662,7 @@ fi
 # Check for stream support via WinPR
 if test "x${have_freerdp}" = "xyes"
 then
-    AC_CHECK_HEADER(winpr/stream.h,,
+    AC_CHECK_HEADER(winpr/stream.h,[RDP_LIBS="$RDP_LIBS -lwinpr-utils"],
```
mike-jumper: This is unfortunately probably wrong.

This section is specific to one aspect of that WinPR library, so adding the library here is odd. It would make more sense to add the WinPR library to `RDP_LIBS` at the point where the final yes/no decision is made regarding whether WinPR-related parts of the build are enabled, a little bit later:

```
if test "x${have_freerdp}" = "xyes" -a "x${have_winpr}" = "xyes"
then
    AC_DEFINE([ENABLE_WINPR],,
              [Whether library support for WinPR types was found])
fi
```

I have not looked into the original issue further, but I'm also unsure whether "libwinpr-utils" is the correct library to add.

mike-jumper: It would make even *more* sense to add a `WINPR_LIBS` variable, and modify `Makefile.am` accordingly. This is the pattern used elsewhere for libraries which may or may not be present. If they are not present/needed, then the variable just stays empty.

necouchman: So, the rationale for adding it with the streams.h section is that this library (winpr-utils) is where the Stream_New and Stream_Free functions actually exist.  If you look at the original ticket, you'll see that when you do not link against winpr-utils, you get two undefined symbols (Stream_New and Stream_Free, in addition to the odd third undefined one) - when you link against winpr-utils, these two symbols are defined.

I'll try to take a look at reworking it in with the final determination of WinPR presence later on, and split out WINPR_LIBS from RDP_LIBS and find the right places in the src/protocols/rdp/Makefile.am to put that in.

necouchman:
Ping myself on this one...haven't made any progress, here....

necouchman:
@mike-jumper: Okay, used a separate WINPR_LIBS variable and broke this functionality out, here.  I kept the definition of that variable at the check for streams.h, since that's the header file that symbol is defined in.  Let me know if anything else needs to be tweaked.

mike-jumper:
@necouchman - looks good, but to confirm:

1. Has it been verified that the originally-reported problem is reproducible without these changes, but cannot be reproduced with these changes applied?
2. What about the other FreeRDP plugins: guacsnd, guacdr, and guacsvc?
3. Does libguac-client-rdp need to be linked with `WINPR_LIBS` as well?

necouchman:
1. Yes, I can confirm it resolves the originally-reported problem, and a couple of other folks tested it and reported success after my original change, so that issue is definitely resolved.
2. I will check on those, but I'm pretty sure the others did not, for whatever reason, need to link directly to the winpr-utils library.  I don't know if they look to the guacai library, but I'll double-check.
3. Same with this one, I will double-check, but am reasonably certain that it was only the guacai library that required this link.

necouchman:
1. Confirmed.
2. The other FreeRDP plugins - snd, dr, and svc - are all linked against the winpr-utils library, already.  I'm not entirely sure how this happens - not sure if it picks it up from guacai, or if a different library it is linking against also pulls it in.
3. No, libguac-client-rdp does not need this link - it appears to be specific to the plugins.

mike-jumper:
Sounds OK, then. We should figure out why exactly that library is problematic only for one specific plugin, but for now I don't see any reason not to merge this.

PR #105, opened 2017-07-13
GUACAMOLE-344: switching to upstream libvncserver because it properly enables VeNCrypt

Metuchen:


Dockerfile
```diff
@@ -77,6 +76,12 @@ COPY . "$BUILD_DIR"
 
 # Build guacamole-server from local source
 RUN yum -y install $BUILD_DEPENDENCIES         && \
+    curl -L https://codeload.github.com/LibVNC/libvncserver/tar.gz/LibVNCServer-0.9.11 | tar -zx && \
```
mike-jumper: Why `codeload.github.com`? Is this URL documented anywhere?

Dockerfile
```diff
@@ -77,6 +76,12 @@ COPY . "$BUILD_DIR"
...
+    cd libvncserver-LibVNCServer-0.9.11 && \
+    ./autogen.sh && \
+    make install && \
+    cd .. && \
+    rm -fr libvncserver-LibVNCServer-0.9.11 && \
```
mike-jumper: Rather than embedding the build steps within the `Dockerfile`, they should instead be within a documented shell script. Besides being cleaner and easier to maintain, this removes the need to `cd ..` to restore the working directory.

See: https://github.com/apache/incubator-guacamole-server/blob/6236eb8f98ff96c495ae94cb4fe3279c202db465/src/guacd-docker/bin/build-guacd.sh

Dockerfile
Metuchen: I got that link by browsing github... but I think this URL might be more appropriate: https://github.com/LibVNC/libvncserver/archive/LibVNCServer-0.9.11.tar.gz

Metuchen:
thanks for the tip...I've integrated your recommendations.

src/guacd-docker/bin/build-libvncserver.sh
```diff
@@ -0,0 +1,36 @@
...
+##     The working directory to download, extract and build the libvncserver distribution
+##
+
+BUILD_DIR="$1"
+
+cd "$BUILD_DIR"
+curl -L https://github.com/LibVNC/libvncserver/archive/LibVNCServer-0.9.11.tar.gz | tar -zx
+cd libvncserver-LibVNCServer-0.9.11
+./autogen.sh
+make install
```
mike-jumper: Missing `ldconfig` after `make install`.

src/guacd-docker/bin/build-libvncserver.sh
```diff
@@ -0,0 +1,36 @@
...
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+#
+
+##
+## @fn build-libvncserver.sh
+##
+## Builds the source of libvncserver
```
mike-jumper: A bit of a nitpick, but to match the style of the rest of our source, there needs to be a period at the end of the sentence here.

src/guacd-docker/bin/build-libvncserver.sh
```diff
@@ -0,0 +1,36 @@
...
+##
+## @param BUILD_DIR
+##     The working directory to download, extract and build the libvncserver distribution
```
mike-jumper: Same here - missing period. Please also wrap your lines at 80 columns where possible (and where doing so doesn't hurt readability). It's sometimes a tough call whether to wrap a line of code, but there are almost no cases where doing so would be bad for the documentation.

See the style of: https://github.com/apache/incubator-guacamole-server/blob/6236eb8f98ff96c495ae94cb4fe3279c202db465/src/guacd-docker/bin/build-guacd.sh#L27-L29

necouchman:
Ping @Metuchen - can you update the PR as requested?

Metuchen:
Sorry for the delay.  I've abandoned this patch because enabling VeNCrypt causes gross instability in the VNC connection.

PR #106, opened 2017-08-13
GUACAMOLE-279: Implement console codes for palette modification.

mike-jumper:
This change implements the OSC used by xterm for modifying the 256-color palette, allowing the default palette to be overridden dynamically. The colors themselves are described using X11 color specs, as defined by `XParseColor()`. In addition to simple RGB specifications, there are also quite a few named colors - I've handled those by doing a binary search over a sorted array of color names, ignoring case and whitespace (see `named-colors.c`).


necouchman:
Looks good to me - no obvious issues jump out to me.  Will merge to master here, shortly.

PR #109, opened 2017-09-24
GUACAMOLE-388: resolve null pointer dereference

chipitsine:


src/protocols/vnc/user.c
```diff
@@ -112,6 +112,8 @@ int guac_vnc_user_leave_handler(guac_user* user) {
 
     guac_vnc_client* vnc_client = (guac_vnc_client*) user->client->data;
 
+    if(!vnc_client || !user || !vnc_client->display || !vnc_client->display->cursor) return 1;
```
necouchman: To quote @mike-jumper: if is not a function.  (Should be a space between the "if" and the opening parenthesis.)

:-)

Also, to be consistent with style used throughout the rest of the code, the return one should be on its own line and indented.

mike-jumper:
@chipitsine, can you clarify exactly what dereference you're trying to fix? I see:

    if (!vnc_client || !user || !vnc_client->display || !vnc_client->display->cursor)
        return 1;

which is very broad. If one or more of those pointers can actually be `NULL` (and is known to be initialized to `NULL` prior to dereference here), then checking those makes sense, however at least `user` cannot be `NULL` by definition (a user can't leave the connection without first existing), and `vnc_client` looks to be in a similar situation (`guac_vnc_user_leave_handler` is only made callable after `vnc_client` has been allocated and assigned to `guac_client->data`).

Checking the [GUACAMOLE-388](https://issues.apache.org/jira/browse/GUACAMOLE-388) issue in JIRA , I don't see any information provided, either, other than a verbatim copy of gdb output.

If there is a problem being solved here, we need to see an analysis of what that problem is, together with a solution which solves that problem specifically (rather than blanketly checking all pointers).

src/protocols/vnc/user.c
```diff
@@ -112,6 +112,9 @@ int guac_vnc_user_leave_handler(guac_user* user) {
 
     guac_vnc_client* vnc_client = (guac_vnc_client*) user->client->data;
 
+    if (!vnc_client || !user || !vnc_client->display || !vnc_client->display->cursor)
```
mike-jumper: Summarizing [my comment from earlier](https://github.com/apache/incubator-guacamole-server/pull/109#issuecomment-331722195), this check seems overly broad, and it's unlikely that all of these pointers need to be verified. We shouldn't be paranoidly checking for `NULL` every time we're about to use a pointer.

If we're going to add checks, the reason for each check should be known, and the cause of [GUACAMOLE-388](https://issues.apache.org/jira/browse/GUACAMOLE-388) needs to be determined.

chipitsine: hello, I'm doing this in my spare time.

I do not mind if you want to determine the root cause of GUACAMOLE-388

what I wanted to do is to add check for null, it still makes sense, i.e. "check for null before dereferencing the pointer". it will help if something will change later on the caller side.


as for "overly broad", I'll make it more narrow soon

mike-jumper: > I do not mind if you want to determine the root cause of GUACAMOLE-388

Excellent, as that's really the first step before attempting any sort of fix. It's not a good idea to simply cast a wide net without knowing the actual cause, and then call it fixed when the problem cannot be reproduced.

> what I wanted to do is to add check for null, it still makes sense, i.e. "check for null before dereferencing the pointer". it will help if something will change later on the caller side.

No, blanket/paranoid checks are not good practice. We shouldn't be checking for `NULL` simply as a matter of course. Things should be written purposefully, and we should know what we're passing in. Coding defensively like that is self-defeating.

chipitsine:
yes, I meant defensive coding. as long as there're null dereferences, such approach makes sense.

I'll spent some time next week in reproducing the above issue

necouchman:
@mike-jumper @chipitsine Looks like this issue might have been resolve with the recently-merged changes from #120??

mike-jumper:
@necouchman Correct.

@chipitsine, when you can, please close this PR, as the necessary changes have been made via #120. I'll go ahead and mark GUACAMOLE-388 as a duplicate of GUACAMOLE-424.

mike-jumper:
@chipitsine, can you close this PR? The issue was addressed through #120.

chipitsine:
closing in favour of https://github.com/apache/guacamole-server/pull/120/files

PR #110, opened 2017-09-24
GUACAMOLE-389: add (basic) travis-ci integration

chipitsine:


necouchman:
I'm good with this, and would also like to see Travis CI support added to the client code.  @mike-jumper @jmuehlner, any objections?

chipitsine:
We encountered a bug in server code, so I started to look at server first

I'll have a look at the rest later

mike-jumper:
Before jumping headlong into using Travis CI, I think this warrants some discussion on the dev list. The ASF already has a Jenkins instance, which would be my personal preference, but others may have opinions here.

I've tried using Travis CI before (for Guacamole and other projects), but have each time regretted and reverted because:

* Maintaining the CI build together with the code is cumbersome and feels like it violates separation of concerns, particularly when credentials and builds of dependencies get involved.
* Though convenient, Travis supports builds only on a strict subset of platforms. Prior to the migration to the Incubator, Guacamole had CI builds for several Linux distributions, FreeBSD, as well as built-from-source versions of FreeRDP. Those builds continue to run on third-party hardware, but really should be migrated over to the ASF.

I'd definitely prefer using a CI system which does not directly couple the CI to the code being integrated.

necouchman:
> Before jumping headlong into using Travis CI, I think this warrants some discussion on the dev list. 

I tried...
http://mail-archives.apache.org/mod_mbox/incubator-guacamole-dev/201708.mbox/browser

:-D

But if there are other systems available, like Jenkins in ASF, then let's absolutely begin to leverage those.  What needs to be done??

mike-jumper:
>>
>> Before jumping headlong into using Travis CI, I think this warrants some discussion on the dev list.
>>
>
> I tried...
> http://mail-archives.apache.org/mod_mbox/incubator-guacamole-dev/201708.mbox/browser
>
> :-D
>

Beware the link you posted isn't to a particular thread, but rather the dev list archives overall.

> But if there are other systems available, like Jenkins in ASF, then let's absolutely begin to leverage those. What needs to be done??

I believe we simply need to get access to the ASF Jenkins instance and begin creating jobs. There's some documentation:

https://cwiki.apache.org/confluence/display/INFRA/Jenkins

Regarding projects in the Incubator, the above states:

> Those of you in the Incubator, there are many PMC Chairs belonging to the Incubator PMC, including but not limited to the Incubator PMC Chair. So if a mentor can't do it, ask around. If no answer within a reasonable time frame, then by all means open an INFRA ticket, mentioning your earlier attempts to get a PMC Chair to do it.

necouchman:
> Beware the link you posted isn't to a particular thread, but rather the dev list archives overall.

Oops...

http://mail-archives.apache.org/mod_mbox/incubator-guacamole-dev/201708.mbox/%3C136680295.3171979.1503836394003@mail.yahoo.com%3E

> I believe we simply need to get access to the ASF Jenkins instance and begin creating jobs. There's some documentation:

Okay.  I'll bump the thread on the Dev discussion, again, and maybe we can get some more feedback...

mike-jumper:
Now that we've graduated, I should actually have the ability to create Jenkins jobs myself (being PMC chair for guac), as well as to grant access to fellow committers.

I'll give it a shot.

mike-jumper:
OK. I managed to get things to work:

https://builds.apache.org/view/E-G/view/Guacamole/

@chipitsine - please close this PR when you can. We are now using Jenkins for CI.

chipitsine:
closing in favour of jenkins

PR #111, opened 2017-09-24
GUACAMOLE-390: add github PR template

chipitsine:


mike-jumper:
Probably should have commented here in the first place, but copying over from JIRA for visibility's sake:

> As we have a `CONTRIBUTING` file in all repositories, GitHub does already show a message prior to opening a PR:
>
> github-pr-message 
>
> How will this improve things further? As it's clear that users often ignore the existing message when opening pull requests for the first time, what is less easy to ignore in this case?
>

chipitsine:
ok, probably "contributing" is enough

PR #113, opened 2017-09-26
GUACAMOLE-396: Fixing ssh socket for IPv6 address

sanhex:
GUACAMOLE-396: guacd cannot connect to ssh servers with IPv6 addresses

Root Cause:
In the ssh library of guacd, the TCP socket for connecting to ssh server is created with AF_INET. So it does not support IPv6 address.

Solution:
When guacd creates the socket for ssh in guac_common_ssh_create_session(), stop using hard coded AF_INET for socket() call, use the address family which is returned from getaddrinfo().

Test:
- Connected successfully via ssh connections with IPv4 and IPv6 hosts.
- No connection error in guacd logs.
- Simulated a connection failure with specifying a ssh server which does not exist. guacd worked well in this case.

### DIFF
--- a/src/common-ssh/ssh.c
+++ b/src/common-ssh/ssh.c
@@ -431,20 +431,11 @@ guac_common_ssh_session* guac_common_ssh_create_session(guac_client* client,
         .ai_protocol = IPPROTO_TCP
     };
 
-    /* Get socket */
-    fd = socket(AF_INET, SOCK_STREAM, 0);
-    if (fd < 0) {
-        guac_client_abort(client, GUAC_PROTOCOL_STATUS_SERVER_ERROR,
-                "Unable to create socket: %s", strerror(errno));
-        return NULL;
-    }
-
     /* Get addresses connection */
     if ((retval = getaddrinfo(hostname, port, &hints, &addresses))) {
         guac_client_abort(client, GUAC_PROTOCOL_STATUS_SERVER_ERROR,
                 "Error parsing given address or port: %s",
                 gai_strerror(retval));
-        close(fd);
         return NULL;
     }
 
@@ -461,6 +452,14 @@ guac_common_ssh_session* guac_common_ssh_create_session(guac_client* client,
             guac_client_log(client, GUAC_LOG_DEBUG,
                     "Unable to resolve host: %s", gai_strerror(retval));
 
+        /* Get socket */
+        fd = socket(current_address->ai_family, SOCK_STREAM, 0);
+        if (fd < 0) {
+            guac_client_abort(client, GUAC_PROTOCOL_STATUS_SERVER_ERROR,
+                    "Unable to create socket: %s", strerror(errno));
+            return NULL;
+        }
+
         /* Connect */
         if (connect(fd, current_address->ai_addr,
                         current_address->ai_addrlen) == 0) {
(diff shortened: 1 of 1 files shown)


src/common-ssh/ssh.c
```diff
@@ -418,7 +418,7 @@ guac_common_ssh_session* guac_common_ssh_create_session(guac_client* client,
 
     int retval;
 
-    int fd;
+    int fd = -1;
```
mike-jumper: Please do not initialize variables unnecessarily.

src/common-ssh/ssh.c
```diff
@@ -479,7 +478,7 @@ guac_common_ssh_session* guac_common_ssh_create_session(guac_client* client,
             guac_client_log(client, GUAC_LOG_DEBUG, "Unable to connect to "
                     "host %s, port %s: %s",
                     connected_address, connected_port, strerror(errno));
-
+        close(fd);
```
mike-jumper: Previously, there was a blank line after this single-line `else` block, visually separating that block from the code that followed. For the sake of readability, that blank line should remain.

I would suggest either restoring the blank line, adding this `close()` call to the `else` block itself, or getting rid of the `else` entirely and rely on the fact that the `if` will `break` out of the loop (and thus the `else` is superfluous anyway).

mike-jumper:
Please also be sure to tag your commit messages with the JIRA issue (see existing git history for examples of this). From the [contribution guidelines](https://github.com/apache/incubator-guacamole-server/blob/master/CONTRIBUTING):

>
> ... Commits should be made in logical units and must reference the JIRA issue number:
>
>     $ git commit -m "GUACAMOLE-123: High-level message describing the changes."
>

sanhex:
Thanks for the quick feedback, Mike. These comments make sense. I updated
the code and committed the patch again.

mike-jumper:
Thanks, @sanhex - looks much better. The most recent commit looks great, but the previous commit is still present:

ssh-ipv6-commits

Can you rebase things into one commit, rather than repeating the same commit message across two commits?

mike-jumper:
I recommend looking into git's interactive rebase feature.

sanhex:
Yeah. I'm new to git. Learning how to make a rebase.

sanhex:
Hi Mike,

I made 2 commits with the web editor on GitHub. After that, I tried to
revert them with GitHub Desktop. So now I can see 4 commits.

I'm following the below official instructions to rebase my commits on Linux
server. When I did "git rebase -i HEAD~4" My commits cannot be shown in the
interactive editor. I also tried "git rebase -i $hash". It looks my branch
is exactly the same with the official master branch. I cannot get my
commits in command line. Could you advise? Or maybe I can close this pull
request and create a new one with my latest code.

https://github.com/edx/edx-platform/wiki/How-to-Rebase-a-Pull-Request

mike-jumper:
Yeah, I would definitely avoid the web editor. Who knows what that would do.

On your local repository within Linux, what is the output of:

    git status

?

sanhex:
Thanks for looking into my problem. Will never use the web editor for the
next time..

jhe@ubuntu:~/github/incubator-guacamole-server$ git status
On branch patch-1
nothing to commit, working directory clean

jhe@ubuntu:~/github/incubator-guacamole-server$ git log -2
commit 4e809609334a30090645dc4318d3617f6b12d6c0
Merge: c5f6743 d2102e5
Author: Nick Couchman <vnick@apache.org>
Date:   Sun Sep 24 08:42:46 2017 -0400

    GUACAMOLE-391: Merge resolve low-impact memory leak.

commit d2102e57059a09d84d1456873d75b3dd66342ab4
Author: Ilya Shipitsin <chipitsine@gmail.com>
Date:   Sun Sep 24 13:42:59 2017 +0500

    GUACAMOLE-391: resolve low impact memory leak

sanhex:
Hi Mike,

I fixed my problem. Now there is only one commit for review. Thanks a lot
for the help.

mike-jumper:
Glad it all worked out. ;)

PR #114, opened 2017-09-27
GUACAMOLE-398: Fix memory leak identified by Coverity

necouchman:
Quick fix to correct a memory leak identified by the latest Coverity scan in the common ssh code.

### DIFF
--- a/src/common-ssh/ssh.c
+++ b/src/common-ssh/ssh.c
@@ -457,7 +457,7 @@ guac_common_ssh_session* guac_common_ssh_create_session(guac_client* client,
         if (fd < 0) {
             guac_client_abort(client, GUAC_PROTOCOL_STATUS_SERVER_ERROR,
                     "Unable to create socket: %s", strerror(errno));
-            free(addresses);
+            freeaddrinfo(addresses);
             return NULL;
         }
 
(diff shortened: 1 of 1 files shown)


src/common-ssh/ssh.c
```diff
@@ -457,6 +457,7 @@ guac_common_ssh_session* guac_common_ssh_create_session(guac_client* client,
         if (fd < 0) {
             guac_client_abort(client, GUAC_PROTOCOL_STATUS_SERVER_ERROR,
                     "Unable to create socket: %s", strerror(errno));
+            free(addresses);
```
mike-jumper: The addresses returned by a call to `getaddrinfo()` actually have to be freed by a call to `freeaddrinfo()`:

https://github.com/apache/incubator-guacamole-server/blob/afb554a0149391693615226060f567a5f2ef7d49/src/common-ssh/ssh.c#L487

Since the result of calling `getaddrinfo()` is a dynamically-allocated linked list, invoking `free()` on the returned list will not necessarily free the entire list, but likely only a single element of that list.

necouchman: ...and if I had looked a few lines down in the code I would have found that.  Should be fixed, now.

PR #115, opened 2017-09-28
GUACAMOLE-400: Fix guacd crash when ssh key fails

sanhex:
Root Cause:
In the ssh library of guacd, function ssh_client_thread(), when guac_ssh_get_user() fails to load private key for ssh authentication, it will return NULL. In this case, the subsequent call to guac_common_ssh_create_session() with parameter 'user=0x0' will cause guacd crash in function guac_common_ssh_authenticate() by accessing 'user->username'.

Solution:
In guac_common_ssh_create_session(), validate parameter 'user'. If it is NULL, abort the ssh session.
Reviewed the logic for other parameters of the function, they look okay and no need to be validated at this point.

Test:
- configured a ssh app with an encrypted private key and a wrong passphrase.
- ran the ssh app from web portal and observed guacd crash.
- applied the fix and reran the ssh app. Observed no crash.

### DIFF
--- a/src/protocols/ssh/ssh.c
+++ b/src/protocols/ssh/ssh.c
@@ -63,7 +63,8 @@
  *     terminal to use when prompting the user.
  *
  * @return
- *     A new user object containing the user's username and other credentials.
+ *     A new user object containing the user's username and other credentials,
+ *     or NULL if fails to import key.
  */
 static guac_common_ssh_user* guac_ssh_get_user(guac_client* client) {
 
@@ -215,6 +216,10 @@ void* ssh_client_thread(void* data) {
 
     /* Get user and credentials */
     ssh_client->user = guac_ssh_get_user(client);
+    if (ssh_client->user == NULL) {
+        /* Already aborted within guac_ssh_get_user() */
+        return NULL;
+    }
 
     /* Open SSH session */
     ssh_client->session = guac_common_ssh_create_session(client,


mike-jumper:
> In guac_common_ssh_create_session(), validate parameter 'user'. If it is NULL, abort the ssh session.

This sounds like a bit of a bandaid solution, particularly since it doesn't make sense to pass in a `NULL` user in the first place.

Why not verify the result of the attempt to create the user in the first place?

sanhex:
It looks this function is considered as a common API and has been called in
several places for different protocols. If we verify the user pointer in
the upstream, may need to add duplicate code in all those places. If
validate 'user' in guac_common_ssh_create_session(), it appears to me the
code could be simpler and safe for future calls.

./protocols/vnc/vnc.c:263:
guac_common_ssh_create_session(client, settings->sftp_hostname,
./protocols/vnc/vnc.c:268:            /* Already aborted within
guac_common_ssh_create_session() */
./protocols/rdp/rdp.c:979:
guac_common_ssh_create_session(client, settings->sftp_hostname,
./protocols/rdp/rdp.c:984:            /* Already aborted within
guac_common_ssh_create_session() */
./protocols/ssh/ssh.c:220:    ssh_client->session =
guac_common_ssh_create_session(client,
./protocols/ssh/ssh.c:223:        /* Already aborted within
guac_common_ssh_create_session() */
./protocols/ssh/ssh.c:260:
guac_common_ssh_create_session(client, settings->hostname,
./protocols/ssh/ssh.c:263:            /* Already aborted within
guac_common_ssh_create_session() */
./common-ssh/common-ssh/ssh.h:100:guac_common_ssh_session*
guac_common_ssh_create_session(guac_client* client,
./common-ssh/ssh.c:416:guac_common_ssh_session*
guac_common_ssh_create_session(guac_client* client,

mike-jumper:
> It looks this function is considered as a common API and has been called in several places for different protocols.

It is.

> If we verify the user pointer in the upstream, may need to add duplicate code in all those places.

Maybe so, but a change avoiding such duplication is only valid if the semantics surrounding that check are guaranteed to be the same across all callers. If the semantics are not guaranteed to be the same, then there really isn't any duplication being avoided, and [separation of concerns](https://en.wikipedia.org/wiki/Separation_of_concerns) is violated.

The issue is not whether code calling this function may need to check for `NULL` (or that this check may occur in multiple places), but whether `NULL` is a value with well-defined meaning at this level - meaning which the function can be expected to correctly interpret on its own.

Take a look at the context of how this function is called in the case where you're seeing the failure:

https://github.com/apache/incubator-guacamole-server/blob/1d0e63b251c548a7b8ddea95e7502ab81283b427/src/protocols/ssh/ssh.c#L216-L221

The `NULL` value is coming from `guac_ssh_get_user()`, a function specific to the SSH plugin, and the desired behavior in this case (cleanly logging an error and aborting the connection) is only desirable because of the context of the call in the case of the failure in question (failure to read/decrypt the private key while establishing a connection which will be useless without that key).

I would say the actual issue here is twofold:

1. The `guac_ssh_get_user()` function does not properly document that `NULL` will be returned if creating the user is unsuccessful, that the connection will automatically be aborted, nor the possible failures that may occur:
    https://github.com/apache/incubator-guacamole-server/blob/1d0e63b251c548a7b8ddea95e7502ab81283b427/src/protocols/ssh/ssh.c#L57-L68
2. The possible `NULL` return is not handled by the caller which knows what `NULL` means in this case (the SSH plugin).

Another good indicator that separation of concerns may be hurt here is the new error message itself:

https://github.com/sanhex/incubator-guacamole-server/blob/6c9c87a04ab9e4e8de454a816cccdd0ebd35af3e/src/common-ssh/ssh.c#L434-L439

The message "Unable to find user credentials." does not provide enough information to the administrator for the problem to be corrected, and really can't be modified to do so, because the context of the error doesn't exist at this level. Only the caller knows why the user is `NULL`, and thus only the caller can properly check, handle, and log the problem.

sanhex:
Thanks for the clarification with all those details. That is helpful for me
to understand.

mike-jumper:
@sanhex, before I merge this, can you correct the commit message to describe the change? It currently describes everything in great detail, but covers your previous approach (checking for `NULL` within `guac_common_ssh_create_session()`).

sanhex:
Ah, forgot to update the comment. Done.

PR #116, opened 2017-10-10
GUACAMOLE-411: Fixed using uninitialized values in guacd_send_fd

jukeks:
Fixed valgrind's complaints when sending fds.
```
==8081== Thread 2:
==8081== Syscall param sendmsg(msg.msg_control) points to uninitialised byte(s)
==8081==    at 0x505EA6D: ??? (syscall-template.S:84)
==8081==    by 0x403F5C: guacd_send_fd (move-fd.c:61)
==8081==    by 0x40390A: guacd_add_user (connection.c:196)
==8081==    by 0x40390A: guacd_route_connection (connection.c:311)
==8081==    by 0x40390A: guacd_connection_thread (connection.c:393)
==8081==    by 0x50556B9: start_thread (pthread_create.c:333)
==8081==    by 0x5A1F3DC: clone (clone.S:109)
==8081==  Address 0xa558d54 is on thread 2's stack
==8081==  in frame #1, created by guacd_send_fd (move-fd.c:34)
==8081==  Uninitialised value was created by a stack allocation
==8081==    at 0x403ED0: guacd_send_fd (move-fd.c:34)
==8081== 
```

### DIFF
--- a/src/guacd/move-fd.c
+++ b/src/guacd/move-fd.c
@@ -44,7 +44,7 @@ int guacd_send_fd(int sock, int fd) {
     message.msg_iovlen = 1;
 
     /* Assign ancillary data buffer */
-    char buffer[CMSG_SPACE(sizeof(fd))];
+    char buffer[CMSG_SPACE(sizeof(fd))] = {0};
     message.msg_control = buffer;
     message.msg_controllen = sizeof(buffer);
 


mike-jumper:
@jukeks, any idea why this warning is occurring? Just initializing everything to `0` to silence valgrind is only a solution if that value is actually correct.

Within `guac_send_fd()`, it looks like all needed parts of the message structure have been set to the required values. With that in mind, there are two possibilities:

1. Valgrind is reporting a false positive, and the uninitialized bytes are not actually used within the syscall.
2. Valgrind is correct, those bytes are used by the syscall, and we need to determine what those bytes should be set to (and why).

jukeks:
If you see `<socket.h>` macros `CMSG_SPACE ` and `CMSG_ALIGN` you notice, the there might be added padding which is not used for data.

```
#define CMSG_ALIGN(len) ( ((len)+sizeof(long)-1) & ~(sizeof(long)-1) )
#define CMSG_SPACE(len) (CMSG_ALIGN(sizeof(struct cmsghdr)) + CMSG_ALIGN(len))
```

```
#include <sys/socket.h>
#include <stdio.h>

int main(void) {
    size_t padding = CMSG_ALIGN(sizeof(int)) - sizeof(int);
    printf("Padding for int: %zu\n", padding);
}
```

For my 64 bit linux machine this prints:
```
Padding for int: 4
```

so in this case there's padding of 4 bytes not initialized.

PR #118, opened 2017-10-18
GUACAMOLE-384: fixing segfault during ssh disconnect

sanhex:
Root Cause:
See the core dump and Valgrind report posted on Jira. guacd was reading a ssh terminal which had been freed. When a ssh connection is terminated, guac_ssh_client_free_handler() will be called from guacd_exec_proc() -> guac_client_free() with pointer client->free_handler. In guac_ssh_client_free_handler(), when ssh_client->term is freed, ssh_client->client_thread may still be using the ssh_client->term. It causes the crash reported in this bug.

The stack trace exposing the problem can be found by running guacd under Valgrind with a ssh test script. The test script repeats doing ssh login and logout for 5000 times.

Solution:
In guac_ssh_client_free_handler(), before calling guac_terminal_free(ssh_client->term), close the stdin pipe of the terminal to stop reading the pipe with guac_terminal_read_stdin() in ssh_input_thread(). So that ssh_input_thread() can be terminated in this case. Call pthread_join() to wait for ssh_client_thread() terminating before freeing the terminal.

After closing the pipe, set the fds to -1 and check them in guac_terminal_free() to avoid closing for twice.

Checking the client running state in ssh_input_thread() and ssh_client_thread() to make sure they can be terminated when the client is stopped in guacd_exec_proc() by another thread.

Test:
- Confirmed ssh connection works normally.
- Observed the child process of guacd exits when ssh connection is terminated.
- Reran the ssh test script. Observed no crash.

### DIFF
--- a/src/protocols/ssh/client.c
+++ b/src/protocols/ssh/client.c
@@ -70,8 +70,12 @@ int guac_ssh_client_free_handler(guac_client* client) {
 
     /* Free terminal (which may still be using term_channel) */
     if (ssh_client->term != NULL) {
-        guac_terminal_free(ssh_client->term);
+        /* Stop the terminal to unblock any pending reads/writes */
+        guac_terminal_stop(ssh_client->term);
+
+        /* Wait ssh_client_thread to finish before freeing the terminal */
         pthread_join(ssh_client->client_thread, NULL);
+        guac_terminal_free(ssh_client->term);
     }
 
     /* Free terminal channel now that the terminal is finished */
--- a/src/protocols/ssh/ssh.c
+++ b/src/protocols/ssh/ssh.c
@@ -162,8 +162,14 @@ void* ssh_input_thread(void* data) {
         pthread_mutex_lock(&(ssh_client->term_channel_lock));
         libssh2_channel_write(ssh_client->term_channel, buffer, bytes_read);
         pthread_mutex_unlock(&(ssh_client->term_channel_lock));
+
+        /* Make sure ssh_input_thread can be terminated anyway */
+        if (client->state == GUAC_CLIENT_STOPPING)
+            break;
     }
 
+    /* Stop the client so that ssh_client_thread can be terminated */
+    guac_client_stop(client);
     return NULL;
 
 }
@@ -336,6 +342,12 @@ void* ssh_client_thread(void* data) {
 
         /* Stop reading at EOF */
         if (libssh2_channel_eof(ssh_client->term_channel)) {
+            pthread_mutex_unlock(&(ssh_client->term_channel_lock));
+            break;
+        }
+
+        /* Client is stopping, break the loop */
+        if (client->state == GUAC_CLIENT_STOPPING) {
             pthread_mutex_unlock(&(ssh_client->term_channel_lock));
             break;
         }
--- a/src/terminal/terminal.c
+++ b/src/terminal/terminal.c
@@ -410,11 +410,23 @@ guac_terminal* guac_terminal_create(guac_client* client,
 
 }
 
+void guac_terminal_stop(guac_terminal* term) {
+
+    /* Close input pipe and set fds to invalid */
+    if (term->stdin_pipe_fd[1] != -1) {
+        close(term->stdin_pipe_fd[1]);
+        term->stdin_pipe_fd[1] = -1;
+    }
+    if (term->stdin_pipe_fd[0] != -1) {
+        close(term->stdin_pipe_fd[0]);
+        term->stdin_pipe_fd[0] = -1;
+    }
+}
+
 void guac_terminal_free(guac_terminal* term) {
 
     /* Close user input pipe */
-    close(term->stdin_pipe_fd[1]);
-    close(term->stdin_pipe_fd[0]);
+    guac_terminal_stop(term);
 
     /* Wait for render thread to finish */
     pthread_join(term->thread, NULL);
--- a/src/terminal/terminal/terminal.h
+++ b/src/terminal/terminal/terminal.h
@@ -492,6 +492,15 @@ int guac_terminal_render_frame(guac_terminal* terminal);
 int guac_terminal_read_stdin(guac_terminal* terminal, char* c, int size);
 
 /**
+ * Manually stop the terminal to forcibly unblock any pending reads/writes,
+ * e.g. forcing guac_terminal_read_stdin() to return and cease all terminal I/O.
+ *
+ * @param term
+ *     The terminal to stop.
+ */
+void guac_terminal_stop(guac_terminal* term);
+
+/**
  * Notifies the terminal that an event has occurred and the terminal should
  * flush itself when reasonable.
  *


src/terminal/terminal.c
```diff
@@ -413,8 +413,10 @@ guac_terminal* guac_terminal_create(guac_client* client,
 void guac_terminal_free(guac_terminal* term) {
 
     /* Close user input pipe */
-    close(term->stdin_pipe_fd[1]);
-    close(term->stdin_pipe_fd[0]);
+    if (term->stdin_pipe_fd[1] != -1)
```
necouchman: Instead of assigning -1 here, would it be better to do a fstat() call and check the return value? It seems a little bit odd to me to be manually assigning -1 to this variable above in client.c and then checking it here.

@mike-jumper Thoughts?

src/protocols/ssh/client.c
```diff
@@ -70,8 +70,14 @@ int guac_ssh_client_free_handler(guac_client* client) {
 
     /* Free terminal (which may still be using term_channel) */
     if (ssh_client->term != NULL) {
-        guac_terminal_free(ssh_client->term);
+        /* Close user input pipe to stop reading in ssh_input_thread */
+        close(ssh_client->term->stdin_pipe_fd[1]);
+        close(ssh_client->term->stdin_pipe_fd[0]);
+        ssh_client->term->stdin_pipe_fd[1] = -1;
+        ssh_client->term->stdin_pipe_fd[0] = -1;
+
```
necouchman: Why not just leave the call to guac_terminal_free here?  Doesn't that function do exactly the same thing that you're manually doing here?  It seems like this is a violation of abstraction (I made that term up) here to be manually modifying the data members of the terminal object rather than letting the methods in terminal.c handle that.

src/terminal/terminal.c
ceharris: `fstat()` is used get a `struct stat` for a file on a filesystem; it isn't of much use with pipe/socket descriptors.

Without weighing in on whether it makes sense to move the calls to `close`, I'd say the use of -1 to mark an unused/closed descriptor field is fairly idiomatic C.

necouchman: Fair enough.  I suggested `fstat()`, not necessarily because of any interest in the object returned, but because you could check the return value of the call for something like EBADF to see if the file descriptor is still valid/open or not.  But, if `-1` is generally accepted, then I have no problem with that.

src/protocols/ssh/client.c
sanhex: I agree this looks really ugly. But if I don't close the pipe here, there will be a dead lock in this case.

The thread here will be waiting for ssh_client_thread() by calling pthread_join(ssh_client->client_thread, NULL), ssh_client_thread() will be waiting for ssh_input_thread(). ssh_input_thread() will be blocked by guac_terminal_read_stdin() show below.
https://github.com/sanhex/incubator-guacamole-server/blob/c6cb1910e3580f9c21fc202f824fc55c95a43e49/src/protocols/ssh/ssh.c#L161

The child process of guacd will never exit after disconnecting the ssh connection. I met this problem.. So I manually closed the pipe to stop the ssh_input_thread() mentioned above.

src/terminal/terminal.c
sanhex: Thanks for the idea. If -1 is acceptable here, maybe I should also do the same in guac_terminal_free(). E.g.

void guac_terminal_free(guac_terminal* term) {

    /* Close user input pipe */
    if (term->stdin_pipe_fd[1] != -1) {
        close(term->stdin_pipe_fd[1]);
        term->stdin_pipe_fd[1] = -1;
    }
    if (term->stdin_pipe_fd[0] != -1) {
        close(term->stdin_pipe_fd[0]);
        term->stdin_pipe_fd[0] = -1;
    }

necouchman: Yeah, that would probably be fine.

src/protocols/ssh/client.c
necouchman: So, you're saying that, even with the changes to the `guac_terminal_free()` function, a call to that function, here, would result in the deadlock?

@ceharris @mike-jumper Thoughts on this one?

sanhex: Sorry for the confusion. The changes in this patch to guac_terminal_free() are only for a better practice of preventing unnecessary double closing. It does nothing with the dead lock. The dead lock will happen if I only switch the guac_terminal_free() and pthread_join(). ssh_input_thread() won't be terminated if I don't manually close the pipe here.

src/terminal/terminal.c
sanhex: Sure. I will take a further look. Good point.

src/protocols/ssh/client.c
necouchman: No need to be sorry - I'm just trying to wrap my head around it, and I'm much more familiar with the guacamole-client code than the server code.  Sometimes I need things explained in small words, repeatedly.

I'm curious what input Mike or Carl have on this - it still makes me a little twitchy to modify these member fds directly in the ssh.c file, but it could be that I'm just being unreasonable.

mike-jumper: > .. it still makes me a little twitchy to modify these member fds directly in the ssh.c file ...

This makes me twitchy, too. I'd much rather that the handling of file descriptor closure be encapsulated within the code that needs to know about it. I don't like the idea of requiring that all users of the terminal manually close the terminal's internal descriptors.

sanhex: @mike-jumper Will it make more sense if I add a new function e.g. guac_terminal_close_pipe() in terminal/terminal.c, call it here and also in guac_terminal_free() ?

mike-jumper: No, that would still expose internals, this time within the name of the function itself.

The new ordering with waiting on the thread prior to freeing the terminal is correct, but it seems to me that it's the client thread's dependence on the terminal that's the issue here. The input thread shouldn't be depending on closure of the terminal at all. The various client-specific threads should take into account the client's own stop signal, rather than relying on `guac_terminal` to expand scope and shoulder that burden.

sanhex: The problem here is the pipe is blocking. The call guac_terminal_read_stdin() in ssh_input_thread() will never return when there is no data in the pipe after ssh disconnecting. Another option is setting the pipe fds to non-blocking and tweaking the below callers to read and sleep.

./protocols/telnet/telnet.c:267:    while ((bytes_read = guac_terminal_read_stdin(telnet_client->term, buffer, sizeof(buffer))) > 0) {
./protocols/ssh/ssh.c:161:    while ((bytes_read = guac_terminal_read_stdin(ssh_client->term, buffer, sizeof(buffer))) > 0) {
./terminal/terminal.c:622:    while (guac_terminal_read_stdin(terminal, &in_byte, 1) == 1) {

mike-jumper: > The problem here is the pipe is blocking. The call guac_terminal_read_stdin() in ssh_input_thread() will never return when there is no data in the pipe after ssh disconnecting.

In that case, I'd suggest something like `guac_terminal_stop()` to provide the needed functionality (forcing reads from `guac_terminal` to fail) without requiring exposure of internals.

sanhex: My first feeling was sending an EOF (maybe defined as -1 on Linux) to the pipe forcing the read to fail. I'm not sure if it will work cos I never try this on a pipe. Any better idea?

mike-jumper: Well:

1. Keep in mind that guacamole-server is meant to be as POSIX-compatible as possible. Unless something cannot be done without platform-specific code (`move-fd.c` is the only current case of this in the codebase), then it needs to be done without platform-specific code.
2. You can't actually send EOF, as EOF isn't data, but a state. If you want reads to fail from one end of the pipe and return EOF, the way to do that is to close the file descriptor on the other end.

Closing the file descriptors to force reads to fail is fine. The problem is requiring the caller to know that this is what is happening. As long as the function naming and documentation do not require the caller to have such internal knowledge of `guac_terminal`, then things are OK.

It makes sense for the caller to need to know:

* Terminal reads/writes may block (I believe this is already documented).
* Some specific function (`guac_terminal_stop()`? `guac_terminal_close()`?) can be used to forcibly unblock any pending reads/writes and cease all terminal I/O.

It doesn't make sense for the caller to need to know:

* `guac_terminal` contains multiple file descriptors for pipes which are used internally by the read/write functions and should be closed to force those reads/writes to unblock and fail.

src/terminal/terminal/terminal.h
```diff
@@ -492,6 +492,14 @@ int guac_terminal_render_frame(guac_terminal* terminal);
 int guac_terminal_read_stdin(guac_terminal* terminal, char* c, int size);
 
 /**
+ * Manually stop the terminal to forcibly unblock any pending reads/writes,
+ * e.g. forcing guac_terminal_read_stdin() to return and cease all terminal I/O.
+ * This function has been called in guac_terminal_free(). It could be useful
```
mike-jumper: > This function has been called in guac_terminal_free().

I would suggest leaving this out as an internal implementation detail. If the intent is to document that `guac_terminal_free()` will also unblock pending reads/writes, then that information should be added to the documentation of `guac_terminal_free()`.

> It could be useful when need to stop the terminal before calling guac_terminal_free().

This doesn't add any information. Really, everything you had prior to this last sentence was perfectly fine. If the caller needs to invoke this function, the first couple of lines documenting that the function forcibly unblocks pending reads/writes are exactly what is needed.

src/protocols/ssh/ssh.c
```diff
@@ -340,6 +346,12 @@ void* ssh_client_thread(void* data) {
             break;
         }
 
+        /* Make sure ssh_client_thread can be terminated anyway */
```
mike-jumper: I'm not sure what you mean here by "make sure ssh_client_thread can be terminated anyway".

src/protocols/ssh/client.c
```diff
@@ -70,8 +70,12 @@ int guac_ssh_client_free_handler(guac_client* client) {
 
     /* Free terminal (which may still be using term_channel) */
     if (ssh_client->term != NULL) {
-        guac_terminal_free(ssh_client->term);
+        /* Stop the terminal to terminate ssh_input_thread */
```
mike-jumper: This is somewhat confusing, as the `pthread_join()` call which immediately follows deals with the client thread. Perhaps there is a way to word this such that it is more relevant?

sanhex:
Fixed the comments. Please help review. Thanks.

The changes to protocols/ssh/ssh.c is not a must for this bug. I considered
them as error handling. In case the read/write in ssh_input_thread() or
ssh_client_thread() fails, we should break the loops and stop the client to
prevent a dead loop.

mike-jumper:
> ... to prevent a dead loop.

Sounds like a must to me.

sanhex:
Yeah. E.g. if the call guac_terminal_read_stdin() in ssh_input_thread()
fails due to an IO error or anything else, we need to stop the client so
that the loop in ssh_client_thread() could break and stop. Vice versa, if
anything wrong break the loop in ssh_client_thread(), it will stop the
client. In this case, the loop in ssh_input_thread() also should be stopped
even if no error in itself.

sanhex:
If there is no I/O error, when the client is normally stopped
by guac_ssh_client_free_handler() in protocols/ssh/client.c, the
call guac_terminal_stop() I added will break the loop in
ssh_input_thread(). The existing call libssh2_channel_send_eof() will break
the loop in ssh_client_thread(). So those 2 threads can be stopped without
my changes to protocols/ssh/ssh.c. This is why I said these changes are not
a must for this bug. In my test, the guacd child process for handing ssh
connection can exit normally without my changes to protocols/ssh/ssh.c.

sanhex:
In the normal exit of ssh connections, when libssh2_channel_send_eof() is
called in guac_ssh_client_free_handler(), the call libssh2_channel_eof()
will break the loop in ssh_client_thread(). So my change is optional in the
normal case.

sanhex:
Updated some of my comments on GitHub to better describe how the ssh_input_thread() and ssh_client_thread() can be terminated when the client exits.

PR #120, opened 2017-10-24
GUACAMOLE-424: Fix null pointer dereference for vnc client display

itsankoff:
More information you can see in the issue description: https://issues.apache.org/jira/browse/GUACAMOLE-424

### DIFF
--- a/src/common/cursor.c
+++ b/src/common/cursor.c
@@ -37,10 +37,12 @@
 
 /**
  * Allocates a cursor as well as an image buffer where the cursor is rendered.
- * If the allocation fails then the function returns NULL.
  *
  * @param client
  *     The client owning the cursor.
+ *
+ * @return
+ *     The newly-allocated cursor or NULL if cursor cannot be allocated.
  */
 guac_common_cursor* guac_common_cursor_alloc(guac_client* client) {
 
--- a/src/common/cursor.c
+++ b/src/common/cursor.c
@@ -34,6 +34,14 @@
 #include <stdlib.h>
 #include <string.h>
 
+
+/**
+ * Allocates a cursor as well as an image buffer where the cursor is rendered.
+ * If the allocation fails then the function returns NULL.
+ *
+ * @param client
+ *     The client owning the cursor.
+ */
 guac_common_cursor* guac_common_cursor_alloc(guac_client* client) {
 
     guac_common_cursor* cursor = malloc(sizeof(guac_common_cursor));
--- a/src/common/display.c
+++ b/src/common/display.c
@@ -99,6 +99,19 @@ static void guac_common_display_free_layers(guac_common_display_layer* layers,
 
 }
 
+/**
+ * Allocates a display and a cursor which are used to represent the remote
+ * display and cursor. If the allocation fails then the function returns NULL.
+ *
+ * @param client
+ *     The client owning the cursor.
+ *
+ * @param width
+ *     The desired width of the display.
+ *
+ * @param height
+ *     The desired height of the display.
+ */
 guac_common_display* guac_common_display_alloc(guac_client* client,
         int width, int height) {
 
@@ -107,13 +120,17 @@ guac_common_display* guac_common_display_alloc(guac_client* client,
     if (display == NULL)
         return NULL;
 
+    /* Allocate shared cursor */
+    display->cursor = guac_common_cursor_alloc(client);
+    if (display->cursor == NULL) {
+        free(display);
+        return NULL;
+    }
+
     pthread_mutex_init(&display->_lock, NULL);
 
     /* Associate display with given client */
     display->client = client;
-
-    /* Allocate shared cursor */
-    display->cursor = guac_common_cursor_alloc(client);
 
     display->default_surface = guac_common_surface_alloc(client,
             client->socket, GUAC_DEFAULT_LAYER, width, height);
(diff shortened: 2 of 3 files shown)


src/protocols/vnc/user.c
```diff
@@ -112,8 +112,10 @@ int guac_vnc_user_leave_handler(guac_user* user) {
 
     guac_vnc_client* vnc_client = (guac_vnc_client*) user->client->data;
 
-    /* Update shared cursor state */
-    guac_common_cursor_remove_user(vnc_client->display->cursor, user);
+    if (vnc_client && vnc_client->display && vnc_client->display->cursor) {
```
mike-jumper: Without justification, this check is too broad. Have you verified which of these pointers is the one that is potentially invalid? For whichever pointer that is, it known that the value will indeed be `NULL` if not valid?

mike-jumper: From `guac_common_display_alloc()`, if the display is successfully allocated at all, it is not possible for the `cursor` to be `NULL`:

https://github.com/apache/incubator-guacamole-server/blob/95be88be19e04e07ac1dafb823993745bee7d146/src/common/display.c#L115-L116

Similarly, the leave handler for the user is assigned only after the `guac_vnc_client` has been allocated and assigned, thus this function will only be invoked after `vnc_client` here is known to be non-`NULL`:

https://github.com/apache/incubator-guacamole-server/blob/95be88be19e04e07ac1dafb823993745bee7d146/src/protocols/vnc/client.c#L55

Checking pointers for `NULL` is not something that should be done blanketly for all pointers just because those pointers are dereferenced. Such checks only make sense when:

1. `NULL` is one of the expected values of the pointer.
2. It is the responsibility of the function in question to validate the pointer.

In this case, both of the above are true for the `display` pointer. It is explicitly initialized to `NULL` due to the `guac_vnc_client` being allocated using `callloc()`, and given that the display is not guaranteed to be allocated as written, it's clear that the handlers assigned within `guac_client_init` need to take this into account before attempting to use the display.

itsankoff: Yeah I totally agree about blindly checking for NULL values. I also agree about the `vnc_client` check and I will remove it. But for the `cursor` there is actually a rare case when cursor allocation can fail and cursor within a valid display to be `NULL` (https://github.com/apache/incubator-guacamole-server/blob/95be88be19e04e07ac1dafb823993745bee7d146/src/common/cursor.c#L40-L41). That's why I think with these thoughts in mind the final check should be:
```
if (vnc_client->display && vnc_client->display->cursor) {
...
}
```

necouchman: I'd say go ahead and update with a commit to remove the vnc_client check but keep the other two.  Mike can weigh in on checking cursor and then we can get this merged.

Thanks!

itsankoff:
Ok, I updated the PR and left only the `NULL` check for client's display and cursor.

mike-jumper:
> Ok, I updated the PR and left only the NULL check for client's display and cursor.

This is an improvement, but the cursor check should be removed as well. From earlier:

> From `guac_common_display_alloc()`, if the display is successfully allocated at all, it is not possible for the `cursor` to be `NULL`:
>
> https://github.com/apache/incubator-guacamole-server/blob/95be88be19e04e07ac1dafb823993745bee7d146/src/common/display.c#L115-L116

Users of `guac_common_display` should be able to rely on the various related objects having been allocated, including the cursor.

It could be argued that the cursor allocation is not guaranteed to succeed due to the nature of `malloc()` and friends. If that's what you're aiming at here, then:

1. The `cursor` allocation check should be within `guac_common_display_alloc()`, to ensure that users of the display can rely on its contents purely based on whether the initial allocation succeeds.
2. The documentation for `guac_common_cursor_alloc()` and `guac_common_display_alloc()` should be updated to note they will return `NULL` if allocation fails.

Either way, in the context of the free handler, only the display need be taken into account. There is no need to re-verify the display's contents. It is better that the display itself be reliable.

itsankoff:
OK, I made the required changes. @mike-jumper let me know if you have any other suggestions?

src/common/cursor.c
```diff
@@ -34,6 +34,14 @@
 #include <stdlib.h>
 #include <string.h>
 
+
+/**
+ * Allocates a cursor as well as an image buffer where the cursor is rendered.
+ * If the allocation fails then the function returns NULL.
```
mike-jumper: This is much better, but please be sure to also use the `@return` tag to document the return value/behavior specifically, not just in the overall description of the function. For example:

https://github.com/apache/incubator-guacamole-server/blob/025fc0525f28b05be206d1471e2d8d488f97cb14/src/pulse/pulse/pulse.h#L92-L94

src/common/display.c
```diff
@@ -99,6 +99,19 @@ static void guac_common_display_free_layers(guac_common_display_layer* layers,
 
 }
 
+/**
+ * Allocates a display and a cursor which are used to represent the remote
+ * display and cursor. If the allocation fails then the function returns NULL.
```
mike-jumper: Same here - much better, but return values should also be documented with `@return`.

src/common/cursor.c
itsankoff: :+1: Done

src/common/display.c
itsankoff: Same here - done

PR #122, opened 2017-11-13
GUACAMOLE-434: Add UK English keymap for RDP.

flangelo:


src/protocols/rdp/keymaps/en_gb_qwerty.keymap
```diff
@@ -0,0 +1,56 @@
...
+map -altgr +shift      0x1E..0x28 0x2B ~ "ASDFGHJKL:@~"
+map -altgr +shift 0x56 0x2C..0x35      ~ "|ZXCVBNM<>?"
+
+#
+# Keys requiring AltGr
+#
+
+map +altgr -shift 0x29 ~ "¦"
+map +altgr -shift 0x05 ~ "€"
+map +altgr -shift 0x12 ~ "é"
```
flangelo: I guess technically the below would be handled via Unicode in RDP...

mike-jumper: They shouldn't be. There is the known issue with dead keys, but that doesn't apply for cases where the desired key can be typed without dead keys. If "é", etc. on a UK keyboard all involve AltGr + some other key, and do not require composition with a dead key, then this will not fall back to Unicode events.

PR #123, opened 2017-11-13
GUACAMOLE-273: Add Brazillian ABNT2 keymap for RDP.

flangelo:


src/protocols/rdp/keymaps/pt_br_qwerty.keymap
```diff
@@ -0,0 +1,58 @@
...
+map -altgr +shift 0x56 0x2C..0x36            ~ "|ZXCVBNM<>:?"
+
+#
+# Keys requiring AltGr
+#
+
+map +altgr -shift 0x02..0x07 0x0D ~ "¹²³£¢¬§"
+map +altgr -shift 0x10..0x11 ~ "/?"
+map +altgr -shift 0x12 ~ "°"
+map +altgr -shift 0x12 ~ "€" # map € to ° so Mac behaves the same as PC
```
mike-jumper: What issue is this meant to work around with PC vs. Mac keyboards?

flangelo: Right... The Mac ABNT2 keyboard flavor isn't laid out like the standard ABNT2 keyboard and produces the Euro symbol in the location the standard ABNT2 keyboard produces the degree symbol. 
In other use-cases, we chose to "normalize" this to standard PC behavior. It was also believed to be less likely someone in Brazil typing Euro character.

mike-jumper: > Arguably - browser specific handling probably should be dealt with in the JS code.

Yep. And, in fact, is. The RDP side of the keyboard mapping is for translating the already-identified key to the keystrokes specific to the RDP side of things. A Mac specific mapping would only be necessary if:

1. The machine serving RDP is a Mac (in this case, we'd probably add a new keyboard layout variant, not merge both variants into one)
2. Or the browser-side handling is failing

If the browser-side mapping is not working correctly, then that needs to be addressed, but probably on the browser side and not within the keymap.

Do things not work on Mac with ABNT2? You should be able to type the Euro symbol using the key combination specific to Mac, and rely on Guacamole to handle the necessary translation for the Windows machine behind RDP.

flangelo: Removing this line behaves fine; Mac users can type the Euro symbol on the Mac ABNT2 keyboard. They should also be able to type degree via Alt/Opt + \ as per Mac ABNT2 layout.

PR #126, opened 2017-12-06
GUACAMOLE-456: use Docker multi-stage build

ceharris:
Since Docker CE 17.05 (edge) and 17.06 (stable), multi-stage builds can be used to create images using a different base images for build versus runtime. This will simplify the process of creating a runtime image for guacamole-server, and will improve build times by making much better use of  Docker's image layer cache.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -21,31 +21,20 @@
 # Dockerfile for guacamole-server
 #
 
-# Start from CentOS base image
-FROM centos:centos7
 
-# Environment variables
-ENV \
-    BUILD_DIR=/tmp/guacd-docker-BUILD \
-    LC_ALL=en_US.UTF-8                \
-    RUNTIME_DEPENDENCIES="            \
-        cairo                         \
-        dejavu-sans-mono-fonts        \
-        freerdp                       \
-        freerdp-plugins               \
-        ghostscript                   \
-        libjpeg-turbo                 \
-        libssh2                       \
-        liberation-mono-fonts         \
-        libtelnet                     \
-        libvorbis                     \
-        libvncserver                  \
-        libwebp                       \
-        pango                         \
-        pulseaudio-libs               \
-        terminus-fonts                \
-        uuid"                         \
-    BUILD_DEPENDENCIES="              \
+# Use CentOS as base for the build
+ARG CENTOS_VERSION=centos7
+FROM centos:${CENTOS_VERSION} AS builder
+
+# Base directory for installed build artifacts.
+# Due to limitations of the Docker image build process, this value is
+# duplicated in an ARG in the second stage of the build.
+#
+ARG PREFIX_DIR=/usr/local/guacamole
+
+# Build arguments
+ARG BUILD_DIR=/tmp/guacd-docker-BUILD
+ARG BUILD_DEPENDENCIES="              \
         autoconf                      \
         automake                      \
         cairo-devel                   \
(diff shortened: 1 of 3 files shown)


src/guacd-docker/bin/build-guacd.sh
```diff
@@ -28,35 +28,22 @@
...
 
 BUILD_DIR="$1"
-
-##
-## Locates the directory in which the FreeRDP libraries (.so files) are
-## located, printing the result to STDOUT.
-##
-where_is_freerdp() {
-    dirname `rpm -ql freerdp-libs | grep 'libfreerdp.*\.so' | head -n1`
-}
```
ceharris: This is now done in the second build stage.

src/guacd-docker/bin/build-guacd.sh
```diff
@@ -28,35 +28,22 @@
...
-#
-# Add FreeRDP plugins to proper path
-#
-
-FREERDP_DIR=`where_is_freerdp`
-FREERDP_PLUGIN_DIR="$FREERDP_DIR/freerdp"
-
-mkdir -p "$FREERDP_PLUGIN_DIR"
-ln -s /usr/local/lib/freerdp/*.so "$FREERDP_PLUGIN_DIR"
-
```
ceharris: This now done in the second build stage.

Dockerfile
```diff
@@ -76,13 +68,63 @@ COPY src/guacd-docker/bin /opt/guacd/bin/
...
+# Copy build artifacts into this stage
+COPY --from=builder ${PREFIX_DIR} ${PREFIX_DIR}
+
+# Link FreeRDP plugins into proper path
+RUN FREERDP_DIR=$(dirname \
+        $(rpm -ql freerdp-libs | grep 'libfreerdp.*\.so' | head -n1)) && \
+    FREERDP_PLUGIN_DIR="${FREERDP_DIR}/freerdp" && \
+    mkdir -p "$FREERDP_PLUGIN_DIR" && \
+    ln -s "$PREFIX_DIR"/lib/freerdp/*.so "$FREERDP_PLUGIN_DIR"
+
```
ceharris: This was previously done in the build script, but needs to be deferred until after build artifacts have been copied into the runtime image. Seems small enough to do inline, but could be moved to an external script.

ceharris:
Just wanted to re-up this to see if there were comments/concerns that need to be addressed. This PR aligns this Docker image build with the approach used in `guacamole-client` using the same multi-stage build approach. It significantly reduces build time by making much better use of the layer cache, and reduces the complexity of the build itself -- no need for a complex single-layer RUN directive that both installs and removes build-time-only dependencies in order to minimize the size of the resulting runtime image.

Dockerfile
mike-jumper: It does seem a little odd to take care of the linking here. Can the creation of the required links not stay with the build script, with the resulting links considered yet more build artifacts?

ceharris: It's a little awkward to copy symlinks. But if you feel really strongly about it, there's certainly a way to get there.

I guess I was thinking that putting these links in place is really just an aspect of installing the build artifacts into the runtime image.

ceharris: Copying symlinks is awkward because copying from the build image to the runtime image is being performed by the Dockerfile COPY directive. If you specify the full path for the symlink, it copies the link target into destination directory rather than copying the link. It will copy symlinks if you specify the directory containing the links (e.g. in this case '/lib64/freerdp) but that copies everything in the directory in addition to the links that would be placed there by Guacamole.

mike-jumper: OK. Sounds like this really is the best we can do, then. It's unfortunate that multi-stage builds don't provide for isolating `COPY` to only the files affected by that build stage.

PR #129, opened 2017-12-18
GUACAMOLE-466: using "-" in place of an input file name with guacenc …

sreid8:
…will cause guacenc to read data from stdin

This is a really small patch that allows guacenc to read data from stdin rather than just from a file. It follows a common Linux practice of reserving the single "-" character to represent stdin if it's used in place of a filename. It also sets a default output filename "output" in the case that stdin is used instead of a file. For encodes that read from a file, the original filename with a ".m4v" is still used as the output filename.

### DIFF
--- a/src/guacenc/encode.c
+++ b/src/guacenc/encode.c
@@ -84,38 +84,44 @@ static int guacenc_read_instructions(guacenc_display* display,
 int guacenc_encode(const char* path, const char* out_path, const char* codec,
         int width, int height, int bitrate, bool force) {
 
-    /* Open input file */
-    int fd = open(path, O_RDONLY);
-    if (fd < 0) {
-        guacenc_log(GUAC_LOG_ERROR, "%s: %s", path, strerror(errno));
-        return 1;
+    int fd;
+    if (strcmp(path, "-") != 0) {
+        /* Open input file */
+        fd = open(path, O_RDONLY);
+        if (fd < 0) {
+            guacenc_log(GUAC_LOG_ERROR, "%s: %s", path, strerror(errno));
+            return 1;
+        }
+
+        /* Lock entire input file for reading by the current process */
+        struct flock file_lock = {
+                .l_type   = F_RDLCK,
+                .l_whence = SEEK_SET,
+                .l_start  = 0,
+                .l_len    = 0,
+                .l_pid    = getpid()
+        };
+
+        /* Abort if file cannot be locked for reading */
+        if (!force && fcntl(fd, F_SETLK, &file_lock) == -1) {
+
+            /* Warn if lock cannot be acquired */
+            if (errno == EACCES || errno == EAGAIN)
+                guacenc_log(GUAC_LOG_WARNING, "Refusing to encode in-progress "
+                        "recording \"%s\" (specify the -f option to override "
+                        "this behavior).", path);
+
+            /* Log an error if locking fails in an unexpected way */
+            else
+                guacenc_log(GUAC_LOG_ERROR, "Cannot lock \"%s\" for reading: %s",
+                        path, strerror(errno));
+
+            close(fd);
+            return 1;
+        }
     }
-
-    /* Lock entire input file for reading by the current process */
-    struct flock file_lock = {
-        .l_type   = F_RDLCK,
-        .l_whence = SEEK_SET,
-        .l_start  = 0,
-        .l_len    = 0,
-        .l_pid    = getpid()
-    };
-
-    /* Abort if file cannot be locked for reading */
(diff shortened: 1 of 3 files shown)


mike-jumper:
Using "-" to represent stdin seems reasonable, but I'm not sure about resorting to a default filename for the output. It is true that "-" is the convention for utilities which write their output to stdout, but I'm not sure how well the convention lends itself to a utility like guacenc which is aimed at processing batches of files.

Do you have any examples handy of utilities with similar behavior (N files of input produces N files of output) and how they handle the stdin case?

What about only reading from stdin if no files are given at all (and writing the output to stdout)?

sreid8:
I do agree that there is a better way to solve the default output name problem. It was a little heavy-handed of me to assume it and set a default. I think a better solution would be to allow guacenc to take output file names as an argument as well. If we supply output file names as well as input file names, though, we very quickly get to a place where the batch functionality of guacenc becomes a hindrance. I’ll list some more examples, but one that immediately comes to mind is that ffmpeg can use an output file extension to know how to write out video time stamping and seek data for an encoded video.
 
Batch functionality doesn’t work well when the user can provide an output location because either a 1-to-1 mapping of inputs to outputs needs to be adhered to which can be frustrating from a user perspective or the tool needs to generate filenames which makes correlation between input and output more difficult.  Examples of existing tools that use batch input are rm, ls, and mkdir which are all tools that take an input, but no file output.

To expand what I’ve already alluded to, I’ve been working on GUACAMOLE-465 as well as this one. It’s an issue I added to Jira with a request to allow guacenc to support more codecs than just mpeg4 and more containers than just m4v. Early in the development of that feature, I’ve been finding that supplying the output file name is essential to give ffmpeg the ability to figure out what container to use during the muxing process of the video and therefore properly write the headers that should be written for that container.

Additionally, the batch functionality of guacenc removes the ability for a set of files to have individual encoding settings. Not all remote systems recorded by guacd will have the same input resolution so in these cases, using batch functionality would be a moot point anyway because the “-s” argument would need to be different for each file so that the video resolution could be different per input file if the user wants the output video to match the remote system resolution.

Finally, I think the batch functionality of guacenc can be replicated much more usefully with shell scripting or using “find -exec” or similar. If guacenc were to lose its ability to batch encode by itself and gained the ability to provide output file names directly, then users can use find or a script to automate the encoding process for all videos in a set of directories and then place the resulting encoded videos wherever they’d like, be that a separate directory for all the videos, in the same directory as the source file, or even stdout assuming we allow “-” to also mean stdout if it’s provided as the output file name. Each of these invocations of guacenc can be done with per-file arguments for bitrate or resolution as well much more easily than can be done now.
 
I’m definitely more than willing to modify this patch so that it better meets what we need guacenc to be. However, I’d really like to discuss how we’d like to see guacenc go because I believe we can get a lot of added functionality from guacenc by letting batching be done by external tools. What do you think?

mike-jumper:
> Hi Mike, can I get some feedback about this? I'd still really like to see something like this in guacenc but I want to do it in a way that's beneficial to the goal of the project

Sure thing. :)

> ... I think a better solution would be to allow guacenc to take output file names as an argument as well. If we supply output file names as well as input file names, though, we very quickly get to a place where the batch functionality of guacenc becomes a hindrance. ...

I agree with both of these points. I wouldn't be against adding a parameter for defining the output file name, but that could be cumbersome for the expected batch usage case.

> Additionally, the batch functionality of guacenc removes the ability for a set of files to have individual encoding settings. ...

Agreed.

> Finally, I think the batch functionality of guacenc can be replicated much more usefully with shell scripting or using “find -exec” or similar. ...

Perhaps, however we must be careful about changing functionality in a way that breaks compatibility. In this case, the breakage would be in a fundamental behavior.

I would say we have two options:

1. Make reasonable attempts to handle the new parameters and deprecate the built-in batching behavior. Issue warnings if the user tries to batch things directly, and warn if parameters are used multiple times in an attempt to apply those parameters in a batch-friendly way. We would then make a note about this deprecation in the release notes of the release in which these changes land.
2. Implement the parameters in a batch-friendly way. Though things may be cumbersome if the parameters are used repeatedly across multiple files on the same command line, this is compatible with both past usages of guacenc and "find -exec".

sreid8:
Thanks for the feedback, Mike! I'll be taking these into consideration as I rework this patch along with the other work I've been doing to guacenc.

sreid8:
I'm closing this request as I think that pull request #159 will handle the good points @mike-jumper brought up in terms of adding stdin to guacenc while also maintaining backward compatibility with the current guacenc command line interface.

PR #130, opened 2018-01-03
GUACAMOLE-448: Add support for configuring bitmap caching.

necouchman:
This change adds support for disabling bitmap caching and offscreen caching, which are enabled by FreeRDP by default.


mike-jumper:
What is the use case(s) for explicitly disabling bitmap/offscreen caching?

PR #139, opened 2018-01-29
GUACAMOLE-352: Clean up keymap lookup/update logic for sake of verifiability.

mike-jumper:
Historically, Guacamole's RDP support has used macros to read and write key definitions within keymaps and similar structures. These macros are difficult to read, difficult to verify, and frequently result in false positives in static analysis builds due to the way each keysym is mapped.

This change replaces these macros with well-documented functions, as well as a more straightforward and explicit mapping scheme. Internal storage of the keymap and current key state has also been unified such that only one set of functions is necessary to address both.

### DIFF
--- a/src/protocols/rdp/keyboard.c
+++ b/src/protocols/rdp/keyboard.c
@@ -188,6 +188,75 @@ static void guac_rdp_send_synchronize_event(guac_rdp_client* rdp_client,
 }
 
 /**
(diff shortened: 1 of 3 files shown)


flangelo:
Looks good to me.

necouchman:
> Looks good to me.

Thanks @flangelo, sorry, didn't mean to leave you out!

PR #153, opened 2018-02-22
GUACAMOLE-269: Add basic support for sending TTY mode encoding over SSH

necouchman:
This is a very basic change that takes care of the problem described in GUACAMOLE-269, specifically in SSH, where the remote system does not correctly register the TTY erase character.  It adds the necessary data structure for passing this through, and the changes to the `libssh2_channel_request_pty_ex()` function to pass the structure through.

I tried to do this in a way that allows it to be extended easily in the future.  Not sure if this is a sane approach or not - it does take care of the issue raised in -269 for the SSH protocol, but I welcome any comments or suggestions on other approaches.

### DIFF
--- a/src/protocols/ssh/ttymode.h
+++ b/src/protocols/ssh/ttymode.h
@@ -1,0 +1,36 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+#ifndef GUAC_SSH_TTYMODE_H
+#define GUAC_SSH_TTYMODE_H
+
+#include "config.h"
+
+#define TTY_OP_END 0
+#define TTY_OP_VERASE 3
+
+#define GUAC_SSH_TERM_DEFAULT_BACKSPACE 127
+
+char guac_tty_modes[] = {
+    TTY_OP_VERASE,
+    0, 0, 0, GUAC_SSH_TERM_DEFAULT_BACKSPACE,
+    TTY_OP_END
+};
+
+#endif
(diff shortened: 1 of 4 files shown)


src/protocols/ssh/ttymode.h
```diff
@@ -0,0 +1,36 @@
...
+#define GUAC_SSH_TTYMODE_H
+
+#include "config.h"
+
+#define TTY_OP_END 0
+#define TTY_OP_VERASE 3
+
+#define GUAC_SSH_TERM_DEFAULT_BACKSPACE 127
+
+char guac_tty_modes[] = {
```
mike-jumper: Global variables within pre-defined values should be defined in the header file using `extern`, with the value of that variable being provided by a corresponding `.c` file. Assuming this is meant to be a constant, the naming should also follow the `UPPERCASE_WITH_UNDERSCORES` convention being used for constants.

For example, see the definition of `GUAC_TERMINAL_INITIAL_PALETTE` within `palette.h`:

https://github.com/apache/guacamole-server/blob/bc5b01d4d8ab0c3c89a08007316d33012261f6b3/src/terminal/terminal/palette.h#L187-L192

whose value is defined in `palette.c`:

https://github.com/apache/guacamole-server/blob/bc5b01d4d8ab0c3c89a08007316d33012261f6b3/src/terminal/palette.c#L23-L34

Again, assuming this is meant as a constant, it should also be declared with a proper `const` qualifier.

src/protocols/ssh/ttymode.h
```diff
@@ -0,0 +1,36 @@
...
+
+#ifndef GUAC_SSH_TTYMODE_H
+#define GUAC_SSH_TTYMODE_H
+
+#include "config.h"
+
+#define TTY_OP_END 0
+#define TTY_OP_VERASE 3
+
+#define GUAC_SSH_TERM_DEFAULT_BACKSPACE 127
```
mike-jumper: This and all other macros/variables declared here need to be documented.

src/protocols/ssh/ttymode.h
```diff
@@ -0,0 +1,36 @@
...
+ * under the License.
+ */
+
+#ifndef GUAC_SSH_TTYMODE_H
+#define GUAC_SSH_TTYMODE_H
+
+#include "config.h"
+
+#define TTY_OP_END 0
+#define TTY_OP_VERASE 3
```
mike-jumper: Unless there is a technical reason to do otherwise, global macros/variables need to be namespaced (in this case, I assume that would be `GUAC_SSH_`).

src/protocols/ssh/ttymode.h
necouchman: Fixed.

src/protocols/ssh/ttymode.h
necouchman: Fixed.

src/protocols/ssh/ttymode.h
necouchman: Documented.

necouchman:
Okay, fixed up those items.

necouchman:
Closing in favor of #156, which includes some of these changes plus the ability to configure the backspace key for both SSH and Telnet sessions.

PR #154, opened 2018-02-23
GUACAMOLE-445: Make Printer Name Configurable

necouchman:
This adds the parameters and changes to rdpdr to make the printer name configurable in the RDP session.

I'm not sure if we had totally settled in the JIRA issue on whether or not this change was actually going to happen, but I went ahead and did it.  I can close the PR if we decide against it.

### DIFF
--- a/src/protocols/rdp/guac_rdpdr/rdpdr_fs_service.c
+++ b/src/protocols/rdp/guac_rdpdr/rdpdr_fs_service.c
@@ -29,27 +29,13 @@
 #include <guacamole/client.h>
 #include <guacamole/protocol.h>
 #include <guacamole/socket.h>
+#include <guacamole/unicode.h>
 
 #ifdef ENABLE_WINPR
 #include <winpr/stream.h>
 #else
 #include "compat/winpr-stream.h"
 #endif
-
-static void guac_rdpdr_device_fs_announce_handler(guac_rdpdr_device* device,
-        wStream* output_stream, int device_id) {
-
-    /* Filesystem header */
-    guac_client_log(device->rdpdr->client, GUAC_LOG_INFO, "Sending filesystem");
-    Stream_Write_UINT32(output_stream, RDPDR_DTYP_FILESYSTEM);
-    Stream_Write_UINT32(output_stream, device_id);
-    Stream_Write(output_stream, "GUAC\0\0\0\0", 8); /* DOS name */
-
-    /* Filesystem data */
-    Stream_Write_UINT32(output_stream, GUAC_FILESYSTEM_NAME_LENGTH);
-    Stream_Write(output_stream, GUAC_FILESYSTEM_NAME, GUAC_FILESYSTEM_NAME_LENGTH);
-
-}
 
 static void guac_rdpdr_device_fs_iorequest_handler(guac_rdpdr_device* device,
         wStream* input_stream, int file_id, int completion_id, int major_func, int minor_func) {
@@ -128,6 +114,9 @@ static void guac_rdpdr_device_fs_iorequest_handler(guac_rdpdr_device* device,
 }
 
 static void guac_rdpdr_device_fs_free_handler(guac_rdpdr_device* device) {
+
+    Stream_Free(device->device_announce, 1);
+    
 }
 
 void guac_rdpdr_register_fs(guac_rdpdrPlugin* rdpdr) {
@@ -143,9 +132,21 @@ void guac_rdpdr_register_fs(guac_rdpdrPlugin* rdpdr) {
     device->rdpdr       = rdpdr;
     device->device_id   = id;
     device->device_name = "Guacamole Filesystem";
+    int device_name_len = guac_utf8_strlen(device->device_name);
+    device->device_type = RDPDR_DTYP_FILESYSTEM;
+    device->dos_name = "GUACFS\0\0";
+
+    /* Set up the device announcement */
+    device->device_announce_len = 20 + device_name_len;
+    device->device_announce = Stream_New(NULL, device->device_announce_len);
+    Stream_Write_UINT32(device->device_announce, device->device_type);
+    Stream_Write_UINT32(device->device_announce, device->device_id);
+    Stream_Write(device->device_announce, device->dos_name, 8);
+    Stream_Write_UINT32(device->device_announce, device_name_len);
+    Stream_Write(device->device_announce, device->device_name, device_name_len);
+    
 
     /* Set handlers */
-    device->announce_handler  = guac_rdpdr_device_fs_announce_handler;
     device->iorequest_handler = guac_rdpdr_device_fs_iorequest_handler;
     device->free_handler      = guac_rdpdr_device_fs_free_handler;
 
--- a/src/protocols/rdp/guac_rdpdr/rdpdr_messages.c
+++ b/src/protocols/rdp/guac_rdpdr/rdpdr_messages.c
@@ -22,9 +22,11 @@
 #include "rdp.h"
 #include "rdpdr_messages.h"
 #include "rdpdr_service.h"
+#include "unicode.h"
 
 #include <freerdp/utils/svc_plugin.h>
 #include <guacamole/client.h>
+#include <guacamole/unicode.h>
 
 #ifdef ENABLE_WINPR
 #include <winpr/stream.h>
@@ -122,20 +124,29 @@ static void guac_rdpdr_send_client_capability(guac_rdpdrPlugin* rdpdr) {
 
 static void guac_rdpdr_send_client_device_list_announce_request(guac_rdpdrPlugin* rdpdr) {
 
-    int i;
-    wStream* output_stream = Stream_New(NULL, 256);
+    /* Calculate number of bytes needed for the stream */
+    int streamBytes = 16;
+    for (int i=0; i < rdpdr->devices_registered; i++)
+        streamBytes += rdpdr->devices[i].device_announce_len;
+
+    /* Allocate the stream */
+    wStream* output_stream = Stream_New(NULL, streamBytes);
 
     /* Write header */
     Stream_Write_UINT16(output_stream, RDPDR_CTYP_CORE);
     Stream_Write_UINT16(output_stream, PAKID_CORE_DEVICELIST_ANNOUNCE);
 
-    /* List devices */
+    /* Get the stream for each of the devices. */
     Stream_Write_UINT32(output_stream, rdpdr->devices_registered);
-    for (i=0; i<rdpdr->devices_registered; i++) {
-        guac_rdpdr_device* device = &(rdpdr->devices[i]);
-        device->announce_handler(device, output_stream, i);
+    for (int i=0; i<rdpdr->devices_registered; i++) {
+        
+        Stream_Write(output_stream,
+                Stream_Buffer(rdpdr->devices[i].device_announce),
+                rdpdr->devices[i].device_announce_len);
+
         guac_client_log(rdpdr->client, GUAC_LOG_INFO, "Registered device %i (%s)",
-                device->device_id, device->device_name);
+                rdpdr->devices[i].device_id, rdpdr->devices[i].device_name);
+        
     }
 
     svc_plugin_send((rdpSvcPlugin*) rdpdr, output_stream);
--- a/src/protocols/rdp/guac_rdpdr/rdpdr_messages.c
+++ b/src/protocols/rdp/guac_rdpdr/rdpdr_messages.c
@@ -273,4 +273,3 @@ void guac_rdpdr_process_prn_cache_data(guac_rdpdrPlugin* rdpdr, wStream* input_s
 void guac_rdpdr_process_prn_using_xps(guac_rdpdrPlugin* rdpdr, wStream* input_stream) {
     guac_client_log(rdpdr->client, GUAC_LOG_INFO, "Printer unexpectedly switched to XPS mode");
 }
-
--- a/src/protocols/rdp/guac_rdpdr/rdpdr_printer.c
+++ b/src/protocols/rdp/guac_rdpdr/rdpdr_printer.c
@@ -133,38 +133,6 @@ void guac_rdpdr_process_print_job_close(guac_rdpdr_device* device,
 
 }
 
-static void guac_rdpdr_device_printer_announce_handler(guac_rdpdr_device* device,
-        wStream* output_stream, int device_id) {
-
-    /* Printer header */
-    guac_client_log(device->rdpdr->client, GUAC_LOG_INFO, "Sending printer");
-    Stream_Write_UINT32(output_stream, RDPDR_DTYP_PRINT);
-    Stream_Write_UINT32(output_stream, device_id);
-    Stream_Write(output_stream, "PRN1\0\0\0\0", 8); /* DOS name */
-
-    /* Printer data */
-    int settings_length = guac_utf8_strlen(device->device_name);
-    int printer_name_length = (settings_length + 1) * 2;
-    char printer_name[printer_name_length];
-    guac_rdp_utf8_to_utf16((const unsigned char*)device->device_name, 
-            settings_length, printer_name, printer_name_length);
-    printer_name[printer_name_length - 2] = '\0';
-    printer_name[printer_name_length - 1] = '\0';
-    Stream_Write_UINT32(output_stream, 24 + GUAC_PRINTER_DRIVER_LENGTH + printer_name_length);
-    Stream_Write_UINT32(output_stream,
-              RDPDR_PRINTER_ANNOUNCE_FLAG_DEFAULTPRINTER
-            | RDPDR_PRINTER_ANNOUNCE_FLAG_NETWORKPRINTER);
-    Stream_Write_UINT32(output_stream, 0); /* reserved - must be 0 */
-    Stream_Write_UINT32(output_stream, 0); /* PnPName length (PnPName is ultimately ignored) */
-    Stream_Write_UINT32(output_stream, GUAC_PRINTER_DRIVER_LENGTH); /* DriverName length */
-    Stream_Write_UINT32(output_stream, printer_name_length);   /* PrinterName length */
-    Stream_Write_UINT32(output_stream, 0);                          /* CachedFields length */
-
-    Stream_Write(output_stream, GUAC_PRINTER_DRIVER, GUAC_PRINTER_DRIVER_LENGTH);
-    Stream_Write(output_stream, printer_name, printer_name_length);
-
-}
-
 static void guac_rdpdr_device_printer_iorequest_handler(guac_rdpdr_device* device,
         wStream* input_stream, int file_id, int completion_id, int major_func, int minor_func) {
 
@@ -196,7 +164,9 @@ static void guac_rdpdr_device_printer_iorequest_handler(guac_rdpdr_device* devic
 }
 
 static void guac_rdpdr_device_printer_free_handler(guac_rdpdr_device* device) {
-    /* Do nothing */
+
+    Stream_Free(device->device_announce, 1);
+
 }
 
 void guac_rdpdr_register_printer(guac_rdpdrPlugin* rdpdr, char* printer_name) {
@@ -210,9 +180,41 @@ void guac_rdpdr_register_printer(guac_rdpdrPlugin* rdpdr, char* printer_name) {
     device->rdpdr       = rdpdr;
     device->device_id   = id;
     device->device_name = printer_name;
+    int device_name_len = guac_utf8_strlen(device->device_name);
+    device->device_type = RDPDR_DTYP_PRINT;
+    device->dos_name = "PRN1\0\0\0\0";
+    
+    /* Set up device announce stream */
+    int prt_name_len = (device_name_len + 1) * 2;
+    device->device_announce_len = 44 + prt_name_len
+            + GUAC_PRINTER_DRIVER_LENGTH;
+    device->device_announce = Stream_New(NULL, device->device_announce_len);
+    
+    /* Write common information. */
+    Stream_Write_UINT32(device->device_announce, device->device_type);
+    Stream_Write_UINT32(device->device_announce, device->device_id);
+    Stream_Write(device->device_announce, device->dos_name, 8);
+    
+    /* DeviceDataLength */
+    Stream_Write_UINT32(device->device_announce, 24 +  prt_name_len + GUAC_PRINTER_DRIVER_LENGTH);
+    
+    /* Begin printer-specific information */
+    Stream_Write_UINT32(device->device_announce,
+              RDPDR_PRINTER_ANNOUNCE_FLAG_DEFAULTPRINTER
+            | RDPDR_PRINTER_ANNOUNCE_FLAG_NETWORKPRINTER); /* Printer flags */
+    Stream_Write_UINT32(device->device_announce, 0); /* Reserved - must be 0. */
+    Stream_Write_UINT32(device->device_announce, 0); /* PnPName Length - ignored. */
+    Stream_Write_UINT32(device->device_announce, GUAC_PRINTER_DRIVER_LENGTH);
+    Stream_Write_UINT32(device->device_announce, prt_name_len);
+    Stream_Write_UINT32(device->device_announce, 0); /* CachedFields length. */
+    
+    Stream_Write(device->device_announce, GUAC_PRINTER_DRIVER, GUAC_PRINTER_DRIVER_LENGTH);
+    guac_rdp_utf8_to_utf16((const unsigned char*) device->device_name,
+            device_name_len + 1, (char*) Stream_Pointer(device->device_announce),
+            prt_name_len);
+    Stream_Seek(device->device_announce, prt_name_len);
 
     /* Set handlers */
-    device->announce_handler  = guac_rdpdr_device_printer_announce_handler;
     device->iorequest_handler = guac_rdpdr_device_printer_iorequest_handler;
     device->free_handler      = guac_rdpdr_device_printer_free_handler;
 
--- a/src/protocols/rdp/guac_rdpdr/rdpdr_printer.c
+++ b/src/protocols/rdp/guac_rdpdr/rdpdr_printer.c
@@ -25,12 +25,14 @@
 #include "rdp.h"
 #include "rdp_print_job.h"
 #include "rdp_status.h"
+#include "unicode.h"
 
 #include <freerdp/utils/svc_plugin.h>
 #include <guacamole/client.h>
 #include <guacamole/protocol.h>
 #include <guacamole/socket.h>
 #include <guacamole/stream.h>
+#include <guacamole/unicode.h>
 #include <guacamole/user.h>
 
 #ifdef ENABLE_WINPR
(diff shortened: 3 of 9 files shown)


src/protocols/rdp/guac_rdpdr/rdpdr_messages.c
```diff
@@ -273,3 +273,15 @@ void guac_rdpdr_process_prn_using_xps(guac_rdpdrPlugin* rdpdr, wStream* input_st
     guac_client_log(rdpdr->client, GUAC_LOG_INFO, "Printer unexpectedly switched to XPS mode");
 }
 
+int guac_rdpdr_encode_utf16(const char* input_string, char* output_string) {
```
necouchman: Not 100% certain this is the right place for this function.  Also, there may be something else that already does this (swprintf?), but I had a hard time finding good examples/documentation for plain C - most everything dealt with the char16_t type which seems specific to C++.

src/protocols/rdp/guac_rdpdr/rdpdr_messages.c
```diff
@@ -273,3 +273,15 @@ void guac_rdpdr_process_prn_using_xps(guac_rdpdrPlugin* rdpdr, wStream* input_st
...
+    int output_length = (strlen(input_string)+ 1) * 2;
+    output_string = realloc(output_string, output_length);
```
necouchman: I'm not sure this is the best way to go about this - feels a little kludgy, but was trying to keep dynamic allocation.

src/protocols/rdp/guac_rdpdr/rdpdr_printer.c
```diff
@@ -141,18 +141,20 @@ static void guac_rdpdr_device_printer_announce_handler(guac_rdpdr_device* device
     Stream_Write(output_stream, "PRN1\0\0\0\0", 8); /* DOS name */
 
     /* Printer data */
-    Stream_Write_UINT32(output_stream, 24 + GUAC_PRINTER_DRIVER_LENGTH + GUAC_PRINTER_NAME_LENGTH);
+    char* utf16_printer_name = malloc(1);
```
necouchman: As with comment above on `realloc()`, this feels a little kludgy, but allows for dynamic allocation to work.

src/protocols/rdp/guac_rdpdr/rdpdr_messages.c
mike-jumper: This function as written would not actually convert things properly, as the input string is UTF-8. Characters in UTF-8 vary in byte length, and simply adding a null byte for each input byte will only work for a very small subset of UTF-8. Thankfully, we've actually done this already within `guac_rdp_utf8_to_utf16()` (defined within `unicode.h`):

https://github.com/apache/guacamole-server/blob/bc5b01d4d8ab0c3c89a08007316d33012261f6b3/src/protocols/rdp/unicode.h#L41-L57

src/protocols/rdp/guac_rdpdr/rdpdr_printer.c
mike-jumper: I am definitely not a fan of dynamically allocating one byte just so we can `realloc()` it immediately after. I suspect we don't actually need to do this though - see below.

src/protocols/rdp/rdp_settings.c
```diff
@@ -726,6 +732,11 @@ guac_rdp_settings* guac_rdp_parse_args(guac_user* user,
         guac_user_parse_args_boolean(user, GUAC_RDP_CLIENT_ARGS, argv,
                 IDX_ENABLE_PRINTING, 0);
 
+    /* Name of redirected printer */
+    settings->printer_name =
+        guac_user_parse_args_string(user, GUAC_RDP_CLIENT_ARGS, argv,
```
mike-jumper: This will need to be freed within `guac_rdp_settings_free()`.

src/protocols/rdp/guac_rdpdr/rdpdr_messages.c
necouchman: Brilliant!  Will use this one, instead.

src/protocols/rdp/rdp_settings.c
necouchman: Ah, yes.

necouchman: Freed.

src/protocols/rdp/guac_rdpdr/rdpdr_printer.c
necouchman: Removed.

src/protocols/rdp/guac_rdpdr/rdpdr_messages.c
necouchman: Removed.

src/protocols/rdp/guac_rdpdr/rdpdr_messages.c
necouchman: Fixed.

src/protocols/rdp/guac_rdpdr/rdpdr_printer.c
```diff
@@ -141,18 +142,23 @@ static void guac_rdpdr_device_printer_announce_handler(guac_rdpdr_device* device
     Stream_Write(output_stream, "PRN1\0\0\0\0", 8); /* DOS name */
 
     /* Printer data */
-    Stream_Write_UINT32(output_stream, 24 + GUAC_PRINTER_DRIVER_LENGTH + GUAC_PRINTER_NAME_LENGTH);
+    int settings_length = strlen(device->device_name);
```
mike-jumper: Unfortunately, this value will not be what you expect if the device name contains any multi-byte characters. You will instead end up with the number of bytes in the string, not the number of characters. You can use `guac_utf8_strlen()`, however, which is part of libguac:

https://github.com/apache/guacamole-server/blob/bc5b01d4d8ab0c3c89a08007316d33012261f6b3/src/libguac/guacamole/unicode.h#L41-L48

src/protocols/rdp/guac_rdpdr/rdpdr_printer.c
```diff
@@ -141,18 +142,23 @@ static void guac_rdpdr_device_printer_announce_handler(guac_rdpdr_device* device
...
+    int printer_name_length = (settings_length + 1) * 2;
+    char* printer_name = malloc(printer_name_length);
+    guac_rdp_utf8_to_utf16((const unsigned char*)device->device_name,
```
mike-jumper: Beware that `guac_rdp_utf8_to_utf16()`, confusingly, does not automatically null-terminate the output string. I'm not sure at the moment whether that is an oversight, or whether things will break if that function is modified to properly null-terminate things, but this will somehow need to be taken into account.

Probably a good idea for us to update the documentation for those utility functions while we're at it, as it's problematic for the null-terminator behavior to be undocumented and different between the two.

src/protocols/rdp/guac_rdpdr/rdpdr_printer.c
necouchman: Aha.  Fixed.

src/protocols/rdp/guac_rdpdr/rdpdr_printer.c
necouchman: Hmmm...okay.  I haven't seen any issue with it behaving improperly, but I've added the null terminator explicitly to the end of the string.  Doesn't feel very elegant, but it appears to work.

necouchman:
@mike-jumper Anything else need attention on this one?

src/protocols/rdp/guac_rdpdr/rdpdr_printer.c
```diff
@@ -141,18 +143,25 @@ static void guac_rdpdr_device_printer_announce_handler(guac_rdpdr_device* device
     Stream_Write(output_stream, "PRN1\0\0\0\0", 8); /* DOS name */
 
     /* Printer data */
-    Stream_Write_UINT32(output_stream, 24 + GUAC_PRINTER_DRIVER_LENGTH + GUAC_PRINTER_NAME_LENGTH);
+    int settings_length = guac_utf8_strlen(device->device_name);
+    int printer_name_length = (settings_length + 1) * 2;
+    char* printer_name = malloc(printer_name_length);
```
mike-jumper: If you will be dynamically-allocating the printer name here, it will need to be freed. Alternatively, statically-allocating the buffer may simplify things.

src/protocols/rdp/guac_rdpdr/rdpdr_printer.c
```diff
@@ -141,18 +143,25 @@ static void guac_rdpdr_device_printer_announce_handler(guac_rdpdr_device* device
...
     Stream_Write_UINT32(output_stream, 0); /* reserved - must be 0 */
     Stream_Write_UINT32(output_stream, 0); /* PnPName length (PnPName is ultimately ignored) */
     Stream_Write_UINT32(output_stream, GUAC_PRINTER_DRIVER_LENGTH); /* DriverName length */
-    Stream_Write_UINT32(output_stream, GUAC_PRINTER_NAME_LENGTH);   /* PrinterName length */
+    Stream_Write_UINT32(output_stream, printer_name_length);   /* PrinterName length */
     Stream_Write_UINT32(output_stream, 0);                          /* CachedFields length */
 
     Stream_Write(output_stream, GUAC_PRINTER_DRIVER, GUAC_PRINTER_DRIVER_LENGTH);
-    Stream_Write(output_stream, GUAC_PRINTER_NAME,   GUAC_PRINTER_NAME_LENGTH);
+    Stream_Write(output_stream, printer_name, printer_name_length);
```
mike-jumper: We may need to do more with respect to verifying the necessary space is actually available. Looking into the dynamic allocation of the stream buffer, we're currently allocating a static 256 bytes to contain all RDPDR devices:

https://github.com/apache/guacamole-server/blob/344ed4f42ea715a2e0e7dfd09867964ecfc320c2/src/protocols/rdp/guac_rdpdr/rdpdr_messages.c#L126

This was sufficient when the length of device names was known and this length could not possible be exceeded even if both drive and printer were enabled, but device names are now arbitrary.

Enforcing length limits on device names, truncating the names, etc. would be one possibility. Dynamically calculating the necessary space would be another (though more complex).

necouchman: Hmmm....okay.  While enforcing limits and truncating names might be easier, I'm a little skeptical it actually will be any easier, and think probably dynamically calculating the necessary space is the way to go.  I'll see what I can figure out, but I have a feeling I'll be needing some guidance on the best way to accomplish this.

necouchman: Okay, took an initial stab at a calculation...I'm sure it needs some work.

mike-jumper: I think some of the complexity we're dealing with is due to the concept of the `announce_handler` and overlapping responsibilities between the devices and the top-level handling of RDPDR and the announce packet.

What if, instead of an `announce_handler`, we:

* Stored the device type (`RDPDR_DTYP_PRINT`, etc.) within the `guac_rdpdr_device`
* Stored the arbitrary device data and the length of that data within the `guac_rdpdr_device`
* Relied upon that within `guac_rdpdr_send_client_device_list_announce_request()`, such that it's truly only the responsibility of that function to assemble the announce packet

I think that should simplify things greatly.

If it helps, I'd be glad to take a stab at the above refactor against the current codebase, though I'm not sure how painful the rebases of these name-related PRs would be after that.

necouchman: Sounds good - I'll be happy to have a go at it, first, and see if I can figure out how to do it - I'll let you know if I need some help.

necouchman: Okay, a couple of clarification questions/concerns, here:
- For tracking the size of arbitrary data within the `guac_rdpdr_device`, are you thinking all of the data - name, type, and then anything in the `data` field?  Or just the size of the `data` field?
- The only potential issue with getting rid of `announce_handler`s entirely is that the complexity of the printer one is significantly larger than that of the filesystem one.  I can still do this, it just means that `guac_rdpdr_send_client_device_list_announce_request()` will end up with either some `if...else if` statements or a `switch` statement to handle initialization that goes with each of the devices.  Or did you envision keeping the `announce_handler`s in some form, just moving a bunch of the code over to the `guac_rdpdr_send_client_device_list_announce_request()` function?

necouchman: Well, I've taken a stab at the implementation of this, as seen in the latest commit.  It feels far, far away from anywhere near good at this point, but you can take a look and let me know if I'm on the right track and make suggestions.  The whole double loop thing just feels wrong, but I'm not sure how else to get the number of bytes before opening the stream and then actually write the stream data.

necouchman:
@mike-jumper Any thoughts on that last bit of work to deal with the stream data size?  Any suggestions on how to handle that more elegantly than looping through twice?

necouchman:
Base changed to staging/1.0.0

necouchman:
@mike-jumper Ping...any suggestions/directions on the best way to get the stream size, here?

mike-jumper:
What about handling the creation of the device-specific portion of that initial announce packet within the device-specific allocation routine, storing the data within a `wStream` on the `guac_rdpdr_device`? So, similar to how there is already `device->device_id`, and `device->device_name`, something like `device->device_announce`?

If `guac_rdpdr_device` had a `wStream*` member which pointed to a `wStream` which the device implementation has the responsibility to allocate, initialize, and (eventually) free, then the top-level bit of code which deals with announcing all devices via RDPDR could rely on the device implementation to supply the full device announce header, and could calculate the total length by adding up the lengths of those headers.

By the way, for reference, the relevant part of the RDPDR spec is:

https://msdn.microsoft.com/en-us/library/cc241357.aspx

That may help with naming and/or determining what parts of things should have responsibility for what. Fair warning: though I'm linking to the HTML docs above, the PDF versions are way more readable if you need to read more than just the linked section (they're just less link-friendly).

necouchman:
> What about handling the creation of the device-specific portion of that initial announce packet within the device-specific allocation routine, storing the data within a wStream on the guac_rdpdr_device? So, similar to how there is already device->device_id, and device->device_name, something like device->device_announce?

So, I went ahead and did this, and as I was implementing it, noticed that it's very similar to how it was implemented before with the announce handler functions :-).

> If guac_rdpdr_device had a wStream* member which pointed to a wStream which the device implementation has the responsibility to allocate, initialize, and (eventually) free, then the top-level bit of code which deals with announcing all devices via RDPDR could rely on the device implementation to supply the full device announce header, and could calculate the total length by adding up the lengths of those headers.

It definitely feels cleaner this way, but doesn't seem to take care of the need for the double-loop - I'm still having to loop twice - once to gather the length from all of the devices so that I have a total size, then allocate the stream, then loop again to copy all of the device-specific streams onto the main stream...

mike-jumper:
Yeah, I don't think you're going to be able to avoid those multiple passes. There are two distinct tasks which must be done separately, and both involve looping over that data.

It's not necessarily in itself a bad thing. Computationally, we're still O(N). Architecturally, there's nice separation of concerns. If this also cleans things up, then I'm a fan.

necouchman:
Fair enough.  Should I stick with the route I've taken, or go back and use the announce handler functions, instead, to handle allocation of this stream?

mike-jumper:
I think you're already well on the right track, and you're right to remove the announce handlers. Allocation of the stream in this case would make the most sense in the same area where the device ID, name, etc. are allocated/initialized:

https://github.com/apache/guacamole-server/blob/67680bd2d51e7949453f0f7ffc7f4234a1136715/src/protocols/rdp/guac_rdpdr/rdpdr_printer.c#L193-L210

https://github.com/apache/guacamole-server/blob/0e5498fc83a05aebbc9558a4503af8c92abfdac6/src/protocols/rdp/guac_rdpdr/rdpdr_fs_service.c#L133-L156

necouchman:
Okay, I think I have it pretty well there.  Let me know what needs attention.  I wasn't sure if there was some place specifically I needed to run `Stream_Free` - the free handler functions are completely empty for all of the device types, so not sure if several things need to be added (Stream cleanup plus `char*` cleanup)?

src/protocols/rdp/guac_rdpdr/rdpdr_printer.c
```diff
@@ -200,10 +182,43 @@ void guac_rdpdr_register_printer(guac_rdpdrPlugin* rdpdr) {
...
+    device->device_name = printer_name;
+    device->device_name_len = guac_utf8_strlen(device->device_name);
+    device->device_type = RDPDR_DTYP_PRINT;
+    device->dos_name = "PRN1\0\0\0\0";
+    
+    /* Set up device announce stream */
+    int prt_name_len = (device->device_name_len + 1) * 2;
+    device->device_announce_len = 44 + prt_name_len
+            + GUAC_PRINTER_DRIVER_LENGTH;
+    char prt_name[prt_name_len];
```
mike-jumper: While C99 does allow dynamic allocation of arrays in the manner, beware that unbounded use of the stack is dangerous. I would normally recommend `malloc()` instead, but perhaps writing directly to the `wStream` buffer would be a better option in this case? There would then be no need to copy things over, leveraging `Stream_Pointer()` and `Stream_Seek()` to accomplish the same through the call to `guac_rdp_utf8_to_utf16()`.

src/protocols/rdp/guac_rdpdr/rdpdr_messages.c
```diff
@@ -122,20 +124,28 @@ static void guac_rdpdr_send_client_capability(guac_rdpdrPlugin* rdpdr) {
...
 
-    /* List devices */
+    /* Get the stream for each of the devices. */
     Stream_Write_UINT32(output_stream, rdpdr->devices_registered);
-    for (i=0; i<rdpdr->devices_registered; i++) {
-        guac_rdpdr_device* device = &(rdpdr->devices[i]);
-        device->announce_handler(device, output_stream, i);
+    for (int i=0; i<rdpdr->devices_registered; i++) {
+        
+        Stream_Copy(output_stream, rdpdr->devices[i].device_announce,
```
mike-jumper: I'm not sure `Stream_Copy()` is the right function/macro to use in this case. Checking the source for FreeRDP 1.1, I see `Stream_Copy()` defined as:

```c
#define Stream_Copy(_dst, _src, _n) do { \
    memcpy(_dst->pointer, _src->pointer, _n); \
    _dst->pointer += _n; \
    _src->pointer += _n; \
    } while (0)
```

It thus depends on the current write position within the buffer, and updates that write position as a result of the call.

Should we be using `Stream_Write()` and `Stream_Buffer()` instead?

src/protocols/rdp/guac_rdpdr/rdpdr_fs_service.c
```diff
@@ -128,6 +114,12 @@ static void guac_rdpdr_device_fs_iorequest_handler(guac_rdpdr_device* device,
 }
 
 static void guac_rdpdr_device_fs_free_handler(guac_rdpdr_device* device) {
+
+    free(device->rdpdr);
+    free((char *) device->device_name);
+    free(device->dos_name);
+    Stream_Free(device->device_announce, 1);
```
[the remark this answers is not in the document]
necouchman: I think this is already present:

https://github.com/apache/guacamole-server/blob/67680bd2d51e7949453f0f7ffc7f4234a1136715/src/protocols/rdp/compat/winpr-stream.h#L68

and:

https://github.com/apache/guacamole-server/blob/67680bd2d51e7949453f0f7ffc7f4234a1136715/src/protocols/rdp/compat/winpr-stream.c#L38

And it looks like the defined `Stream_Free()` function takes two arguments??

src/protocols/rdp/guac_rdpdr/rdpdr_fs_service.c
```diff
@@ -128,6 +114,12 @@ static void guac_rdpdr_device_fs_iorequest_handler(guac_rdpdr_device* device,
 }
 
 static void guac_rdpdr_device_fs_free_handler(guac_rdpdr_device* device) {
+
+    free(device->rdpdr);
```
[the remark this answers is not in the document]
necouchman: Got it, will remove.

src/protocols/rdp/guac_rdpdr/rdpdr_fs_service.c
```diff
@@ -128,6 +114,12 @@ static void guac_rdpdr_device_fs_iorequest_handler(guac_rdpdr_device* device,
...
+    free((char *) device->device_name);
+    free(device->dos_name);
```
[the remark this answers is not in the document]
necouchman: Ah, makes sense.  My and my hang-up with static vs. dynamic allocation.  I think I always equate pointers to dynamic allocation...

src/protocols/rdp/guac_rdpdr/rdpdr_printer.c
necouchman: Okay, will give it a shot.

src/protocols/rdp/guac_rdpdr/rdpdr_messages.c
necouchman: I'll take a look - I might need some guidance figuring this out, but I'll give it a try.

src/protocols/rdp/guac_rdpdr/rdpdr_printer.c
```diff
@@ -165,9 +165,6 @@ static void guac_rdpdr_device_printer_iorequest_handler(guac_rdpdr_device* devic
 
 static void guac_rdpdr_device_printer_free_handler(guac_rdpdr_device* device) {
 
-    free(device->rdpdr);
-    free((char *) device->device_name);
```
mike-jumper: `device->device_name` *is* dynamically allocated in this case (printer) due to the changes in this PR. Only the filesystem one currently remains static and shouldn't be freed.

necouchman: Ah, yes.  Of course.

src/protocols/rdp/guac_rdpdr/rdpdr_printer.c
necouchman: Okay, I think I've managed to implement it the way it's supposed to be :-).

src/protocols/rdp/guac_rdpdr/rdpdr_messages.c
necouchman: Well, clearly I've messed something up - not sure if it's this or the printer naming thing, but segfaults every time...

src/protocols/rdp/guac_rdpdr/rdpdr_messages.c
```diff
@@ -138,10 +138,10 @@ static void guac_rdpdr_send_client_device_list_announce_request(guac_rdpdrPlugin
 
     /* Get the stream for each of the devices. */
     Stream_Write_UINT32(output_stream, rdpdr->devices_registered);
-    for (int i=0; i<rdpdr->devices_registered; i++) {
+    for (int i=0; i < rdpdr->devices_registered; i++) {
         
         Stream_Write(output_stream,
-                Stream_Buffer(rdpdr->devices[i].device_announce),
+                rdpdr->devices[i].device_announce,
```
mike-jumper: Intrepid.

src/protocols/rdp/guac_rdpdr/rdpdr_messages.c
mike-jumper: :shipit:

mike-jumper: I'll pull what you've got so far and see if something jumps out.

mike-jumper: ...am reverting c30fd7a first, though. Writing the raw data structure is a bit too scary, especially since it only contains a pointer to the desired buffer. It shouldn't work, and I'm not sure what I'd do if it _did_ work. ;)

mike-jumper: aHA! You're missing the `Stream_New()` and corresponding assignment to `device->device_announce` for the printer device. It's there for filesystem, but not for printer:

https://github.com/apache/guacamole-server/blob/e2994778c3fed6e71f809aa589afe2db32b2e0ae/src/protocols/rdp/guac_rdpdr/rdpdr_fs_service.c#L144

mike-jumper: Quickly added the necessary allocation and assignment for the sake of testing, and it seems to work nicely:

renamed-printer

necouchman: :man_facepalming: Doh.  Yeah, that usually helps.  Okay, I've initialized this and reverted my bad try below.  Should be good, now.

src/protocols/rdp/compat/winpr-stream.h
```diff
@@ -60,6 +60,7 @@
 #define Stream_Buffer                  stream_get_head
 #define Stream_Pointer                 stream_get_tail
 #define Stream_Length                  stream_get_size
+#define Stream_Copy                    stream_copy
```
mike-jumper: Is this still needed as of the latest changes?

src/protocols/rdp/guac_rdpdr/rdpdr_printer.h
```diff
@@ -35,7 +35,7 @@
  * Registers a new printer device within the RDPDR plugin. This must be done
  * before RDPDR connection finishes.
  */
-void guac_rdpdr_register_printer(guac_rdpdrPlugin* rdpdr);
+void guac_rdpdr_register_printer(guac_rdpdrPlugin* rdpdr, char* printer_name);
```
mike-jumper: Please document the parameters of this function, now that the function is being updated.

src/protocols/rdp/guac_rdpdr/rdpdr_service.h
```diff
@@ -74,23 +74,43 @@ struct guac_rdpdr_device {
...
+
+    /**
+     * The type of RDPDR device that this represents.
+     */
+    uint32_t device_type;
+
+    /**
+     * The DOS name of the device.  Max 8 bytes, including terminator.
+     */
+    char *dos_name;
```
mike-jumper: Like `device_name`, this should probably also be `const char*`, unless you think they should both be `char*`.

src/protocols/rdp/guac_rdpdr/rdpdr_printer.c
```diff
@@ -200,10 +180,43 @@ void guac_rdpdr_register_printer(guac_rdpdrPlugin* rdpdr) {
...
+            | RDPDR_PRINTER_ANNOUNCE_FLAG_NETWORKPRINTER); /* Printer flags */
+    Stream_Write_UINT32(device->device_announce, 0); /* Reserved - must be 0. */
+    Stream_Write_UINT32(device->device_announce, 0); /* PnPName Length - ignored. */
+    Stream_Write_UINT32(device->device_announce, GUAC_PRINTER_DRIVER_LENGTH);
+    Stream_Write_UINT32(device->device_announce, prt_name_len);
+    Stream_Write_UINT32(device->device_announce, 0); /* CachedFields length. */
+    
+    Stream_Write(device->device_announce, GUAC_PRINTER_DRIVER, GUAC_PRINTER_DRIVER_LENGTH);
+    guac_rdp_utf8_to_utf16((const unsigned char*) device->device_name,
+            device->device_name_len,
```
mike-jumper: This should be `device_name_len + 1` to include the null terminator.

src/protocols/rdp/guac_rdpdr/rdpdr_service.h
```diff
@@ -74,23 +74,43 @@ struct guac_rdpdr_device {
...
      */
     const char* device_name;
 
     /**
-     * Handler which will be called when the RDPDR plugin is forming the client
-     * device announce list.
+     * The length of the device name, used for generating stream.
      */
-    guac_rdpdr_device_announce_handler* announce_handler;
+    int device_name_len;
```
mike-jumper: Is this useful within `guac_rdpdr_device`? Or is the length really only an internal consideration of the code generating `device_announce`?

mike-jumper:
This is beginning to look pretty great.

src/protocols/rdp/compat/winpr-stream.h
necouchman: Nope.

src/protocols/rdp/guac_rdpdr/rdpdr_printer.h
necouchman: Got it.

src/protocols/rdp/guac_rdpdr/rdpdr_service.h
necouchman: `const` is fine with me

src/protocols/rdp/guac_rdpdr/rdpdr_printer.c
necouchman: Got it.

src/protocols/rdp/guac_rdpdr/rdpdr_service.h
necouchman: Not really - best to just make it local.

necouchman: Localized.

src/protocols/rdp/guac_rdpdr/rdpdr_printer.c
necouchman: +1'd.

src/protocols/rdp/guac_rdpdr/rdpdr_service.h
necouchman: const'd.

src/protocols/rdp/guac_rdpdr/rdpdr_printer.h
necouchman: Documented.

src/protocols/rdp/compat/winpr-stream.h
necouchman: Removed.

src/protocols/rdp/guac_rdpdr/rdpdr_printer.c
necouchman: So, actually, the `device->device_name` is passed by reference from the `settings->printer_name` pointer, so if I try to free both `device->device_name` (in this function) and `settings->printer_name` (in the rdp_settings.c free function), I get a double free or corruption error.  So I've removed it out of here.

necouchman:
> This is beginning to look pretty great.

I'll take that as a compliment :smiley:.  Anything else need to be cleaned up, here?  I'll probably squash some of the commits together so there aren't quite so many of those, but otherwise should be getting close...

necouchman:
Okay, squashed it down to a more reasonable number of commits.

PR #155, opened 2018-02-23
GUACAMOLE-446: Allow redirected drive name to be configured

necouchman:
Allows for the name of the redirected drive to be configured when pushing a filesystem through via RDP.

Note that this change not only allows for the name to be configured, but also *removes UTF-16 encoding* of the drive name as it appears that Windows does not expect the filesystem name to be UTF-16 encoded.

### DIFF
--- a/src/protocols/rdp/guac_rdpdr/rdpdr_fs_service.h
+++ b/src/protocols/rdp/guac_rdpdr/rdpdr_fs_service.h
@@ -41,8 +41,14 @@
 /**
  * Registers a new filesystem device within the RDPDR plugin. This must be done
  * before RDPDR connection finishes.
+ * 
+ * @param rdpdr
+ *     The RDP device redirection plugin with which to register the device.
+ * 
+ * @param drive_name
+ *     The name of the redirected drive to display in the RDP connection.
  */
-void guac_rdpdr_register_fs(guac_rdpdrPlugin* rdpdr);
+void guac_rdpdr_register_fs(guac_rdpdrPlugin* rdpdr, char* drive_name);
 
 #endif
 
--- a/src/protocols/rdp/guac_rdpdr/rdpdr_fs_messages_vol_info.c
+++ b/src/protocols/rdp/guac_rdpdr/rdpdr_fs_messages_vol_info.c
@@ -25,6 +25,7 @@
 #include "rdp_status.h"
 
 #include <freerdp/utils/svc_plugin.h>
+#include <guacamole/unicode.h>
 
 #ifdef ENABLE_WINPR
 #include <winpr/stream.h>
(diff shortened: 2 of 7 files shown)


mike-jumper:
> Note that this change not only allows for the name to be configured, but also *removes UTF-16 encoding* of the drive name as it appears that Windows does not expect the filesystem name to be UTF-16 encoded.

This will break things. The RDP protocol requires that these names are encoded in UTF-16. For example, from [the RDPDR specification for the `DR_DEVICELIST_ANNOUNCE` message](https://msdn.microsoft.com/en-us/library/cc241357.aspx):

> **DeviceAnnounce (variable):** ... If the client supports DRIVE_CAPABILITY_VERSION_02 in the Drive Capability Set, then the full name MUST also be specified in the DeviceData field, as a null-terminated Unicode string. ...

Earlier in the spec, "Unicode string" is defined as being specifically UTF-16LE.

Not encoding things properly as UTF-16 is what caused the main problem in Amarjeet's modifications noted on the now-closed [GUACAMOLE-512](https://issues.apache.org/jira/browse/GUACAMOLE-512) issue, where his attempt to change the drive name to "Cloud Storage" was interpreted by the RDP server as "汃畯瑓牯条e".

Removing the UTF-16 violates spec, even if the RDP server is in some cases able to work regardless, and is definitely the cause of the above problem.

necouchman:
@mike-jumper So, I think we're good with having removed UTF-16 encoding here, right?  Anything else that needs to be changed on this one.

mike-jumper:
Yes, I believe so. From further investigation and from the discussion on the mailing lists, it seems clear that RDP does not use UTF-16 for the drive name (despite the spec and despite requiring UTF-16 for the other device names).

src/protocols/rdp/guac_rdpdr/rdpdr_fs_service.h
```diff
@@ -42,7 +42,7 @@
  * Registers a new filesystem device within the RDPDR plugin. This must be done
  * before RDPDR connection finishes.
  */
-void guac_rdpdr_register_fs(guac_rdpdrPlugin* rdpdr);
+void guac_rdpdr_register_fs(guac_rdpdrPlugin* rdpdr, char* drive_name);
```
mike-jumper: The documentation for this function should be brought up-to-date now that it's being updated. Other than that, I think these changes are pretty much good to go.

mike-jumper:
As mentioned on #154, I just noticed that the space available for the device announcement is hard-coded to 256 bytes. Now that device names are dynamic, we will need to somehow ensure that we will not exceed available space.

necouchman:
So, I've attempted to take care of dynamic allocation in #154 - if that PR gets merged ahead of this one it should be okay.

src/protocols/rdp/guac_rdpdr/rdpdr_fs_service.h
necouchman: Updated.

necouchman:
Base changed to staging/1.0.0

necouchman:
This will get rebased on top of the changes in #154 once that gets merged.

PR #156, opened 2018-02-26
GUACAMOLE-269: Allow backspace key behavior to be configured

necouchman:
Server-side changes that allow the backspace key to be configured, and, for the SSH protocol, also pass through that behavior to the SSH server via the TTY Mode Encoding options.  This replaces #153.


src/terminal/terminal.c
```diff
@@ -1594,7 +1599,7 @@ static int __guac_terminal_send_key(guac_terminal* term, int keysym, int pressed
         /* Non-printable keys */
         else {
 
-            if (keysym == 0xFF08) return guac_terminal_send_string(term, "\x7F"); /* Backspace */
+            if (keysym == 0xFF08) return guac_terminal_send_string(term, &term->backspace); /* Backspace */
```
mike-jumper: `guac_terminal_send_string()` requires a null-terminated string. As `&term->backspace` is a pointer to a single character, there is no guarantee of null terminator, and this may read without bound, resulting in connection closure or dumping of memory data to the remote terminal.

src/protocols/ssh/settings.h
```diff
@@ -223,6 +223,11 @@ typedef struct guac_ssh_settings {
      */
     int server_alive_interval;
 
+    /**
+     * The decismal ASCII code of the command to send for backspace.
```
mike-jumper: "decismal" is a typo, but that may be moot as this is technically incorrect. While the string from which this `int` was parsed may have been decimal, at this point it's just an `int`.

src/protocols/ssh/ttymode.c
```diff
@@ -0,0 +1,78 @@
...
+ */
+
+#include "config.h"
+#include "ttymode.h"
+
+#include <stdlib.h>
+#include <string.h>
+
+guac_ssh_ttymodes init_ttymodes() {
+    // Simple allocation for a placeholder
```
mike-jumper: Comments for C source within Guacamole should use `/* ... */` syntax, even if only a single line.

src/protocols/ssh/ttymode.c
```diff
@@ -0,0 +1,78 @@
...
+    guac_ssh_ttymode* ttymode_array = malloc(1);
+    
+    // Set up the initial data structure.
+    guac_ssh_ttymodes empty_modes = {
+        ttymode_array,
+        0
+    };
+
+    return empty_modes;
```
mike-jumper: Assigning/returning a struct by value like this actually does an implicit copy in C.

mike-jumper:
Before I continue with this review, do you think that `init_ttymodes()` / `add_ttymode()` / `sizeof_ttymodes()` / `ttymodes_to_array()` might be overkill for the problem at hand? There's a lot of complex logic and dynamic allocation happening there for what boils down to a single opcode.

necouchman:
> Before I continue with this review, do you think that init_ttymodes() / add_ttymode() / sizeof_ttymodes() / ttymodes_to_array() might be overkill for the problem at hand? There's a lot of complex logic and dynamic allocation happening there for what boils down to a single opcode.

You're probably right - it definitely felt a little overkill-ish while writing it; however, I was trying to both solve the problem of reading the new SSH setting into the array while also making it so that this could be expanded and reused easily in the future.  While this particular JIRA issue deals with the behavior of the backspace key, there are several other TTY Mode Encoding options specified in the SSH RFC, and I didn't want to make the assumption that this is the only one we'd ever be interested in implementing.

However, if you think that's a problem we should worry about when the second opcode comes along, that's fine with me - I can back off this and just write a function that generates the pointer array with the VERASE opcode and value and the ending opcode.  Let me know which route you think is best and I'll be happy to do it.

src/protocols/ssh/ttymode.c
necouchman: So, I take it this is bad?  Would it be preferable to pass by reference or something like that?

mike-jumper:
> You're probably right - it definitely felt a little overkill-ish while writing it; however, I was trying to both solve the problem of reading the new SSH setting into the array while also making it so that this could be expanded and reused easily in the future. ...

Future-proofing these changes is not a bad idea, but let's do so without complex dynamic allocation. In this case:

1. We know ahead of time the necessary size of the mode array.
2. The mode array itself needs to live only for the duration of that initial call which allocates the PTY.

For the sake of simplicity, I'd say we should really just use a statically-allocated array for now. If you feel that dynamic allocation would be more appropriate, then we should make use of our knowledge of the necessary size and allocate the necessary space ahead of time (without resizing) and be careful to free the allocated data once it is no longer needed.

necouchman:
@mike-jumper 
Well, I stuck with dynamic allocation, mainly because I was having a silly amount of trouble figuring out how to properly handle standard arrays in the data structure.  But, I'm allocating everything ahead of time with known sizes, so hopefully that addresses any concerns.  I'm happy to take other suggestions, though.

I also added some calls to free the data structures - I think I got everything, but worth double-checking to make sure.

mike-jumper:
Hm ... there has to be a cleaner way. The current implementation is:

1. An internal representation of the TTY modes array is dynamically allocated with `guac_ssh_ttymodes_init()`
2. The single TTY mode is added to that array dynamically with `guac_ssh_ttymodes_add()`
3. That internal representation is converted to the representation required by SSH through more dynamic allocation via `guac_ssh_ttymodes_size()` and `guac_ssh_ttymodes_to_array()`
4. All memory allocated above is manually freed via calls to `free()`.

I have a few problems with this. We are dynamically allocating an intermediate representation which really has no purpose other than to be converted into another dynamically-allocated representation, both of which have the same lifespan. Abstraction is a great thing, but the complexity of this abstraction seems to outweigh the intended benefit.

I suggest:

* Look for ways to avoid so much dynamic allocation. Perhaps there is a way to avoid the intermediate representation which ultimately gets thrown away?
* Look for ways to abstract things such that they (1) simplify your current implementation and (2) make future development easier. Perhaps the intermediate representation could be passed in directly without dynamic allocation? Might variadic functions achieve this better? Something like:

      size = guac_ssh_ttymodes_init(&some_static_buffer, sizeof(some_static_buffer),
                  GUAC_SSH_TTY_OP_VERASE, settings->backspace);

  would abstract things away, require only a single function call, and would not depend on dynamic allocation.

necouchman:
> Hm ... there has to be a cleaner way.

Oh, I have no doubt at all there is a cleaner way! :-)

> Perhaps there is a way to avoid the intermediate representation which ultimately gets thrown away?

Yeah, I'll see what I can do about this.  I was initially trying to create a data structure that could be cast into a `char *` and passed directly to the `libssh2_channel_request_pty_ex()` function.  I thought I remembered from somewhere in my past that doing this:

    struct ttymode {
        char opcode;
        uint32_t value;
    } ttymode;

Would result in memory being allocated in such a way that it could then be cast into `char *` and you'd end up with the single opcode byte and the four value bytes all together, which is what the function expects.  Doesn't seem to work out, though.  If that is supposed to work, that seems like the simplest route to be, because the members can then be accessed easily (array[0].opcode and array[0].value) but then the entire thing can be passed directly to the function.  Am I remembering that correctly, or is there something else I'm not factoring in to how that allocation works?

mike-jumper:
Unfortunately, that isn't safe across different architectures.

A C `struct` provides a guarantee regarding _order_ in memory (`opcode` will definitely come before `value`), but it does not provide guarantees regarding _alignment_ (there may be padding before or after those members). Some systems may have alignment requirements that mean `opcode` will not be directly adjacent to `value` in memory.

There are also no guarantees regarding byte order for things like a `uint32_t`. Mathematical operations will work transparently, but the byte order used by the architecture for 32-bit integers may differ from the byte order required by the SSH server.

necouchman:
Okay, I've removed all of the dynamic allocation and most of the functions, aside from init.  I'm using a variadic function, with a custom data structure to simplify counting and passing arguments through to that function.  Let me know how this looks.

As usual, Mike, thanks for the patience and help working through this...

necouchman:
@mike-jumper Any other changes needed to wrap this one up?

src/protocols/ssh/ttymode.c
```diff
@@ -0,0 +1,59 @@
...
+       We need one extra byte at the end for the ending opcode. */
+    if ((num_opcodes * GUAC_SSH_TTY_OPCODE_SIZE) >= (array_size))
+        return 1;
+
+    for (int i = 0; i < num_opcodes; i++) {
+        /* Calculate offset in array */
+        int offset = i * GUAC_SSH_TTY_OPCODE_SIZE;
+
+        /* Get the next argument to this function */
+        guac_ssh_ttymode ttymode = va_arg(args, guac_ssh_ttymode);
```
mike-jumper: The structure in this case is rather small, so this does not make much of a difference in this case, but beware that passing structures by value will result in a full copy of that structure, with this assignment resulting in another full copy.

src/protocols/ssh/ttymode.c
```diff
@@ -0,0 +1,59 @@
...
+
+        /* Place opcode and value in array */
+        opcode_array[offset] = ttymode.opcode;
+        opcode_array[offset + 1] = (ttymode.value >> 24) & 0xFF;
```
mike-jumper: You don't need to manually calculate the offset for each byte written. If you have a pointer to the next byte to be written within the array, this could be as simple as:

    *(current++) = whatever_you_want_to_write;

src/protocols/ssh/ttymode.c
```diff
@@ -0,0 +1,59 @@
...
+        opcode_array[offset + 2] = (ttymode.value >> 16) & 0xFF;
+        opcode_array[offset + 3] = (ttymode.value >> 8) & 0xFF;
+        opcode_array[offset + 4] = ttymode.value & 0xFF;
+    }
+
+    /* Put the end opcode in the last opcode space */
+    opcode_array[num_opcodes * GUAC_SSH_TTY_OPCODE_SIZE] = GUAC_SSH_TTY_OP_END;
```
mike-jumper: To confirm: the final "END" does not need the full 5-byte opcode and value? Just the opcode?

necouchman: That's correct - as soon as it encounters the 0 opcode in one of the expected positions (5 byte offsets, basically) it stops.

src/protocols/ssh/ttymode.c
necouchman: Is it possible to do this by pointer/reference rather than value?  I'm not familiar with the va_arg macro and how to use it - do I just change the argument to `&guac_ssh_ttymode` and the assignment to `guac_ssh_ttymode* ttymode`?

src/protocols/ssh/ttymode.c
```diff
@@ -0,0 +1,59 @@
...
+
+    return 0;
```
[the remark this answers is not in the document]
necouchman: Documentation has been fixed.

src/protocols/ssh/ttymode.h
```diff
@@ -0,0 +1,80 @@
...
+
+/**
+ * The SSH TTY mode encoding opcode that configures
+ * the TTY erase code to configure the server
+ * backspace key.
+ */
+#define GUAC_SSH_TTY_OP_VERASE 3
+
+/**
+ * Simple structure that defines a opcode and
```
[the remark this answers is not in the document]
necouchman: Took a stab at some more thorough documentation.

src/protocols/ssh/ttymode.h
```diff
@@ -0,0 +1,80 @@
...
+ * value pair.
+ */
+typedef struct guac_ssh_ttymode {
+    char opcode;
+    uint32_t value;
```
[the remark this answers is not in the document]
necouchman: Documented.

src/protocols/ssh/ttymode.h
```diff
@@ -0,0 +1,80 @@
...
+} guac_ssh_ttymode;
+
+/**
+ * Initialize an array with the specified opcode/value
```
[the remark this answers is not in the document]
necouchman: Beefed up this documentation.

src/terminal/terminal.c
```diff
@@ -1594,7 +1598,13 @@ static int __guac_terminal_send_key(guac_terminal* term, int keysym, int pressed
         /* Non-printable keys */
         else {
 
-            if (keysym == 0xFF08) return guac_terminal_send_string(term, "\x7F"); /* Backspace */
+            /* Backspace can vary based on configuration of terminal by client. */
+            if (keysym == 0xFF08) {
+                char* backspace_str = malloc(sizeof(char) * 2);
```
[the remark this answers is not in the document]
necouchman: Switched to static.

src/terminal/terminal.c
```diff
@@ -1594,7 +1598,13 @@ static int __guac_terminal_send_key(guac_terminal* term, int keysym, int pressed
...
+                backspace_str[0] = term->backspace;
+                backspace_str[1] = '\0';
+                return guac_terminal_send_string(term, backspace_str);
```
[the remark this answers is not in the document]
necouchman: Staticized.

src/protocols/telnet/settings.c
```diff
@@ -328,6 +335,11 @@ guac_telnet_settings* guac_telnet_parse_args(guac_user* user,
         guac_user_parse_args_boolean(user, GUAC_TELNET_CLIENT_ARGS, argv,
                 IDX_CREATE_RECORDING_PATH, false);
 
+    /* Parse backspae key code */
```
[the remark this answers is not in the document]
necouchman: Speling corectid.

src/protocols/ssh/ttymode.h
```diff
@@ -0,0 +1,80 @@
...
+ * pairs, and return the size, in bytes, of the array.
```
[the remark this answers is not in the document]
necouchman: Documentation updated.

src/protocols/ssh/ttymode.c
necouchman: Okay, took a stab at this, and I *think* I got it working correctly - at least, the code runs and behaves as expected.  Might be worth someone looking it over, though, just to make sure I did it in a way that isn't going to hit some corner case at some point down the road.

src/protocols/ssh/ttymode.c
necouchman: I think I got this going, too - again, probably worth someone taking a look just to make sure.  Seems to work, though.

necouchman:
Okay, @mike-jumper, next round of updates done - let me know what else needs attention!

src/protocols/ssh/ttymode.c
```diff
@@ -0,0 +1,58 @@
...
+        *(current++) = (ttymode->value >> 24) & 0xFF;
+        *(current++) = (ttymode->value >> 16) & 0xFF;
+        *(current++) = (ttymode->value >> 8) & 0xFF;
+        *(current++) = ttymode->value & 0xFF;
+    }
+
+    /* Put the end opcode in the last opcode space */
+    *(current) = GUAC_SSH_TTY_OP_END;
+
+    return 0;
```
mike-jumper: I recommend instead returning the size of the data written to the array, with zero indicating failure. This would remove the need for the caller to know ahead of time the size of the data this function will write (which is an odd requirement for what is otherwise an abstraction).

src/terminal/terminal.c
```diff
@@ -1594,7 +1598,13 @@ static int __guac_terminal_send_key(guac_terminal* term, int keysym, int pressed
         /* Non-printable keys */
         else {
 
-            if (keysym == 0xFF08) return guac_terminal_send_string(term, "\x7F"); /* Backspace */
+            /* Backspace can vary based on configuration of terminal by client. */
+            if (keysym == 0xFF08) {
+                char backspace_str[2];
+                backspace_str[0] = term->backspace;
+                backspace_str[1] = '\0';
```
mike-jumper: You can clean this up as:

    char backspace_str[2] = { term->backspace, '\0' };

or even:

    char backspace_str[] = { term->backspace, '\0' };

src/protocols/ssh/ssh.c
```diff
@@ -296,9 +301,15 @@ void* ssh_client_thread(void* data) {
 
     }
 
+    /* Set up the ttymode array prior to requesting the PTY */
+    if (guac_ssh_ttymodes_init(ssh_ttymodes, sizeof(ssh_ttymodes),
+            num_tty_opcodes, (guac_ssh_ttymode){ GUAC_SSH_TTY_OP_VERASE, settings->backspace}))
+        guac_client_abort(client, GUAC_PROTOCOL_STATUS_SERVER_ERROR, "Error configuring TTY mode encoding.");
```
mike-jumper: What do you mean by "configuring TTY mode encoding"? Is there a way to phrase this which would better represent what is failing?

src/protocols/ssh/ttymode.c
```diff
@@ -0,0 +1,58 @@
...
+#include <stdarg.h>
+#include <stdlib.h>
+#include <string.h>
+
+int guac_ssh_ttymodes_init(char opcode_array[], const int array_size,
+        const int num_opcodes, ...) {
+
+    /* Initialize the variable argument list. */
+    va_list args;
+    va_start(args, num_opcodes);
```
mike-jumper: There must be a matching call to `va_end()` before the end of this function.

src/protocols/ssh/ssh.c
```diff
@@ -296,9 +301,15 @@ void* ssh_client_thread(void* data) {
 
     }
 
+    /* Set up the ttymode array prior to requesting the PTY */
+    if (guac_ssh_ttymodes_init(ssh_ttymodes, sizeof(ssh_ttymodes),
+            num_tty_opcodes, (guac_ssh_ttymode){ GUAC_SSH_TTY_OP_VERASE, settings->backspace}))
```
mike-jumper: Looking at the source for `guac_ssh_ttymodes_init()`, doesn't this need to be a pointer to a `guac_ssh_ttymode`? ie: `&((guac_ssh_ttymode) { GUAC_SSH_TTY_OP_VERASE, settings->backspace })` ?

Alternatively, there's no need for all these parameters to be of the same type. It would also be possible to do something like:

    if (guac_ssh_ttymodes_init(ssh_ttymodes, sizeof(ssh_ttymodes),
            GUAC_SSH_TTY_OP_VERASE, settings->backspace, GUAC_SSH_TTY_OP_END))

The code is your oyster.

src/protocols/ssh/ssh.c
```diff
@@ -191,6 +192,10 @@ void* ssh_client_thread(void* data) {
         return NULL;
     }
 
+    /* Initialize a ttymode array */
+    const int num_tty_opcodes = 1;
+    char ssh_ttymodes[(GUAC_SSH_TTY_OPCODE_SIZE * num_tty_opcodes) + 1];
```
mike-jumper: Not really a dealbreaker, but if you want to ensure this is a constant expression while also abstracting away the calculations required, this could be handled via a macro defined in `ttymode.h`. For example:

    #define GUAC_SSH_TTYMODES_SIZE(num_opcodes) ((GUAC_SSH_TTY_OPCODE_SIZE * num_opcodes) + 1)

You could then rely upon:

    char ssh_ttymodes[GUAC_SSH_TTYMODES_SIZE(1)];

... and, of course, documenting within `guac_ssh_ttymodes_init()` how the minimum size for the provided buffer should be calculated.

Magic.

src/protocols/ssh/ssh.c
```diff
@@ -191,6 +192,10 @@ void* ssh_client_thread(void* data) {
         return NULL;
     }
 
+    /* Initialize a ttymode array */
```
mike-jumper: Point of clarification - nothing is being initialized here, only declared. With that much being obvious (that the array is being declared), rather than change "Initialize" to "Declare", I'd say just kill this line.

src/terminal/terminal.c
necouchman: Of course.  Done & thanks.

src/protocols/ssh/ttymode.c
necouchman: Sounds good - is there a more elegant way than just adding an integer byte counter and incrementing along the way?

src/protocols/ssh/ttymode.c
necouchman: Added.

src/protocols/ssh/ttymode.c
necouchman: I've gone the byte counter route - if you have a suggestion for another way to do it, let me know.

src/protocols/ssh/ssh.c
necouchman: Tried to clean up this error message bit.

src/protocols/ssh/ssh.c
necouchman: Refactored the `guac_ssh_ttymodes_init()` function.  I may have introduced some other bugs or conditions that need to be handled, so let me know what you think about it.

src/protocols/ssh/ssh.c
```diff
@@ -296,9 +299,17 @@ void* ssh_client_thread(void* data) {
 
     }
 
+    /* Set up the ttymode array prior to requesting the PTY */
+    int ttymodeBytes = guac_ssh_ttymodes_init(ssh_ttymodes,
+            GUAC_SSH_TTY_OP_VERASE, settings->backspace, GUAC_SSH_TTY_OP_END);
+    if (ttymodeBytes < 1)
+        guac_client_abort(client, GUAC_PROTOCOL_STATUS_SERVER_ERROR, "Error storing TTY mode encoding \
```
mike-jumper: This will actually log:

    Error storing TTY mode encoding                 opcodes and values in array.

as the whitespace at the beginning of the following line is part of the string.

To split strings in C, all you need to do is place two strings next to each other. They're automatically concatenated by the compiler:

    guac_client_abort(client, GUAC_PROTOCOL_STATUS_SERVER_ERROR, "Error storing TTY mode encoding "
            "opcodes and values in array.");

As this message is meant for the user/admin, I suggest rephrasing to describe things from that perspective. For example:

    Unable to set TTY modes. Backspace may not work as expected.

at the warning level may make more sense.

src/protocols/ssh/ttymode.c
```diff
@@ -0,0 +1,64 @@
...
+        *(current++) = (value >> 16) & 0xFF;
+        *(current++) = (value >> 8) & 0xFF;
+        *(current++) = value & 0xFF;
+        bytes += sizeof(uint32_t);
+    }
+
+    /* We're done processing arguments. */
+    va_end(args);
+
+    return bytes;
```
mike-jumper: FYI - you can actually do:

    current - opcode_array

src/protocols/ssh/ttymode.h
```diff
@@ -0,0 +1,81 @@
...
+ */
+#define GUAC_SSH_TTY_OP_VERASE 3
+
+/**
+ * Macro for calculating the number of bytes required
+ * to pass a given number of opcodes, which calculates
+ * the size of the number of opcodes plus the single byte
+ * end opcode.
+ */
+#define GUAC_SSH_TTYMODES_SIZE(num_opcodes) ((GUAC_SSH_TTY_OPCODE_SIZE * num_opcodes) + 1)
```
mike-jumper: Please document `num_opcodes` with `@param`, just as you would for a function.

src/protocols/ssh/ssh.c
necouchman: Changed.

necouchman: Changed to warning.

src/protocols/ssh/ttymode.c
necouchman: Yes, of course.

src/protocols/ssh/ttymode.h
necouchman: Documented that, and also documented the return value with `@returns` -  not sure if that should be done, too?

necouchman:
Next round done.

src/protocols/ssh/ttymode.h
mike-jumper: Yup, definitely. Good catch.

PR #161, opened 2018-03-29
GUACAMOLE-42: Add support for specifying log level to guacd in docker

necouchman:
Small changes to allow the guacd log level to be passed through as an environment variable.  Not entirely sure this is the right way to go - I know the guacamole-client code for taking and parsing variables is significantly more complex.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -82,6 +82,7 @@ ARG PREFIX_DIR=/usr/local/guacamole
 
 # Runtime environment
 ENV LC_ALL=en_US.UTF-8
+ENV GUACD_LOG_LEVEL=info
 
 ARG RUNTIME_DEPENDENCIES="            \
         cairo                         \
@@ -126,5 +127,5 @@ EXPOSE 4822
 # Note the path here MUST correspond to the value specified in the 
 # PREFIX_DIR build argument.
 #
-CMD [ "/usr/local/guacamole/sbin/guacd", "-b", "0.0.0.0", "-f" ]
+CMD [ "/usr/local/guacamole/sbin/guacd", "-b", "0.0.0.0", "-L", "$GUACD_LOG_LEVEL", "-f" ]
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -127,5 +127,5 @@ EXPOSE 4822
 # Note the path here MUST correspond to the value specified in the 
 # PREFIX_DIR build argument.
 #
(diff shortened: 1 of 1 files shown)


mike-jumper:
> Not entirely sure this is the right way to go - I know the guacamole-client code for taking and parsing variables is significantly more complex.

It is true that the Docker documentation specifies that the shell form of `CMD` is not the preferred way, but I'm not sure that applies in this case. The alternative would be to move this into a shell script similar to the guacamole image, which might be overkill and would still result in a shell instance.

Assuming this does work as intended, I have no technical objections to the approach.

mike-jumper:
Looking a bit further, there is a difference between the way the shells are handled between these changes and the guacamole image which may worth taking into account:

https://github.com/apache/guacamole-client/blob/4f27a03adfe54d18772c81553c730f5cdc9115e4/guacamole-docker/bin/start.sh#L336

That `exec` will ensure that there isn't an unnecessary instance of the shell hanging around while the only process we care about continues to run as its child process. Perhaps we should do similar here? We don't necessarily need a full-fledged script - adding an `exec` to the existing `CMD` may make things equivalent.

necouchman:
> Assuming this does work as intended, I have no technical objections to the approach.

I did test it and it seems to work as intended.

> We don't necessarily need a full-fledged script - adding an exec to the existing CMD may make things equivalent.

I'll give that a shot and see what happens.  So, sounds like I should be looking for some instance of either sh or bash getting left around as a parent process to guacd?

mike-jumper:
> So, sounds like I should be looking for some instance of either sh or bash getting left around as a parent process to guacd?

In its current state, yes, I believe so, unless Docker magically adds `exec` on its own.

necouchman:
Looks like it may do it automagically:

    [root@guachost1 ~]# ps -ef|grep 9305
    root      9305 55586  0 13:44 ?        00:00:00 docker-containerd-shim -namespace moby -workdir /var/lib/docker/containerd/daemon/io.containerd.runtime.v1.linux/moby/b65c98beca10cf4df6fddd898677a1cbb7ae15cb3d6c679ae03795bef05238b8 -address /var/run/docker/containerd/docker-containerd.sock -containerd-binary /usr/bin/docker-containerd -runtime-root /var/run/docker/runtime-runc
    root      9321  9305  0 13:44 ?        00:00:00 /usr/local/guacamole/sbin/guacd -b 0.0.0.0 -L debug -f

That's all I see, with the 55586 parent process as the docker-containerd.  No extra shells, as far as I can tell.

mike-jumper:
Alrighty. Good enough for me.

PR #162, opened 2018-04-02
GUACAMOLE-407: Refactor guacd Docker image to use "debian:stable" base.

mike-jumper:
These changes migrate from a CentOS-based image to a Debian-based one, updating dependencies accordingly.

Part of the idea behind this migration is to allow the Docker image to be easily rebuilt to take advantage of upstream updates, thus rather than hard-code runtime dependencies, the image now dynamically determines the packages required by listing the packages which provide the libraries that end up linked.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -22,9 +22,9 @@
 #
 
 
-# Use CentOS as base for the build
-ARG CENTOS_VERSION=centos7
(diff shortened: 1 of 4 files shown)


necouchman:
No complaints from anyone else, so pushing forward with the merge.

PR #163, opened 2018-04-02
GUACAMOLE-533: Wait at most 5 seconds for connection processes to terminate following disconnect.

mike-jumper:
This change refactors guacd such that `guac_client_free()` is invoked in a dedicated thread, while guacd waits for the free/disconnect operation to signal completion. If free/disconnect does not complete in a timely manner, guacd takes control, forcibly kills the connection process, and logs a warning.

In addition, even if the `guac_client` is freed successfully and within a reasonable time, guacd now verifies that all child processes of that connection have also terminated. If any child processes remain, they are forcibly terminated by guacd, and a warning is logged.

### DIFF
--- a/src/guacd/proc.c
+++ b/src/guacd/proc.c
@@ -33,11 +33,15 @@
 #include <guacamole/user.h>
 
 #include <errno.h>
+#include <pthread.h>
+#include <signal.h>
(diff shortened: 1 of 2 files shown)


flangelo:
Looks good to me.

PR #164, opened 2018-04-06
GUACAMOLE-527: Implement SSH host key checking against known hosts

necouchman:
This series of commits implements changes to guacd necessary to perform checks of the SSH host keys.

Aside from some specific comments on some of the code, it's worth noting that this is a fairly significant change in behavior, as connections will fail, by default, if a matching host key is not found.  I believe this is the preferred operation, but current users will need to be aware of the significance of this change.


necouchman:
@mike-jumper @ceharris @jmuehlner Any comments on this?

src/common-ssh/ssh.c
```diff
@@ -518,6 +520,64 @@ guac_common_ssh_session* guac_common_ssh_create_session(guac_client* client,
...
+
+    libssh2_knownhost_free(ssh_known_hosts);
+
+    switch (kh_check) {
+        case LIBSSH2_KNOWNHOST_CHECK_MATCH:
+            guac_client_log(client, GUAC_LOG_DEBUG,
+                "Host key match found for %s", hostname);
+            break;
+        case LIBSSH2_KNOWNHOST_CHECK_NOTFOUND:
+            guac_client_abort(client, GUAC_PROTOCOL_STATUS_SERVER_ERROR,
```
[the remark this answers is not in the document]
mike-jumper: To the extent that the failure to validate the host is due to misconfiguration, `GUAC_PROTOCOL_STATUS_SERVER_ERROR` is probably correct. That is the status code we use elsewhere when an error with connection parameters is encountered.

Other options for the case where the host key doesn't match could be:

* `GUAC_PROTOCOL_STATUS_UPSTREAM_ERROR` (the documentation states "The operation was unsuccessful due to an error or otherwise unexpected condition of the upstream server", and this certainly seems an unexpected condition of the upstream server).
* `GUAC_PROTOCOL_STATUS_UPSTREAM_NOT_FOUND` (we were not able to find the expected remote desktop server - we found some other server instead)

src/protocols/rdp/rdp_settings.c
```diff
@@ -822,6 +838,32 @@ guac_rdp_settings* guac_rdp_parse_args(guac_user* user,
...
+        char* str_host_key_type = guac_user_parse_args_string(user, GUAC_RDP_CLIENT_ARGS, argv,
+                    IDX_SFTP_HOST_KEY_TYPE, "ssh-rsa");
+        
+        if (strcmp(str_host_key_type, "ssh-rsa") == 0)
+            settings->sftp_host_key_type = LIBSSH2_KNOWNHOST_KEY_SSHRSA;
+        else if (strcmp(str_host_key_type, "ssh-dss") == 0)
+            settings->sftp_host_key_type = LIBSSH2_KNOWNHOST_KEY_SSHDSS;
+        else if (strcmp(str_host_key_type, "rsa1") == 0)
+            settings->sftp_host_key_type = LIBSSH2_KNOWNHOST_KEY_RSA1;
+        else {
```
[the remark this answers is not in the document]
mike-jumper: Leveraging [`libssh2_knownhost_readline()`](https://www.libssh2.org/libssh2_knownhost_readline.html) might be a better alternative.

src/common-ssh/ssh.c
```diff
@@ -518,6 +520,64 @@ guac_common_ssh_session* guac_common_ssh_create_session(guac_client* client,
         return NULL;
     }
 
+    /* Check known_hosts, start by getting known_hosts file of user running guacd */
+    struct passwd *pw = getpwuid(getuid());
+    const char *known_hosts = strcat(pw->pw_dir, "/.ssh/known_hosts");
+    LIBSSH2_KNOWNHOSTS *ssh_known_hosts = libssh2_knownhost_init(session);
+    libssh2_knownhost_readfile(ssh_known_hosts, known_hosts, LIBSSH2_KNOWNHOST_FILE_OPENSSH);
+
```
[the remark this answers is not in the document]
necouchman: I agree that this is a departure from backward-compatibility, but I have two arguments for why I think it should be done that way, anyway:
- It makes Guacamole behave more like other SSH clients.  Most SSH clients require that you at least verify the host key on initial connection and agree to add it to the known_hosts file, and throw an error if it doesn't match.  This gets closer to that, though without prompting within the client, we can't ask if it doesn't initially exist.
- It makes SSH support in Guacamole behave more like RDP.  In RDP, you have to explicitly tell Guacamole to ignore server certificates, else it fails the connection unless you have correctly configured SSL on the system running guacd such that the certificates are trusted.  So, the behavior is more consistent within itself.

One potential compromise might be this: if the host key does not already exist in the known_hosts file, add it and allow the connection to continue instead of failing.  If the host key exists but does not match, fail.

Thoughts?  I definitely understand the importance of maintaining backward-compatibility; however, this seems like a case where perhaps breaking that would be acceptable, in some form or another.

src/common-ssh/ssh.c
```diff
@@ -414,7 +415,8 @@ static int guac_common_ssh_authenticate(guac_common_ssh_session* common_session)
 }
 
 guac_common_ssh_session* guac_common_ssh_create_session(guac_client* client,
-        const char* hostname, const char* port, guac_common_ssh_user* user, int keepalive) {
+        const char* hostname, const char* port, guac_common_ssh_user* user, int keepalive,
+        const int host_key_type, const char* host_key) {
```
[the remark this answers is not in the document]
necouchman: Yeah, probably a better option.

src/common-ssh/ssh.c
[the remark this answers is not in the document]
necouchman: Okay, I've gone with the approach that known_hosts of the user running guacd will be read and checked against (in addition to any provided host key in the configuration parameters).  If there is a mismatch, the connection will fail.  If there is no match, guacd will warn but continue the connection.  Is that acceptable behavior?  Or would you still prefer an explicit enable parameter for checking the known_hosts file at all?

src/common-ssh/ssh.c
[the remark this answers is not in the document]
necouchman: I'll just leave it as it is.

src/protocols/rdp/rdp_settings.c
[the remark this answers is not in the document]
necouchman: Good call - switched over to this, which makes configuration a little more straight-forward.

necouchman:
@mike-jumper Changed as requested, both here and on the guacamole-client code.  Let me know if anything else needs to be tweaked, particularly with regard to the known_hosts file check.

src/common-ssh/ssh.c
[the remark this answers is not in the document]
mike-jumper: > It makes Guacamole behave more like other SSH clients. Most SSH clients require that you at least verify the host key on initial connection and agree to add it to the known_hosts file, and throw an error if it doesn't match.

Closer, yes, but breaking SSH functionality for all existing users of Guacamole would be bad. The result of upgrading Guacamole from one version to another shouldn't be that all existing SSH connections (and VNC/RDP connections using SFTP) need to be reconfigured.

> It makes SSH support in Guacamole behave more like RDP.

I don't necessarily think that the built-in host checking of the FreeRDP library is the gold standard here, partly because in our case it universally requires that the keys be stored on the server running guacd. In fact, we should probably look toward implementing our own host checking within RDP, overriding FreeRDP's default behavior if a host key is specified, so that ignoring the certificate becomes unnecessary.

> One potential compromise might be this: if the host key does not already exist in the known_hosts file, add it and allow the connection to continue instead of failing.

I believe that would be a security vulnerability. The protections added through host checking would be able to be bypassed by a malicious user if they happen to use a guacd node which does not yet know about the SSH host in question.

> Thoughts? I definitely understand the importance of maintaining backward-compatibility; however, this seems like a case where perhaps breaking that would be acceptable, in some form or another.

My thoughts are:

* Existing VNC/RDP/SSH connections shouldn't break as a result of these changes
* If host checking is enabled (through providing a `known_hosts` file or through specifying the host key as a connection parameter), then strict host checking should be performed, with the connection being aborted if the host key is unknown or does not match.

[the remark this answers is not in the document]
mike-jumper: > Okay, I've gone with the approach that known_hosts of the user running guacd will be read and checked against (in addition to any provided host key in the configuration parameters).

Any reason behind checking against two sources? If a host key is specified via the connection parameters, shouldn't that override a local `known_hosts` entirely?

> Or would you still prefer an explicit enable parameter for checking the known_hosts file at all?

What about only checking local `known_hosts` if (1) the host key parameter is not provided and (2) the file exists?

src/common-ssh/ssh.c
```diff
@@ -518,6 +520,63 @@ guac_common_ssh_session* guac_common_ssh_create_session(guac_client* client,
         return NULL;
     }
 
+    /* Check known_hosts, start by getting known_hosts file of user running guacd */
+    struct passwd *pw = getpwuid(getuid());
+    const char *known_hosts = strcat(pw->pw_dir, "/.ssh/known_hosts");
```
mike-jumper: According to the documentation of `strcat()`:

> ... The strings may not overlap, and the `dest` string must have enough space for the  result.   If  `dest`  is  not large  enough, program behavior is unpredictable; buffer overruns are a favorite avenue for attacking secure programs.

Unless we know that `pw->pw_dir` will have sufficient space for the additional "/.ssh/known_hosts", this may not be safe as-written.

What about using a location specific to guacd like `/etc/guacamole/ssh/known_hosts`? That would avoid inheriting known hosts from the user running the process, and would effectively require explicit steps to enable the checking.

src/common-ssh/ssh.c
```diff
@@ -518,6 +520,63 @@ guac_common_ssh_session* guac_common_ssh_create_session(guac_client* client,
...
+    LIBSSH2_KNOWNHOSTS *ssh_known_hosts = libssh2_knownhost_init(session);
+    libssh2_knownhost_readfile(ssh_known_hosts, known_hosts, LIBSSH2_KNOWNHOST_FILE_OPENSSH);
+
+    /* Add host key provided from settings */
+    if (host_key && strcmp(host_key, "") > 0) {
```
mike-jumper: When testing for inequality, `strcmp(...) > 0` is odd. Did you mean `strcmp(...) != 0`?

src/common-ssh/ssh.c
necouchman: > What about using a location specific to guacd like /etc/guacamole/ssh/known_hosts?

I'm willing to go that route.  Will rewrite it as such.

src/common-ssh/ssh.c
necouchman: Yeah, not sure how I arrived at that.  Will change it.

src/common-ssh/ssh.c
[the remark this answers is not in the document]
necouchman: > Any reason behind checking against two sources? If a host key is specified via the connection parameters, shouldn't that override a local known_hosts entirely?
> What about only checking local known_hosts if (1) the host key parameter is not provided and (2) the file exists?

Sounds good to me...will make it so.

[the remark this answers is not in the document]
necouchman: Okay, I've implemented it as suggested and verified that it works.  The code seems a little clunky to me, but let me know what you think.  I was trying to avoid duplicating a lot of code checking it, but maybe the host key checking needs to be split into its own function?

src/common-ssh/ssh.c
necouchman: Okay, implemented as suggested.  Again, this might be a little clunky - let me know if you think it's worth moving this to its own function - maybe even in the key.h/key.c files within the SSH code?

src/common-ssh/ssh.c
necouchman: Fixed.

necouchman:
Okay @mike-jumper, switched it up a little bit per your suggestions.  The code for checking the host keys feels a little clunky (was trying to avoid duplicating code) and like it might belong in its own function, maybe even in the `key.h`/`key.c` files?  I'll be happy to move that out if you think that makes more sense than having it all remain in this single SSH session creation function.  On the other hand, I don't really expect it to be used in many (any) other places, so it isn't as if there's a lot gained by pulling it out, aside from readability.

src/common-ssh/ssh.c
```diff
@@ -518,6 +520,71 @@ guac_common_ssh_session* guac_common_ssh_create_session(guac_client* client,
...
+        int fp_type;
+        const char *fingerprint = libssh2_session_hostkey(session, &fp_len, &fp_type);
+
+        if (!fingerprint)
+            guac_client_abort(client, GUAC_PROTOCOL_STATUS_SERVER_ERROR,
+                    "Failed to get fingerprint for host %s", hostname);
+
+        /* Check fingerprint against known hosts */
+        struct libssh2_knownhost *host;
+        int kh_check = libssh2_knownhost_checkp(ssh_known_hosts, hostname, atoi(port),
```
mike-jumper: What happens here if `fingerprint` is `NULL`, as checked earlier?

src/common-ssh/ssh.c
```diff
@@ -518,6 +520,71 @@ guac_common_ssh_session* guac_common_ssh_create_session(guac_client* client,
...
+    /* Add host key provided from settings */
+    if (host_key && strcmp(host_key, "") != 0) {
+
+        int kh_add = libssh2_knownhost_readline(ssh_known_hosts, host_key, strlen(host_key),
+                LIBSSH2_KNOWNHOST_FILE_OPENSSH);
+        num_known_hosts++;
+
+        if (kh_add)
+            guac_client_log(client, GUAC_LOG_WARNING, "Failed to add provided host key"
+                    " to known hosts store for %s.  Error was %d", hostname, kh_add);
```
mike-jumper: Is there any way to get a human-readable error? The integer value of `kh_add` isn't likely to have any meaning to the user reading the logs, let alone to us when reported as a bug.

mike-jumper:
Readability may be reason enough to split things out. If the logic is divided into less complex, easier to verify functions, opportunities to de-clunk things may present themselves.

src/common-ssh/ssh.c
necouchman: The check for `fingerprint` aborts the connection if it is null, so this should never be reached in that case, correct?  The thought is that, if the fingerprint of the server cannot be calculated, something else worse than just a host key check has gone wrong and we should bail out entirely.

necouchman:
> Readability may be reason enough to split things out.

Sounds good - I'll put this/these function(s) in the `key.c` file and see where it goes from there.

src/common-ssh/ssh.c
mike-jumper: The connection will be aborted, yes, but `guac_client_abort()` does not stop the remainder of the function from executing. It's a combination of `guac_client_stop()`, `guac_client_log()`, and `guac_protocol_send_error()`, flagging the `guac_client` as stopped (which eventually terminates the loops involved), reporting the error to the client, and logging an error to guacd's logs.

It's a cooperative signal, coupled with logging:

https://github.com/apache/guacamole-server/blob/master/src/libguac/guacamole/client.h#L314-L317

necouchman: Got it.  Returning NULL now with the aborts in those areas.

necouchman:
Okay, have moved the host key verification into the `key.h`/`key.c` files.  I originally tried to split it into two functions, one to load the known_hosts, either from argument or from the file, and the other to do the verification, but I'm not sure it really added anything to the code, so I just combined them into a single function.  Let me know how that looks.

necouchman:
Okay, a couple more minor tweaks and I think this is ready to review, again.

necouchman:
Base changed to staging/1.0.0.

necouchman:
@mike-jumper Ping...

src/common-ssh/key.c
```diff
@@ -245,3 +246,84 @@ int guac_common_ssh_key_sign(guac_common_ssh_key* key, const char* data,
...
+        if (access(guac_known_hosts, F_OK) != -1)
+            known_hosts = libssh2_knownhost_readfile(ssh_known_hosts, guac_known_hosts, LIBSSH2_KNOWNHOST_FILE_OPENSSH);
+
+    }
+
+    /* If there's an error provided, abort connection and return that. */
+    if (known_hosts < 0) {
+
+        guac_client_log(client, GUAC_LOG_ERROR,
+            "Failure trying to load SSH host keys.");
```
mike-jumper: Is there any way to log what the specific failure is?

necouchman: I'll have a look - I believe there is a function in the libssh2 API for translating the error to text.  I think I actually had it implemented that way at one point, but it must have gotten lost in one of my changes.

necouchman: Okay, updated it and tested out some errors, and it looks like it shows the correct error message.

src/common-ssh/ssh.c
```diff
@@ -518,6 +520,40 @@ guac_common_ssh_session* guac_common_ssh_create_session(guac_client* client,
...
+    /* Get fingerprint of host we're connecting to */
+    size_t fp_len;
+    int fp_type;
+    const char *fingerprint = libssh2_session_hostkey(session, &fp_len, &fp_type);
+
+    /* Failure to generate a fingerprint means we should abort */
+    if (!fingerprint) {
+        guac_client_abort(client, GUAC_PROTOCOL_STATUS_SERVER_ERROR,
+            "Failed to get fingerprint for host %s", hostname);
+        return NULL;
```
mike-jumper: This will leak `common_session` and `fd`, allocated earlier in this function. For example, see:

https://github.com/apache/guacamole-server/blob/c120aa0274ee2d11465c80e1d3052c725db44a17/src/common-ssh/ssh.c#L512-L519

src/common-ssh/ssh.c
```diff
@@ -518,6 +520,40 @@ guac_common_ssh_session* guac_common_ssh_create_session(guac_client* client,
...
+
+        if (known_host_check < 0)
+            guac_client_abort(client, GUAC_PROTOCOL_STATUS_SERVER_ERROR,
+                "Error occurred attempting to check host key: %s", err_msg);
+
+        if (known_host_check > 0)
+            guac_client_abort(client, GUAC_PROTOCOL_STATUS_SERVER_ERROR,
+                "Host fingerprint did not match any provided known host keys. %s", err_msg);
+
+        return NULL;
```
mike-jumper: Same here - `common_session` and `fd` are resources allocated by this function which will not be freed elsewhere if `NULL` is returned. They need to be freed prior to return.

src/common-ssh/key.c
```diff
@@ -245,3 +246,86 @@ int guac_common_ssh_key_sign(guac_common_ssh_key* key, const char* data,
...
+    /* No host keys were loaded, so we bail out checking and continue the connection. */
+    else if (known_hosts == 0) {
+        guac_client_log(client, GUAC_LOG_WARNING,
+            "No known host keys provided, host identity will not be verified.");
+        libssh2_knownhost_free(ssh_known_hosts);
+        return known_hosts;
+    }
+
+
+    /* Check fingerprint against known hosts */
```
mike-jumper: Is this the fingerprint? I see that the variable is named `fingerprint`, but so far the functions receiving that value seem to deal only with keys, or at least that's what I gather from the libssh2 documentation.

src/common-ssh/common-ssh/key.h
```diff
@@ -166,5 +169,52 @@ void guac_common_ssh_key_free(guac_common_ssh_key* key);
...
+ *     /etc/guacamole/ssh_known_hosts file and load entries from it.
+ *
+ * @param hostname
+ *     The hostname or IP of the server that is being verified.
+ *
+ * @param port
+ *     The port number of the server being verified.
+ *
+ * @param fingerprint
+ *     The fingering of the server being verified.
```
mike-jumper: fingerprint*, I believe, assuming this is indeed a fingerprint and not a key (see below).

src/common-ssh/ssh.c
necouchman: Leak plugged.

src/common-ssh/ssh.c
necouchman: Lead plugged.

src/common-ssh/key.c
necouchman: Changed the nomenclature, here.

src/common-ssh/common-ssh/key.h
necouchman: Renamed.

src/common-ssh/ssh.c
```diff
@@ -518,6 +520,43 @@ guac_common_ssh_session* guac_common_ssh_create_session(guac_client* client,
...
+    /* SSH known host key checking. */
+    int known_host_check = guac_common_ssh_verify_host_key(session, client, host_key,
+                                                           hostname, atoi(port), remote_hostkey,
+                                                           remote_hostkey_len);
+
+    /* Abort on any error codes */
+    if (known_host_check != 0) {
+        char* err_msg;
+        int err_len;
+        libssh2_session_last_error(session, &err_msg, &err_len, 0);
```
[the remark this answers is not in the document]
necouchman: NULLed.

PR #166, opened 2018-05-16
GUACAMOLE-564: Hide APC escape sequence.

darchons:
I see some programs emitting these sequences. An APC escape sequence contains an arbitrary string command between (ESC _) and (ESC \) sequences. While we don't support any APC commands, we
should ignore any commands that we do encounter so they're not echoed onto the screen.

### DIFF
--- a/src/terminal/terminal_handlers.c
+++ b/src/terminal/terminal_handlers.c
@@ -367,6 +367,10 @@ int guac_terminal_escape(guac_terminal* term, unsigned char c) {
             guac_terminal_reset(term);
             break;
 
+        case '_':
+            term->char_handler = guac_terminal_apc;
+            break;
+
         default:
             guac_client_log(term->client, GUAC_LOG_DEBUG,
                     "Unhandled ESC sequence: %c", c);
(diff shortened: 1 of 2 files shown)


src/terminal/terminal_handlers.c
```diff
@@ -1373,3 +1377,22 @@ int guac_terminal_ctrl_func(guac_terminal* term, unsigned char c) {
 
 }
 
+/**
+ * xterm does not implement APC functions and neither do we.
+ */
+int guac_terminal_apc(guac_terminal* term, unsigned char c) {
+
+    /* Look for the "ESC \" (string terminator) sequence,
```
mike-jumper: Is `ESC \` the only sequence that is valid as the string terminator for APC? Other sequences have involved other characters for the string terminator:

https://github.com/apache/guacamole-server/blob/b61a6ab758177b4d4cec4a3a0a2f51a3e76c3772/src/terminal/terminal_handlers.c#L1335-L1337

src/terminal/terminal_handlers.c
```diff
@@ -1373,3 +1377,22 @@ int guac_terminal_ctrl_func(guac_terminal* term, unsigned char c) {
 
 }
 
+/**
```
mike-jumper: `guac_terminal_apc()` is correctly documented at the header level. There shouldn't be a duplicate Doxygen comment attached to the function body which provides additional information. The comment should either be moved within the function, or the comment should be deleted and the information instead added to the existing header-level comment.

src/terminal/terminal_handlers.c
```diff
@@ -1373,3 +1377,22 @@ int guac_terminal_ctrl_func(guac_terminal* term, unsigned char c) {
...
+ * xterm does not implement APC functions and neither do we.
+ */
+int guac_terminal_apc(guac_terminal* term, unsigned char c) {
+
+    /* Look for the "ESC \" (string terminator) sequence,
+     * while ignoring other chars. */
+    static char escaping = 0;
```
mike-jumper: If representing a boolean value, please use the semantically-correct `bool`.

We have to live with using `int` within public APIs like libguac to avoid issues with third-party code that redefines `bool` in a non-standard way, as well as within anything dealing with FreeRDP (an example of such third-party code), but the internal terminal emulator is definitely safe to use `bool`.

darchons:
Updated commits to address comments.

PR #167, opened 2018-05-16
GUACAMOLE-565: Add terminal-type parameter for SSH and Telnet.

darchons:
Add a terminal-type parameter for SSH and Telnet connections, to specify the terminal emulator type that is passed to programs. If not specified, the default type of "linux" is used in keep with existing behavior.

Some programs make use of the terminal type to determine some of the terminal's capabilities. For example, tmux seems to only have good 256-color support when the terminal type is something like `screen-256color` instead of `linux`.

### DIFF
--- a/src/protocols/ssh/settings.c
+++ b/src/protocols/ssh/settings.c
@@ -57,6 +57,7 @@ const char* GUAC_SSH_CLIENT_ARGS[] = {
     "read-only",
     "server-alive-interval",
     "backspace",
+    "terminal-type",
     NULL
 };
 
@@ -217,6 +218,12 @@ enum SSH_ARGS_IDX {
      */
     IDX_BACKSPACE,
 
+    /**
+     * The terminal emulator type that is passed to the remote system (e.g.
+     * "xterm" or "xterm-256color"). "linux" is used if unspecified.
+     */
+    IDX_TERMINAL_TYPE,
+
     SSH_ARGS_COUNT
 };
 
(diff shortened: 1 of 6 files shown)


necouchman:
Just a note that this will also require a minor update on the guacamole-client side to provide the parameter for the SSH and Telnet connections.  Pretty straight-forward.

darchons:
Thanks for the reminder! I added apache/guacamole-client#288

src/protocols/ssh/settings.c
```diff
@@ -217,6 +218,12 @@ enum SSH_ARGS_IDX {
      */
     IDX_BACKSPACE,
 
+    /**
+     * The terminal emulator type that is passed to programs (e.g. "xterm" or
```
necouchman: So, I have one minor nitpick about this comment - Guacamole is not actually passing the terminal emulator type to any particular program(s), this simply passes it to the system - it's up to the system to make it available to the programs.  Perhaps "passed to the remote system" would be better?

src/protocols/ssh/settings.h
```diff
@@ -228,6 +228,11 @@ typedef struct guac_ssh_settings {
      */
     int backspace;
 
+    /**
+     * The terminal emulator type that is passed to programs.
```
necouchman: As noted above - perhaps "passed to remote system" would be better.

src/protocols/telnet/settings.c
```diff
@@ -181,6 +182,12 @@ enum TELNET_ARGS_IDX {
      */
     IDX_BACKSPACE,
 
+    /**
+     * The terminal emulator type that is passed to programs (e.g. "xterm" or
```
necouchman: Same thing as noted above.

src/protocols/telnet/settings.h
```diff
@@ -214,6 +214,11 @@ typedef struct guac_telnet_settings {
      */
     int backspace;
 
+    /**
+     * The terminal emulator type that is passed to programs.
```
necouchman: And, again...

mike-jumper:
With the exception of the points raised by @necouchman regarding program vs. remote system, things look good to me here.

darchons:
Update commit to address comments.

PR #169, opened 2018-06-02
GUACAMOLE-422: Add support for timezone redirection

necouchman:
These are the server-side changes necessary to support timezone redirection for RDP and SSH.  I originally attempted to implement for Telnet, as well; however, I quickly found that at least the default Linux telnet server implementation completely ignores arbitrary environment variables and only pays attention to specific ones (USER, REMOTEHOST, DISPLAY, etc.).

### DIFF
--- a/src/protocols/rdp/rdp_settings.c
+++ b/src/protocols/rdp/rdp_settings.c
@@ -1281,15 +1281,8 @@ void guac_rdp_push_settings(guac_rdp_settings* guac_settings, freerdp* rdp) {
 #endif
 
     /* Device redirection */
-    if (guac_settings->timezone) {
-
-        /* Set the TZ env variable */
-        if (setenv("TZ", guac_settings->timezone, 1)) {
-            guac_user_log(user, GUAC_LOG_WARNING, "Could not set TZ "
-                "variable.  Received error %i", errno);
-        }
-
-    }
+    if (guac_settings->timezone)
+        setenv("TZ", guac_settings->timezone, 1)
 
 #ifdef LEGACY_RDPSETTINGS
 #ifdef HAVE_RDPSETTINGS_DEVICEREDIRECTION
--- a/src/protocols/rdp/rdp_settings.c
+++ b/src/protocols/rdp/rdp_settings.c
@@ -1280,8 +1280,15 @@ void guac_rdp_push_settings(guac_rdp_settings* guac_settings, freerdp* rdp) {
 #endif
 
     /* Device redirection */
-    if (guac_settings->timezone)
-        setenv("TZ", guac_settings->timezone, 1);
+    if (guac_settings->timezone) {
+
+        /* Set the TZ env variable */
+        if (setenv("TZ", guac_settings->timezone, 1)) {
+            guac_user_log(user, GUAC_LOG_WARNING, "Could not set TZ "
+                "variable.  Received error %i", errno);
+        }
+
+    }
 
 #ifdef LEGACY_RDPSETTINGS
 #ifdef HAVE_RDPSETTINGS_DEVICEREDIRECTION
--- a/src/protocols/rdp/rdp_settings.c
+++ b/src/protocols/rdp/rdp_settings.c
@@ -1169,7 +1169,8 @@ static char* guac_rdp_strdup(const char* str) {
 
 }
 
-void guac_rdp_push_settings(guac_rdp_settings* guac_settings, freerdp* rdp) {
+void guac_rdp_push_settings(guac_client* client,
+        guac_rdp_settings* guac_settings, freerdp* rdp) {
 
     BOOL bitmap_cache = !guac_settings->disable_bitmap_caching;
     rdpSettings* rdp_settings = rdp->settings;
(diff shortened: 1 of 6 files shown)


necouchman:
Rebased to current master...anyone care to review?

mike-jumper:
> Rebased to current master...anyone care to review?

Sure - I'll take a look.

> I quickly found that at least the default Linux telnet server implementation completely ignores arbitrary environment variables and only pays attention to specific ones (USER, REMOTEHOST, DISPLAY, etc.).

I encountered similar when trying to set `LANG` via telnet for [GUACAMOLE-649](https://issues.apache.org/jira/browse/GUACAMOLE-649). The standard does define a mechanism for setting other variables that aren't the well-known set, but I found that the telnet server I was testing against didn't correctly implement that part of the standard, instead interpreting those variables as part of `USER` and breaking auth. Finally gave up and restricted scope to SSH.

src/protocols/rdp/rdp_settings.c
```diff
@@ -1265,6 +1277,9 @@ void guac_rdp_push_settings(guac_rdp_settings* guac_settings, freerdp* rdp) {
 #endif
 
     /* Device redirection */
+    if (guac_settings->timezone)
+        setenv("TZ", guac_settings->timezone, 1);
```
mike-jumper: A warning should be logged if this cannot be set, since an expectation of the behavior of the connection is going unfulfilled.

src/protocols/ssh/ssh.c
```diff
@@ -256,6 +256,10 @@ void* ssh_client_thread(void* data) {
         return NULL;
     }
 
+    /* Set the client timezone */
+    if (settings->timezone != NULL)
+        libssh2_channel_setenv(ssh_client->term_channel, "TZ", settings->timezone);
```
mike-jumper: Same here - a warning should be logged if this cannot be set.

src/protocols/ssh/settings.h
```diff
@@ -253,6 +253,10 @@ typedef struct guac_ssh_settings {
      * environment variable.
      */
     char* locale;
+    /**
```
mike-jumper: Missing a blank line separating the block comment and the code above.

src/protocols/ssh/settings.c
```diff
@@ -246,6 +247,15 @@ enum SSH_ARGS_IDX {
      * variable to be set.
      */
     IDX_LOCALE,
+     
+    /**
+     * The timezone that is passed from the client system to the
+     * remote server, or null if not specified.  If set, and allowed
```
mike-jumper: "... or null if not specified" is good to know, but not actually applicable to the connection parameter. This is an implementation detail which goes along with the internal representation within the settings struct. The documentation of the connection parameter should be from the perspective of a user considering providing a value for that parameter.

src/protocols/rdp/rdp_settings.c
```diff
@@ -356,6 +357,11 @@ enum RDP_ARGS_IDX {
      */
     IDX_PRECONNECTION_BLOB,
 
+    /**
+     * The timezone to pass through to the RDP connection.
```
mike-jumper: What format are these timezones in? Is there some reference users can consult to determine what legal values are available?

src/protocols/rdp/rdp_settings.c
necouchman: It appears that, in order to do this, I would need to pass the entire `client` object through to this `guac_rdp_push_settings()` function?

src/protocols/ssh/ssh.c
necouchman: Done.

src/protocols/ssh/settings.h
necouchman: Added.

src/protocols/ssh/settings.c
necouchman: Reworded, hopefully it's good.  Added reference to Wikipedia page.

src/protocols/rdp/rdp_settings.c
necouchman: Added reference to Wikipedia page.

src/protocols/rdp/rdp_settings.c
mike-jumper: A pointer to it, yeah.

src/protocols/rdp/rdp_settings.c
```diff
@@ -1265,6 +1281,9 @@ void guac_rdp_push_settings(guac_rdp_settings* guac_settings, freerdp* rdp) {
 #endif
 
     /* Device redirection */
```
mike-jumper: This comment looks to belong to the block that used to follow directly.

src/protocols/rdp/rdp_settings.c
```diff
@@ -1264,6 +1281,14 @@ void guac_rdp_push_settings(guac_rdp_settings* guac_settings, freerdp* rdp) {
 #endif
 #endif
 
+    /* Timezone redirection */
+    if (guac_settings->timezone) {
+        if(setenv("TZ", guac_settings->timezone, 1)) {
```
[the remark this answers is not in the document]
necouchman: Wow, and I thought I was past that.

src/protocols/rdp/rdp_settings.c
```diff
@@ -1264,6 +1281,14 @@ void guac_rdp_push_settings(guac_rdp_settings* guac_settings, freerdp* rdp) {
...
+            guac_client_log(client, GUAC_LOG_WARNING,
+                "Unable to set TZ variable, error %i", errno);
```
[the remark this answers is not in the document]
necouchman: Used your suggestion.

src/protocols/ssh/ssh.c
```diff
@@ -256,6 +256,17 @@ void* ssh_client_thread(void* data) {
         return NULL;
     }
 
+    /* Set the client timezone */
+    if (settings->timezone != NULL) {
+        if (libssh2_channel_setenv(ssh_client->term_channel, "TZ",
+                    settings->timezone)) {
+            guac_client_log(client, GUAC_LOG_WARNING,
+                    "Unable to set the timzeone: SSH server "
```
mike-jumper: timzeone

necouchman: Speling corected.

mike-jumper: Tnahks. Loks gud.

PR #173, opened 2018-06-17
GUACAMOLE-574: Allow input for SSH/telnet to come from "STDIN" pipe.

mike-jumper:
This change adds support to SSH and telnet for reading user input from a pipe stream named "STDIN". When the client side opens this pipe stream, user input through keyboard, clipboard, etc. is blocked, and data from the pipe stream is sent directly over the SSH/telnet connection. User input is unblocked once the "STDIN" pipe is closed.

### DIFF
--- a/src/terminal/terminal/terminal.h
+++ b/src/terminal/terminal/terminal.h
@@ -201,6 +201,14 @@ struct guac_terminal {
     int stdin_pipe_fd[2];
 
     /**
+     * The currently-open pipe stream from which all terminal input should be
+     * read, if any. If no pipe stream is open, terminal input will be received
+     * through keyboard, clipboard, and mouse events, and this value will be
+     * NULL.
+     */
+    guac_stream* input_stream;
+
+    /**
      * The currently-open pipe stream to which all terminal output should be
      * written, if any. If no pipe stream is open, terminal output will be
      * written to the terminal display, and this value will be NULL.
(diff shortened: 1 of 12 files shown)


src/terminal/terminal/terminal.h
```diff
@@ -568,17 +576,97 @@ int guac_terminal_printf(guac_terminal* terminal, const char* format, ...);
...
  * Handles the given key event, sending data, scrolling, pasting clipboard
- * data, etc. as necessary.
+ * data, etc. as necessary. If terminal input is currently coming from a
+ * stream due to a prior call to guac_terminal_send_stream(), any input
+ * which would normally result from the key event is dropped.
+ *
+ * @param term
+ *     The terminal which should receive the given data on STDIN.
+ *
+ * @param user
```
necouchman: I don't see the `user` parameter actually in the function stub?

src/terminal/terminal/terminal.h
```diff
@@ -568,17 +576,97 @@ int guac_terminal_printf(guac_terminal* terminal, const char* format, ...);
...
+ *     The X coordinate of the mouse within the display when the event
+ *     occurred, in pixels. This value is not guaranteed to be within the
+ *     bounds of the display area.
+ *
+ * @param y
+ *     The Y coordinate of the mouse within the display when the event
+ *     occurred, in pixels. This value is not guaranteed to be within the
+ *     bounds of the display area.
+ *
+ * @param button_mask
```
necouchman: `button_mask` or `mask`?

mike-jumper:
OK - comments should be fixed.

mike-jumper:
Regarding testing this and similar low-level things, the easiest way I've found so far is to access the `Guacamole.Client` instance via the wrapping `ManagedClient` on the scope of the client page:

    var managedClient = $('.client-view').scope().client;
    var client = managedClient.client;

You can then do things like:

    var stdin = new Guacamole.StringWriter(client.createPipeStream('text/plain', 'STDIN'));
    stdin.sendText('some test input');
    stdin.sendEnd();

It'd be good to standardize on a way to access the underlying `Guacamole.Client` at some point (for the sake of extensions which need to do so), but the above will work as things stand.

necouchman:
> Regarding testing this and similar low-level things, the easiest way I've found so far is to access the Guacamole.Client instance via the wrapping ManagedClient on the scope of the client page:

Very cool - just tried that out and it worked great.

PR #176, opened 2018-07-18
GUACAMOLE-573: Correct scrollback buffer bounds checks.

mike-jumper:
The new bounds checks added for copying text from the scrollback buffer are incorrect for the case where the number of rows in scrollback has not yet reached at least the size of the terminal window. These changes remove those incorrect checks, instead modifying scrollback buffer row retrieval such that it is guaranteed to stay within bounds.


mike-jumper:
b0b0b18 is my new favorite commit hash.

PR #183, opened 2018-09-10
GUACAMOLE-620: Fix instruction loss when buffer is too short

changkun:
`guac_parser_read` moves `unparsed_end` and `unparsed_start` to the middle of `parser->__instructionbuf` after `GUAC_PARSE_COMPLETE` state, and so on. They will be reset only if `unparsed_end == buffer_end`.

This PR fixes an instruction parse error when a provided buffer to `read` call is shorter than the expected instruction data, then the next `guac_parser_append` will read incomplete instruction, which leads a `GUAC_PARSE_ERROR` eventually.

### DIFF
--- a/src/libguac/parser.c
+++ b/src/libguac/parser.c
@@ -222,7 +222,11 @@ int guac_parser_read(guac_parser* parser, guac_socket* socket, int usec_timeout)
             retval = guac_socket_select(socket, usec_timeout);
             if (retval <= 0)
                 return -1;
-           
+            
+            /* Reset pointers if instruction buf len is less than max instruction len */
+            if (buffer_end - unparsed_end < GUAC_INSTRUCTION_MAX_LENGTH)
+                unparsed_end = unparsed_start = parser->__instructionbuf;
+            
             /* Attempt to fill buffer */
             retval = guac_socket_read(socket, unparsed_end,
                     buffer_end - unparsed_end);
--- a/src/libguac/parser.c
+++ b/src/libguac/parser.c
@@ -167,7 +167,6 @@ int guac_parser_read(guac_parser* parser, guac_socket* socket, int usec_timeout)
 
     char* unparsed_end   = parser->__instructionbuf_unparsed_end;
     char* unparsed_start = parser->__instructionbuf_unparsed_start;
-    char* instr_start    = parser->__instructionbuf_unparsed_start;
     char* buffer_end     = parser->__instructionbuf + sizeof(parser->__instructionbuf);
 
     /* Begin next instruction if previous was ended */
@@ -187,35 +186,9 @@ int guac_parser_read(guac_parser* parser, guac_socket* socket, int usec_timeout)
 
             /* If no space left to read, fail */
             if (unparsed_end == buffer_end) {
-
-                /* Shift backward if possible */
-                if (instr_start != parser->__instructionbuf) {
-
-                    int i;
-
-                    /* Shift buffer */
-                    int offset = instr_start - parser->__instructionbuf;
-                    memmove(parser->__instructionbuf, instr_start,
-                            unparsed_end - instr_start);
-
-                    /* Update tracking pointers */
-                    unparsed_end -= offset;
-                    unparsed_start -= offset;
-                    instr_start = parser->__instructionbuf;
-
-                    /* Update parsed elements, if any */
-                    for (i=0; i < parser->__elementc; i++)
-                        parser->__elementv[i] -= offset;
-
-                }
-
-                /* Otherwise, no memory to read */
-                else {
-                    guac_error = GUAC_STATUS_NO_MEMORY;
-                    guac_error_message = "Instruction too long";
-                    return -1;
-                }
-
+                guac_error = GUAC_STATUS_NO_MEMORY;
+                guac_error_message = "Instruction too long";
+                return -1;
             }
 
             /* No instruction yet? Get more data ... */


mike-jumper:
> `guac_parser_read` moves `unparsed_end` and `unparsed_start` to the middle of `parser->__instructionbuf` after `GUAC_PARSE_COMPLETE` state, and so on. They will be reset only if `unparsed_end == buffer_end`.

It could be argued that these pointers should be reset earlier to ensure subsequent reads can use a larger buffer for efficiency's sake, however this will not result in a parse error. There is no loss of data here.

> This PR fixes an instruction parse error when a provided buffer to read call is shorter than the expected instruction data, then the next `guac_parser_append` will read incomplete instruction, which leads a `GUAC_PARSE_ERROR` eventually.

This is incorrect. It is expected that `guac_socket_read()` will read chunks which are incomplete portions of instructions, and it is expected that `guac_parser_append()` will be given incomplete portions of instructions. What matters is that each chunk, when concatenated together, ultimately forms a complete instruction at some point. The purpose of the instruction buffer is to provide the space for this assembly.

Running out of space in the instruction buffer will not result in parse errors. The in-progress instruction will be moved to the beginning of the buffer to make room (the condition you cite here).

The relevant contents of the instruction buffer are preserved if space runs out through the use of `memmove()` and manual updates to existing parsed parts of the instruction:

https://github.com/apache/guacamole-server/blob/332e187813595fc2e769f3e29c0582b7ec726ea1/src/libguac/parser.c#L196-L208

If the `unparsed_start` and `unparsed_end` pointers are simply reset to the beginning of the buffer without performing these updates (as in this PR), that *will* result in data loss / corruption.

src/libguac/parser.c
```diff
@@ -222,7 +222,11 @@ int guac_parser_read(guac_parser* parser, guac_socket* socket, int usec_timeout)
             retval = guac_socket_select(socket, usec_timeout);
             if (retval <= 0)
                 return -1;
-           
+            
+            /* Reset pointers if instruction buf len is less than max instruction len */
+            if (buffer_end - unparsed_end < GUAC_INSTRUCTION_MAX_LENGTH)
+                unparsed_end = unparsed_start = parser->__instructionbuf;
```
mike-jumper: This can result in the corruption of the in-progress instruction. There is no guarantee that the sequence of reads which trigger the `if` condition will not (for example) split part of an instruction element. While the normal pattern of updates to `unparsed_start` and `unparsed_end` would guarantee that the sequence of reads ultimately builds a contiguous instruction, that guarantee is lost if those pointers are updated in this way - reads are suddenly no longer contiguous.

mike-jumper: `GUAC_STATUS_NO_MEMORY` happens only if it is impossible to free up space within the buffer using `memmove()`, a fatal condition. The check `unparsed_end == buffer_end` is not about any individual `read()` call; it's about the union of all `read()` calls since the buffer was last reset.

mike-jumper: @changkun, if you are confident that these changes are correct, can you perhaps provide a sequence of `read()`s that demonstrate that the parser (without these changes) incorrectly generates a parse error?

changkun: @mike-jumper All description lies in the jira ticket: https://issues.apache.org/jira/projects/GUACAMOLE/issues/GUACAMOLE-620

Frist, all TCP traffic contains full instruction.

Second, the following outputs:

```
__write_all: 4.sync,8.76258350;
__write_all: 5.mouse,3.445,3.261,1.1;
guac_socket_fd_read_handler: 4.sync,8.76258�, length: 14
guac_socket_fd_read_handler: 5.mouse,3.445,3.261,1.1;eo, length: 24
```

`__write_all` in `guacd` successfully writes full instruction, somehow (no idea with system calls behavior),  `guac_socket_fd_read_handler` lost part of the instruction, and then reads next instruction.

mike-jumper: > First, all TCP traffic contains full instruction.

What matters when considering the validity of these changes and the underlying cause of what you're seeing are the specific results of each call to `read()`. Each call to `read()` will populate the buffer with an unknown portion of that stream. The size of each chunk in the sequence matters.

If it's been established that the TCP stream is correct in itself, that's good, but that's not the whole picture.

> ... somehow (no idea with system calls behavior), `guac_socket_fd_read_handler` lost part of the instruction, and then reads next instruction.

That isn't possible. Each call to `read()` is not guaranteed to return all data written to the socket - it may only return part of that data, but the next call to `read()` should return yet more of the data, and so on. It's not possible for data to appear to just vanish between reads unless:

1. Multiple threads are both consuming the data from the same file descriptor.
2. `__write_all()` is not actually writing everything.

You are testing against unmodified guacamole-server master, correct? You're not including your changes from #179?

changkun: > That isn't possible. Each call to read() is not guaranteed to return all data written to the socket - it may only return part of that data, but the next call to read() should return yet more of the data, and so on.

I know it is not possible, socketpair has its own buffer on both side of sockets, they cache unread data. Magically, I somehow meet this problem.

> __write_all() is not actually writing everything.

So far, I observed from the output log, `__write_all` write entire buffer array successfully, in the end, before parse error.

> You are testing against unmodified guacamole-server master, correct? You're not including your changes from #179?

No, I do not including #179 . All tests and tracking are based on vanilla tag 0.9.14.

mike-jumper: Here, I've added debug logging for each block written by `write()` within `__write_all()` as well as each block read via `guac_socket_read()` within the parser:

https://github.com/mike-jumper/guacamole-server/tree/debug-io

The logging added above should be stable in the sense that:

1. Reads and writes are written to separate files and thus the log shouldn't become corrupt if they happen to overlap.
2. The log routine does not expect a null terminator, relying instead on a supplied length, so there shouldn't be garbage / stale bytes in the logs that weren't actually read/written at the time logged.

Reads are logged to `read-log.txt` within the current directory of guacd and writes are logged to `write-log.txt` within the current directory of guacd. The logged data is in hex dump format and will generally look like this:

```
--- 25 bytes:
00000000:  34 2E 73 69 7A 65 2C 34 2E 31 31 35 34 2C 33 2E  4.size,4.1154,3.
00000010:  36 35 34 2C 32 2E 39 36 3B                       654,2.96;
--- 209 bytes:
00000000:  35 2E 61 75 64 69 6F 2C 38 2E 61 75 64 69 6F 2F  5.audio,8.audio/
00000010:  4C 38 2C 39 2E 61 75 64 69 6F 2F 4C 31 36 3B 35  L8,9.audio/L16;5
00000020:  2E 76 69 64 65 6F 3B 35 2E 69 6D 61 67 65 2C 31  .video;5.image,1
00000030:  30 2E 69 6D 61 67 65 2F 6A 70 65 67 2C 39 2E 69  0.image/jpeg,9.i
00000040:  6D 61 67 65 2F 70 6E 67 2C 31 30 2E 69 6D 61 67  mage/png,10.imag
00000050:  65 2F 77 65 62 70 3B 37 2E 63 6F 6E 6E 65 63 74  e/webp;7.connect
00000060:  2C 39 2E 6C 6F 63 61 6C 68 6F 73 74 2C 34 2E 35  ,9.localhost,4.5
00000070:  39 30 31 2C 30 2E 2C 30 2E 2C 30 2E 2C 30 2E 2C  901,0.,0.,0.,0.,
00000080:  30 2E 2C 30 2E 2C 30 2E 2C 30 2E 2C 30 2E 2C 30  0.,0.,0.,0.,0.,0
00000090:  2E 2C 30 2E 2C 30 2E 2C 30 2E 2C 30 2E 2C 30 2E  .,0.,0.,0.,0.,0.
000000A0:  2C 30 2E 2C 30 2E 2C 30 2E 2C 30 2E 2C 30 2E 2C  ,0.,0.,0.,0.,0.,
000000B0:  30 2E 2C 30 2E 2C 30 2E 2C 30 2E 2C 30 2E 2C 30  0.,0.,0.,0.,0.,0
000000C0:  2E 2C 30 2E 2C 30 2E 2C 30 2E 2C 30 2E 2C 30 2E  .,0.,0.,0.,0.,0.
000000D0:  3B                                               ;
--- 30 bytes:
00000000:  39 2E 63 6C 69 70 62 6F 61 72 64 2C 31 2E 30 2C  9.clipboard,1.0,
00000010:  31 30 2E 74 65 78 74 2F 70 6C 61 69 6E 3B        10.text/plain;
--- 76 bytes:
00000000:  33 2E 65 6E 64 2C 31 2E 30 3B 35 2E 61 75 64 69  3.end,1.0;5.audi
00000010:  6F 2C 31 2E 30 2C 33 31 2E 61 75 64 69 6F 2F 4C  o,1.0,31.audio/L
00000020:  31 36 3B 72 61 74 65 3D 34 34 31 30 30 2C 63 68  16;rate=44100,ch
00000030:  61 6E 6E 65 6C 73 3D 32 3B 34 2E 73 79 6E 63 2C  annels=2;4.sync,
00000040:  39 2E 38 33 30 34 34 37 36 30 32 3B              9.830447602;
...
```

This should hopefully give us cleaner, more thorough logs.

Please try to reproduce the parse error with the above branch and provide the resulting `write-log.txt` and `read-log.txt` for the connection that failed. Be sure to remove the `write-log.txt` and `read-log.txt` files between connection attempts, as the debug code just repeatedly appends to those files. If the files exist at the beginning of a connection, the old log data will still be there.

changkun: Log files are uploaded.

mike-jumper: Well, the good news is that your servlet is absolved. There can be no doubt that the protocol data is indeed reaching guacd correctly and is not being clobbered by shared use of a common socket as I originally suspected. [Your provided `write_log.txt`](https://issues.apache.org/jira/secure/attachment/12939404/write-log.txt) clearly demonstrates this:

```
...
--- 23 bytes:
00000000:  35 2E 6D 6F 75 73 65 2C 33 2E 35 34 31 2C 32 2E  5.mouse,3.541,2.
00000010:  32 36 2C 31 2E 31 3B                             26,1.1;
--- 23 bytes:
00000000:  35 2E 6D 6F 75 73 65 2C 33 2E 35 34 32 2C 32 2E  5.mouse,3.542,2.
00000010:  32 36 2C 31 2E 31 3B                             26,1.1;
```

The bad news is that these changes are definitely incorrect, and they do indeed break the parser. The following are the blocks ultimately read by simple, sequential calls to `guac_socket_read()` within the parser according to [your provided `read.txt`](https://issues.apache.org/jira/secure/attachment/12939405/read-log.txt):

```
...
--- 20 bytes:
00000000:  35 2E 6D 6F 75 73 65 2C 33 2E 35 34 31 2C 32 2E  5.mouse,3.541,2.
00000010:  32 36 2C 31                                      26,1
--- 23 bytes:
00000000:  35 2E 6D 6F 75 73 65 2C 33 2E 35 34 32 2C 32 2E  5.mouse,3.542,2.
00000010:  32 36 2C 31 2E 31 3B                             26,1.1;
```

Any changes which allow the parser to accept "5.mouse,3.541,2.26,15.mouse,3.542,2.26,1.1;" are incorrect, as that should result in a parse error. It is invalid protocol data.

The issue is not that the parser is behaving incorrectly - it is correctly bailing out with a parse error when given invalid data. The issue is that bytes are somehow vanishing from the stream between otherwise normal calls to `read()`.

changkun: As expected and as same as what I wrote before.

Two suggestions if you do what make changes from this PR:

1. Carefully handling conditions when instruction incomplete, for example, reconstruction mechanism in a transmission context.
2. Redesigning protocol, add probably one instruction which allows `client/server` retransmission, as far as I see, there is no error handling for transmission. Also, `LENGTH.VALUE` is a redundant design and it can be simplified with `VALUE` only.

mike-jumper:
Careful handling of conditions is exactly why that error is occurring. Garbage in, garbage out.

The error itself is not a bug, and the protocol is correctly designed. Without getting into why the length prefixes are necessary and why the protocol is designed as it is, the problem which _does_ exist is the loss of data between the write and the read. Let's focus on resolving that.

Changes to the parser to the protocol are non-starters. Let's please not continue down that path of thought.

mike-jumper:
Theoretically, the socket created by that call to `socketpair()` is a `SOCK_STREAM`, where stream reliability is guaranteed. If that's not actually happening, what we need to determine is why, and then address that.

changkun:
> If that's not actually happening, what we need to determine is why, and then address that.

This is an asbsolute kernel level issue and I have no interests to put effort on that.

Since this PR reliablely supports the connection round few days, I am going to close this if it is not recommended merge to master.

mike-jumper:
For the sake of summary, let me break down the investigation so far:

* You have through [GUACAMOLE-620](https://issues.apache.org/jira/browse/GUACAMOLE-620) reported a case where an instruction parse error occurs.
* Through much investigation on your side, the error has been confirmed to be correct. The received data is incorrect when the parser receives it, and the parser correctly reports that it is incorrect.
* Our original expectation was that the error in the instruction data was due to modifications that you had made on the web application side. This has since been shown to not be the case.
* Rather unexpectedly, it has been demonstrated that there is dataloss along the socketpair connecting the receiving side of guacd with the parser that feeds the connection.

The issue here is not that invalid data should be accepted by the parser; it's invalid data. Modifying the parser such that it does accept that data may appear to solve the issue in your case, but it is actually introducing a bug.

Given that the socketpair involved is `SOCK_STREAM`, the byte sequence coming out _should_ be reliable and lossless. The fact that this is not happening in your case could be due to:

1. A kernel-level bug (rather unlikely)
2. A known quirk of the platform you're using (worth investigating)
3. Parameters being set incorrectly on the socketpair such that it in fact is not reliable (this would be a bug in Guacamole)

It would be a shame if you give up on investigating further, particularly since you are so far the only individual who has encountered this. The best way to benefit both yourself and the community would be to continue investigating until the true underlying cause is known and resolved.

Beyond that, no, we definitely can't accept these changes as they stand. I encourage you to continue working on this from another direction (determining why the socketpair is dropping data in your case). If you don't wish to do this, again, that would be a shame, but it's your prerogative.

changkun:
The environment is based on a common docker image, which is based on `debian:stretch`, here is part of the docker file relates to guacamole-server:

```Docker
FROM buildpack-deps:stretch-scm
LABEL maintainer="Ou Changkun <hi@changkun.us>"

# guacamole dependencies
RUN apt-get update && apt-get install -y \
    libcairo2-dev                        \
    libjpeg62-turbo-dev                  \
    libpng-dev                           \
    libossp-uuid-dev                     \
    libavcodec-dev                       \
    libavutil-dev                        \
    libswscale-dev                       \
    freerdp                              \
    libfreerdp-dev                       \
    libpango1.0-dev                      \
    libssh2-1-dev                        \
    libtelnet-dev                        \
    libvncserver-dev                     \
    libpulse-dev                         \
    libssl-dev                           \
    libvorbis-dev                        \
    libwebp-dev
# compile guacamole from github
# RUN cd /go/src/github.com/sjtushi/guacamole-go/guacamole && \
RUN git clone https://github.com/apache/guacamole-server.git /guacamole && cd /guacamole && \
    git checkout 0.9.14                                                     && \
    autoreconf -fi                                                          && \
    ./configure --with-init-dir=/etc/init.d                                 && \
    make                                                                    && \
    make install                                                            && \
    ldconfig                                                                && \
    ln -s /usr/local/lib/freerdp/guac* /usr/lib/x86_64-linux-gnu/freerdp/
```

changkun:
The environment is based on a common docker image, which is based on `debian:stretch`, here is part of the docker file relates to guacamole-server:

```Docker
FROM buildpack-deps:stretch-scm
LABEL maintainer="Ou Changkun <hi@changkun.us>"

# guacamole dependencies
RUN apt-get update && apt-get install -y \
    libcairo2-dev                        \
    libjpeg62-turbo-dev                  \
    libpng-dev                           \
    libossp-uuid-dev                     \
    libavcodec-dev                       \
    libavutil-dev                        \
    libswscale-dev                       \
    freerdp                              \
    libfreerdp-dev                       \
    libpango1.0-dev                      \
    libssh2-1-dev                        \
    libtelnet-dev                        \
    libvncserver-dev                     \
    libpulse-dev                         \
    libssl-dev                           \
    libvorbis-dev                        \
    libwebp-dev
# compile guacamole from github
RUN git clone https://github.com/apache/guacamole-server.git /guacamole && cd /guacamole && \
    git checkout 0.9.14                                                     && \
    autoreconf -fi                                                          && \
    ./configure --with-init-dir=/etc/init.d                                 && \
    make                                                                    && \
    make install                                                            && \
    ldconfig                                                                && \
    ln -s /usr/local/lib/freerdp/guac* /usr/lib/x86_64-linux-gnu/freerdp/
```

mike-jumper:
On platform is the Docker host system?

changkun:
I close the issue due to the following reason:

1. After carefully reading the parser, it indeed implemented correctly if an instruction won't be lost. The fix handles if the data accidentally lost somehow, however it is unlikely the case of how libguac implements. Therefore the problem appears in a special operating system (or kernel), and the platform information is presented in above, including how an environment is built and what the host is.
In summary, the instruction loss problem seems appears in docker based Debian stretch builds, further investigation requires but extremely time-consuming and hopeless (each change required a rebuild).

2. The platform does not only cause the instruction loss issue, but it also presents a TLS problem in VNC connections (tigerVNC server), see [GUACAMOLE-414](https://issues.apache.org/jira/projects/GUACAMOLE/issues/GUACAMOLE-414) for more details.

A provided practical solution is to switch to non-Debian builds, which stable for a month without any changes, CentOS 7+ is recommended.

PR #184, opened 2018-09-11
GUACAMOLE-712: Add Danish to RDP

netromnetrom:
Danish keyboard layout

### DIFF
--- a/src/protocols/rdp/keymaps/dk_dk_qwerty.keymap
+++ b/src/protocols/rdp/keymaps/dk_dk_qwerty.keymap
@@ -1,0 +1,64 @@
+#
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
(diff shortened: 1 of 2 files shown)


mike-jumper:
The file itself looks good, but please see our [contribution guidelines](https://github.com/apache/guacamole-server/blob/master/CONTRIBUTING). In particular, there needs to be a JIRA issue associated with your changes, the issue needs to be tagged in the commit message (see the example in the guidelines), and the message itself should describe the high-level change you are making.

Simply stating that a particular file was created or updated doesn't add more information than what version control already provides automatically. You should make use of the commit message to describe the high-level details that the automated system can't tell you.

Please also consider what you're separating into different commits / pull requests and why. Separate pull requests are necessary by nature for different repositories, but it would make more sense if #184 and #185 were the same pull request, and even _more_ sense if they were the same commit.

Same goes for apache/guacamole-client#316 and apache/guacamole-client#317.

src/protocols/rdp/keymaps/dk_dk_qwerty.keymap
```diff
@@ -0,0 +1,64 @@
...
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+#
+
+parent  "base"
+name    "dk-dk-qwerty"
```
mike-jumper: The format used for the name is similar to that used for locales: the two-letter language code, two-letter country code, and finally keyboard variant. The final "-dk-qwerty" looks correct, but perhaps this should be "da-dk-qwerty"? "dk" is the country code for Denmark, but not the language code for Danish.

PR #199, opened 2018-11-05
GUACAMOLE-649: Add support for setting LANG environment variable via SSH.

mike-jumper:


src/protocols/ssh/ssh.c
```diff
@@ -320,6 +320,17 @@ void* ssh_client_thread(void* data) {
         return NULL;
     }
 
+    /* Forward specified locale */
+    if (settings->locale != NULL) {
+        if (libssh2_channel_setenv(ssh_client->term_channel, "LANG",
+                    settings->locale)) {
+            guac_client_abort(client, GUAC_PROTOCOL_STATUS_UPSTREAM_ERROR,
```
necouchman: Should this really be a fatal error in terms of the overall connection?  I ask because in general this is not fatal in terms of other SSH clients - for example, if you use the "ssh" command on pretty much any platform and ask to forward a variable that the server does not allow, it is simply silently ignored and does not abort the connection.

PR #201, opened 2018-11-11
GUACAMOLE-547: Add support for SSH NONE authentication method

necouchman:
I'm not 100% certain that this is all that's needed, because I don't have a SSH server to test this against, but, according to the libssh2 documentation, this may be all that's required to allow the session to continue when there's no authentication required.

https://libssh2.org/libssh2_userauth_list.html

### DIFF
--- a/src/common-ssh/common-ssh/ssh.h
+++ b/src/common-ssh/common-ssh/ssh.h
@@ -36,6 +36,9 @@
  *     The name of the credential being requested, which will be shared
  *     with the client in order to generate a meaningful prompt.
  * 
+ * @return
+ *     The credential provided by the user, which should be a dynamically-
+ *     allocated such that it can be freed as required.
  */
 typedef char* guac_ssh_credential_handler(guac_client* client, char* cred_name);
 
--- a/src/common-ssh/common-ssh/ssh.h
+++ b/src/common-ssh/common-ssh/ssh.h
@@ -37,8 +37,8 @@
  *     with the client in order to generate a meaningful prompt.
  * 
  * @return
- *     The credential provided by the user, which should be a dynamically-
- *     allocated such that it can be freed as required.
+ *     A newly-allocated string containing the credentials provided by
+ *     the user, which must be freed by a call to free().
  */
 typedef char* guac_ssh_credential_handler(guac_client* client, char* cred_name);
 
@@ -129,7 +129,8 @@ void guac_common_ssh_uninit();
  * 
  * @param credential_handler
  *     The handler function for retrieving additional credentials from the user
- *     as required by the SSH server.
+ *     as required by the SSH server, or NULL if the user will not be asked
+ *     for additional credentials.
  *
  * @return
  *     A new SSH session if the connection and authentication succeed, or NULL
--- a/src/common-ssh/common-ssh/ssh.h
+++ b/src/common-ssh/common-ssh/ssh.h
@@ -26,6 +26,11 @@
 #include <libssh2.h>
 
 /**
+ * Handler for retrieving additional credentials.
+ */
+typedef char* guac_ssh_credential_handler(guac_client* client, char* credName);
+
+/**
  * An SSH session, backed by libssh2 and associated with a particular
  * Guacamole client.
  */
@@ -50,6 +55,11 @@ typedef struct guac_common_ssh_session {
      * The file descriptor of the socket being used for the SSH connection.
      */
     int fd;
+    
+    /**
+     * Callback function to retrieve credentials.
+     */
+    guac_ssh_credential_handler* credential_handler;
 
 } guac_common_ssh_session;
 
@@ -99,7 +109,7 @@ void guac_common_ssh_uninit();
  */
 guac_common_ssh_session* guac_common_ssh_create_session(guac_client* client,
         const char* hostname, const char* port, guac_common_ssh_user* user, int keepalive,
-        const char* host_key);
+        const char* host_key, guac_ssh_credential_handler* credential_callback);
 
 /**
  * Disconnects and destroys the given SSH session, freeing all associated
--- a/src/common-ssh/common-ssh/ssh.h
+++ b/src/common-ssh/common-ssh/ssh.h
@@ -27,8 +27,17 @@
 
 /**
  * Handler for retrieving additional credentials.
+ * 
+ * @param client
+ *     The Guacamole Client associated with this need for additional
+ *     credentials.
+ * 
+ * @param cred_name
+ *     The name of the credential being requested, which will be shared
+ *     with the client in order to generate a meaningful prompt.
+ * 
  */
-typedef char* guac_ssh_credential_handler(guac_client* client, char* credName);
+typedef char* guac_ssh_credential_handler(guac_client* client, char* cred_name);
 
 /**
  * An SSH session, backed by libssh2 and associated with a particular
@@ -102,14 +111,31 @@ void guac_common_ssh_uninit();
  *
  * @param user
  *     The user to authenticate as, once connected.
+ * 
+ * @param keepalive
+ *     How frequently the connection should send keepalive packets, in
+ *     seconds.  Zero disables keepalive packets, and 2 is the minimum
+ *     configurable value.
+ * 
+ * @param host_key
+ *     The known public host key of the server, as provided by the client.  If
+ *     provided the identity of the server will be checked against this key,
+ *     and a mis-match between this and the server identity will cause the
+ *     connection to fail.  If not provided, no checks will be done and the
+ *     connection will proceed.
+ * 
+ * @param credential_handler
+ *     The handler function for retrieving additional credentials from the user
+ *     as required by the SSH server.
  *
  * @return
  *     A new SSH session if the connection and authentication succeed, or NULL
  *     if the connection or authentication were not successful.
  */
 guac_common_ssh_session* guac_common_ssh_create_session(guac_client* client,
-        const char* hostname, const char* port, guac_common_ssh_user* user, int keepalive,
-        const char* host_key, guac_ssh_credential_handler* credential_callback);
+        const char* hostname, const char* port, guac_common_ssh_user* user,
+        int keepalive, const char* host_key,
+        guac_ssh_credential_handler* credential_handler);
 
 /**
  * Disconnects and destroys the given SSH session, freeing all associated
--- a/src/common-ssh/ssh.c
+++ b/src/common-ssh/ssh.c
@@ -304,20 +304,18 @@ static int guac_common_ssh_authenticate(guac_common_ssh_session* common_session)
     LIBSSH2_SESSION* session = common_session->session;
 
     /* Get user credentials */
-    char* username = user->username;
-    char* password = user->password;
     guac_common_ssh_key* key = user->private_key;
 
     /* Validate username provided */
-    if (username == NULL) {
+    if (user->username == NULL) {
         guac_client_abort(client, GUAC_PROTOCOL_STATUS_CLIENT_UNAUTHORIZED,
                 "SSH authentication requires a username.");
         return 1;
     }
 
     /* Get list of supported authentication methods */
-    char* user_authlist = libssh2_userauth_list(session, username,
-            strlen(username));
+    char* user_authlist = libssh2_userauth_list(session, user->username,
+            strlen(user->username));
 
     /* If auth list is NULL, then authentication has succeeded with NONE */
     if (user_authlist == NULL) {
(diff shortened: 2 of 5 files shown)


src/common-ssh/ssh.c
```diff
@@ -321,6 +321,10 @@ static int guac_common_ssh_authenticate(guac_common_ssh_session* common_session)
     guac_client_log(client, GUAC_LOG_DEBUG,
             "Supported authentication methods: %s", user_authlist);
 
+    /* If auth list is NULL, then authentication has succeeded with NONE */
+    if (user_authlist == NULL)
```
mike-jumper: If `user_authlist` is `NULL`, what happens to that `guac_client_log()` call above that tries to print the value?

mike-jumper:
I'll try to spin up a server and test.

src/common-ssh/ssh.c
necouchman: Good point.  Perhaps that NULL check should be located elsewhere...

necouchman:
> I'll try to spin up a server and test.

If you're able to, can you provide documentation on how to do this?  I spent some time on the Google machine trying to find instructions for it, and could never figure out how to configure OpenSSH to support NONE authentication method.

src/common-ssh/ssh.c
necouchman: Relocated and logged the NONE success.

mike-jumper:
OK - I think I figured it out:

1. Prepare a server that you're OK with having essentially no protection.
2. Edit `/etc/ssh/sshd_config` such that the following are set to `yes`:

       PasswordAuthentication yes
       PermitEmptyPasswords yes

3. Create an arbitrary user and use `passwd` to delete their password (making the account passwordless):

       $ sudo useradd test
       $ sudo passwd -d test
       Removing password for user test.
       passwd: Success
       $

Authentication for that specific user is then disabled for SSH and you should be able to log in as that user with any SSH client that supports the "NONE" method.

I tried the above on a fresh and isolated CentOS 7 machine and can reproduce the same behavior described on [GUACAMOLE-547](https://issues.apache.org/jira/browse/GUACAMOLE-547), and can confirm that your changes here allow authentication to succeed.

... **BUT:** With these changes in place, Guacamole still prompts the user for a password, unlike the OpenSSH client which just allows the session to begin. Things probably need further rearranging for Guacamole's behavior to be truly 100% correct for SSH NONE.

mike-jumper:
If it helps, I can open up the test SSH server for you to test against. I'd need to know the IP address or subnet you'll be testing from, though.

necouchman:
> With these changes in place, Guacamole still prompts the user for a password, unlike the OpenSSH client which just allows the session to begin. Things probably need further rearranging for Guacamole's behavior to be truly 100% correct for SSH NONE.

Okay, I'll spend a little more time trying to rearrange things to see what I can figure out.  I'm guessing the `libssh2_userauth_list()` call just needs to be moved up some (or the call for the password needs to be moved down), so I'll try to rearrange things.

> If it helps, I can open up the test SSH server for you to test against. I'd need to know the IP address or subnet you'll be testing from, though.

Appreciate the offer, I'll let you know if I need it.  I think with your instructions above I should be able to set something up in my test environment that I can use to reproduce this.

mike-jumper:
> Okay, I'll spend a little more time trying to rearrange things to see what I can figure out. I'm guessing the `libssh2_userauth_list()` call just needs to be moved up some (or the call for the password needs to be moved down), so I'll try to rearrange things.

Since the password prompt is currently started by the SSH protocol implementation (`src/protocols/ssh/`) and not by the common SSH client (`src/common-ssh/`), I think we may need to provide some sort of callback to allow `guac_common_ssh_authenticate()` to request additional credentials which the caller can then satisfy.

Some similar means for the caller to respond to additional credential requests from the SSH server would also eventually be necessary for [GUACAMOLE-141](https://issues.apache.org/jira/browse/GUACAMOLE-141) ("Complete support for keyboard interactive authentication").

necouchman:
> Since the password prompt is currently started by the SSH protocol implementation (src/protocols/ssh/) and not by the common SSH client (src/common-ssh/), I think we may need to provide some sort of callback to allow guac_common_ssh_authenticate() to request additional credentials which the caller can then satisfy.

Okay, sounds like a good challenge :-).  I'll see what I can figure out.

necouchman:
Followed your instructions for configuring a user that results in NONE authentication, so will try to figure out the best way to organize things to avoid the password prompt.  Presumably it would also be nice to do this in a way that the SFTP support would work, as well, both for SSH connections and for VNC or RDP using SFTP for file transfer.

necouchman:
Well, I'm trying to make this work, but having trouble figuring out the best place to put the callback function.  I was trying to re-use the existing `guac_terminal_prompt` function and put it in the `guac_common_ssh_session` struct, but then I have to have the terminal, which means retrieving the `guac_ssh_client` struct from within the common SSH code, which seems like something that shouldn't be done.  I'll keep playing with it, but, in the meantime, I'd be happy to take any advice or direction.

necouchman:
Okay, this now works with the call-back function, skipping the password prompt for the NONE authentication method but prompting for a password if NONE and public key methods fail, and then using that password for password or keyboard-interactive.  It also sets the credential for use with the SFTP connection, and should correctly skip the SFTP connection on other protocols if a password is not provided (though I believe this will still kill the overall connection).

I'm sure it needs some fixing up, but hopefully it's going in the right direction?

necouchman:
@mike-jumper What else needs to be fixed, here?

src/protocols/ssh/ssh.c
```diff
@@ -130,26 +130,36 @@ static guac_common_ssh_user* guac_ssh_get_user(guac_client* client) {
...
-
     /* Clear screen of any prompts */
     guac_terminal_printf(ssh_client->term, "\x1B[H\x1B[J");
 
     return user;
 
 }
 
+/**
+ * A call-back function used to gather additional credentials from a client
```
mike-jumper: This looks like documentation intended for the `guac_ssh_credential_handler` type. It makes sense from a generic perspective, but the information provided is only sufficient for a developer trying to implement such a callback, not for a developer potentially wanting to use this existing implementation.

It should be clear from the documentation of the function why this particular implementation is appropriate from the perspective of a developer that may want to use this. It prompts the user for credentials using the terminal emulator associated with the SSH connection via `guac_client`? Probably good to note.

src/protocols/ssh/ssh.c
```diff
@@ -130,26 +130,36 @@ static guac_common_ssh_user* guac_ssh_get_user(guac_client* client) {
...
+ *     where additional credentials are required.
+ * 
+ * @param credName
+ *     The prompt text to display to the screen when prompting for the
+ *     additional credentials.
+ * 
+ * @return 
+ *     The string of credentials gathered from the user.
+ */
+char* guac_ssh_get_credential(guac_client *client, char* credName) {
```
mike-jumper: See above - the C convention for variables is `lowercase_with_underscores`.

src/common-ssh/ssh.c
```diff
@@ -416,7 +427,7 @@ static int guac_common_ssh_authenticate(guac_common_ssh_session* common_session)
 
 guac_common_ssh_session* guac_common_ssh_create_session(guac_client* client,
         const char* hostname, const char* port, guac_common_ssh_user* user, int keepalive,
-        const char* host_key) {
+        const char* host_key, guac_ssh_credential_handler* credential_handler) {
```
mike-jumper: This is called `credential_callback` in the prototype. Either is fine with me, but the prototype and function body should be consistent.

src/common-ssh/common-ssh/ssh.h
```diff
@@ -25,6 +25,11 @@
 #include <guacamole/client.h>
 #include <libssh2.h>
 
+/**
+ * Handler for retrieving additional credentials.
```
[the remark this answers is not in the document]
necouchman: Documented.

src/common-ssh/common-ssh/ssh.h
```diff
@@ -25,6 +25,11 @@
...
+ */
+typedef char* guac_ssh_credential_handler(guac_client* client, char* credName);
```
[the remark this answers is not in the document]
necouchman: good_now

src/common-ssh/common-ssh/ssh.h
```diff
@@ -99,7 +109,7 @@ void guac_common_ssh_uninit();
  */
 guac_common_ssh_session* guac_common_ssh_create_session(guac_client* client,
         const char* hostname, const char* port, guac_common_ssh_user* user, int keepalive,
-        const char* host_key);
+        const char* host_key, guac_ssh_credential_handler* credential_callback);
```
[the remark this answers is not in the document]
necouchman: Documented the new one, plus all of the missing ones.

src/common-ssh/ssh.c
```diff
@@ -352,14 +359,18 @@ static int guac_common_ssh_authenticate(guac_common_ssh_session* common_session)
 
     }
 
+    /* Down to username + password authentication. */
```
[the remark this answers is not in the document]
necouchman: Down with the rewording.

src/protocols/ssh/ssh.c
necouchman: Reworded this a bit and added the notes as suggested.

src/common-ssh/ssh.c
necouchman: Went with `credential_handler` across the board.

src/common-ssh/ssh.c
```diff
@@ -304,23 +304,30 @@ static int guac_common_ssh_authenticate(guac_common_ssh_session* common_session)
...
+    if (user_authlist == NULL) {
+        guac_client_log(client, GUAC_LOG_DEBUG,
+            "SSH NONE authentication succeeded.");
+        return 0;
+    }
+
     guac_client_log(client, GUAC_LOG_DEBUG,
             "Supported authentication methods: %s", user_authlist);
 
+
```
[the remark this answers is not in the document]
necouchman: Ummm....because...

Removed.

src/protocols/ssh/ssh.c
necouchman: Changed, here, too.

necouchman:
@mike-jumper Okay, fixed up the stuff you requested - let me know if anything else needs work.  Also rebased on top of current `staging/1.1.0` branch.

src/common-ssh/common-ssh/ssh.h
```diff
@@ -25,6 +25,20 @@
...
+ * Handler for retrieving additional credentials.
+ * 
+ * @param client
+ *     The Guacamole Client associated with this need for additional
+ *     credentials.
+ * 
+ * @param cred_name
+ *     The name of the credential being requested, which will be shared
+ *     with the client in order to generate a meaningful prompt.
+ *
```
mike-jumper: Missing docs for the return value.

If the returned string will eventually be freed with a call to `free()` and is thus required to be dynamically allocated, that should be documented as well.

necouchman: Added that documentation.

mike-jumper: Looks better, but:

> ... which should be a dynamically-allocated such that it can be freed as required.

More than that, the returned value _must_ be dynamically-allocated, as the caller will eventually attempt to free it with a call to `free()`. For example:

https://github.com/apache/guacamole-server/blob/5e2ddb890a90e840a419f4878381b0c511cc6f25/src/terminal/terminal/terminal.h#L664-L670

and

https://github.com/apache/guacamole-server/blob/6dad6cd91933a19bc18095854e93354b6764d763/src/protocols/rdp/tests/fs/normalize_path.c#L174-L179

While we're here: is there a meaning for returning `NULL`? If not, that's fine, and the documentation covers what it needs to cover. Well-behaved implementations won't return `NULL` unless it's actually documented as a legal return value. However, if an implementation returning `NULL` is intended and has a specific meaning, that should be documented here, too.

src/common-ssh/common-ssh/ssh.h
```diff
@@ -92,14 +114,31 @@ void guac_common_ssh_uninit();
...
+ * @param host_key
+ *     The known public host key of the server, as provided by the client.  If
+ *     provided the identity of the server will be checked against this key,
+ *     and a mis-match between this and the server identity will cause the
+ *     connection to fail.  If not provided, no checks will be done and the
+ *     connection will proceed.
+ * 
+ * @param credential_handler
+ *     The handler function for retrieving additional credentials from the user
+ *     as required by the SSH server.
```
mike-jumper: Probably worth adding "or NULL if ...", as that is an intended use.

src/common-ssh/common-ssh/ssh.h
necouchman: Okay, reworked.

src/common-ssh/common-ssh/ssh.h
necouchman: Added that bit in.

PR #204, opened 2018-11-21
GUACAMOLE-354: Add Swiss-German keymap for RDP

andrin-jost:


mike-jumper:
Looks good to me.

src/protocols/rdp/keymaps/de_ch_qwertz.keymap
```diff
@@ -0,0 +1,58 @@
...
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+#
+
+parent  "base"
+name    "ch-de-qwertz"
```
mike-jumper: The keyboard name should be LANG-COUNTRY-VARIANT, same as the filename except with dashes instead of underscores. Is "de-ch-qwertz" correct here or "ch-de-qwertz"?

andrin-jost: Oh, yes you're right

PR #205, opened 2018-11-26
GUACAMOLE-667: Enable loading configuration file from other locations than default.

andrzejdoro:
These commits enable setting path to the configuration file during **guacd** startup:

- by "-c path/to/conf-file" command line argument
- by "GUACD_CONF_FILE" environment variable

Ccommand line arg has precedence. If none of above is set then default location is used.

---

After accepting these changes the manual should be updated (both _man guacd_ and _man guacd.conf_).

### DIFF
--- a/src/guacd/conf-args.c
+++ b/src/guacd/conf-args.c
@@ -32,10 +32,15 @@ int guacd_conf_parse_args(guacd_config* config, int argc, char** argv) {
 
     /* Parse arguments */
     int opt;
-    while ((opt = getopt(argc, argv, "l:b:p:L:C:K:fv")) != -1) {
+    while ((opt = getopt(argc, argv, "c:l:b:p:L:C:K:fv")) != -1) {
+
+        /* -c: Configuration file */
+        if (opt == 'c') {
+            continue;
+        }
 
         /* -l: Bind port */
-        if (opt == 'l') {
+        else if (opt == 'l') {
             free(config->bind_port);
             config->bind_port = strdup(optarg);
         }
@@ -102,6 +107,7 @@ int guacd_conf_parse_args(guacd_config* config, int argc, char** argv) {
         else {
 
             fprintf(stderr, "USAGE: %s"
+                    " [-c CONFIG_FILE]"
                     " [-l LISTENPORT]"
                     " [-b LISTENADDRESS]"
                     " [-p PIDFILE]"
@@ -122,3 +128,15 @@ int guacd_conf_parse_args(guacd_config* config, int argc, char** argv) {
 
 }
 
+const char* guacd_conf_get_path(int argc, char** argv) {
+
+    /* Find -c option */
+    int i;
+    for(i=1; i<argc-1; i++) {
+        if(strcmp("-c", argv[i]) == 0) return argv[i+1];
+    }
+
+    /* Not found */
+    return NULL;
+
+}
--- a/src/guacd/conf-file.c
+++ b/src/guacd/conf-file.c
@@ -189,7 +189,10 @@ guacd_config* guacd_conf_load(const char* conf_file_path) {
 #endif
 
     /* Determine path of configuration file */
-    if(conf_file_path == NULL) conf_file_path = GUACD_CONF_FILE;
+    if(conf_file_path == NULL) {
+        conf_file_path = getenv("GUACD_CONF_FILE");
+        if(conf_file_path == NULL) conf_file_path = GUACD_CONF_FILE;
+    }
 
     /* Read configuration from file */
     int fd = open(conf_file_path, O_RDONLY);
--- a/src/guacd/conf-file.c
+++ b/src/guacd/conf-file.c
@@ -169,7 +169,7 @@ int guacd_conf_parse_file(guacd_config* conf, int fd) {
 
 }
 
-guacd_config* guacd_conf_load() {
+guacd_config* guacd_conf_load(const char* conf_file_path) {
 
     guacd_config* conf = malloc(sizeof(guacd_config));
     if (conf == NULL)
@@ -188,15 +188,18 @@ guacd_config* guacd_conf_load() {
     conf->key_file = NULL;
 #endif
 
+    /* Determine path of configuration file */
+    if(conf_file_path == NULL) conf_file_path = GUACD_CONF_FILE;
+
     /* Read configuration from file */
-    int fd = open(GUACD_CONF_FILE, O_RDONLY);
+    int fd = open(conf_file_path, O_RDONLY);
     if (fd > 0) {
 
         int retval = guacd_conf_parse_file(conf, fd);
         close(fd);
 
         if (retval != 0) {
-            fprintf(stderr, "Unable to parse \"" GUACD_CONF_FILE "\".\n");
+            fprintf(stderr, "Unable to parse \"%s\".\n", conf_file_path);
             free(conf);
             return NULL;
         }
@@ -205,7 +208,7 @@ guacd_config* guacd_conf_load() {
 
     /* Notify of errors preventing reading */
     else if (errno != ENOENT) {
-        fprintf(stderr, "Unable to open \"" GUACD_CONF_FILE "\": %s\n", strerror(errno));
+        fprintf(stderr, "Unable to open \"%s\": %s\n", conf_file_path, strerror(errno));
         free(conf);
         return NULL;
     }
--- a/src/guacd/conf-file.h
+++ b/src/guacd/conf-file.h
@@ -32,10 +32,11 @@
 int guacd_conf_parse_file(guacd_config* conf, int fd);
 
 /**
- * Loads the configuration from any of several default locations, if found. If
- * parsing fails, NULL is returned, and an error message is printed to stderr.
+ * Loads the configuration from the given path or from the default location, if
+ * NULL is passed. Configuration is loaded only if the specified file is found.
+ * If parsing fails, NULL is returned, and an error message is printed to stderr.
  */
-guacd_config* guacd_conf_load();
+guacd_config* guacd_conf_load(const char* conf_file_path);
 
 #endif
 
--- a/src/guacd/conf-args.h
+++ b/src/guacd/conf-args.h
@@ -30,5 +30,11 @@
  */
 int guacd_conf_parse_args(guacd_config* config, int argc, char** argv);
 
+/**
+ * Finds a path to a configuration file in the given arguments. The path is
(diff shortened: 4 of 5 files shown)


src/guacd/conf-file.c
```diff
@@ -188,15 +188,21 @@ guacd_config* guacd_conf_load() {
     conf->key_file = NULL;
 #endif
 
+    /* Determine path of configuration file */
+    if(conf_file_path == NULL) {
```
necouchman: Two issues, here:
* If is not a function...

`if (conf_file_path == NULL) {`

* Why do this particular check, here?  Why not always make sure this function is fed a path, and then upstream feed either the provided location or the constant `GUACD_CONF_FILE`?

src/guacd/conf-file.c
```diff
@@ -188,15 +188,21 @@ guacd_config* guacd_conf_load() {
...
+        conf_file_path = getenv("GUACD_CONF_FILE");
```
necouchman: Why the change from the constant `GUACD_CONF_FILE` to `getenv("GUACD_CONF_FILE")`?  I believe this changes from using the constant to assuming that it is defined within the environment??

src/guacd/conf-args.c
```diff
@@ -122,3 +128,15 @@ int guacd_conf_parse_args(guacd_config* config, int argc, char** argv) {
...
+const char* guacd_conf_get_path(int argc, char** argv) {
+
+    /* Find -c option */
+    int i;
+    for(i=1; i<argc-1; i++) {
+        if(strcmp("-c", argv[i]) == 0) return argv[i+1];
+    }
+
+    /* Not found */
+    return NULL;
```
necouchman: See comments below, but I would argue for making this function return either the path provided on the command line with the `-c /path/to/config/file` argument, or the `GUACD_CONF_FILE` constant.

src/guacd/conf-args.c
```diff
@@ -122,3 +128,15 @@ int guacd_conf_parse_args(guacd_config* config, int argc, char** argv) {
 
 }
 
+const char* guacd_conf_get_path(int argc, char** argv) {
+
+    /* Find -c option */
+    int i;
+    for(i=1; i<argc-1; i++) {
+        if(strcmp("-c", argv[i]) == 0) return argv[i+1];
```
mike-jumper: Manually re-parsing the command line arguments seems a poor approach. Why not parse within `guacd_conf_parse_args()` fully leveraging `getopt()`? Couldn't this setting live on the `guacd_config` struct like the others?

src/guacd/conf-file.c
mike-jumper: > Why do this particular check, here? Why not always make sure this function is fed a path, and then upstream feed either the provided location or the constant `GUACD_CONF_FILE`?

I agree with this. This doesn't seem the place for a `NULL` check and additional semantics. It would be far better if the default value for the config file path were handled like other defaults.

I'd argue that it's the current handling of defaults that should be restructured - bringing that out of `guacd_conf_load()` such that the process becomes:

1. Populate `guacd_config` with defaults.
2. Apply any settings from config file to that `guacd_config`, overriding default values.
3. Apply any settings from command line to that `guacd_config`, overriding default values and those from the config file.

See:

https://github.com/apache/guacamole-server/blob/bbb6afaf462f6930a589f962302806c52f350c0b/src/guacd/conf-file.c#L178-L184

src/guacd/conf-args.c
andrzejdoro: Yes, I haven't thought about it. It would be better approach.
I will prepare a new commit.

src/guacd/conf-file.c
andrzejdoro: This is extension - one can define `"GUACD_CONF_FILE"` environment variable, and if it is not present there, then `GUACD_CONF_FILE` constant is used (as it was done before) - see one line below.

This enables specifying path to config file by "-c" argument or by environment variable (both approaches are quite common, we can discard one of them).

src/guacd/conf-file.c
andrzejdoro: I agree that this process could be improved.
Should it be done in separate pull-request or can I just add a new commit?

src/guacd/conf-args.c
andrzejdoro: > Why not parse within `guacd_conf_parse_args()` fully leveraging `getopt()`?

We need to find out what is the path to the configuration file before we parse command line arguments, since we have to apply any settings from config file before we apply any settings from command line (the latter overrides the former).

Running `while(getopt(...)...)` may cause permutation of arguments, that's why I've decided to manually parse arguments list.

> Couldn't this setting live on the `guacd_config` struct like the others?

I believe that `guacd_config` struct should store contents of a guacd configuration file so storing path to the configuration file is not needed inside it. We use it only once - when loading configuration file.

src/guacd/conf-file.c
necouchman: Ah, yes, seems I overlooked that part of the original message.

src/guacd/conf-args.c
mike-jumper: I really don't like the idea of ignoring `-c` initially and manually re-parsing the arguments later. There has to be a better way.

What about parsing any file specified with `-c` immediately upon reading that `-c`? Arguments parsed before the file would be overridden by that file, while arguments which follow would override further. This would also open up the possibility of reading multiple files through specifying `-c` multiple times.

src/guacd/conf-file.c
mike-jumper: Either way would probably be fine. If the changes would be distractingly complicated compared to the base goal of GUACAMOLE-667, then separating that refactor into its own pull request would be better.

src/guacd/conf-args.c
andrzejdoro: > I really don't like the idea of ignoring `-c` initially and manually re-parsing the arguments later. There has to be a better way.

More precisely, it's the other way round: at first we manually search for `-c` (then load config) and then we parse all the arguments. Nevertheless, I'm not happy with this solution either, but it seems that _getopt()_ is not helping.

> What about parsing any file specified with `-c` immediately upon reading that `-c`?

That's a nice solution! I'm not sure whether it's quite common behaviour, but we can try it. :+1:

andrzejdoro:
The proposed solution is ready:
- handling of creation and setting defaults is now independent of the loading of configuration file
- after each `-c` option, following configuration file is parsed

Two questions - I'll put them as comments in code view.

src/guacd/conf-file.h
```diff
@@ -24,6 +24,11 @@
 
 #include "conf.h"
 
+/**
+ * Creates a new configuration structure, filled with default settings.
+ */
+guacd_config* guacd_conf_create();
+
```
andrzejdoro: After changes, guacd_config creation is independent of parsing configuration files.
Should I put this function here or in `conf.h` (and implementation in **new** `conf.c` file)?

src/guacd/conf-file.h
```diff
@@ -32,10 +37,17 @@
 int guacd_conf_parse_file(guacd_config* conf, int fd);
```
andrzejdoro: What about hiding `guacd_conf_parse_file()` by making it static?
Now, `guacd_conf_load` provides this functionality on higher level.

src/guacd/conf-file.c
```diff
@@ -169,48 +192,40 @@ int guacd_conf_parse_file(guacd_config* conf, int fd) {
...
-    return conf;
+    /* Load successfully */
+    return 0;
 
 }
 
+int guacd_conf_load_default(guacd_config* conf) {
+
+    /* Determine path to the default configuration file and load it */
+    char *path = getenv("GUACD_CONF_FILE");
```
necouchman: This may be a bit of a nit-pick, but is loading the location of the config file from an environment variable really loading a default configuration file?

src/guacd/conf-file.h
necouchman: Even though it's independent of the parsing, it still seems like it's related to pulling values from the files, so I'm fine with it staying where it is.

src/guacd/conf-file.h
necouchman: My only comment here would be to stay consistent with how this is done elsewhere in the code.  Are other functions hidden using static, or just not stubbed in header files?  Go with the route that provides the most consistency, unless there's a really good reason to do otherwise.

src/guacd/conf-file.c
mike-jumper: I think, arguably, this accurately refers to the file as default in the sense that it's where the default configuration values come from.

src/guacd/conf-file.h
```diff
@@ -24,6 +24,11 @@
 
 #include "conf.h"
 
+/**
+ * Creates a new configuration structure, filled with default settings.
+ */
+guacd_config* guacd_conf_create();
```
mike-jumper, line 30: Please document the return value of this new function.

src/guacd/conf-file.h
```diff
@@ -32,10 +37,17 @@
...
+ * error message is printed to stderr.
+ */
+int guacd_conf_load(guacd_config* conf, const char* conf_file_path);
+
+/**
+ * Determines path to the default configuration file (environment variable or
+ * predefined path) and loads settings from it. On success, zero is returned.
  */
-guacd_config* guacd_conf_load();
+int guacd_conf_load_default(guacd_config* conf);
```
mike-jumper, line 50: While the functions defined here previously were not documented in this way, new code should use Doxygen-style comments to document the return values and parameters accepted by functions.

An example of the style used for new code is:

https://github.com/apache/guacamole-server/blob/ca1fbd5e980392dc91b52ddff473698035d8f357/src/protocols/rdp/keyboard.h#L207-L219

Another example, within guacd:

https://github.com/apache/guacamole-server/blob/ca1fbd5e980392dc91b52ddff473698035d8f357/src/guacd/connection.h#L103-L119

mike-jumper (changes requested):
With the exception of updating the documentation of new functions to match the new style, I think this looks fine.

The man page will need to be updated to cover the new option and environment variable. If up to it, I suggest doing this, as well:

https://github.com/apache/guacamole-server/blob/ca1fbd5e980392dc91b52ddff473698035d8f357/src/guacd/man/guacd.8.in

As the man page for `guacd.conf` specifically notes its location as `/etc/guacamole/guacd.conf`, documenting the environment variable there would be also good:

https://github.com/apache/guacamole-server/blob/ca1fbd5e980392dc91b52ddff473698035d8f357/src/guacd/man/guacd.conf.5.in

PR #206, opened 2018-12-01
GUACAMOLE-663: guacd SEGVs intermittently on systems with small(er) thread stack sizes

rvs:


necouchman:
This looks okay to me, but I'm not confident in my ability to review the configure.ac changes.  @mike-jumper anything that needs to be fixed before this one goes in?

src/guacd/daemon.c
```diff
@@ -19,6 +19,12 @@
 
 #include "config.h"
 
+#ifdef PTHREAD_STACK_SIZE
+#define _GNU_SOURCE 1
```
mike-jumper: I'm concerned over explicitly setting `_GNU_SOURCE` and using the non-portable `pthread_setattr_default_np()`. There's no guarantee that `pthread_setattr_default_np()` will be available, or that it will be available if `_GNU_SOURCE` is set.

Testing for these within configure would be an improvement, but the non-portability is still concerning. Is there no portable way to do this?

src/guacd/daemon.c
```diff
@@ -275,6 +281,17 @@ int main(int argc, char* argv[]) {
...
     int retval;
 
+#ifdef PTHREAD_STACK_SIZE
+    /* Set default stack size */
+    pthread_attr_t default_pthread_attr;
+    if (pthread_attr_init(&default_pthread_attr) ||
+        pthread_attr_setstacksize(&default_pthread_attr, PTHREAD_STACK_SIZE) ||
+        pthread_setattr_default_np(&default_pthread_attr)) {
+       fprintf(stderr, "Couldn't set requested pthread stack size of %d\n", PTHREAD_STACK_SIZE);
+       exit(EXIT_FAILURE);
```
mike-jumper: Please use 4-space indents per level. See: http://guacamole.apache.org/guac-style/#general-style

configure.ac
```diff
@@ -140,6 +140,46 @@ AC_SUBST([COMMON_SSH_INCLUDE], '-I$(top_srcdir)/src/common-ssh')
 AC_SUBST([TERMINAL_LTLIB],   '$(top_builddir)/src/terminal/libguac_terminal.la')
 AC_SUBST([TERMINAL_INCLUDE], '-I$(top_srcdir)/src/terminal $(PANGO_CFLAGS) $(PANGOCAIRO_CFLAGS) $(COMMON_INCLUDE)')
 
+# pthread stack size
+AC_ARG_WITH(pthread_stack,
+            [AS_HELP_STRING([--with-pthread_stack=<size in bytes>],
+                            [explicitly set pthread stack size (8MB is recommended)])
```
mike-jumper: This may be confusing as "8MB" is not a value accepted by this option.

configure.ac
```diff
@@ -140,6 +140,46 @@ AC_SUBST([COMMON_SSH_INCLUDE], '-I$(top_srcdir)/src/common-ssh')
...
+            ],pthread_stack_size=$withval
+              AC_DEFINE_UNQUOTED([PTHREAD_STACK_SIZE], [$pthread_stack_size], [pthread stack size (8MB is recommended)])
+)
+if test "x$ac_cv_lib_pthread_pthread_create" = xyes; then
+        AC_MSG_CHECKING([whether default pthread stack is larger than 8MB])
+        ac_save_libs="$LIBS"
+        LIBS="$LIBS -lpthread"
```
mike-jumper: Not all platforms use libpthread for POSIX threads. Some have this built into libc. If necessary, you might be able to use `$PTHREAD_LIBS` which is set by an earlier test:

https://github.com/apache/guacamole-server/blob/bbb6afaf462f6930a589f962302806c52f350c0b/configure.ac#L61-L64

configure.ac
```diff
@@ -140,6 +140,46 @@ AC_SUBST([COMMON_SSH_INCLUDE], '-I$(top_srcdir)/src/common-ssh')
...
 AC_SUBST([TERMINAL_INCLUDE], '-I$(top_srcdir)/src/terminal $(PANGO_CFLAGS) $(PANGOCAIRO_CFLAGS) $(COMMON_INCLUDE)')
 
+# pthread stack size
+AC_ARG_WITH(pthread_stack,
+            [AS_HELP_STRING([--with-pthread_stack=<size in bytes>],
+                            [explicitly set pthread stack size (8MB is recommended)])
+            ],pthread_stack_size=$withval
+              AC_DEFINE_UNQUOTED([PTHREAD_STACK_SIZE], [$pthread_stack_size], [pthread stack size (8MB is recommended)])
+)
+if test "x$ac_cv_lib_pthread_pthread_create" = xyes; then
```
mike-jumper: Where does `ac_cv_lib_pthread_pthread_create` come from?

configure.ac
```diff
@@ -140,6 +140,46 @@ AC_SUBST([COMMON_SSH_INCLUDE], '-I$(top_srcdir)/src/common-ssh')
...
+                return stacksize < 8*1024*1024;
+                #endif
+            }
+
+          ]])],
+          [AC_MSG_RESULT([yes])],
+          [AC_MSG_RESULT([no])
+           AC_MSG_WARN([
+  --------------------------------------------
+   Default pthread stack size is less than 8MB
```
mike-jumper: Do we know for certain that 8MB satisfies our stack requirements?

configure.ac
```diff
@@ -140,6 +140,46 @@ AC_SUBST([COMMON_SSH_INCLUDE], '-I$(top_srcdir)/src/common-ssh')
...
+            [AS_HELP_STRING([--with-pthread_stack=<size in bytes>],
+                            [explicitly set pthread stack size (8MB is recommended)])
+            ],pthread_stack_size=$withval
+              AC_DEFINE_UNQUOTED([PTHREAD_STACK_SIZE], [$pthread_stack_size], [pthread stack size (8MB is recommended)])
+)
+if test "x$ac_cv_lib_pthread_pthread_create" = xyes; then
+        AC_MSG_CHECKING([whether default pthread stack is larger than 8MB])
+        ac_save_libs="$LIBS"
+        LIBS="$LIBS -lpthread"
+        AC_RUN_IFELSE([AC_LANG_SOURCE([[
```
mike-jumper: [The documentation for `AC_RUN_IFELSE`](https://www.gnu.org/software/autoconf/manual/autoconf-2.68/html_node/Runtime.html) states:

> Avoid running test programs if possible, because this prevents people from configuring your package for cross-compiling.

Is there any other way to achieve this which would not hurt cross-compilation?

src/guacd/daemon.c
```diff
@@ -275,6 +281,17 @@ int main(int argc, char* argv[]) {
     /* General */
     int retval;
 
+#ifdef PTHREAD_STACK_SIZE
+    /* Set default stack size */
+    pthread_attr_t default_pthread_attr;
+    if (pthread_attr_init(&default_pthread_attr) ||
+        pthread_attr_setstacksize(&default_pthread_attr, PTHREAD_STACK_SIZE) ||
+        pthread_setattr_default_np(&default_pthread_attr)) {
+       fprintf(stderr, "Couldn't set requested pthread stack size of %d\n", PTHREAD_STACK_SIZE);
```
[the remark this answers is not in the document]
rvs: Good point. Will fix both of these.

src/guacd/daemon.c
rvs: pthread_setattr_default_np() is available for quite sometime with _GNU_SOURCE so I don't think this should be that big of a concern. To make it reliable I can add testing for pthread_setattr_default_np to the configure part of this patch. Consider this done.

As for the more portable way of doing this -- the only way I know of is to add add explicit setting of the stack size right before each call of pthread_create. Please let me know if you'd prefer this way of doing it (I can see myself wrapping it up in a single function of course).

src/guacd/daemon.c
rvs: Got it! Will fix.

configure.ac
rvs: Would you prefer it to read "8388608 (8MB)"

configure.ac
rvs: I'm not sure I follow this comment -- this is exactly why I'm using.

configure.ac
rvs: See the comment above. It is automagically created by autoconf because of the test you quoted (testing for pthread_create in a library called pthread).

configure.ac
rvs: Yes we do. This is the default with glibc.

configure.ac
rvs: I honestly don't know if there is. I couldn't come up with one. That said -- this is basically a runtime check for a warning message. As such I can remove it (while leaving the rest of the patch). It would, however, be *very* unfortunate to do so since in my personal example it cost me a day to figure out that it was the stack size that was killing guacd on alpine. Anybody building guacd in the future on a similar platform will surely appreciate getting this type of a message early.

ON THE OTHER HAND -- if you're willing to make it a firm guarantee -- not an option that this size of a stack is alway requested -- perhaps we can simplify this patch quite a bit. Is this something that would be better discussed on the mailing list?

mike-jumper: > ON THE OTHER HAND -- if you're willing to make it a firm guarantee -- not an option that this size of a stack is alway requested -- perhaps we can simplify this patch quite a bit.

Making this a firm guarantee sounds like the best idea, particularly since you've demonstrated things simply don't work unless the stack size is at least 8MB. I'd suggest:

1. If default stack size is less than 8MB, guacd should attempt to set the default stack size to 8MB.
2. If the default stack size cannot be set (either because setting the attribute fails or `pthread_setattr_default_np()` is unavailable, a warning should be logged.

> Is this something that would be better discussed on the mailing list?

If there are other angles to consider, hopping onto the mailing list is always a good thing.

configure.ac
mike-jumper: What I mean is the failure of this test (whether `pthread_create()` is defined within libpthread) does not mean `pthread_create()` is not present at all, nor that the test for `pthread_setattr_default_np()` need not be performed.

This test is essentially a check for whether libpthread needs to be linked in. If it fails, `pthread_create()`, etc. are still used; POSIX threads are then simply assumed to be part of libc.

configure.ac
mike-jumper: That would probably be better (to avoid users trying things like `--with-pthread_stack=8MB`), though this may no longer be necessary based on your latest suggestion to simplify everything. I agree that it doesn't make sense to have a low-level option to tweak the stack size if it's known that the stack size absolutely must be at least 8MB.

configure.ac
necouchman: Ping @rvs - Any progress on this?

allquixotic:
My `ulimit -s` returns `8192` (I think it's Kbytes, so 8 MB) and I get general protection faults in session `guacd`s (not the parent `guacd`, though) somewhat intermittently. I set it to 65536 (for both hard and soft limit) and I still get intermittent GPFs. It's an "error:0 in libc-2.2.7.so" which could very well be a stack overflow, or it might be something else. Not sure.

Would setting `ulimit -s 65536` before running `guacd` potentially fix this problem while this code is still being updated/reviewed? If so, I guess having this problem means my build either has a stack memory leak, or guacd actually needs much more than 8 MiB of stack to be stable, or `ulimit -s` isn't applying to guacd.

mike-jumper:
@allquixotic, please use the dev@ or user@ mailing list if you have questions:

http://guacamole.apache.org/support/#mailing-lists

There are no known issues with the behavior you describe, and it's unlikely that what you are seeing is related to this change. It's also unlikely that a bug as fundamental as complete instability in the RDP support exists, and more likely that the cause is something else:

http://guacamole.apache.org/faq/#probably-not-a-bug

There are other factors which may cause a segfault, including accidentally mixing binaries from different guacamole-server builds. If there is a bug causing what you're seeing, it's worth chasing down, but this pull request is not the place for that.

jmuehlner:
Should we retarget this at `main`? It hasn't had any updates for 5 years so it's unclear if it's actually ready for 1.6.0 yet or not.

necouchman:
@jmuehlner I think this actually may be closed by #360, which was merged a couple of weeks ago.

mike-jumper:
Per above, closing as superseded by #360.

PR #207, opened 2018-12-04
GUACAMOLE-673: Create automatically needed directory for file transfer.

dejan032:
Automatically create transfer folder in guacd file system when user is logged into rdp, so that the transfer network drive can be used immediately without needing to open it first one time to start copying things into and out of it.

### DIFF
--- a/src/protocols/rdp/guac_rdpdr/rdpdr_fs_service.c
+++ b/src/protocols/rdp/guac_rdpdr/rdpdr_fs_service.c
@@ -153,5 +153,42 @@ void guac_rdpdr_register_fs(guac_rdpdrPlugin* rdpdr, char* drive_name) {
     /* Init data */
     device->data = rdp_client->filesystem;
 
+	/* Init directory for file transfer */
+	int create_fs = 0;
+
+	if (rdp_client) {
+		guac_rdp_settings* settings = rdp_client->settings;
+		if (settings) {
+			create_fs = settings->create_drive_path;
+		}
+		else {
+			guac_client_log(device->rdpdr->client, GUAC_LOG_ERROR, "No settings.");
+		}
+	}
+	else {
+		guac_client_log(device->rdpdr->client, GUAC_LOG_ERROR, "No rdp client.");
+	}
+
+	if (create_fs) {
+
+		/* Get filesystem, return error if no filesystem */
+		guac_rdp_fs* fs = rdp_client->filesystem;
+		if (fs == NULL) {
+			guac_client_log(device->rdpdr->client, GUAC_LOG_ERROR, "No filesystem.");
+			return;
+		}
+
+		int file_id;
+		char* file_path = "\\";
+
+		/* create file path */
+		file_id = guac_rdp_fs_open(fs, file_path, ACCESS_GENERIC_ALL, 0, DISP_FILE_OPEN_IF, FILE_DIRECTORY_FILE);
+		if (file_id < 0) {
+			guac_client_log(device->rdpdr->client, GUAC_LOG_ERROR, "Init client remote directory failed.");
+		}
+		else {
+			guac_client_log(device->rdpdr->client, GUAC_LOG_INFO, "Init client remote directory succeeded.");
+		}
+	}
 }
 


src/protocols/rdp/guac_rdpdr/rdpdr_fs_service.c
```diff
@@ -153,5 +153,42 @@ void guac_rdpdr_register_fs(guac_rdpdrPlugin* rdpdr, char* drive_name) {
     /* Init data */
     device->data = rdp_client->filesystem;
 
+	/* Init directory for file transfer */
```
necouchman: Just off the bat, please follow the style used throughout the rest of the code - each tab should be 4 spaces.

See: http://guacamole.apache.org/guac-style/

src/protocols/rdp/guac_rdpdr/rdpdr_fs_service.c
```diff
@@ -153,5 +153,42 @@ void guac_rdpdr_register_fs(guac_rdpdrPlugin* rdpdr, char* drive_name) {
...
+	int create_fs = 0;
+
+	if (rdp_client) {
```
necouchman: Not sure this check is necessary, here - rdp_client is used previously in here under the assumption that it is not null, so this is probably not necessary.

src/protocols/rdp/guac_rdpdr/rdpdr_fs_service.c
```diff
@@ -153,5 +153,42 @@ void guac_rdpdr_register_fs(guac_rdpdrPlugin* rdpdr, char* drive_name) {
...
+		}
+		else {
+			guac_client_log(device->rdpdr->client, GUAC_LOG_ERROR, "No settings.");
+		}
+	}
+	else {
+		guac_client_log(device->rdpdr->client, GUAC_LOG_ERROR, "No rdp client.");
+	}
+
+	if (create_fs) {
```
necouchman: My guess is that this and a bunch of the code above could be simplified into something like this:

    if (rdp_client->settings->create_drive_path)

Since both `rdp_client` and `rdp_client->settings` should be initialized and defined by this point in the code.

src/protocols/rdp/guac_rdpdr/rdpdr_fs_service.c
```diff
@@ -153,5 +153,42 @@ void guac_rdpdr_register_fs(guac_rdpdrPlugin* rdpdr, char* drive_name) {
...
+
+		/* Get filesystem, return error if no filesystem */
+		guac_rdp_fs* fs = rdp_client->filesystem;
+		if (fs == NULL) {
+			guac_client_log(device->rdpdr->client, GUAC_LOG_ERROR, "No filesystem.");
+			return;
```
necouchman: A couple of issues, here.  First, the `rdp_client->filesystem` is already assigned to `device->data` above - not sure there's any reason to re-assign it, here?  Second, I don't know that the null check is required, here, either.  I'd need to go back and look at the code that calls this `guac_rdpdr_register_fs` function, but I think it might be previously checked?

mike-jumper:
@dejan032, can you explain why this is necessary? I've read through your JIRA issue but what you describe sounds like the very point of the `create_drive_path` flag, which is handled the moment that the filesystem is allocated:

https://github.com/apache/guacamole-server/blob/bbb6afaf462f6930a589f962302806c52f350c0b/src/protocols/rdp/rdp_fs.c#L47-L59

That allocation happens before the connection is fully established, so the path will indeed exist before anything within the RDP session gains the ability to write to that path via the drive:

https://github.com/apache/guacamole-server/blob/0d435e243579f15c84ddca1d44bbb33c7269e39a/src/protocols/rdp/rdp.c#L914-L926

Outside of that, I also agree with at @necouchman: assuming these changes are necessary, please remove the paranoid `NULL` checks. We should be checking for `NULL` only when things are actually known to possibly be `NULL`.

dejan032:
@mike-jumper After reading your comment I've re-tested latest version of guacd and it works as expected. I had these issues with symlink and guacd transfer drive while code was still in incubation period and I 'fixed' it with my code. At least I thought so :), probably it was something Windows related on the remote machine. But as you said these changes are unnecessary. 
Thank you for you time and for clearing things up :)

PR #209, opened 2019-01-07
GUACAMOLE-637: Migrate strncpy(), strcat(), etc. to safer libguac implementations.

mike-jumper:
This change adds platform-independent libguac implementations of the `strlcpy()` and `strlcat()` functions, `guac_strlcpy()` and `guac_strlcat()`, provided via a new `guacamole/string.h` header. These functions will use the versions provided by the platform's libc, if available.

Unlike `strncpy()` and `strncat()`, the `strlcpy()` and `strlcat()` variants guarantee that the destination buffer is null-terminated, even if the result had to be truncated: https://www.freebsd.org/cgi/man.cgi?query=strlcpy&sektion=3&manpath=FreeBSD+6.2-RELEASE

This change also adds a similar convenience function, `guac_strljoin()`, which provides an implementation of the familiar "join" operation for strings.

Unit tests have been added for each of the new functions.


src/libguac/tests/string/strlcat.c
```diff
@@ -0,0 +1,153 @@
...
+ */
+
+#include <CUnit/CUnit.h>
+#include <guacamole/string.h>
+
+#include <stdlib.h>
+#include <string.h>
+
+/**
+ * Verify guac_strlcat() behavior when string fits buffer without truncation.
```
jmuehlner: It feels like there is a word or two from this sentence.

src/libguac/tests/string/strlcat.c
```diff
@@ -0,0 +1,153 @@
...
+ * The return value of each call should be the length of the resulting string.
+ * Each resulting string should contain the full result of the contatenation,
```
jmuehlner: "contatenation" is a fun word!

src/libguac/tests/string/strlcat.c
mike-jumper: [Why say lot word when few word do trick?](https://www.youtube.com/watch?v=_K-L9uhsBLM)

mike-jumper: ... but I'll fix it.

src/libguac/tests/string/strlcat.c
```diff
@@ -0,0 +1,153 @@
...
+
+}
+
+/**
+ * Verify guac_strlcat() behavior with unterminated buffers. With respect to
+ * the return value, the length of the string in the buffer should be
+ * considered equal to the size of the buffer, however the resulting buffer
+ * should not be null-terminated.
+ */
+void test_string__strlcat_nonull() {
```
jmuehlner: I'm not sure I understand this test. The function documentation requires that the buffer parameter MUST be null-terminated, and also guarantees that the buffer will be null-terminated. It sounds like at least that second part is being violated here.

src/libguac/tests/string/strlcpy.c
```diff
@@ -0,0 +1,100 @@
...
+ */
+
+#include <CUnit/CUnit.h>
+#include <guacamole/string.h>
+
+#include <stdlib.h>
+#include <string.h>
+
+/**
+ * Verify guac_strlcpy() behavior when string fits buffer without truncation.
```
jmuehlner: Words may missing from this sentence too.

src/libguac/tests/string/strlcat.c
mike-jumper: This tests a failsafe behavior of `strlcat()`. While the caller is required to ensure that the destination string is null-terminated, the function is also required to never overflow the buffer. In the case that the end of the buffer is encountered before a null terminator is found, the function must bail out rather than overrun.

This behavior is documented for [the FreeBSD version of `strlcat()`](https://www.freebsd.org/cgi/man.cgi?query=strlcat&sektion=3&apropos=0&manpath=FreeBSD+6.2-RELEASE), so we should probably do similar here:

> Note however, that if `strlcat()` traverses size characters without finding a NUL, the length of the string is considered to be size and the destination string will not be NUL-terminated (since there was no space for the NUL).  This keeps `strlcat()` from running off the end of a string.  In practice this should not happen (as it means that either size is incorrect or that dst is not a proper "C'' string).  The check exists to prevent potential security problems in incorrect code.

mike-jumper: Actually, with respect to the above, it seems our implementation differs from that implemented elsewhere. We should also update `guac_strlcat()` such that it returns size in this case, not size + length.

mike-jumper: It might be that the FreeBSD docs are not entirely correct. Checking elsewhere (OpenBSD?), I see the same behavior we implement (size + length), which does seem more consistent:

http://www.delorie.com/djgpp/doc/libc/libc_762.html

I'll try building on the BSDs and see whether (1) the platform's own version is used and (2) whether our unit tests succeed.

src/protocols/rdp/rdp_fs.c
```diff
@@ -607,11 +608,10 @@ const char* guac_rdp_fs_read_dir(guac_rdp_fs* fs, int file_id) {
 int guac_rdp_fs_normalize_path(const char* path, char* abs_path) {
 
     int i;
-    int path_depth = 0;
+    int path_depth = 1;
```
jmuehlner: What is this change about?

mike-jumper: I will check.

I should probably add unit tests for the path manipulation bits, too, honestly.

src/libguac/tests/string/strlcat.c
mike-jumper: Confirmed that FreeBSD has the same behavior as our implementations. The unit tests pass whether the platform's own `strlcpy()` / `strlcat()` are used or not. I had to make some additional changes for the sake of portability, but the implementation itself seems correct.

I'll clarify in the documentation for libguac.

src/libguac/tests/string/strlcat.c
mike-jumper: Fixed to be less fun.

src/libguac/tests/string/strlcat.c
mike-jumper: This specific safeguard is now documented for `guac_strlcat()`.

src/protocols/rdp/rdp_fs.c
mike-jumper: > What did you find?

The additional leading `""` path component is used to ensure the final, normalized path has a leading backslash following `guac_strljoin()`. I'm adding documentation to that effect.

> Are you planning on adding more unit tests in this PR?

Yes, to verify that these various normalization functions continue to function as intended.

necouchman:
Changed merge base to `staging/1.1.0` as this is slated to go into release 1.1.0.

src/protocols/rdp/rdp_fs.c
mike-jumper: OK - I've cleaned up the SFTP and RDP filesystem normalization logic and removed the need for setting `path_depth = 1`, which the unit tests I've just added revealed to be incorrect.

jmuehlner:
Ok, I think everything looks good now.

PR #213, opened 2019-02-09
GUACAMOLE-729: Bump version to 1.1.0.

mike-jumper:
As an interface has been changed (a new member has been added to the `guac_user` struct), the libtool version info had to be bumped from 16:0:0 to 17:0:0.


mike-jumper:
Good catch - I'll update it. `guacctl` actually did change following 1.0.0.

PR #214, opened 2019-02-19
GUACAMOLE-381: Add parameters for disabling clipboard copy/paste.

mike-jumper:
This change adds `disable-copy` and `disable-paste` parameters to all protocols, allowing copy/paste to be selectively disabled.

These parameters affect only the transfer of clipboard data through Guacamole. Within the remote desktop or terminal, text can still be copied or pasted independently.

### DIFF
--- a/src/protocols/rdp/rdp_settings.c
+++ b/src/protocols/rdp/rdp_settings.c
@@ -118,6 +118,8 @@ const char* GUAC_RDP_CLIENT_ARGS[] = {
     "load-balance-info",
 #endif
 
+    "disable-copy",
+    "disable-paste",
     NULL
 };
 
@@ -544,6 +546,20 @@ enum RDP_ARGS_IDX {
      */
     IDX_LOAD_BALANCE_INFO,
 #endif
+
+    /**
+     * Whether outbound clipboard access should be blocked. If set to "true",
+     * it will not be possible to copy data from the remote desktop to the
+     * client using the clipboard. By default, clipboard access is not blocked.
+     */
+    IDX_DISABLE_COPY,
+
+    /**
+     * Whether inbound clipboard access should be blocked. If set to "true", it
+     * will not be possible to paste data from the client to the remote desktop
+     * using the clipboard. By default, clipboard access is not blocked.
+     */
+    IDX_DISABLE_PASTE,
 
     RDP_ARGS_COUNT
 };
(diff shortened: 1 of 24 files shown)


src/protocols/rdp/rdp_settings.c
```diff
@@ -1007,6 +1023,16 @@ guac_rdp_settings* guac_rdp_parse_args(guac_user* user,
                 IDX_LOAD_BALANCE_INFO, NULL);
 #endif
 
+    /* Parse clipboard copy disable flag */
+    settings->disable_copy =
+        guac_user_parse_args_boolean(user, GUAC_RDP_CLIENT_ARGS, argv,
+                IDX_DISABLE_COPY, 0);
```
necouchman: Any particular reason, for the RDP flags, that you're parsing as `boolean` but using `0` instead of `false` for the default?  Looking at the other protocols you've used `false` - not sure why this is different?

src/protocols/rdp/rdp_settings.h
```diff
@@ -268,6 +268,20 @@ typedef struct guac_rdp_settings {
      */
     char** svc_names;
 
+    /**
+     * Whether outbound clipboard access should be blocked. If set, it will not
+     * be possible to copy data from the remote desktop to the client using the
+     * clipboard.
+     */
+    int disable_copy;
```
necouchman: As with the parsing, the other protocols use a `bool` type rather than an `int` - any particular reason for this?

mike-jumper:
> Also, I assume these changes do not impact compatibility across versions?

Correct.

src/protocols/rdp/rdp_settings.c
mike-jumper: Yep - the FreeRDP library defines its own boolean type which breaks any code using `stdbool.h`. We thus can't use the standard `bool` type in any code that touches FreeRDP.

This is why the `enable_sftp` flag is an `int` in the RDP support:

https://github.com/apache/guacamole-server/blob/802e5b5547dc26f1fa7d0abba91570f87cfefb32/src/protocols/rdp/rdp_settings.h#L353

but is `bool` elsewhere:

https://github.com/apache/guacamole-server/blob/802e5b5547dc26f1fa7d0abba91570f87cfefb32/src/protocols/ssh/settings.h#L161

src/protocols/rdp/rdp_settings.h
mike-jumper: Yep - [see above](https://github.com/apache/guacamole-server/pull/214#discussion_r259652831).

ibuler:
guacamole version  1.1.0  has no option set in page, and using api set it not work

mike-jumper:
@ibuler, it's not part of 1.1.0. See the JIRA issue and the 1.1.0 release notes.

ibuler:
@mike-jumper thanks

franckyz0r:
Hello @mike-jumper ,
Is there any way to disable copy-paste both ways (at least for RDP) without rebuilding the whole server files?
Or at least, can I build the server on another machine and replace only some files? Some hints would be great.

We have a production env and I don't really want to mess with it too much.

Thanks a lot,
Regards,

necouchman:
@franckyz0r: Please avoid responding on closed pull requests - if you have a question, it is best posed on the mailing list - see http://guacamole.apache.org/support/.

To answer your question, on, you cannot just replace files on the system to get this functionality - you will need to upgrade the software to the version that supports it, and these changes are required mainly on the guacamole-server (guacd) side, although parameters are also required on the server side, as well.

PR #216, opened 2019-02-23
GUACAMOLE-707: Build FreeRDP into guacd container

necouchman:
Since FreeRDP 1.1.0 doesn't contain support for RD Connection Brokers, and Debian Stable doesn't have packages for 1.2.0, I've taken a shot as resolving this by building FreeRDP in the container with guacd.  Not sure if it's a good or ideal solution, but it seems to work.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -31,15 +31,45 @@ FROM debian:${DEBIAN_VERSION} AS builder
 # duplicated in an ARG in the second stage of the build.
 #
 ARG PREFIX_DIR=/usr/local/guacamole
+ARG BUILD_DIR=/tmp/guacd-docker-BUILD
+ARG FREERDP_BUILD=/tmp/freerdp-BUILD
+
+# Add configuration scripts
+COPY src/guacd-docker/bin "${PREFIX_DIR}/bin/"
+
+# Copy source to container for sake of build
+COPY . "$BUILD_DIR"
+
+# FreeRDP build information
+ARG FREERDP_DEPS="              \
+        build-essential         \
+        cmake                   \
+        curl                    \
+        git                     \
+        libcups2-dev            \
+        libgsm1-dev             \
+        libjpeg-dev             \
+        libpulse-dev            \
+        libssl1.0-dev           \
+        libxml2-dev             \
+        libz-dev                \
+        make                    \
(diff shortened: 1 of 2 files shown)


avwx:
You should also apply all patches that Debian applies against 1.1.0. This includes critical security patches as well as some features like TLS 1.2. You can download them here http://deb.debian.org/debian/pool/main/f/freerdp/freerdp_1.1.0~git20140921.1.440916e+dfsg1-13+deb9u3.debian.tar.xz. FreeRDP really is a mess...

necouchman:
@avwx: I'll be happy to give them a shot - are we sure they apply to the 1.2.x code?

avwx:
I doubt they will apply cleanly. Your time might be better invested in supporting FreeRDP 2.0 😃

necouchman:
@avwx: Mike is actively working that.

mike-jumper:
@necouchman, what are your thoughts on these changes vs. instead relying on the FreeRDP 2.0 support?

necouchman:
@mike-jumper: Yeah, this is a sticky issue, for a variety of reasons.  I think FreeRDP 2.0 support would be ideal, if we can get that; however, I'm not sure that FreeRDP 2.0 is going to be available in the `debian:stable` Docker image that we base guacd on?

FreeRDP version (1.1 vs. 2.0) aside, I'm also not really crazy about the idea of compiling from source some other project's code in order to support guacd.  On the other hand, have this regression in our project really sucks.

So, as you can tell, really on the fence about this.  I'm not tied to this particular route at all, and open to other suggestions on which direction would be the best to go for Guacamole, mainly from a maintenance perspective.

lorengordon:
The approach in this pr looks pretty stable to me... It's checking out a specific tag, then applying a specific patch. Absent the source url just not being available, seems like that ought to result in a stable, reproducible build each time.

How close is guacd to supporting FreeRDP 2.0?

Or, alternatively, is there any particular reason to stick with the Debian container? The Ubuntu LTS repos have a working FreeRDP 1.2 pkg...

necouchman:
Closing, as per @mike-jumper this issue will be resolved with the changes coming in GUACAMOLE-249.

PR #217, opened 2019-03-10
GUACAMOLE-414: Implement TLS Write Locking Callbacks

necouchman:
This pull request implements the `LockWriteToTLS()` and `UnlockWriteToTLS()` callbacks used by LibVNCClient to avoid race conditions when TLS is in use that cause the VNC connection to fail/crash.

Note that support for these callbacks was released in LibVNCServer/Client 0.9.11, so it will only work against 0.9.11 or later- the bug documented in GUACAMOLE-414 will still be present when using libvncserver/client prior to 0.9.11.

### DIFF
--- a/src/protocols/vnc/vnc.c
+++ b/src/protocols/vnc/vnc.c
@@ -77,7 +77,8 @@ static rfbBool guac_vnc_lock_write_to_tls(rfbClient* rfb_client) {
     /* Lock write access */
     int retval = pthread_mutex_lock(&(vnc_client->tls_lock));
     if (retval) {
-        guac_client_log(gc, GUAC_LOG_ERROR, "Error locking TLS write mutex: %d", retval);
+        guac_client_log(gc, GUAC_LOG_ERROR, "Error locking TLS write mutex: %s",
+                strerror(retval));
         return FALSE;
     }
     return TRUE;
@@ -106,7 +107,8 @@ static rfbBool guac_vnc_unlock_write_to_tls(rfbClient* rfb_client) {
     /* Unlock write access */
     int retval = pthread_mutex_unlock(&(vnc_client->tls_lock));
     if (retval) {
-        guac_client_log(gc, GUAC_LOG_ERROR, "Error unlocking TLS write mutex: %d", retval);
+        guac_client_log(gc, GUAC_LOG_ERROR, "Error unlocking TLS write mutex: %d",
+                strerror(retval));
         return FALSE;
     }
     return TRUE;
--- a/src/protocols/vnc/vnc.c
+++ b/src/protocols/vnc/vnc.c
@@ -47,6 +47,7 @@
 #include <guacamole/socket.h>
 #include <guacamole/timestamp.h>
 #include <rfb/rfbclient.h>
+#include <rfb/rfbconfig.h>
 #include <rfb/rfbproto.h>
 
 #include <stdlib.h>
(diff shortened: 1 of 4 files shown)


necouchman:
Okay, updated to check for rfbClient data structure members within configure.ac.

src/protocols/vnc/vnc.c
```diff
@@ -182,6 +247,11 @@ void* guac_vnc_client_thread(void* data) {
     rfbClientLog = guac_vnc_client_log_info;
     rfbClientErr = guac_vnc_client_log_error;
 
+#ifdef ENABLE_VNC_TLS_LOCKING
+    /* Initialize the write lock */
+    pthread_mutex_init(&(vnc_client->tls_lock), NULL);
```
mike-jumper: There should be a corresponding `pthread_mutex_destroy()` when the mutex is no longer needed. This may be easier to manage if the mutex is initialized when the `guac_vnc_client` structure is allocated (to ensure `pthread_mutex_destroy()` will never be called on `vnc_client->tls_lock` before it's been initialized):

https://github.com/apache/guacamole-server/blob/5e2ddb890a90e840a419f4878381b0c511cc6f25/src/protocols/vnc/client.c#L47-L52

src/protocols/vnc/vnc.h
```diff
@@ -32,6 +32,7 @@
 #include <guacamole/client.h>
 #include <guacamole/layer.h>
 #include <rfb/rfbclient.h>
+#include <rfb/rfbconfig.h>
```
mike-jumper: With the TLS locking callback check now happening within `configure.ac`, is this header necessary?

src/protocols/vnc/vnc.c
```diff
@@ -55,6 +56,64 @@
...
+ * @param rfb_client
+ *     The rfbClient for which to lock the TLS mutex.
+ *
+ * @returns
+ *     rfbBool FALSE if an error occurs locking the mutex, otherwise
+ *     TRUE.
+ */
+static rfbBool guac_vnc_lock_write_to_tls(rfbClient* rfb_client) {
+
+    // Retrieve the Guacamole data structures
```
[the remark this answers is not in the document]
necouchman: Fixed.

src/protocols/vnc/vnc.c
necouchman: Moved this to client.c and added `pthread_mutex_destroy()` call.

src/protocols/vnc/vnc.h
necouchman: Nope, removed.

src/protocols/vnc/vnc.c
```diff
@@ -55,6 +56,64 @@
...
+    guac_client* gc = rfbClientGetClientData(rfb_client, GUAC_VNC_CLIENT_KEY);
+    guac_vnc_client* vnc_client = (guac_vnc_client*) gc->data;
+
+    // Lock write access
+    int retval = pthread_mutex_lock(&(vnc_client->tls_lock));
+    if (retval) {
+        guac_client_log(gc, GUAC_LOG_ERROR, "Error locking TLS write mutex: %d", retval);
```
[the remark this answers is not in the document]
necouchman: Converted to use strings output from `strerror()` upon failure.

necouchman:
@mike-jumper: Fixed up those items - let me know if there is anything else that needs tweaking.

src/protocols/vnc/vnc.h
mike-jumper: GitHub shows the `#include <rfb/rfbconfig.h>` still present here. Is there a commit missing?

necouchman: Okay, really should be removed this time...

¯\_(ツ)_/¯

necouchman:
@mike-jumper: Ping - anything else, here?

PR #218, opened 2019-03-10
GUACAMOLE-296: Add checks for Stream functions in winpr libraries.

necouchman:
This PR is, yet another, attempt to fix the transient linking issue that happens across various versions of FreeRDP related to the `Stream_New()` and `Stream_Free()` functions.  I think I've correctly written the checks in the `configure.ac` script such that this will work across versions of FreeRDP that implement winpr, but won't cause problems for 1.0.0.


configure.ac
```diff
@@ -719,6 +720,31 @@ then
...
+  ------------------------------------------
+  Unable to locate stream functions in winpr
+  libraries.  RDP will be disabled.
+  ------------------------------------------])
+                                    have_freerdp=no])])
+fi
+
+if test "x${have_freerdp}" = "xyes"
+then
+    AC_CHECK_LIB([freerdp-client], [freerdp_channels_new],
```
mike-jumper: I think this might be copypasta of one of the existing checks:

https://github.com/apache/guacamole-server/blob/7024b290abc2744168bd93de9ceb781f62959e60/configure.ac#L561-L570

necouchman: Haha - oops.  I think I may have copied it down for referenced.  Removed.

necouchman:
@mike-jumper: Anything else, here?

PR #219, opened 2019-03-25
GUACAMOLE-422: Add timezone to client handshake

necouchman:
This adds remaining pieces of the server code needed to make the timezone part of the client handshake.

### DIFF
--- a/src/libguac/guacamole/protocol.h
+++ b/src/libguac/guacamole/protocol.h
@@ -38,7 +38,18 @@
 #include <cairo/cairo.h>
 #include <stdarg.h>
 
-/* The protocol version */
+/**
+ * This defines the overall protocol version that this build of libguac
+ * supports.  The protocol version is used to provide compatibility between
+ * potentially different versions of Guacamole server and clients.  The
+ * version number is a MAJOR_MINOR_PATCH version that matches the versioning
+ * used throughout the components of the Guacamole project.  This version
+ * will not necessarily increment with the other components, unless additional
+ * functionality is introduced that affects compatibility.
+ * 
+ * This version is passed by the __guac_protocol_send_args() function from the
+ * server to the client during the client/server handshake.
+ */
 #define GUACAMOLE_PROTOCOL_VERSION "VERSION_1_1_0"
 
 /* CONTROL INSTRUCTIONS */
--- a/src/libguac/guacamole/protocol.h
+++ b/src/libguac/guacamole/protocol.h
@@ -38,6 +38,9 @@
 #include <cairo/cairo.h>
 #include <stdarg.h>
 
+/* The protocol version */
+#define GUACAMOLE_PROTOCOL_VERSION "VERSION_1_1_0"
+
 /* CONTROL INSTRUCTIONS */
 
 /**
--- a/src/libguac/guacamole/user.h
+++ b/src/libguac/guacamole/user.h
@@ -88,6 +88,13 @@ struct guac_user_info {
      * stated resolution of the display size request is recommended.
      */
     int optimal_resolution;
+    
+    /**
+     * The timezone of the remote system.  If the client does not provide
+     * a specific timezone then this will be NULL.  The format of the timezone
+     * is the standard tzdata naming convention.
+     */
+    const char** timezone;
 
 };
 
--- a/src/libguac/guacamole/user.h
+++ b/src/libguac/guacamole/user.h
@@ -506,79 +506,6 @@ struct guac_user {
 };
 
 /**
- * Handler for Guacamole protocol opcode specific to the handshake that
- * happens between client and server at the beginning of the connection.  The
- * handler will be invoked when the matching opcode is received during the
- * handshake process.
- * 
- * @param user
- *     The user that initiated the handshake.
- * 
- * @param parser
- *     The parser allocated for parsing the data provided by the client.
- * 
- * @param timeout
- *     The timeout, in microseconds, for parsing the value.
- * 
- * @return
- *     Zero if the handshake instruction is successfully parsed; otherwise
- *     false.
- */
-typedef int __guac_handshake_handler(guac_user* user, int argc, char** argv);
-
-/**
- * Structure that maps opcodes received during the handshake phase of the
- * connection to callback functions used when those opcodes are received.
- */
-typedef struct __guac_handshake_mapping {
-    
-    /**
-     * The instruction opcode which maps to the handler.
-     */
-    char* opcode;
-    
-    /**
-     * The handler function used when specified opcode is received.
-     */
-    __guac_handshake_handler* handler;
-    
-} __guac_handshake_mapping;
-
-/**
- * Internal handler function that is called when the size instruction is
- * received during the handshake process.
- */
-__guac_handshake_handler __guac_handshake_size_handler;
-
-/**
- * Internal handler function that is called when the audio instruction is
- * received during the handshake process, specifying the audio mimetypes
- * available to the client.
- */
-__guac_handshake_handler __guac_handshake_audio_handler;
-
-/**
- * Internal handler function that is called when the video instruction is
- * received during the handshake process, specifying the video mimetypes
- * available to the client.
- */
-__guac_handshake_handler __guac_handshake_video_handler;
-
-/**
- * Internal handler function that is called when the image instruction is
- * received during the handshake process, specifying the image mimetypes
- * available to the client.
- */
-__guac_handshake_handler __guac_handshake_image_handler;
-
-/**
- * Internal handler function that is called when the timezone instruction is
- * received during the handshake process, specifying the timezone of the
- * client.
- */
-__guac_handshake_handler __guac_handshake_timezone_handler;
-
-/**
  * Allocates a new, blank user, not associated with any specific client or
  * socket.
  *
@@ -616,32 +543,6 @@ void guac_user_free(guac_user* user);
  *     connection from being handled properly.
  */
 int guac_user_handle_connection(guac_user* user, int usec_timeout);
-
-/**
- * Call the appropriate handler defined by the given user for the given
- * instruction. A comparison is made between the instruction opcode and the
- * initial handler lookup table defined in user-handlers.c. The intial handlers
- * will in turn call the user's handler (if defined).
- *
- * @param user
- *     The user whose handlers should be called.
- *
- * @param opcode
- *     The opcode of the instruction to pass to the user via the appropriate
- *     handler.
- *
- * @param argc
- *     The number of arguments which are part of the instruction.
- *
- * @param argv
- *     An array of all arguments which are part of the instruction.
- *
- * @return
- *     Non-negative if the instruction was handled successfully, or negative
- *     if an error occurred.
- */
-int guac_user_handle_instruction(guac_user* user, const char* opcode,
-        int argc, char** argv);
 
 /**
  * Allocates a new stream. An arbitrary index is automatically assigned
--- a/src/libguac/guacamole/user.h
+++ b/src/libguac/guacamole/user.h
@@ -506,6 +506,79 @@ struct guac_user {
 };
 
 /**
+ * Handler for Guacamole protocol opcode specific to the handshake that
+ * happens between client and server at the beginning of the connection.  The
+ * handler will be invoked when the matching opcode is received during the
+ * handshake process.
+ * 
+ * @param user
+ *     The user that initiated the handshake.
+ * 
+ * @param parser
+ *     The parser allocated for parsing the data provided by the client.
+ * 
+ * @param timeout
+ *     The timeout, in microseconds, for parsing the value.
+ * 
+ * @return
+ *     Zero if the handshake instruction is successfully parsed; otherwise
+ *     false.
+ */
+typedef int __guac_handshake_handler(guac_user* user, int argc, char** argv);
+
+/**
+ * Structure that maps opcodes received during the handshake phase of the
+ * connection to callback functions used when those opcodes are received.
+ */
+typedef struct __guac_handshake_mapping {
+    
+    /**
+     * The instruction opcode which maps to the handler.
+     */
+    char* opcode;
+    
+    /**
+     * The handler function used when specified opcode is received.
+     */
+    __guac_handshake_handler* handler;
+    
+} __guac_handshake_mapping;
+
+/**
+ * Internal handler function that is called when the size instruction is
+ * received during the handshake process.
+ */
+__guac_handshake_handler __guac_handshake_size_handler;
+
+/**
+ * Internal handler function that is called when the audio instruction is
+ * received during the handshake process, specifying the audio mimetypes
+ * available to the client.
+ */
+__guac_handshake_handler __guac_handshake_audio_handler;
+
+/**
+ * Internal handler function that is called when the video instruction is
+ * received during the handshake process, specifying the video mimetypes
+ * available to the client.
+ */
+__guac_handshake_handler __guac_handshake_video_handler;
+
+/**
+ * Internal handler function that is called when the image instruction is
+ * received during the handshake process, specifying the image mimetypes
+ * available to the client.
+ */
+__guac_handshake_handler __guac_handshake_image_handler;
+
+/**
+ * Internal handler function that is called when the timezone instruction is
+ * received during the handshake process, specifying the timezone of the
+ * client.
+ */
+__guac_handshake_handler __guac_handshake_timezone_handler;
+
+/**
  * Allocates a new, blank user, not associated with any specific client or
  * socket.
  *
--- a/src/libguac/guacamole/user.h
+++ b/src/libguac/guacamole/user.h
@@ -545,6 +545,32 @@ void guac_user_free(guac_user* user);
 int guac_user_handle_connection(guac_user* user, int usec_timeout);
 
 /**
+ * Call the appropriate handler defined by the given user for the given
+ * instruction. A comparison is made between the instruction opcode and the
+ * initial handler lookup table defined in user-handlers.c. The initial handlers
+ * will in turn call the user's handler (if defined).
+ *
+ * @param user
+ *     The user whose handlers should be called.
+ *
+ * @param opcode
+ *     The opcode of the instruction to pass to the user via the appropriate
+ *     handler.
+ *
+ * @param argc
+ *     The number of arguments which are part of the instruction.
+ *
+ * @param argv
+ *     An array of all arguments which are part of the instruction.
+ *
+ * @return
+ *     Non-negative if the instruction was handled successfully, or negative
+ *     if an error occurred.
+ */
+int guac_user_handle_instruction(guac_user* user, const char* opcode,
+        int argc, char** argv);
+
+/**
  * Allocates a new stream. An arbitrary index is automatically assigned
  * if no previously-allocated stream is available for use.
  *
--- a/src/libguac/guacamole/user.h
+++ b/src/libguac/guacamole/user.h
@@ -94,7 +94,7 @@ struct guac_user_info {
      * a specific timezone then this will be NULL.  The format of the timezone
      * is the standard tzdata naming convention.
      */
-    const char** timezone;
+    const char* timezone;
 
 };
 
--- a/src/libguac/user-handlers.c
+++ b/src/libguac/user-handlers.c
@@ -622,6 +622,8 @@ int __guac_handshake_size_handler(guac_user* user, int argc, char** argv) {
 
 int __guac_handshake_audio_handler(guac_user* user, int argc, char** argv) {
 
+    guac_free_mimetypes((char **) user->info.audio_mimetypes);
+    
     /* Store audio mimetypes */
     user->info.audio_mimetypes = (const char**) guac_copy_mimetypes(argv, argc);
     
@@ -631,6 +633,8 @@ int __guac_handshake_audio_handler(guac_user* user, int argc, char** argv) {
 
 int __guac_handshake_video_handler(guac_user* user, int argc, char** argv) {
 
+    guac_free_mimetypes((char **) user->info.video_mimetypes);
+    
     /* Store video mimetypes */
     user->info.video_mimetypes = (const char**) guac_copy_mimetypes(argv, argc);
     
@@ -639,6 +643,8 @@ int __guac_handshake_video_handler(guac_user* user, int argc, char** argv) {
 }
 
 int __guac_handshake_image_handler(guac_user* user, int argc, char** argv) {
+    
+    guac_free_mimetypes((char **) user->info.image_mimetypes);
     
     /* Store image mimetypes */
     user->info.image_mimetypes = (const char**) guac_copy_mimetypes(argv, argc);
@@ -680,6 +686,9 @@ char** guac_copy_mimetypes(char** mimetypes, int count) {
 
 void guac_free_mimetypes(char** mimetypes) {
 
+    if (mimetypes == NULL)
+        return;
+    
     char** current_mimetype = mimetypes;
 
     /* Free all strings within NULL-terminated mimetype array */
--- a/src/libguac/user-handlers.c
+++ b/src/libguac/user-handlers.c
@@ -662,6 +662,9 @@ int __guac_handshake_timezone_handler(guac_user* user, int argc, char** argv) {
     if (argc > 0 && strcmp(argv[0], ""))
         user->info.timezone = (const char*) strdup(argv[0]);
     
+    else
+        user->info.timezone = NULL;
+    
     return 0;
     
 }
--- a/src/libguac/user-handlers.c
+++ b/src/libguac/user-handlers.c
@@ -693,7 +693,7 @@ void guac_free_mimetypes(char** mimetypes) {
 
 }
 
-int guac_user_handle_instruction(__guac_instruction_handler_mapping* map,
+int __guac_user_call_opcode_handler(__guac_instruction_handler_mapping* map,
         guac_user* user, const char* opcode, int argc, char** argv) {
 
     /* For each defined instruction */
--- a/src/libguac/user-handlers.c
+++ b/src/libguac/user-handlers.c
@@ -720,7 +720,7 @@ int __guac_user_call_opcode_handler(__guac_instruction_handler_mapping* map,
     }
 
     /* If unrecognized, log and ignore */
-    guac_user_log(user, GUAC_LOG_WARNING, "Handler not found for \"%s\"",
+    guac_user_log(user, GUAC_LOG_DEBUG, "Handler not found for \"%s\"",
             opcode);
     return 0;
 
--- a/src/libguac/user-handlers.c
+++ b/src/libguac/user-handlers.c
@@ -31,6 +31,7 @@
 #include <stdio.h>
 #include <stdint.h>
 #include <stdlib.h>
+#include <string.h>
 
 /* Guacamole instruction handler map */
 
@@ -51,6 +52,17 @@ __guac_instruction_handler_mapping __guac_instruction_handler_map[] = {
    {"audio",      __guac_handle_audio},
    {"argv",       __guac_handle_argv},
    {NULL,         NULL}
+};
+
+/* Guacamole handshake handler map */
+
+__guac_instruction_handler_mapping __guac_handshake_handler_map[] = {
+    {"size",     __guac_handshake_size_handler},
+    {"audio",    __guac_handshake_audio_handler},
+    {"video",    __guac_handshake_video_handler},
+    {"image",    __guac_handshake_image_handler},
+    {"timezone", __guac_handshake_timezone_handler},
+    {NULL,       NULL}
 };
 
 /**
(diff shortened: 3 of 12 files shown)


src/libguac/user-handshake.c
```diff
@@ -384,6 +384,31 @@ int guac_user_handle_connection(guac_user* user, int usec_timeout) {
...
+    if (parser->argc < 1) {
+        guac_user_log(user, GUAC_LOG_ERROR, "Received \"timezone\" instruction "
+                "lacked required arguments.");
+        guac_parser_free(parser);
+        return 1;
+    }
+    
+    /* Store timezone */
+    char* timezone = parser->argv[0];
+    user->info.timezone = (const char*) timezone;
```
mike-jumper: So, the documentation for `user->info.timezone` currently states:

> ... If the client does not provide a specific timezone then this will be NULL. ...

As written, this will not be the case. Assuming not providing a timezone means providing the empty string (`8.timezone,0.;`), then this will actually contain the empty string.

src/protocols/rdp/rdp_settings.c
```diff
@@ -854,6 +854,10 @@ guac_rdp_settings* guac_rdp_parse_args(guac_user* user,
     settings->timezone =
         guac_user_parse_args_string(user, GUAC_RDP_CLIENT_ARGS, argv,
                 IDX_TIMEZONE, NULL);
+    
+    /* If no timezone explicitly set, try to pull from tunnel */
```
mike-jumper: guacd doesn't know about the tunnel. ;)

This is good to note, but will need to be noted differently to be correct from the perspective of this part of the stack.

src/protocols/rdp/rdp_settings.c
```diff
@@ -854,6 +854,10 @@ guac_rdp_settings* guac_rdp_parse_args(guac_user* user,
...
+    if (settings->timezone == NULL)
+        settings->timezone = user->info.timezone;
```
mike-jumper: Later code expects that `settings->timezone` is dynamically allocated. If assigned directly from `user->info.timezone`, this will not be the case and you'll end up with the same memory being freed twice.

Rather than implementing this logic manually, you can instead rely on `guac_user_parse_args_string()`, which allows for default values and will always dynamically allocate its return value when not `NULL`:

https://guacamole.apache.org/doc/libguac/user_8h.html#a7530b31b5a621a4f68fdce4bee2bdc5f

src/libguac/user-handshake.c
necouchman: Should be fixed...hopefully this was the correct way...

src/protocols/rdp/rdp_settings.c
necouchman: Removed comment, as the section got removed and used default value.

src/protocols/rdp/rdp_settings.c
necouchman: Updated to put the `user->info.timezone` member into the default value, which should pull either the value or null if it hasn't been set.

src/libguac/user-handshake.c
```diff
@@ -384,6 +384,32 @@ int guac_user_handle_connection(guac_user* user, int usec_timeout) {
...
+    if (parser->argc < 1) {
+        guac_user_log(user, GUAC_LOG_ERROR, "Received \"timezone\" instruction "
+                "lacked required arguments.");
+        guac_parser_free(parser);
+        return 1;
+    }
+    
+    /* Store timezone, if present */
+    char* timezone = parser->argv[0];
+    if (timezone != NULL && !strcmp(timezone, ""))
```
mike-jumper: The values within `argv` can't be `NULL`.

src/protocols/ssh/settings.c
```diff
@@ -425,6 +425,10 @@ guac_ssh_settings* guac_ssh_parse_args(guac_user* user,
     settings->timezone =
         guac_user_parse_args_string(user, GUAC_SSH_CLIENT_ARGS, argv,
                 IDX_TIMEZONE, NULL);
+    
+    /* If timezone not explicitly set, try to pull from tunnel */
```
mike-jumper: Same here as with RDP - guacd doesn't know about the tunnel.

src/protocols/ssh/settings.c
```diff
@@ -425,6 +425,10 @@ guac_ssh_settings* guac_ssh_parse_args(guac_user* user,
...
+    if (settings->timezone == NULL)
+        settings->timezone = user->info.timezone;
```
mike-jumper: Same here as with RDP - later code relies on `settings->timezone` being dynamically allocated. It would be better to leverage the default value logic already provided by `guac_user_parse_args_string()`.

src/protocols/ssh/settings.c
necouchman: Fixed.

src/protocols/ssh/settings.c
necouchman: Fixed.

src/libguac/user-handshake.c
necouchman: Removed null check.

necouchman:
@mike-jumper: This has been reworked in line with your comments in apache/guacamole-client#348 for allowing arbitrary order of incoming opcodes for the handshake.  I'm sure it needs some work, but it does working, including with an older version of the client that does not pass in the `timezone` instruction.

src/libguac/user-handshake.c
```diff
@@ -171,6 +171,80 @@ static void guac_free_mimetypes(char** mimetypes) {
...
+        user->info.optimal_resolution = 96;
+    
+    return 0;
+    
+}
+
+int __guac_handshake_audio_handler(guac_user* user, int argc, char** argv) {
+    
+    /* Store audio mimetypes */
+    user->info.audio_mimetypes = (const char**) guac_copy_mimetypes(argv, argc);
```
mike-jumper: If `audio` is received twice, this will leak memory. The previous value, if any, will need to be freed before assigning a new value.

src/libguac/user-handshake.c
```diff
@@ -171,6 +171,80 @@ static void guac_free_mimetypes(char** mimetypes) {
...
+    
+    return 0;
+    
+}
+
+int __guac_handshake_video_handler(guac_user* user, int argc, char** argv) {
+    
+    /* Store video mimetypes */
+    user->info.video_mimetypes = (const char**) guac_copy_mimetypes(argv, argc);
```
mike-jumper: If `video` is received twice, this will leak memory. The previous value, if any, will need to be freed before assigning a new value.

src/libguac/user-handshake.c
```diff
@@ -171,6 +171,80 @@ static void guac_free_mimetypes(char** mimetypes) {
...
+    
+    return 0;
+    
+}
+
+int __guac_handshake_image_handler(guac_user* user, int argc, char** argv) {
+    
+    /* Store image mimetypes */
+    user->info.image_mimetypes = (const char**) guac_copy_mimetypes(argv, argc);
```
mike-jumper: If `image` is received twice, this will leak memory. The previous value, if any, will need to be freed before assigning a new value.

src/libguac/guacamole/user.h
```diff
@@ -498,6 +505,79 @@ struct guac_user {
...
+ *     The parser allocated for parsing the data provided by the client.
+ * 
+ * @param timeout
+ *     The timeout, in microseconds, for parsing the value.
+ * 
+ * @return
+ *     Zero if the handshake instruction is successfully parsed; otherwise
+ *     false.
+ */
+typedef int __guac_handshake_handler(guac_user* user, int argc, char** argv);
```
mike-jumper: I recommend moving these internal functions and structures to a non-public header, perhaps [`user-handlers.h`](https://github.com/apache/guacamole-server/blob/ffe0b57faa97272ee1287bcea74c5d787e221953/src/libguac/user-handlers.h).

In fact, if that's done, you might be able to leverage the existing `__guac_instruction_handler_mapping`, `__guac_instruction_handler`, etc. functions. Their counterparts here seem to essentially be duplicates.

Assuming that's the path forward, perhaps there is a way to separate out the core of `guac_user_handle_instruction()` such that it's possible to pass in arbitrary mappings (even said core is kept internal to `user-handlers.h` / `user-handlers.c`)?

src/libguac/user-handshake.c
```diff
@@ -305,96 +379,53 @@ int guac_user_handle_connection(guac_user* user, int usec_timeout) {
...
-
-    /* Get args from connect instruction */
-    if (guac_parser_expect(parser, socket, usec_timeout, "connect")) {
-
-        /* Log error */
-        guac_user_log_handshake_failure(user);
-        guac_user_log_guac_error(user, GUAC_LOG_DEBUG,
-                "Error reading \"connect\"");
+    /* Handle each of the opcodes. */
+    while (1) {
```
mike-jumper: Conditionless `while` loops make me itchy ... but I imagine this may vanish regardless if the duplicate code between the main user instruction handlers and the handshake user instruction handlers is separated out. If unavoidable, then unavoidable, but it would be better to have an explicit bound on the loop.

src/libguac/user-handshake.c
necouchman: Freed.

src/libguac/user-handshake.c
necouchman: Freed.

src/libguac/user-handshake.c
necouchman: Freed.

src/libguac/user-handshake.c
necouchman: Well, I tried calling `guac_free_mimetypes()` here at the start of this function, but this is resulting in some...odd...behavior if `user->info.audio_mimetypes` (and image/video) haven't been initialized at all.

necouchman: Yeah, tried a couple of other things, here, and every time I try to run `guac_free_mimetypes()` within the handler function it just hangs.  Looks like probably an endless loop, but it's hard to say - the `size` instruction gets processed as it should, but as soon as it hits the `audio` instruction it hangs and doesn't go any further.  If I take out the `guac_free_mimetypes()` call it behaves fine, but that doesn't resolve memory leak concerns.  I also tried settings those mimetypes arrays to NULL at the beginning of the `guac_handle_connection()` function, but that didn't seem to help, either.

I'm sure I'm missing something simple...

src/libguac/user-handshake.c
necouchman: Should be able to consolidate stuff.  I actually had it done, but was running into issues and realized I needed to change fewer things concurrently.  So, now I'll go back and do this.  I was having some issues with includes, so I might end up needing some help - it's a little challenging with the maps to get everything in the correct header files to be seen where it needs to be.

necouchman: Well, didn't manage to make the conditionless while loop disappear, yet, so I'll be happy to take a suggestion or two on the best way to do this.

src/libguac/guacamole/user.h
necouchman: Okay, I think I've sufficiently accomplished this.

necouchman:
@mike-jumper: Okay, took care of some of it, but will need some guidance on the remaining items....

src/libguac/guacamole/user.h
```diff
@@ -537,32 +544,6 @@ void guac_user_free(guac_user* user);
...
- *     The number of arguments which are part of the instruction.
- *
- * @param argv
- *     An array of all arguments which are part of the instruction.
- *
- * @return
- *     Non-negative if the instruction was handled successfully, or negative
- *     if an error occurred.
- */
-int guac_user_handle_instruction(guac_user* user, const char* opcode,
```
mike-jumper: This function is part of the public API of libguac. I doubt it's in wide use, but we shouldn't delete a public function without a deprecation period.

Why is this function being removed from `user.h`?

src/protocols/kubernetes/settings.c
```diff
@@ -25,6 +25,14 @@
 
 /* Client plugin arguments */
 const char* GUAC_KUBERNETES_CLIENT_ARGS[] = {
+    /**
+     * This first argument defines the protocol version in use so that the
+     * client knows how to handle talking to different versions of guacd.
+     * If this is omitted the client may choose not to enable certain
+     * features.
+     */
+    "VERSION_1_1_0",
```
mike-jumper: Thoughts on handling this automatically within `guac_protocol_send_args()`?

mike-jumper: Any particular reason behind the formatting of the version string?

src/libguac/guacamole/user.h
necouchman: If I don't move this, and leave it the way I have it implemented (passing a `__guac_instruction_handler_mapping` pointer in as the array for it to search through), then I have to include `user-handlers.h` in `user.h` in order for it to have that pointer type defined.  This seemed like something worth avoiding, both because `user-handlers.h` is outside the public API for libguac, and because `user-handlers.c` already includes `user.h`, making it almost a circular set of includes.

Should I just copy it instead of move it, and mark this for deprecation?  Or is there some other route I should go with where each of these items should be to avoid this conundrum?  Mainly I was trying to re-use the code in `guac_user_handle_instruction`, and it seems like the best way to do that is to allow the caller to specify which array of opcode/handler pairs should be referenced.

src/protocols/kubernetes/settings.c
necouchman: Fine with me, I'll take a look...

necouchman: It matches the `enum` on the Java side of things that I implemented, so it makes it easy to use `valueOf` in Java to pull the matching value.  I'm not particularly tied to that - originally I had something like `version:1.1.0` and could go back to that, or whatever else we decide we want to do for that argument.

mike-jumper: `VERSION_1_1_0` seems fine to me, and would indeed be convenient for parsing on the Java side.

This would remain `VERSION_1_1_0` even after guacamole-server is at 1.2.0, 1.3.999, etc., being a reference to the protocol variation introduced by the release having that same version number?

src/libguac/guacamole/user.h
mike-jumper: > Should I just copy it instead of move it ...

I think the best route would be:

* Keep the new version of `guac_user_handle_instruction()` but rename it such that it can exist as an internal function which doesn't collide with the existing `guac_user_handle_instruction()`.
* Invoke the new version from the old function such that `guac_user_handle_instruction()` continues to work, yet primarily exists in one place.

> ... and mark this for deprecation?

I'm not 100% on that. If we will be removing this function, it does need to be deprecated first, but should it be removed? It seems a useful counterpart to `guac_parser`.

src/protocols/kubernetes/settings.c
necouchman: > This would remain VERSION_1_1_0 even after guacamole-server is at 1.2.0, 1.3.999, etc., being a reference to the protocol variation introduced by the release having that same version number?

I don't have a strong preference either way, but that would be fine with me.  Unless you think it adds confusion to have the protocol version follow the product version?  I think it would be okay, as it would establish the minimum software version for each of the protocol features, which seems reasonably to me.

necouchman: Okay,  I think I've managed to accomplish this.  Only odd thing at this point is that, when both client and server are updated, I see the following message in the server log:

    guacd[35020]: DEBUG:	Parameter "(null)" omitted. Using default value of "America/New_York".

Is there something else I need to do on the receiving side to pull this off correctly so that it doesn't end up as a `null` parameter?  It seems to work okay - the timezone is getting set - but the message indicates something isn't quite right.

src/libguac/guacamole/user.h
necouchman: Okay, I've done this.  Hopefully naming of the new version of the function is acceptable.  I also left the documentation as-is - I'm guessing how it is handled internally within that function is not the concern of someone using the public API?

src/protocols/kubernetes/settings.c
mike-jumper: For a call like:

```c
guac_user_parse_args_string(user, GUAC_RDP_CLIENT_ARGS, argv,
         IDX_TIMEZONE, user->info.timezone);
```

With the new timezone parameter being at the end, the `IDX_TIMEZONE` is likely off by one, resulting in the `NULL` value at the end of `GUAC_*_CLIENT_ARGS` being pulled in for logging rather than the actual name of the parameter. This is probably due to the addition of `IDX_PROTOCOL_VERSION`.

Just as `guac_protocol_send_args()` has been modified to automatically include the protocol version, I suggest also automatically parsing out the protocol version within the code handling `connect`, storing the value somewhere within the `guac_user` just as we do the other parts of the handshake. This would avoid the need for protocol implementations to be aware of this detail.

Could be as simple as:

* Verifying receipt of the expected number of arguments
* Storing `argv[0]` somewhere sensible in `guac_user`
* Using `argc - 1` and `argv + 1` for the call to `guac_client_add_user()`

Existing protocol implementations would then need zero modification, except those that wish to check the protocol version.

src/libguac/user-handshake.c
mike-jumper: >  I also tried settings those mimetypes arrays to NULL at the beginning of the `guac_handle_connection()` function, but that didn't seem to help, either.

That should do the trick. What failed when you went that route?

src/protocols/kubernetes/settings.c
necouchman: > Could be as simple as:
> Verifying receipt of the expected number of arguments
> Storing argv[0] somewhere sensible in guac_user
> Using argc - 1 and argv + 1 for the call to guac_client_add_user()

Is there a way to do this outside of the protocol-specific parameter parser?  The expected number of arguments will vary protocol-to-protocol, so is there a way to know prior to the `guac_client_add_user()` call what the `parser->argc` value should be, and whether or not to subtract from `argc` and do `argv + 1`?

mike-jumper: > Is there a way to do this outside of the protocol-specific parameter parser?

Yes. There's no protocol specific parser - the Guacamole protocol requires that the `connect` instruction include a value for each argument declared in `args`, in the same order as specified in `args`.

Regardless of the protocol implementation, the Guacamole protocol would guarantee one of the following has to be true:

1. The client side is an up-to-date implementation and will populate that initial `VERSION_1_1_0` argument with its own protocol version.
2. The client side is older and will interpret `VERSION_1_1_0` as the literal name of an argument. As that argument presumably does not exist and does not have a value, it will be blank in the reponse.
3. The client does not correctly implement the Guacamole protocol and the argument count does not match (the server should close the connection at this point as all bets are off).

mike-jumper: Side note: even if it proves inconvenient to fully verify the argument count outside the protocol implementation (doing so would require iterating through the args until we find the `NULL`), we _do_ know that any conforming Guacamole implementation will provide at least one argument value because `guac_protocol_send_args()` will now always send at least one argument value (the version), so lacking at least one value within the received `args` would be a protocol violation.

We could, of course, choose to be tolerant of that, leave total rejection up to the protocol implementation, and handle the lack of even one argument value as equivalent to leaving that initial value blank.

necouchman: Okay, so I've played around with this a bit more and got it to the point where it works across various combinations of server and client versions in a consistent and expected way.  As it is right now, I ended up just removing the first argument on the guacd side.  The client implementation does not currently populate a value for the protocol version, it simply examines the protocol that the server sends and adjusts accordingly (see apache/guacamole-client#348).

The various scenarios are:
* Server 1.1.0, Client 1.1.0: Server sends protocol version and args to client, client examines and sets protocol version, adjusts handshake, and sends back an empty value for the protocol version argument.  The server then ignores this argument when joining the user to the connection by decrementing `argc` and passing through `argv + 1` to the join function.
* Server 1.1.0, Client 1.0.0 (or lower): Server sends protocol version and args to client, client has no knowledge of protocol version, so does not adjust handshake at all, and sends back an empty value for the protocol version argument.  The server ignores/strips the argument when joining using the same method as above.
* Server 1.0.0 (or lower), Client 1.1.0: Server sends arguments to client, without any protocol version.  The client checks for protocol version and does not see it, so it does not add any of the additional protocol-related functionality and simple passes through the supported handshake and arguments.  The server processes all of the arguments as usual.

So, my two questions would be:
* Is there some use-case I'm missing here where checking the number of provided/expected arguments in the handshake actually helps anything out?  As it is, it seems to work with various combinations of the server and client, so I'm not sure this is valuable?
* I currently have the client where it simply sends an empty argument back for the protocol version - is there any reason to have it populate something else in that argument, like the actual protocol version it's using?  It doesn't seem to need it right now in order to function properly, but maybe it's worth just implementing it this way, anyway, to future-proof it?  Or just leave it as-is?

src/libguac/user-handshake.c
necouchman: Okay, well, made the changes, again, plus one slight tweak to the `guac_free_mimetypes()` function, and I have this working.

src/protocols/kubernetes/settings.c
necouchman: > I currently have the client where it simply sends an empty argument back for the protocol version

I went back and reworked this on the client-side a little bit, and I'm now sending back the negotiated protocol version from client to server as an argument.  I'm still not doing anything with it on the server-side, but I'm no longer just passing an empty argument.

src/libguac/user-handlers.c
```diff
@@ -581,3 +593,131 @@ int __guac_handle_disconnect(guac_user* user, int argc, char** argv) {
...
+}
+
+int __guac_handshake_timezone_handler(guac_user* user, int argc, char** argv) {
+    
+    /* Free any past value */
+    free((char *) user->info.timezone);
+    
+    /* Store timezone, if present */
+    if (argc > 0 && strcmp(argv[0], ""))
+        user->info.timezone = (const char*) strdup(argv[0]);
```
mike-jumper: In the case of the following sequence:

1. `8.timezone,6.foobar;`
2. `8.timezone;`

`user->info.timezone` will contain a stale pointer to the old, freed `"foobar"` memory, rather than `NULL`. I think this should explicitly assign `NULL` in the blank/omitted case.

src/libguac/user-handlers.h
```diff
@@ -186,4 +220,73 @@ __guac_instruction_handler __guac_handle_disconnect;
...
+ *     handler.
+ *
+ * @param argc
+ *     The number of arguments which are part of the instruction.
+ *
+ * @param argv
+ *     An array of all arguments which are part of the instruction.
+ *
+ * @return
+ *     Non-negative if the instruction was handled successfully, or negative
```
mike-jumper: In at least one case, the return value of this function is presumed to be successful only if zero, with any non-zero value indicating an error:

https://github.com/apache/guacamole-server/blob/98cb7ccf6773f9e55645ccfe542de31ab2bc136c/src/libguac/user-handshake.c#L273-L275

The above usage matches with the underlying definition of `__guac_instruction_handler`:

https://github.com/apache/guacamole-server/blob/98cb7ccf6773f9e55645ccfe542de31ab2bc136c/src/libguac/user-handlers.h#L52-L55

As `__guac_user_call_opcode_handler()` passes through the return value of the matching `__guac_instruction_handler`, this should also be zero if successful, non-zero otherwise.

src/libguac/user-handshake.c
```diff
@@ -290,7 +234,11 @@ int guac_user_handle_connection(guac_user* user, int usec_timeout) {
 
     guac_socket* socket = user->socket;
     guac_client* client = user->client;
-
+    
+    user->info.audio_mimetypes = NULL;
+    user->info.image_mimetypes = NULL;
+    user->info.video_mimetypes = NULL;
```
mike-jumper: What about `user->info.timezone`?

src/libguac/user-handshake.c
mike-jumper: I think the usual structure would be (pseudocode):

    while (instruction successfully read) {
        handle_success;
    }

    handle_failure;

In our case, though, we have more like:

    while (handshake_in_progress) {
        handle_handshake_instruction;
    }

    this_might_be_failure_or_success;

I'm not sure myself how this could be arranged as-is to remove the unbounded loop, given that the exit condition of the loop would be ambiguously either success or failure. Things might be easier/cleaner if the actual internal handshake handling were its own function. Then you'd end up with something like:

    if (init_user_with_handshake(stuff))
        handle_failure;

    add_user_to_connection

with the actual function being the expected:

    while (instruction successfully read) {

        if (instruction_is_connect)
            return 0;

        handle_instruction;
    }

    handle_failure;
    return 1;

I'm sure there's a way to rearrange the logic such that we have a net gain in readability, part of that readability being an explicit loop bound.

src/libguac/guacamole/protocol.h
```diff
@@ -38,6 +38,9 @@
 #include <cairo/cairo.h>
 #include <stdarg.h>
 
+/* The protocol version */
```
mike-jumper: This public API documentation for a macro like this needs to be more specific.

src/libguac/user-handshake.c
necouchman: Okay, I'll go another round with that one and see what I can do.

src/libguac/user-handshake.c
necouchman: We can definitely do that, here.  It doesn't break things to not have it, here, because the `free()` call is just a standard call rather than a loop that goes through a pointer array, as in the case of the audio, image, and video mimetypes.  But it's trivial to initialize it, so I'll add that.

src/libguac/guacamole/protocol.h
necouchman: Will beef it up a little.

src/libguac/user-handlers.c
necouchman: Would it be sufficient to set this to `NULL` in the beginning of the handshake function (as noted below), or do I need an `else` clause here that sets it to `NULL` if nothing is found?

src/libguac/user-handlers.h
necouchman: Adjusted comment/docs to match `__guac_instruction_handler()`

src/libguac/user-handshake.c
necouchman: Added a `NULL` assignment for `user->info.timezone`, here, to match the others.

src/libguac/guacamole/protocol.h
necouchman: Hopefully the comment is adequate and accurate, now.

src/libguac/user-handshake.c
necouchman: Okay, this should be implemented, now, in a way that gets rid of the infinite loop, breaking the handshake out into a separate function.

src/libguac/user-handlers.c
necouchman: Went ahead and added an `else` statement that nulls it out if nothing is received.

necouchman:
@mike-jumper: Okay, latest round of issues addressed.  Next?

src/libguac/user-handshake.c
```diff
@@ -305,104 +310,20 @@ int guac_user_handle_connection(guac_user* user, int usec_timeout) {
...
 
     /* Acknowledge connection availability */
     guac_protocol_send_ready(socket, client->connection_id);
     guac_socket_flush(socket);
-
-    /* Attempt join */
-    if (guac_client_add_user(client, user, parser->argc, parser->argv))
+    
+    /* Attempt to join user to connection. */
+    if (guac_client_add_user(client, user, (parser->argc - 1), parser->argv + 1))
```
mike-jumper: Is `argc` verified anywhere prior to advancing forward by one? I looked around and didn't see any check prior to this call, but I might be missing something.

While it is true that the Guacamole protocol requires the client to send sufficient argument values to match each argument given with `args`, a broken client may well not.

src/libguac/user-handshake.c
```diff
@@ -219,7 +162,8 @@ static void* guac_user_input_thread(void* data) {
         guac_error_message = NULL;
 
         /* Call handler, stop on error */
-        if (guac_user_handle_instruction(user, parser->opcode, parser->argc, parser->argv) < 0) {
+        if (__guac_user_call_opcode_handler(__guac_instruction_handler_map, 
+                user, parser->opcode, parser->argc, parser->argv) < 0) {
```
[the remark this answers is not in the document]
necouchman: Yeah, probably so...

src/libguac/user-handshake.c
```diff
@@ -286,11 +230,72 @@ static int guac_user_start(guac_parser* parser, guac_user* user,
...
+        guac_user_log(user, GUAC_LOG_DEBUG, "Processing instruction: %s",
+                parser->opcode);
+        
+        /* Run instruction handler for opcode with arguments. */
+        if (__guac_user_call_opcode_handler(__guac_handshake_handler_map, user,
+                parser->opcode, parser->argc, parser->argv)) {
+            
+            guac_user_log_handshake_failure(user);
+            guac_user_log_guac_error(user, GUAC_LOG_DEBUG,
+                    "Error handling handling opcode during handshake.");
```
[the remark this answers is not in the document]
necouchman: Yep yep

src/libguac/user-handshake.c
```diff
@@ -305,104 +310,20 @@ int guac_user_handle_connection(guac_user* user, int usec_timeout) {
...
-    if (guac_parser_expect(parser, socket, usec_timeout, "size")) {
-
-        /* Log error */
+    /* Perform the handshake with the client. */
+    if (__guac_user_handshake(user, parser, usec_timeout)) {
         guac_user_log_handshake_failure(user);
         guac_user_log_guac_error(user, GUAC_LOG_DEBUG,
-                "Error reading \"size\"");
-
+                "Error while reading opcode instruction.");
```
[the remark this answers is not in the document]
necouchman: Okay, I'll just pull it out.  But, I think I forgot a return statement that should probably be added.

src/libguac/user-handlers.c
```diff
@@ -581,3 +593,134 @@ int __guac_handle_disconnect(guac_user* user, int argc, char** argv) {
...
+
+        /* If recognized, call handler */
+        if (strcmp(opcode, current->opcode) == 0)
+            return current->handler(user, argc, argv);
+
+        current++;
+    }
+
+    /* If unrecognized, ignore */
+    return 0;
```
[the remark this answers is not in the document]
necouchman: Yep, good idea.

src/libguac/user-handshake.c
```diff
@@ -286,11 +230,72 @@ static int guac_user_start(guac_parser* parser, guac_user* user,
...
+                    parser->opcode);
+
+            guac_parser_free(parser);
+            return 1;
+            
+        }
+        
+    }
+    
+    /* If we get here something has gone wrong. */
```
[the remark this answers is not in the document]
necouchman: Yep, good point.

src/libguac/user-handshake.c
necouchman: Ah, that would be why I should have stuck with counting them :smile:.

src/libguac/user-handshake.c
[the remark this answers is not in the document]
necouchman: Fixed.

src/libguac/user-handshake.c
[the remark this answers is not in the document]
necouchman: Fixed.

src/libguac/user-handlers.c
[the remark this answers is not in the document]
necouchman: Documented and debugged.

src/libguac/user-handshake.c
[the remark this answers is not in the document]
necouchman: Documented and logged.

src/libguac/user-handshake.c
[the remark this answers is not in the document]
necouchman: Logged.

necouchman: Okay, should be verified, now.

src/libguac/user-handshake.c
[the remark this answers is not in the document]
necouchman: Resolved.

necouchman:
@mike-jumper: Okay, latest round should be handled.

src/libguac/user-handlers.c
```diff
@@ -581,3 +593,136 @@ int __guac_handle_disconnect(guac_user* user, int argc, char** argv) {
...
+
+        /* If recognized, call handler */
+        if (strcmp(opcode, current->opcode) == 0)
+            return current->handler(user, argc, argv);
+
+        current++;
+    }
+
+    /* If unrecognized, log and ignore */
+    guac_user_log(user, GUAC_LOG_WARNING, "Handler not found for \"%s\"",
```
mike-jumper: I think you mean `GUAC_LOG_DEBUG`.

src/libguac/user-handshake.c
```diff
@@ -286,11 +230,78 @@ static int guac_user_start(guac_parser* parser, guac_user* user,
...
     guac_client* client = user->client;
-
+    
+    user->info.audio_mimetypes = NULL;
+    user->info.image_mimetypes = NULL;
+    user->info.video_mimetypes = NULL;
+    user->info.timezone = NULL;
+    
+    /* Count number of arguments. */
+    int numArgs;
```
mike-jumper: Per existing style here and http://guacamole.apache.org/guac-style/#naming, please use `lowercase_with_underscores` rather than `headlessCamelCase` for variable names in C.

The `headlessCamelCase` convention is used only for the Java and JavaScript of parts of Guacamole

src/libguac/user-handlers.c
necouchman: Uh, sure :-).  I went with warning because I thought maybe that was a warn-able event, but debug is fine with me.

PR #221, opened 2019-04-08
GUACAMOLE-772: Reducing image Docker size for Guacd

IT-Experte:
This pull request is connected to [JIRA CR-772](https://issues.apache.org/jira/browse/GUACAMOLE-772).


necouchman:
The commit message needs to be tagged with "GUACAMOLE-772:" the same as the Pull Request title.  See other pull requests for examples.

PR #225, opened 2019-06-02
Adding libs for guacenc

lshahar:


sreid8:
I've built a docker image with guacenc in it too, but I wonder if this were to be merged, whether it more sent to keep the building of guacenc in a separate dockerfile, like a Dockerfile.guacenc or something. I'm not convinced that the vast majority of deployments need guacenc in the image, so for most people it would just make their image bigger for no reason.

lshahar:
We can consider to add small image that includes only the guacenc ?
u know if there is any plan to adding the playback view ? once the playback view will be part of the interface the guacenc will be unnecessary at all.

necouchman:
@lshahar: Thanks for the contribution.  A couple of notes for you:
- All pull requests need to have an associated JIRA issue with them as part of the process of tracking the issue - bug, enhancement, etc.  These can be created at the Guacamole JIRA page (https://issues.apache.org/jira/browse/GUACAMOLE) and you can see full contribution guidelines on our main web page (http://guacamole.apache.org/open-source/).
- I tend to agree with @sreid8 in the observation that most people using guacd in Docker will probably not want guacenc, as well.  I don't feel strongly about that, just seems like most using Docker will be using it in a very non-interactive way, and probably don't need an interactive decoding tool.  Having a separate container for it might be an option.
- I believe there are some playback examples - see https://github.com/apache/guacamole-client/tree/master/doc/guacamole-playback-example.

mike-jumper:
I would agree with @sreid8, as well. I don't believe the guacd Docker image is the proper home for a build of the guacenc tool. The guacd image should just contain guacd and its required libraries.

A separate, automated container could be interesting.

mike-jumper:
@lshahar, as an aside, if you do decide to re-approach this as a separate image, please be sure to describe the nature of your commits at a high level going forward. Commit messages should be written with the intent of being useful to someone who wants to understand the purpose of your changes, particularly when looking at the git history sometime in the future.

A message like "Update Dockerfile" (commit ef0561e) just duplicates what git already provides automatically. It doesn't add any information and doesn't capture your intent behind the change.

It's particularly important to not fall into the trap of assuming the change itself is simple enough that anyone would understand your intent and reasoning by looking at the patch. That can only be true if the change is correct, and *that* is never a safe assumption.

sreid8:
@mike-jumper Does the current build process have any way to build just libguac and guacenc? I admit I haven't looked too closely myself, but whenever I've needed to build guacenc, it seems like I also build guacd whether I wanted to or not.

mike-jumper:
> Does the current build process have any way to build just libguac and guacenc? I admit I haven't looked too closely myself, but whenever I've needed to build guacenc, it seems like I also build guacd whether I wanted to or not.

Yes - take a look at the output of `./configure --help`. Both specific binaries and dependencies can be individually excluded through additional options given to `configure`. The option to not build guacd is `--disable-guacd`.

sreid8:
Perfect. Then I think it would be interesting to see a specific guacenc Dockerfile rather than pollute guacd's

mike-jumper:
Closed given the overall consensus that including guacenc within the guacd Docker image would be the wrong way forward.

@lshahar, if you'd like to rework things such that guacenc is provided as a separate Docker image, presumably automated in some fashion, feel free to open a different PR or to discuss things further on the dev@guacamole.apache.org list.

PR #226, opened 2019-07-05
GUACAMOLE-547: Do not ignore password from settings.

fgiuba:
The password set for a SSH connection is always ignored.

Every SSH password set as connection parameter is now ignored, even if it is not NULL.
The issue is quite simple to reproduce:
1. Set up a working SSH connection with a password parameter with the latest stable (1.0.0).
2. Try the same connection compiling master with GUACAMOLE-547 commits. The terminal prompt the password request, ignoring the configured one.

### DIFF
--- a/src/protocols/ssh/ssh.c
+++ b/src/protocols/ssh/ssh.c
@@ -131,7 +131,7 @@ static guac_common_ssh_user* guac_ssh_get_user(guac_client* client) {
     } /* end if key given */
 
     /* If available, get password from settings */
-    else if (settings->password != NULL){
+    else if (settings->password != NULL) {
         guac_common_ssh_user_set_password(user, settings->password);
     }
 
--- a/src/protocols/ssh/ssh.c
+++ b/src/protocols/ssh/ssh.c
@@ -130,6 +130,11 @@ static guac_common_ssh_user* guac_ssh_get_user(guac_client* client) {
 
     } /* end if key given */
 
+    /* If available, get password from settings */
+    else if (settings->password != NULL){
+        guac_common_ssh_user_set_password(user, settings->password);
+    }
+
     /* Clear screen of any prompts */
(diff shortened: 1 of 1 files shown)


necouchman:
This change is not required - parsing the password setting is handled perfectly fine within the code.  I have tested with the current git master and it seems to work fine.

src/protocols/ssh/ssh.c
```diff
@@ -130,6 +130,11 @@ static guac_common_ssh_user* guac_ssh_get_user(guac_client* client) {
 
     } /* end if key given */
 
+    /* If available, get password from settings */
+    else if (settings->password != NULL){
```
necouchman: Slight style nitpick, here - should be a space between the `)` and the `{`.

mike-jumper:
Also, watch out - there's a bunch of commits from `master` that have appeared following the change of merge base.

`staging/1.1.0` is correct, as GUACAMOLE-547 is part of the 1.1.0 scope, but the relevant commit(s) will need to be rebased.

mike-jumper:
Looking at commit 9cf04eb5d1268c75cda25f9b7edfbf802a588ffe directly (can't look at the changes tab directly due to the merge base issue noted above), I concur on `){` vs `) {`. It's worth removing that splinter.

So, with the exception of:

 * Needs rebasing on top of `staging/1.1.0`
 * Adding a space will make me feel better

LGTM

fgiuba:
Oops, missing space added.

necouchman:
@fgiuba: As Mike mentions, you need to rebase this pull request on the staging/1.1.0 branch rather than master, since this is going into the 1.1.0 release.  Once you do that this will be ready to merge.

fgiuba:
Just rebased the branch, now it should be ready to be merged.

necouchman:
@fgiuba: Thanks for the fix!

PR #228, opened 2019-07-17
GUACAMOLE-221: Implement guacd support for prompting

necouchman:
Okay, here we go...this pull request implements support for two major components of getting Guacamole to support prompting:
* Support for sending the "required" instruction to the client, along with whatever is being required.
* Support for reading back the required input from the client via the argv instruction, including various handlers for the protocols to handle (mostly authentication) data sent back.

I've tested where I can to try to make sure everything works - RDP and SSH seem to be good (along with some client changes that I'll need to clean up and submit, soon).  I don't have a great VNC setup to try that out with, but I tested what I could.

Also, at this moment this change does not handle collecting and sending multiple required parameters, and then reading them back concurrently - it does things one at a time.  I know that was part of the discussion somewhere (other PR, mailing list - something), so I'm happy to take a stab at that, too, but wanted to get some other eyes on this code, first.

Also, I'm not sure if this particular change warrants a rev of the protocol version introduced in 1.1.0 - it does add the `required` instruction??


mike-jumper:
Should the new `required` instruction allow for multiple parameters, similar in principle to `args`?

It feels awkward to request credentials one at a time, particularly in the common username+password case.

necouchman:
@mike-jumper: Yep, definitely.  I'll continue to work that...

necouchman:
@mike-jumper: Okay, protocol side has been reworked to act like `send_args` where the type is a `const char**` and it loops through a `NULL`-terminated array gather up everything.

I retooled the RDP credential request function to leverage it - it feels a bit clunky and probably needs some work.

I'm not sure how feasible it is on the SSH side to gather up all of the required credentials and send them all at once - the various requests for credentials are spread out quite a bit through different pieces of the code.  Or maybe it's just an opportunity to do things differently :-).

necouchman:
@mike-jumper: Well, most of that was pretty straight-forward on the server side, but not entirely sure how to handle multiple values being sent.  The RDP code is probably the best current example of this - the way it's constructed right now:
* guacd sends parameters to client that need to be input.
* RDP authentication callback kicks off `pthread_cond_wait` waiting for request to be filled.
* Prompts are presented to the client, which sends them back one at a time.
* argv handler triggers completion via the `pthread_cond_broadcast` as soon one of the parameters is received.

I don't think it's going to work the way I currently have it implemented because the thread is going to continue as soon as the first value is received, and not wait for the remaining ones.  Any suggestions on how I should proceed that would satisfy all of the following:
* Allow multiple parameters to be both requested (guacd -> client) and received (client->guacd) in a single shot
* Allow callback functions of the various libraries to be used.
* Allow for the returned username, password, and/or domain (in the case of RDP) to be empty strings.

??

necouchman:
Okay, think I figured something out for this - implemented some flags for the conditional stuff, maybe that'll work...

src/libguac/protocol.c
```diff
@@ -961,6 +961,37 @@ int guac_protocol_send_rect(guac_socket* socket,
 
 }
 
+static int __guac_protocol_send_required(guac_socket* socket,
```
mike-jumper: Please document this function.

Also if sending a NULL-terminated array of strings as a series of instruction elements is common here (it looks like we've done this with at least `args` and `connect`, as well), I think it's worth considering whether we should extract the commonality so there is one internal function leveraged by all three, similar to how we have `__guac_socket_write_length_string()` and `__guac_socket_write_length_int()`.

src/protocols/rdp/argv.c
```diff
@@ -0,0 +1,192 @@
...
+static int guac_rdp_argv_end_handler(guac_user* user,
+        guac_stream* stream) {
+
+    guac_client* client = user->client;
+    guac_rdp_client* rdp_client = (guac_rdp_client*) client->data;
+    guac_rdp_settings* settings = rdp_client->settings;
+    
+    /* Append null terminator to value */
+    guac_rdp_argv* argv = (guac_rdp_argv*) stream->data;
+    argv->buffer[argv->length] = '\0';
```
mike-jumper: Is there a guarantee that there will be space for this null terminator within `argv->buffer`?

src/protocols/rdp/argv.c
```diff
@@ -0,0 +1,192 @@
...
+
+    /* Apply changes to chosen setting */
+    switch (argv->setting) {
+
+        /* Update RDP username. */
+        case GUAC_RDP_ARGV_SETTING_USERNAME:
+            free(settings->username);
+            settings->username = malloc(strlen(argv->buffer) * sizeof(char));
+            strcpy(settings->username, argv->buffer);
```
mike-jumper: `strdup()` could be easier than `malloc(strlen(...))` followed by `strcpy()`. That said, it looks like already have the length with `argv->length`, and so manually terminating the input, recalculating the length with `strlen()`, etc. feels superfluous.

Perhaps `strndup(..., argv->length)` would solve everything?

src/protocols/rdp/argv.c
```diff
@@ -0,0 +1,192 @@
...
+            rdp_client->rdp_cond_flags ^= GUAC_RDP_COND_FLAG_USERNAME;
```
mike-jumper: Is the intent here to always set, always clear, or toggle the flag?

src/protocols/rdp/rdp.h
```diff
@@ -48,6 +48,21 @@
...
+
+/**
+ * A flag for tracking if we are waiting conditionally on a password.
+ */
+#define GUAC_RDP_COND_FLAG_PASSWORD 2
+
+/**
+ * A flag for tracking if we are waiting conditionally on a domain.
+ */
+#define GUAC_RDP_COND_FLAG_DOMAIN 3
```
mike-jumper: Since 3 is represented in binary as `11`, it cannot be safely used as the value of a bitwise flag intended to represent a state independent of the other flags. They would all need to be powers of two. I think you meant 4 here.

src/protocols/rdp/rdp.h
```diff
@@ -166,6 +181,17 @@ typedef struct guac_rdp_client {
      * reallocated during reconnection.
      */
     pthread_mutex_t rdp_lock;
+    
+    /**
+     * Condition which is used when the pthread needs to wait for a
+     * particular condition to be met.
```
mike-jumper: That particular condition needs to be pretty well defined, at least on our part, for waiting on this condition to make sense. If this condition and the associated flags are to be used for requesting arguments, for requesting credentials, etc. we can make this much clearer by specifically intending them for that use and documenting them as such.

src/protocols/ssh/ssh.h
```diff
@@ -89,6 +89,12 @@ typedef struct guac_ssh_client {
      * Lock dictating access to the SSH terminal channel.
      */
     pthread_mutex_t term_channel_lock;
+    
+    /**
+     * Condition used when SSH client thread needs to wait for Guacamole
+     * client response.
```
mike-jumper: Same here - I think the usage of this condition should much more specific than an arbitrary client response of some unknown kind. It would be better to be explicit in the naming and documentation of this and its corresponding flags.

src/protocols/rdp/argv.c
necouchman: Intent would be to clear this particular flag if the username is received - same with the ones below this.

src/protocols/rdp/rdp.h
necouchman: Should be fixed.

src/protocols/ssh/ssh.h
necouchman: Okay, I've switched this to be more specific.

src/protocols/rdp/rdp.h
necouchman: Okay, I've redefined this to be more specific to credential gathering.

src/protocols/rdp/argv.c
necouchman: Okay, after some quick Google reading, I think I've managed to do this correctly...

src/protocols/rdp/argv.c
necouchman: Less code, I like it.  Fixed.

src/protocols/rdp/argv.c
necouchman: I haven't looked, but my assumption (yes, I know) is that when `argv->length` is calculated, it cannot be higher than the length of the buffer, so the worst that would happen is this kills the last character of the provided string?

Also, wasn't sure if you were saying in the other comment that using `strndup()` makes this unnecessary??

mike-jumper: If `buffer` points to a buffer of exactly `length` bytes, then `buffer[length]` is outside the buffer, not the last byte of the buffer.

src/libguac/protocol.c
necouchman: Documentation added.  Also implemented common function `__guac_socket_write_array()` for writing out a NULL-terminated array.

src/protocols/rdp/argv.c
necouchman: Ah, yes, that makes sense.  Is adding the NULL-terminator necessary?  I'm guessing I copied it from other code - probably from the code that updates terminal values SSH/Telnet sessions.  I'm happy to remove it completely if there's no need for it.

mike-jumper: Yep, definitely necessary. Nothing handling the received data as a string will be able to function safely without that null terminator.

The beauty of `strndup(foo, length)` is that it will automatically add that null terminator and ensure that there is space for it within the newly-allocated buffer. Assuming the string within the buffer is indeed `length` characters long, `strndup(..., length)` would allocate buffer space of at least `length + 1` bytes.

necouchman: Okay, so do the changes I've made address this, since I've moved it to `strndup()`, or does the null terminator need to be added to the original `argv->buffer` string?

necouchman: From re-reading your above comments, adding it manually is not required when using `strndup()`

necouchman:
Okay, made some changes, fixed up some style and such, and has been rebased after RDP updates.

src/protocols/rdp/rdp.c
```diff
@@ -223,10 +223,53 @@ static BOOL rdp_freerdp_authenticate(freerdp* instance, char** username,
...
+    if (settings->domain == NULL || strcmp(settings->domain, "") == 0) {
+        params[i] = "domain";
+        rdp_client->rdp_credential_flags |= GUAC_RDP_CRED_FLAG_DOMAIN;
+        i++;
+    }
+    
+    /* NULL-terminate the array. */
+    params[i] = NULL;
+    
+    if (i > 0) {
```
knacktim, line 252: @necouchman Would it be reasonable for us to put an extra parameter here to tell what client what form to request if it doesn't have one?  That would circumvent the CONNECTION_GROUP issue

necouchman: Maybe so, but I don't think we necessarily need anything on the server-side of this, I think it can all be accomplished on the guacamole-client side (I think).  My original code had it looking up the protocol for the connection and pulling down the form manually, but I opted for this route based on some of the changes Mike had made for the hidden menu and making changes on-the-fly.  I'm sure there's a way to do this elegantly :-).

knacktim: Alrighty, I'll keep poking around on the UI!

src/protocols/rdp/argv.c
```diff
@@ -0,0 +1,188 @@
...
+    if (length > remaining)
+        length = remaining;
+
+    /* Append received data to end of buffer */
+    memcpy(argv->buffer + argv->length, data, length);
+    argv->length += length;
+
+    return 0;
+
+}
```
mike-jumper, line 101: Thoughts on replacing repeated use of this pattern with a convenience function at the libguac level? I can imagine something like `guac_user_receive_string()` that takes over the `guac_stream` and accepts a callback for the final value would result in a fair amount of code deletion.

src/protocols/rdp/rdp.c
```diff
@@ -227,10 +227,53 @@ static BOOL rdp_freerdp_authenticate(freerdp* instance, char** username,
...
+        params[i] = "password";
+        rdp_client->rdp_credential_flags |= GUAC_RDP_CRED_FLAG_PASSWORD;
+        i++;
+    }
+    
+    if (settings->domain == NULL || strcmp(settings->domain, "") == 0) {
+        params[i] = "domain";
+        rdp_client->rdp_credential_flags |= GUAC_RDP_CRED_FLAG_DOMAIN;
+        i++;
+    }
```
mike-jumper, line 251: Can these ever be the empty string? It looks to me that argument parsing will ensure that an omitted parameter will be represented by `NULL`:

https://github.com/apache/guacamole-server/blob/c10ceab7e80f9d6eae9ad98254bf97f41a0de112/src/protocols/rdp/settings.c#L760-L773

If the concern here is an empty `argv` stream, perhaps the handling of the stream should ensure that it consistently assigns `NULL` for the relevant setting in those cases, such that code further downstream need not be aware of those concerns?

src/protocols/rdp/rdp.c
```diff
@@ -227,10 +227,53 @@ static BOOL rdp_freerdp_authenticate(freerdp* instance, char** username,
...
+    if (i > 0) {
+        /* Lock the client thread. */
+        pthread_mutex_lock(&(rdp_client->rdp_credential_lock));
+        
+        /* Send require params and flush socket. */
+        guac_protocol_send_required(client->socket, (const char**) params);
+        guac_socket_flush(client->socket);
+        
+        /* Wait for condition. */
+        pthread_cond_wait(&(rdp_client->rdp_credential_cond), &(rdp_client->rdp_credential_lock));
```
mike-jumper, line 265: We'll need to be careful here - this will potentially block indefinitely. There are additional cases that aren't yet being handled that we need to consider:

* If the connection is closed while this is waiting, is there anything to ensure it stops waiting? Does the blockage need to stop prior to attempting to free FreeRDP structures?
* This lock and the condition need to be properly destroyed in the free handler. I see some new cleanup within the VNC free handler, but not for RDP.
* `pthread_cond_destroy()` is documented as potentially failing with `EBUSY` if another thread is using the condition with `pthread_cond_wait()`: https://pubs.opengroup.org/onlinepubs/009695399/functions/pthread_cond_destroy.html

There is failsafe code in guacd which will forcibly terminate the process if the connection closes yet fails to clean itself up in a reasonable time, but we should aim to avoid that needing to kick in.

src/protocols/rdp/rdp.h
```diff
@@ -153,6 +168,25 @@ typedef struct guac_rdp_client {
...
+    
+    /**
+     * Condition which is used when the pthread needs to wait for one or more
+     * credentials to be provided by the client.
+     */
+    pthread_cond_t rdp_credential_cond;
+    
+    /**
+     * Flags for tracking events related to the rdp_credential_cond
+     * pthread condition.
```
mike-jumper, line 186: Recommend pointing to the specific flags with `@see`.

src/protocols/rdp/rdp.c
```diff
@@ -227,10 +227,53 @@ static BOOL rdp_freerdp_authenticate(freerdp* instance, char** username,
...
+    
+    /* NULL-terminate the array. */
+    params[i] = NULL;
+    
+    if (i > 0) {
+        /* Lock the client thread. */
+        pthread_mutex_lock(&(rdp_client->rdp_credential_lock));
+        
+        /* Send require params and flush socket. */
+        guac_protocol_send_required(client->socket, (const char**) params);
```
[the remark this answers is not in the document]
necouchman: Yeah, that makes sense.  However, question on the implementation of that related to the above comment about the `guac_socket_require_keep_alive()` call...
* Presumably this is done using the `guac_client_for_owner()` function, passing the `client` object, `guac_protocol_send_required` as the callback, and then `(const char**) params` as the data?
* Should the call to `guac_socket_require_keep_alive()` be limited to the owner, or is it safe to call that on the `client->socket` object, here?

src/libguac/protocol.c
```diff
@@ -961,6 +975,50 @@ int guac_protocol_send_rect(guac_socket* socket,
...
+ *     to the client.
+ *
+ * @return
+ *     Zero if successful, non-zero on error.
+ */
+static int __guac_protocol_send_required(guac_socket* socket,
+        const char** required) {
+
+    // The socket should be kept alive while waiting for user response.
+    guac_socket_require_keep_alive(socket);
```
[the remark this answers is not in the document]
necouchman: So, had a question related to this below on the owner thread, but, question, here - it looks like at least SSH and VNC already set this in the main thread to handle lengthy connects.  Presumably once it's set on a socket it does not have to be re-set - so the usages toward the beginning of the `guac_vnc_client_thread()` and `ssh_client_thread()` functions set it for the duration of the SSH session, correct?  Or do the sockets come and go for a particular connection and it would have to be set, again, during other phases of the connection?

If my assumptions are correct...
* Is there any reason not to just add it to the main RDP thread, prior to authentication?
* With 3 out of 5 protocols implementing the requirement for keep-alives very early on in the connection process, and assuming the keep-alives are then active for the duration of the connections, is there any reason not to just set it across the board?

src/common-ssh/ssh.c
```diff
@@ -360,7 +360,7 @@ static int guac_common_ssh_authenticate(guac_common_ssh_session* common_session)
 
     /* Attempt authentication with username + password. */
     if (user->password == NULL && common_session->credential_handler)
-            user->password = common_session->credential_handler(client, "Password: ");
+        common_session->credential_handler(client, "password");
```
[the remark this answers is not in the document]
necouchman: Sounds good, I'll set up an include file for that and implement.

src/protocols/rdp/argv.c
necouchman: Yeah, seems like a good idea - I think they are all reasonably similar enough that it makes sense to combine them.

src/protocols/rdp/rdp.c
necouchman: Good point - and, there could be legitimate cases where users want an empty string versus a `NULL` value.  However, I'm not sure that there's currently a way to pass an empty string through??  But I think we can at least just simplify it to `NULL` checks and skip the empty string checks.

src/protocols/rdp/rdp.c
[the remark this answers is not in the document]
necouchman: Okay, I think I figured this out - looks like `guac_client_for_owner()` won't work because it sends the `client->__owner` to the callback and does not send a `socket`.  So, I used `client->__owner->socket` for both the `guac_socket_require_keep_alive()` and the `guac_protocol_send_required()` call, along with the flush.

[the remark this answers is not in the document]
mike-jumper: Alrighty, a couple of issues there:

1. Nothing should be referencing the private portions of libguac structures except libguac. To safely access the owner of a connection, you really do need to use `guac_client_for_owner()`.

   It should work just fine. In any case that you would be tempted to reference `client->__owner->whatever`, you can instead do `user->whatever` for the user passed to the callback.

2. You probably actually want to use the broadcast socket for `guac_socket_require_keep_alive()`.

   The utility of that function is that it ensures the other side of the connection doesn't drop out while waiting for data from the server. It is true that we probably only want to request `argv` from the owner, but if the connection is shared, we don't want _any_ of those connections to drop out. A user sharing the connection shouldn't get disconnected just because the owner of the connection is taking their time to enter their credentials.

src/common-ssh/ssh.c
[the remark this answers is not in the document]
necouchman: Okay, I've implemented those constants for all of the parameters in each of the protocols that might be adjusted via `required` and/or `argv` instructions.  Let me know if you think I should go all the way with all of the parameters...

src/protocols/rdp/rdp.c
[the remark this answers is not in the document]
necouchman: > Using guac_client_for_owner() should work just fine whenever you need a guac_user. In any case that you would be tempted to reference client->__owner->whatever, you can instead do user->whatever for the user passed to the callback.

So, should the `guac_protocol_send_required()` function be re-worked to take `guac_user` instead of a `guac_socket`?  Or should I implement some intermediate function in `user.c` that uses this `guac_protocol_send_required()` function and just passes through the `guac_user->socket`?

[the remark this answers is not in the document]
mike-jumper: > So, should the `guac_protocol_send_required()` function be re-worked to take `guac_user` instead of a `guac_socket`?

No, I think the `guac_protocol_send_*()` family of functions should remain unaware of the concerns of `guac_user`, `guac_client`, etc. They are intended as low-level functions for sending instructions over a `guac_socket`.

> Or should I implement some intermediate function in `user.c` that uses this `guac_protocol_send_required()` function and just passes through the `guac_user->socket`?

This would be the way, yes.

As this would be a common need, if you're looking to avoid having to do this everywhere, I think it could make sense to provide a convenience function like `guac_client_owner_send_required()` or similar.

src/libguac/protocol.c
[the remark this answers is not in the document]
mike-jumper: > Presumably once it's set on a socket it does not have to be re-set - so the usages toward the beginning of the `guac_vnc_client_thread()` and `ssh_client_thread()` functions set it for the duration of the SSH session, correct?

Correct.

> Is there any reason not to just add it to the main RDP thread, prior to authentication?

Nope. It imposes very minimal overhead, and is already designed to not bother sending pings unless nothing has been sent along the `guac_socket` in a while.

> With 3 out of 5 protocols implementing the requirement for keep-alives very early on in the connection process, and assuming the keep-alives are then active for the duration of the connections, is there any reason not to just set it across the board?

Actually ... I think this would be fine, too. That fact would need to be documented, but it seems like that would be best. We can then mark `guac_socket_require_keep_alive()` as deprecated and note that it is now a no-op.

src/protocols/rdp/rdp.c
mike-jumper: For `settings->foo` coming from `guac_user_parse_args_string(..., NULL)`, yes, this will always be `NULL`. It would only perhaps be `""` if the call were `guac_user_parse_args_string(..., "")`.

BUT, it is likely possible for things to be `""` if the value is set via an empty `argv` stream. If that's the case, then I would suggest ensuring that handling of the `argv` stream uses `NULL` rather than `""` if it receives zero bytes overall.

src/protocols/rdp/argv.c
mike-jumper: BTW, an idea just occurred to me that may work well here. We currently provide a number of convenience functions for parsing arguments like `guac_user_parse_args_string()`:

https://github.com/apache/guacamole-server/blob/master/src/libguac/guacamole/user.h#L875-L876

Perhaps we should provide analogous `argv` variants?

necouchman: Sounds good - I'll take a look and give it a shot.

src/protocols/rdp/rdp.c
necouchman: Got it - I've adjusted everything to behave that way.

src/protocols/rdp/rdp.c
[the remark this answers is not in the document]
necouchman: Okay, shot has been taken at an implementation of `guac_client_owner_send_required()`.  Guessing it'll need some tweaking, but it's a start.

src/libguac/protocol.c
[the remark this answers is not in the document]
necouchman: Is there a central location where the keep alive should be enabled when a socket starts up, or just do it at the beginning of each protocol implementation?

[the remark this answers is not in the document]
necouchman: I think I got this in the right spot...

src/protocols/rdp/rdp.c
necouchman: Okay - I'll dig around and see what I can find to help figure out how to clean it up.  I added the destroy call for the RDP cleanup handler, but I may need further guidance on how to make sure the `pthread_cond` is not stuck waiting.  I'll take a run at it, though.

src/protocols/rdp/rdp.c
```diff
@@ -227,10 +227,53 @@ static BOOL rdp_freerdp_authenticate(freerdp* instance, char** username,
...
+        guac_socket_flush(client->socket);
```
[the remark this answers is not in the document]
necouchman: Okay, I have taken a wild stab at implementing support for this on the server-side.  It compiles, now, but I haven't made the client-side changes that will be necessary to make it work.

I've no doubt that it needs some polishing, so let me know what needs to be changed.

src/libguac/protocol.c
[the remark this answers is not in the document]
mike-jumper: Hm ... looking over 9bfcbbf3a576054f82ce9bce6ef2a0497304ca8a, I'm not sure `guac_socket` allows for this to be done safely by default:

1. The current architecture of `guac_socket` relies on it being impossible for that `guac_socket` to be used before its handlers have been initialized. This is because a pointer to the new `guac_socket` simply does not exist outside the internals of the various `guac_socket` implementations until they have finished their init process.

   If `guac_socket` will now create this thread as part of its own init (which happens first), then there is no guarantee that the handlers will be properly initialized by the implementation  before the thread starts trying to write `nop` instructions, leading to unstable behavior.

2. Creating the keep-alive ping on the broadcast socket is nice and efficient as it only results in a new single thread, with that thread covering all users of a connection. Creating it across the board would result in a new thread for the broadcast socket plus new threads for each of the per-user sockets it writes to, a bit of a waste.

It would be safe for implementations to start the keep-alive ping on their own, but I don't think `guac_socket` can do this on behalf of all implementations. There's just no way for those implementations to let `guac_socket` know that initialization is complete.

My knee-jerk thoughts on the above would be to say that we should start the keep-alive by default within the broadcast socket implementation and leave it at that.

What do you think?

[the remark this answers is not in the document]
mike-jumper: In fact ... given that there are definitely cases where a `guac_socket` does not need a `nop` ping (the tee socket writing screen recordings to a file, for example), perhaps the keep-alive should only be set by default where it is known to be correct, like right after initialization of the `guac_client` via `guac_client_load_plugin()` succeeds within guacd?

Even assuming that the broadcast socket needs a keep-alive ping might be overly presumptuous on the utility of a socket that broadcasts to all users.

src/protocols/ssh/ssh.c
```diff
@@ -57,6 +58,38 @@
...
+ * 
+ * @param cred_name
+ *     The name of the parameter to prompt for in the client.
+ */
+static void guac_ssh_get_credential(guac_client *client, char* cred_name) {
+
+    guac_ssh_client* ssh_client = (guac_ssh_client*) client->data;
+    
+    /* If the client does not support the "require" instruction, just return. */
+    if (!guac_client_supports_require(client))
```
necouchman, line 78: Hmmm...I think we should probably revert to prompting via the terminal in this case...

src/libguac/protocol.c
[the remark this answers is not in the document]
necouchman: So, somewhere around here?

https://github.com/apache/guacamole-server/blob/c10ceab7e80f9d6eae9ad98254bf97f41a0de112/src/guacd/proc.c#L316-L329

Is it safe to assume, there, that the socket is correctly initialized and ready to have the `nop` ping started?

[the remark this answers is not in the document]
necouchman: Went ahead and implemented it, there...let me know if there's somewhere else.  And, thanks, as always, for the patience with me :-).

src/protocols/rdp/rdp.c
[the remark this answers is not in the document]
necouchman: Just a quick update to this - I'm sure you can gather most of it from looking at the code, but the idea I have is that:
*Dduring the handshake process, another instruction will be addded - `protocol` - which the client will use to communicate its protocol level to the server
* guacd will store that version, and, if it doesn't receive a version, will have a default one.  My thought is that the default will be VERSION_1_1_0, since that's the latest version that doesn't require any special instructions client-side.
* The rest of the process will proceed per usual - client and server will negotiate the latest version common to them, and then both sides will store that information for the duration of the session.

src/common-ssh/ssh.c
```diff
@@ -360,7 +361,7 @@ static int guac_common_ssh_authenticate(guac_common_ssh_session* common_session)
 
     /* Attempt authentication with username + password. */
     if (user->password == NULL && common_session->credential_handler)
-            user->password = common_session->credential_handler(client, "Password: ");
+        common_session->credential_handler(client, GUAC_SSH_PARAMETER_NAME_PASSWORD);
```
mike-jumper, line 364: For SSH/telnet, this should probably support both the old-style terminal prompt and the prompt leveraging `required`, internally selecting between these options depending on owner support.

src/guacd/proc.c
```diff
@@ -340,6 +340,9 @@ static void guacd_exec_proc(guacd_proc* proc, const char* protocol) {
         owner = 0;
 
     }
+    
+    /* Enable keep alive on the user socket */
```
mike-jumper, line 344: This is on the broadcast socket.

src/libguac/client.c
```diff
@@ -478,6 +478,29 @@ int guac_client_load_plugin(guac_client* client, const char* protocol) {
 
 }
 
+int guac_client_owner_send_required(guac_client* client, const char** required) {
+
+    /* Don't send require instruction if client does not support it. */
+    if (!guac_client_supports_require(client))
+        return -1;
```
mike-jumper, line 485: Is it `required` or `require`?

src/libguac/client.c
```diff
@@ -478,6 +478,29 @@ int guac_client_load_plugin(guac_client* client, const char* protocol) {
...
+
+    pthread_rwlock_rdlock(&(client->__users_lock));
+
+    /* Invoke callback with current owner */
+    retval = guac_protocol_send_required(client->__owner->socket, required);
+    
+    /* Flush the socket */
+    guac_socket_flush(client->__owner->socket);
+
+    pthread_rwlock_unlock(&(client->__users_lock));
```
mike-jumper, line 497: Rather than manually lock `__users_lock`, we should leverage `guac_client_for_owner()`.

src/libguac/guacamole/protocol-types.h
```diff
@@ -276,5 +278,50 @@ typedef enum guac_line_join_style {
...
+    VERSION_1_1_0 = 110,
+            
+    /**
+     * Protocol version 1.2.0, which supports the "required" instruction,
+     * allowing connections in guacd to request information from the client and
+     * await a response.
+     */
+    VERSION_1_2_0 = 120
+
+} guac_protocol_version;
```
mike-jumper, line 306: With 1.2.0 now being released, and the new `required` instruction being part of 1.3.0, we'll need a constant for that version, as well.

mike-jumper:
> During the handshake process, another instruction will be addded - protocol - which the client will use to communicate its protocol level to the server

Why not use the value communicated by the client via `connect`?

src/libguac/guacamole/socket.h
```diff
@@ -136,11 +136,25 @@ void guac_socket_free(guac_socket* socket);
  * to ensure neither side of the socket times out while the socket is open.
  * This ping will take the form of a "nop" instruction.
  *
+ * @deprecated
+ *     Manually starting the keep alive process on sockets is no longer
+ *     necessary, as the sockets enable the "nop" ping by default.
+ *
```
mike-jumper, line 142: This is no longer the case, correct? guacd will turn on the keep-alive ping automatically for the only socket that should generally require it (the broadcast socket), but this function no longer needs to be deprecated?

src/libguac/guacamole/socket.h
```diff
@@ -136,11 +136,25 @@ void guac_socket_free(guac_socket* socket);
...
 
+/**
+ * Enables the keep-alive ping functionality on the given socket, causing it
+ * to automatically send a "nop" instruction ping periodically while the
+ * socket is open.
+ * 
+ * @param socket
+ *     The guac_socket on which to enable the automatic keep-alive ping.
+ */
+void guac_socket_enable_keep_alive(guac_socket* socket);
```
mike-jumper, line 156: Why a new function rather than `guac_socket_require_keep_alive()`?

src/libguac/guacamole/protocol-constants.h
```diff
@@ -38,7 +38,7 @@
  * This version is passed by the __guac_protocol_send_args() function from the
  * server to the client during the client/server handshake.
  */
-#define GUACAMOLE_PROTOCOL_VERSION "VERSION_1_1_0"
+#define GUACAMOLE_PROTOCOL_VERSION "VERSION_1_2_0"
```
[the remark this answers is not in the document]
necouchman: Okay, I wasn't sure if it should just increment, or increment to match the version, but that sounds fine to me.

src/libguac/guacamole/protocol-types.h
```diff
@@ -276,5 +278,50 @@ typedef enum guac_line_join_style {
...
+     * to the server.
+     */
+    VERSION_1_1_0 = 110,
+            
+    /**
+     * Protocol version 1.2.0, which supports the "required" instruction,
+     * allowing connections in guacd to request information from the client and
+     * await a response.
+     */
+    VERSION_1_2_0 = 120
```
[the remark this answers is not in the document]
necouchman: Aha, yes.

src/common-ssh/ssh.c
necouchman: Yep, agreed.

src/libguac/guacamole/protocol-types.h
necouchman: So, I guess my question, here, is: is there actually a protocol version 1.2.0?  Since, protocol-wise, 1.2.0 is identical to 1.1.0, and we're making the changes in 1.3.0?

src/libguac/guacamole/socket.h
necouchman: I was thinking of just changing the name, since it isn't that we "require" keep alive on a particular socket, we're just enabling it across the board.

src/libguac/guacamole/protocol-types.h
mike-jumper: Ah, I see what you're saying. Yes, that's correct, there would be no protocol version 1.2.0.

src/libguac/guacamole/socket.h
mike-jumper: Well ... internally, yes, I agree it's more of an "enable" than a "require".

The idea behind the original `guac_socket_require_keep_alive()` naming was to define that function as a means of declaring a requirement, rather than imposing a specific method for socket implementations to satisfy that requirement, decoupling the statement of need from how/whether the need is met.

Admittedly, the wording in the documentation of `guac_socket_require_keep_alive()` which explicitly specifies a "keep-alive ping" and states that "this ping will take the form of a "nop" instruction" probably defeats that intent.

src/libguac/client.c
```diff
@@ -633,6 +656,12 @@ static void* __webp_support_callback(guac_user* user, void* data) {
 }
 #endif
 
+int guac_client_supports_require(guac_client* client) {
+    
+    return guac_user_supports_require(client->__owner);
```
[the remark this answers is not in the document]
necouchman: Okay, I think I've managed to do this correctly.

src/libguac/client.c
necouchman: Should be `required` - I believe I've updated everything correctly, but will take another look through to make sure.

src/libguac/guacamole/socket.h
necouchman: Okay, I just kept "require" and undid the deprecation of it for the time being.  We can change it, if that makes sense, but I'm also all for not making changes when there isn't a great reason to change anything.

src/libguac/client.c
necouchman: Okay, went through and updated the comments in various places as well - this should be consistently documented as `required`.

src/libguac/client.c
necouchman: So, the only thing that I'm struggling with here is the callback - the current specification for the callback allows for the one `void *` data field, which I need to use to pass the required parameters through.  I haven't figured out a good way to get the `int` return value of the underlying `guac_protocol_send_required()` function to pass through the `void *` return value of the callback function.

FWIW, I'm not sure how much it really matters to track the return value of `guac_protocol_send_required()` - in the all of the places where it was used directly in the past I don't think I ever made any use of the return value, but it makes me a little twitchy to have a function that returns something that just never gets looked at.

Hints or opinions on the best way to handle this, all-around?

src/protocols/ssh/ssh.c
mike-jumper: Yep.

src/libguac/client.c
mike-jumper: Yep - there's a standard C99 way to losslessly transform between `int` and ` void*` and back: `intptr_t`, provided by `stdint.h`.

Here's an example of `int` to `void*`:

https://github.com/apache/guacamole-server/blob/c10ceab7e80f9d6eae9ad98254bf97f41a0de112/src/guaclog/keydef.c#L166

And here's an example of `void*` to `int`:

https://github.com/apache/guacamole-server/blob/c10ceab7e80f9d6eae9ad98254bf97f41a0de112/src/guaclog/keydef.c#L143

If `intptr_t` weren't a possibility, the next option would be to pass everything that the callback needs within a `struct`, including a location to store what would otherwise be its return value.

necouchman: Brilliant.  I think I've managed to implement this correctly with `intptr_t`.

src/libguac/protocol.c
```diff
@@ -961,6 +1004,26 @@ int guac_protocol_send_rect(guac_socket* socket,
...
+    
+    guac_socket_instruction_begin(socket);
+    
+    if (guac_socket_write_string(socket, "8.required"))
+        return -1;
+
+    if (guac_socket_write_array(socket, required))
+        return -1;
+
+    ret_val = guac_socket_write_string(socket, ";");
```
mike-jumper, line 1019: FYI, given the definition of the return value of this function, this is equivalent to:

```
ret_val =
       guac_socket_write_string(socket, "8.required")
    || guac_socket_write_array(socket, required)
    || guac_socket_write_string(socket, ";");
```

src/libguac/protocol.c
```diff
@@ -1241,3 +1304,34 @@ int guac_protocol_decode_base64(char* base64) {
...
+        
+        if (current->version == version) {
+            return (const char*) current->version_string;
+        }
+        
+    }
+    
+    return NULL;
+    
+}
```
mike-jumper, line 1336: Looks good, but I would suggest adding unit tests for these, now that our unit testing tooling is more reasonable.

https://github.com/apache/guacamole-server/blob/c10ceab7e80f9d6eae9ad98254bf97f41a0de112/README-unit-testing.md

src/protocols/vnc/vnc.h
```diff
@@ -133,6 +143,27 @@ typedef struct guac_vnc_client {
...
+    
+    /**
+     * A condition to use for signaling the thread when credentials have been
+     * retrieved from the client.
+     */
+    pthread_cond_t vnc_credential_cond;
+    
+    /**
+     * A field to track flags related to retrieving required credentials
+     * from the client.
```
mike-jumper, line 161: The nature of that relation should be documented. Are these flags the parameters that are required but have not yet been provided? Are they the parameters that have been provided so far?

src/protocols/rdp/rdp.h
```diff
@@ -153,6 +168,29 @@ typedef struct guac_rdp_client {
...
```
mike-jumper, line 186: As with VNC, the nature of this value with respect to those flags should be documented here.

src/protocols/ssh/ssh.c
```diff
@@ -281,6 +285,7 @@ void* ssh_client_thread(void* data) {
     }
 
     pthread_mutex_init(&ssh_client->term_channel_lock, NULL);
+    pthread_cond_init(&(ssh_client->ssh_credential_cond), NULL);
```
mike-jumper, line 288: I don't think `term_channel_lock` should reused for `ssh_credential_cond`. It's specific to guarding access to the SSH terminal channel.

A pthread condition needs to have its own mutex which guards modification to the value that is changed when the condition is signaled.

src/libguac/protocol.c
necouchman: I like that a lot better!

src/libguac/client.c
```diff
@@ -478,6 +478,44 @@ int guac_client_load_plugin(guac_client* client, const char* protocol) {
...
+    
+}
+
+int guac_client_owner_send_required(guac_client* client, const char** required) {
+
+    /* Don't send required instruction if client does not support it. */
+    if (!guac_client_owner_supports_required(client))
+        return -1;
+    
+    return *(int*) guac_client_for_owner(client, guac_client_owner_send_required_callback, required);
```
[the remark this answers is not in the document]
necouchman: As it was written, yes, but I think I've since updated that - hopefully it's better, now?!

src/libguac/guacamole/protocol-types.h
```diff
@@ -20,6 +20,8 @@
 #ifndef _GUAC_PROTOCOL_TYPES_H
 #define _GUAC_PROTOCOL_TYPES_H
 
+#include <stddef.h>
+
```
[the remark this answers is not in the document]
necouchman: I was getting an error that `NULL` wasn't defined - that seems to be at least one of the places it is usually defined?

src/libguac/protocol.c
necouchman: Okay, I *think* I've got these implemented correctly with the latest commit...

src/protocols/ssh/ssh.c
necouchman: Okay, well, I think I have this working, but it ain't pretty.  Happy to take advice, pointers, or whatever code you might be cooking up...

necouchman:
> Why not use the value communicated by the client via connect?

image

mike-jumper:
> The one thing I'm still not real happy with right now is the SSH code handling of credential retrieval - the prompting through the required instruction is adding complexity that makes it a little messy right now. ...

Might be worth just leaving the existing text-based SSH credential prompting as it is for now, and limiting scope of the initial changes to just VNC and RDP.

necouchman:
> Might be worth just leaving the existing text-based SSH credential prompting as it is for now, and limiting scope of the initial changes to just VNC and RDP.

You're probably right - would keep it consistent with the other terminal-based protocols (Telnet and Kubernetes).  I was hoping to make sure the Keyboard-Interactive support got resolved with it, but might not be worth complicating things at tihs point.

necouchman:
Okay @mike-jumper - I've rolled back the SSH changes and left that with just terminal prompting, as before.

mike-jumper:
Nice - I'll take a look.

src/protocols/vnc/auth.c
```diff
@@ -19,27 +19,110 @@
...
+            i++;
+            vnc_client->vnc_credential_flags |= GUAC_VNC_COND_FLAG_USERNAME;
+        }
+        
+        /* Check if password is null or empty. */
+        if (settings->password == NULL) {
+            params[i] = GUAC_VNC_ARGV_PASSWORD;
+            i++;
+            vnc_client->vnc_credential_flags |= GUAC_VNC_COND_FLAG_PASSWORD;
+        }
```
mike-jumper, line 101: `vnc_credential_flags` should only be modified/checked while the `vnc_credential_lock` is acquired.

src/protocols/vnc/argv.c
```diff
@@ -0,0 +1,59 @@
...
+
+    guac_client* client = user->client;
+    guac_vnc_client* vnc_client = (guac_vnc_client*) client->data;
+    guac_vnc_settings* settings = vnc_client->settings;
+
+    /* Update username */
+    if (strcmp(name, GUAC_VNC_ARGV_USERNAME) == 0) {
+        free(settings->username);
+        settings->username = strdup(value);
+        vnc_client->vnc_credential_flags &= GUAC_VNC_COND_FLAG_USERNAME;
```
mike-jumper, line 43: `vnc_credential_flags` should only be modified/checked while the `vnc_credential_lock` is acquired.

src/protocols/vnc/auth.c
```diff
@@ -19,27 +19,110 @@
...
+        if (i > 0) {
+            /* Lock the thread. */
+            pthread_mutex_lock(&(vnc_client->vnc_credential_lock));
+            
+            /* Send required parameters to owner. */
+            guac_client_owner_send_required(gc, (const char**) params);
+            
+            /* Wait for the parameters to be returned. */
+            pthread_cond_wait(&(vnc_client->vnc_credential_cond),
+                    &(vnc_client->vnc_credential_lock));
```
mike-jumper, line 113: Why not `guac_argv_await()`?

mike-jumper (changes requested):
Before going too much further, I think it might be worth considering the `guac_argv_await()`, `guac_argv_register()`, etc. family of functions that were recently added as part of GUACAMOLE-221 to make this sort of logic easier. It may remove the need for dedicated conditionals, flags, etc. entirely, and allows disconnects in the middle of waiting to be handled safely with `guac_argv_stop()`.

src/protocols/vnc/argv.c
```diff
@@ -0,0 +1,59 @@
...
+    }
+    
+    /* Update password */
+    else if (strcmp(name, GUAC_VNC_ARGV_PASSWORD) == 0) {
+        free(settings->password);
+        settings->password = strdup(value);
+        vnc_client->vnc_credential_flags &= GUAC_VNC_COND_FLAG_USERNAME;
+    }
```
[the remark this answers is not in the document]
necouchman: Ah, so then I should keep the flags I just yanked out as part of the move to the new `argv` functions...lol.

src/protocols/vnc/argv.c
```diff
@@ -0,0 +1,59 @@
...
+        free(settings->username);
+        settings->username = strdup(value);
+        vnc_client->vnc_credential_flags &= GUAC_VNC_COND_FLAG_USERNAME;
+    }
+    
+    /* Update password */
+    else if (strcmp(name, GUAC_VNC_ARGV_PASSWORD) == 0) {
+        free(settings->password);
+        settings->password = strdup(value);
+        vnc_client->vnc_credential_flags &= GUAC_VNC_COND_FLAG_USERNAME;
```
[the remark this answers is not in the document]
necouchman: uh, yeah.

src/protocols/vnc/auth.c
necouchman: Yeah, I think that was half the point of implementing the `argv` common stuff...

necouchman:
> I think it might be worth considering the guac_argv_await(), guac_argv_register(), etc. family of functions that were recently added as part of GUACAMOLE-221 to make this sort of logic easier

Yeah, that makes perfect sense, seeing as it was one of the driving reasons for adding them.

Follow-up question, though - if we move to that, is there still any need for the `pthread_cond` and flags? I removed those as part of moving RDP and VNC to these new `argv` functions, but some of the comments you made above make it sounds like it's still worth keeping the flags around and checking that the parameters cannot be manipulated by someone with nefarious intentions?  Or does the move to the `argv` handlers also take care of this concern?

src/libguac/client.c
```diff
@@ -633,6 +669,44 @@ static void* __webp_support_callback(guac_user* user, void* data) {
...
+ * @return
+ *     Always NULL.
+ */
+static void* guac_owner_supports_required_callback(guac_user* user, void* data) {
+    
+    int* required_supported = (int *) data;
+    
+    /* Check if owner supports required */
+    if (*required_supported)
+        *required_supported = guac_user_supports_required(user);
```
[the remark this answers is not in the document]
necouchman: I think I've got this cleaned up a bit - let me know if it looks okay or you had something different in mind.

src/libguac/protocol.c
```diff
@@ -1241,3 +1302,34 @@ int guac_protocol_decode_base64(char* base64) {
...
+    }
+    
+    return GUAC_PROTOCOL_VERSION_UNKNOWN;
+    
+}
+
+const char* guac_protocol_version_to_string(guac_protocol_version version) {
+    
+    guac_protocol_version_mapping* current = guac_protocol_version_table;
+    while (current->version) {
```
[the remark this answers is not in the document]
necouchman: I believe I'm doing this correctly, now.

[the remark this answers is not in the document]
necouchman: Yeah, that would be good...

src/libguac/tests/protocol/guac_protocol_version.c
```diff
@@ -0,0 +1,69 @@
...
+ * Test which verifies that conversion of the guac_protocol_version enum to
+ * string values succeeds and produces the expected results.
+ */
+void test_guac_protocol__version_to_string() {
+    
+    guac_protocol_version version_a = GUAC_PROTOCOL_VERSION_1_3_0;
+    guac_protocol_version version_b = GUAC_PROTOCOL_VERSION_1_0_0;
+    guac_protocol_version version_c = GUAC_PROTOCOL_VERSION_UNKNOWN;
+    
+    CU_ASSERT_STRING_EQUAL_FATAL(guac_protocol_version_to_string(version_a), "VERSION_1_3_0");
```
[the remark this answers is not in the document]
necouchman: I'm just not all that familiar with the CUnit framework, so I went for broke.

mike-jumper:
> Follow-up question, though - if we move to that, is there still any need for the `pthread_cond` and flags? I removed those as part of moving RDP and VNC to these new `argv` functions, but some of the comments you made above make it sounds like it's still worth keeping the flags around and checking that the parameters cannot be manipulated by someone with nefarious intentions? Or does the move to the `argv` handlers also take care of this concern?

Selectively invoking `guac_argv_register()` and migrating to `guac_argv_handler` would take care of the concern, yep:

https://github.com/apache/guacamole-server/blob/382d72a26a04008aa936680deed00e3a31086b1e/src/libguac/guacamole/argv.h#L119-L125

necouchman:
> Selectively invoking guac_argv_register() and migrating to guac_argv_handler would take care of the concern, yep

Okay, I think I've got that taken care of, then...let me know what else needs to be fixed!

src/protocols/vnc/auth.c
```diff
@@ -19,27 +19,96 @@
...
+            guac_argv_register(GUAC_VNC_ARGV_PASSWORD, guac_vnc_argv_callback, NULL, 0);
+            params[i] = GUAC_VNC_ARGV_PASSWORD;
+            i++;
+        }
+        
+        /* If we have empty parameters, request them. */
+        if (i > 0) {
+            
+            /* Send required parameters to owner. */
+            guac_client_owner_send_required(gc, (const char**) params);
```
mike-jumper, line 103: `guac_client_owner_send_required()` requires a null-terminated array. This will be the case here if only one parameter is requested, but because the length of `params` is 2, it cannot be the case if two parameters are requested, and the function may overrun array bounds.

I think your `params` array needs to be one element longer to ensure there is room for the NULL terminator.

src/protocols/rdp/rdp.c
```diff
@@ -227,10 +234,60 @@ static BOOL rdp_freerdp_authenticate(freerdp* instance, char** username,
...
+        
+        /* Send required parameters to the owner. */
+        guac_client_owner_send_required(client, (const char**) params);
+        
+        guac_argv_await((const char**) params);
+        
+        /* Get new values from settings. */
+        *username = settings->username;
+        *password = settings->password;
+        *domain = settings->domain;
```
mike-jumper, line 286: There may already be pointers to allocated memory assigned to `*username`, `*password`, etc. which will orphaned through these assignments. The existing values should first be freed.

As FreeRDP will also automatically free the assigned strings, and we already free our copies of these strings stored within `settings`, these should be duplicated so that they aren't double-freed. This is how things are currently done within `guac_rdp_push_settings()`:

https://github.com/apache/guacamole-server/blob/382d72a26a04008aa936680deed00e3a31086b1e/src/protocols/rdp/settings.c#L1296-L1299

src/protocols/vnc/user.c
```diff
@@ -98,6 +99,9 @@ int guac_vnc_user_join_handler(guac_user* user, int argc, char** argv) {
         /* Inbound (client to server) clipboard transfer */
         if (!settings->disable_paste)
             user->clipboard_handler = guac_vnc_clipboard_handler;
+        
+        /* Updates to connection parameters */
+        user->argv_handler = guac_argv_handler;
```
mike-jumper, line 104: Same here - I believe this should be set only for the connection owner.

src/libguac/client.c
```diff
@@ -478,6 +478,41 @@ int guac_client_load_plugin(guac_client* client, const char* protocol) {
...
+ * 
+ * @return
+ *     The integer status of the operation, cast as void*.
+ */
+static void* guac_client_owner_send_required_callback(guac_user* user, void* data) {
+    
+    const char** required = (const char **) data;
+    
+    /* Send required parameters to owner. */
+    return (void*) ((intptr_t) guac_protocol_send_required(user->socket, required));
```
mike-jumper, line 502: Beware that `user` can be `NULL` if the owner has left the connection. See:

https://guacamole.apache.org/doc/1.1.0/libguac/client_8h.html#af3f4ed85d98b16376e2cdc031ff1b44a

src/libguac/client.c
```diff
@@ -633,6 +668,40 @@ static void* __webp_support_callback(guac_user* user, void* data) {
...
+ *     A pointer to an int containing the status for support of the "required"
+ *     instruction.  This will be 0 if the owner does not support this
+ *     instruction, or 1 if the owner does support it.
+ * 
+ * @return
+ *     Always NULL.
+ */
+static void* guac_owner_supports_required_callback(guac_user* user, void* data) {
+    
+    data = (void *) ((intptr_t) guac_user_supports_required(user));
```
mike-jumper, line 689: Same here - beware that `user` can be `NULL` if the owner has left the connection.

mike-jumper (changes requested):

src/libguac/client.c
necouchman: I think this occurrence is okay, as there is no de-reference of a pointer within this function, and the called function `guac_user_supports_required()` does check to make sure `user` is not `NULL`.

Or would you prefer to handle the case of `user` being `NULL` within this function, as well?

src/protocols/rdp/rdp.c
necouchman: As `guac_rdp_strdup()` is only defined within the context of `settings.c`, is using `strdup()` acceptable, here, or do we need to pull `guac_rdp_strdup()` up into a header file and use that?

src/libguac/client.c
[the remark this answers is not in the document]
necouchman: I believe I've handled this correctly.

mike-jumper: Ah, OK. If `guac_user_supports_required()` already handles this case, I have no issue with this.

src/protocols/rdp/rdp.c
mike-jumper: I think `guac_rdp_strdup()` is needed here. The standard `strdup()` will segfault if given `NULL`, whereas `guac_rdp_strdup()` will simply and safely return `NULL`.

The definition for `guac_rdp_strdup()` could be moved to `settings.h`. Or we could consider adding a `guac_strdup()` to libguac's `string.h`.

necouchman: Went ahead and moved it to libguac's `string.h` and rename it `guac_strdup()`.

src/protocols/rdp/rdp.c
```diff
@@ -227,10 +236,69 @@ static BOOL rdp_freerdp_authenticate(freerdp* instance, char** username,
...
+        params[i] = GUAC_RDP_ARGV_PASSWORD;
+        i++;
+    }
+    
+    /* If the domain is undefined, add it to the requested parameters. */
+    if (settings->domain == NULL) {
+        guac_argv_register(GUAC_RDP_ARGV_DOMAIN, guac_rdp_argv_callback, NULL, 0);
+        params[i] = GUAC_RDP_ARGV_DOMAIN;
+        i++;
+    }
```
mike-jumper, line 273: If an administrator provides:

| Parameter | Value |
| --------- | ----- |
| `domain` | (blank) |
| `username` | `foo` |
| `password` | (blank) |

I think the domain is effectively specified - the administrator intends that the user log in to the default domain / local machine, and that the user should be prompted for their password. I doubt that the administrator intends that the user be prompted for their domain and password, with only the username set in stone.

Perhaps we should prompt for the domain _only_ if the username is also blank?

src/libguac/string.c
```diff
@@ -81,6 +81,17 @@ size_t guac_strlcat(char* restrict dest, const char* restrict src, size_t n) {
 
 }
 
+char* guac_strdup(const char* str) {
```
mike-jumper, line 84: Looks good, but we should add a unit test for this.

mike-jumper (changes requested):

src/protocols/rdp/rdp.c
necouchman: Yeah, I'm good with that.

src/protocols/vnc/auth.c
```diff
@@ -19,27 +19,100 @@
...
+                guac_argv_await((const char**) params);
+
+                return creds;
+
+            }
+        }
+        
+        /* Copy the values and return the credential set. */
+        creds->userCredential.username = strdup(settings->username);
+        creds->userCredential.password = strdup(settings->password);
```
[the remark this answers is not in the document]
necouchman: Okay, so a few follow-up questions:

1) Do you know if there's a particular reason why the VNC settings use an empty string instead of `NULL` to indicate that a value has not been provided? I'm not familiar enough with the internals of VNC to know if this is required by libvncclient or if the VNC servers expect empty strings rather than no value at all? It looks like most (though certainly not all) of the other places throughout the settings code among the various protocols use `NULL`.  I'm guessing there's history and reason?
2) The `settings->username` and `settings->password` value are freed by `libvncclient`, but what about `creds`? I suppose that `settings` and all of its members are going to be around as long as the connection is active, so there's not really any risk that `settings->username` will get freed prior to being used as part of `creds`, but what about the other way around.  If I remember my C memory handling correctly, doing the above (`creds->userCredential.username = settings->username`) just points `creds->userCredential.username` at the same spot in memory where `settings->username` is pointed - what if `libvncclient` frees `creds` and its members before the connection closes? Presumably that would have adverse impacts on the continued ability for the client to read `settings->username`? Or am I totally off in the weeds, here?
3) If `strdup` is used to copy these values, presumably `creds` would also need to be cleaned up at some point - either within `libvncclient` or manually, correct?

src/libguac/string.c
necouchman: Okay, wrote a couple of basic tests - not sure if there are other cases I should test in there??

src/protocols/vnc/auth.c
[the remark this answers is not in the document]
mike-jumper: Briefly checking the libvncclient source, it looks like it always checks for `NULL` for both the username and password before taking any action, and so it should be safe for us to migrate to using `NULL` instead of `""`.

Good point on the `rfbCredential` that we allocate here. I had not checked this before, but it is indeed freed by libvncclient, just after libvncclient frees the username and password, as part of `FreeUserCredential()` - an internal function invoked by the parts of libvncclient that handle auth after they are done handling auth.

If there is concern that we may need to continue referencing the username/password ourselves, then yes, I agree they should be strdup'd and we should manually free our copies when disconnect happens. It looks like libvncclient frees the memory associated with credentials immediately after use, so any attempt to use those vales after they have already been used by libvncclient will segfault (or worse).

src/protocols/rdp/rdp.c
```diff
@@ -227,10 +236,70 @@ static BOOL rdp_freerdp_authenticate(freerdp* instance, char** username,
...
+        
+        /* Free old values and get new values from settings. */
+        if (*username != NULL)
+            free(*username);
+        
+        if (*password != NULL)
+            free(*password);
+        
+        if (*domain != NULL)
+            free(*domain);
```
mike-jumper, line 294: By the way, there's no need to check a pointer against `NULL` before invoking `free()`. If the pointer passed to `free()` is `NULL`, no operation is performed.

src/libguac/string.c
mike-jumper: The tests you've added look good to me.

src/protocols/rdp/rdp.c
necouchman: Okay - I couldn't remember which scenarios you had to worry about the double free corruption issues, so I checked it.  I can remove the checks.

src/protocols/vnc/auth.c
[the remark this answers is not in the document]
mike-jumper: With the switch to using `NULL`, I think this is likely fine. Just need to now:

* Use the `NULL`-friendly `guac_strdup()`.
* Free `settings->username` and `settings->password` within `guac_vnc_settings_free()`.

src/protocols/vnc/settings.c
```diff
@@ -392,11 +393,11 @@ guac_vnc_settings* guac_vnc_parse_args(guac_user* user,
 
     settings->username =
         guac_user_parse_args_string(user, GUAC_VNC_CLIENT_ARGS, argv,
-                IDX_USERNAME, ""); /* NOTE: freed by libvncclient */
+                IDX_USERNAME, NULL); /* NOTE: freed by libvncclient */
     
     settings->password =
         guac_user_parse_args_string(user, GUAC_VNC_CLIENT_ARGS, argv,
-                IDX_PASSWORD, ""); /* NOTE: freed by libvncclient */
+                IDX_PASSWORD, NULL); /* NOTE: freed by libvncclient */
```
mike-jumper, line 400: I believe these notes regarding these strings being freed by libvncclient are no longer correct.

src/protocols/vnc/auth.c
```diff
@@ -19,27 +19,101 @@
...
+        
+        /* Send the request for password to the owner. */
+        guac_client_owner_send_required(gc, params);
+                
+        /* Wait for the arguments to be returned */
+        guac_argv_await(params);
+
+    }
+    
+    return settings->password;
```
mike-jumper, line 62: As this will also be freed by libvncclient, we'll need a `guac_strdup()` here just as with `guac_vnc_get_credentials()`.

mike-jumper (changes requested):
Almost ... there ...

necouchman:
Yeah, it's a bit of a process with me...lol.

necouchman:
Okay, did another swipe through the code and fixed up a handful of things.

mike-jumper (approved):
:trophy:

PR #229, opened 2019-07-20
GUACAMOLE-847: Fix severe memory leak when using audio with RDP

fhriley:
This fixes a memory leak that leaks 4+ GBs over an 8 hour period when continuous audio is being played through RDP.


fhriley:
Done, I believe the style matches other code now.

There are 2 confirmations that this is the correct fix:

1. I ran it overnight with no issues and no memory leaks.
2. [The equivalent call in FreeRDP's rdpsnd does a free](https://github.com/FreeRDP/FreeRDP/blob/stable-1.1/channels/rdpsnd/client/rdpsnd_main.c#L515).

PR #232, opened 2019-08-03
GUACAMOLE-514: Implement additional VNC authentication support

necouchman:
This pull request implements the callback and parameters required to support additional authentication mechanisms by VNC servers that don't do just password-based authentication.  Ideally this should implement both username/password support as well as X509 support.  I have tested the username/password support and it appears to work - I don't have a way of testing X509 right now so I'm not 100% certain that it will work; however, since I think most of that is handled internally in libvncserver(client), it's reasonable that it will work.

### DIFF
--- a/src/protocols/vnc/auth.c
+++ b/src/protocols/vnc/auth.c
@@ -41,9 +41,12 @@ rfbCredential* guac_vnc_get_credentials(rfbClient* client, int credentialType) {
         creds->userCredential.password = settings->password;
         return creds;
     }
-    
-    guac_client_log(gc, GUAC_LOG_ERROR,
+
+    guac_client_abort(gc, GUAC_PROTOCOL_STATUS_SERVER_ERROR,
             "Unsupported credential type requested.");
+    guac_client_log(gc, GUAC_LOG_DEBUG,
+            "Unable to provided requested credential %d.",
+            credentialType);
     return NULL;
     
 }
--- a/src/protocols/vnc/auth.c
+++ b/src/protocols/vnc/auth.c
@@ -34,18 +34,49 @@ char* guac_vnc_get_password(rfbClient* client) {
 rfbCredential* guac_vnc_get_credentials(rfbClient* client, int credentialType) {
     guac_client* gc = rfbClientGetClientData(client, GUAC_VNC_CLIENT_KEY);
     rfbCredential *creds = malloc(sizeof(rfbCredential));
+    guac_vnc_settings* settings = ((guac_vnc_client*) gc->data)->settings;
     
     if (credentialType == rfbCredentialTypeUser) {
-        creds->userCredential.username = ((guac_vnc_client*) gc->data)->settings->username;
-        creds->userCredential.password = ((guac_vnc_client*) gc->data)->settings->password;
+        creds->userCredential.username = settings->username;
+        creds->userCredential.password = settings->password;
         return creds;
     }
     
     else if (credentialType == rfbCredentialTypeX509) {
-        creds->x509Credential.x509ClientCertFile = ((guac_vnc_client*) gc->data)->settings->client_cert;
-        creds->x509Credential.x509ClientKeyFile = ((guac_vnc_client*) gc->data)->settings->client_key;
-        creds->x509Credential.x509CACertFile = ((guac_vnc_client*) gc->data)->settings->ca_cert;
-        creds->x509Credential.x509CACrlFile = ((guac_vnc_client*) gc->data)->settings->ca_crl;        
+        char* template = "guac_XXXXXX";
+        
+        if (settings->client_cert != NULL) {
+            settings->client_cert_temp = strdup(template);
+            int cert_fd = mkstemp(settings->client_cert_temp);
+            write(cert_fd, settings->client_cert, strlen(settings->client_cert));
+            close(cert_fd);
+            creds->x509Credential.x509ClientCertFile = settings->client_cert_temp;
+        }
+        
+        if (settings->client_key != NULL) {
+            settings->client_key_temp = strdup(template);
+            int key_fd = mkstemp(settings->client_key_temp);
+            write(key_fd, settings->client_key, strlen(settings->client_key));
+            close(key_fd);
+            creds->x509Credential.x509ClientKeyFile = settings->client_key_temp;
+        }
+        
+        if (settings->ca_cert != NULL) {
+            settings->ca_cert_temp = strdup(template);
+            int ca_fd = mkstemp(settings->ca_cert_temp);
+            write(ca_fd, settings->ca_cert, strlen(settings->ca_cert));
+            close(ca_fd);
+            creds->x509Credential.x509CACertFile = settings->ca_cert_temp;
+        }
+        
+        if (settings->ca_crl != NULL) {
+            settings->ca_crl_temp = strdup(template);
+            int crl_fd = mkstemp(settings->ca_crl_temp);
+            write(crl_fd, settings->ca_crl, strlen(settings->ca_crl));
+            close(crl_fd);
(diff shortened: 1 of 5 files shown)


src/protocols/vnc/auth.c
```diff
@@ -31,3 +31,25 @@ char* guac_vnc_get_password(rfbClient* client) {
...
+    rfbCredential *creds = malloc(sizeof(rfbCredential));
+    
+    if (credentialType == rfbCredentialTypeUser) {
+        creds->userCredential.username = ((guac_vnc_client*) gc->data)->settings->username;
+        creds->userCredential.password = ((guac_vnc_client*) gc->data)->settings->password;
+        return creds;
+    }
+    
+    else if (credentialType == rfbCredentialTypeX509) {
+        creds->x509Credential.x509ClientCertFile = ((guac_vnc_client*) gc->data)->settings->client_cert;
```
mike-jumper: Is there no way to read these keys from memory? It feels odd to have to have the certificates live within files on the guacd server, with these parameters being paths rather than PEM keys.

necouchman: I'm not sure.  The way I've implemented it, between the Client and Server parameters, is assuming that the contents of the file will be passed through, not the location of the file, but I'm not sure one way or the other.

Unfortunately, if those values represent the locations of a file, rather than the contents, I'm not sure there's much to do about it within the libVNCclient implementation, since that's internal to that implementation.

I can think of one way to work around it - we could have guacd write out the contents of the file to a location on-disk, have libvncclient read it in, then immediately delete the file.  This is a little risky, as it has the potential to expose the contents of the private key to locations where it could be read by others on the system.

I'm open to suggestions, even if it's "let's not implement this right now."

necouchman: Well, after digging into the source code, it does indeed look like libvncclient expects this to be the path to the file, not the contents of the file.  Which stinks.

https://github.com/LibVNC/libvncserver/blob/f997b5a75fa171d79c5e568b7157fba83c8d8355/libvncclient/tls_gnutls.c#L416-L425

Which calls `gnutls_certificate_set_x509_key_file()`, which takes file locations as an argument.

Aside from writing out a temporary file, are there any creative ways to point a function like `gnutls_certificate_set_x509_key_file()` to read a location in memory rather than a filesystem location?

necouchman: Well, I took a stab at writing the output to temp files.  Feels like the block of code that would be next to the word "kludge" in a programming dictionary...

mike-jumper: Yeah, I agree that writing credentials to temporary files on disk is not the way to go.

Perhaps the username half of this should be split away from the X.509 half? That would narrow scope to something that's definitely solvable. Supporting X.509 might need to wait for improvements to the underlying library.

src/protocols/vnc/auth.c
```diff
@@ -31,3 +31,56 @@ char* guac_vnc_get_password(rfbClient* client) {
...
+            write(crl_fd, settings->ca_crl, strlen(settings->ca_crl));
+            close(crl_fd);
+            creds->x509Credential.x509CACrlFile = settings->ca_crl_temp;
+        }
+        
+        return creds;
+    }
+    
+    guac_client_log(gc, GUAC_LOG_ERROR, "Unknown credential type requested.");
+    return NULL;
```
mike-jumper: This will leak the memory allocated for `creds` if an unknown credential type is received.

src/protocols/vnc/auth.c
necouchman: Yeah, I'm fine with not doing the X509 bits for now.  I'll go through and just pull those out, and we'll focus this on the username + password portion.

src/protocols/vnc/auth.c
necouchman: Fixed.

necouchman:
@mike-jumper: Okay, x509 bits removed, should be just username + password, now.

src/protocols/vnc/auth.c
```diff
@@ -31,3 +31,19 @@ char* guac_vnc_get_password(rfbClient* client) {
...
+    
+    if (credentialType == rfbCredentialTypeUser) {
+        rfbCredential *creds = malloc(sizeof(rfbCredential));
+        creds->userCredential.username = settings->username;
+        creds->userCredential.password = settings->password;
+        return creds;
+    }
+    
+    guac_client_log(gc, GUAC_LOG_ERROR,
+            "Unsupported credential type requested.");
```
mike-jumper: I haven't looked into the behavior of libvncclient with respect to this callback. Is a request for an unsupported credential type an error? Or is it something which is expected to occur when a VNC server supports multiple authentication mechanisms, not all of which are necessarily supported by the client?

src/protocols/vnc/settings.c
```diff
@@ -28,13 +28,15 @@
 #include <stdlib.h>
 #include <string.h>
 #include <time.h>
+#include <unistd.h>
```
mike-jumper: What brings this in?

src/protocols/vnc/auth.c
necouchman: I believe for the callback function it is an error, because this will be called when the client detects the need for additional credentials, and, if the requested credentials are not provided, this constitutes an error that the client cannot handle - that is, the client cannot fulfill the required credentials.

Does that make sense?

src/protocols/vnc/settings.c
necouchman: That probably was for one of the x509 bits - let me pull it out and make sure it isn't required anymore.

src/protocols/vnc/auth.c
mike-jumper: Sure, makes perfect sense.

In that case, I would recommend:

1. Using `guac_client_abort()`, so this is explicitly fatal to the connection.
2. Additionally logging the credential type that's unsupported, at least at the debug level.

necouchman: Okay, I think I did this correctly.  There doesn't seem to be an easy way to go from the `int` type requested credential back to some sort of human-readable name, so, while that's less-than-ideal, hopefully that's okay...

necouchman:
@mike-jumper Anything else, here, or is this good to go, now?

src/protocols/vnc/auth.c
```diff
@@ -31,3 +31,22 @@ char* guac_vnc_get_password(rfbClient* client) {
...
+        rfbCredential *creds = malloc(sizeof(rfbCredential));
+        creds->userCredential.username = settings->username;
+        creds->userCredential.password = settings->password;
+        return creds;
+    }
+
+    guac_client_abort(gc, GUAC_PROTOCOL_STATUS_SERVER_ERROR,
+            "Unsupported credential type requested.");
+    guac_client_log(gc, GUAC_LOG_DEBUG,
+            "Unable to provided requested credential %d.",
```
mike-jumper: "Unable to provided" ;)

It would be better IMHO to refer to the `%d` as a credential type than a credential, but the only dealbreaker from my perspective is the funky wording.

necouchman: I can do good English.  LoL.  Should be corrected, and added "type of" in there.

necouchman:
@mike-jumper: Okay, weird English has been untangled.

limenghui614:
Are the new supported validations account authentication for Unix/Linux systems or Windows system authentication?

necouchman:
@limenghui614: It should not be platform-specific, so any VNC server that sends back a challenge for username and password should accept the credentials.  However, this may depend on how the VNC server sends that challenge - the changes simply allow Guacamole to pull both a username and password for authentication if the VNC server so requests.  If the VNC server is requesting something additional - like an AD domain, or a different handshake, it may not work.  YMMV.

limenghui614:
Thank you very much. Your answer was very helpful

PR #237, opened 2019-09-07
GUACAMOLE-871: Add support for hidden cursor (DECTECM)

mawww:
As requested on JIRA, here is the patch as a pull request. Please not that I did not test this as I dont know how to put all the pieces in place to check that. It can be easily tested with:

```
printf '\e[?25l' # cursor should be hidden after this
printf '\e[?25h' # cursor should be visible again after this
```

### DIFF
--- a/src/terminal/terminal.c
+++ b/src/terminal/terminal.c
@@ -202,6 +202,7 @@ void guac_terminal_reset(guac_terminal* term) {
     /* Reset cursor location */
     term->cursor_row = term->visible_cursor_row = term->saved_cursor_row = 0;
     term->cursor_col = term->visible_cursor_col = term->saved_cursor_col = 0;
+    term->cursor_visible = true;
 
     /* Clear scrollback, buffer, and scroll region */
     term->buffer->top = 0;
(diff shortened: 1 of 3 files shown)


src/terminal/terminal.c
```diff
@@ -777,31 +778,43 @@ void guac_terminal_commit_cursor(guac_terminal* term) {
...
+    if (term->visible_cursor_row != -1 && term->visible_cursor_col != -1)
+    {
+        /* Get old row with cursor */
+        row = guac_terminal_buffer_get_row(term->buffer, term->visible_cursor_row, term->visible_cursor_col+1);
+
+        guac_char = &(row->characters[term->visible_cursor_col]);
+        guac_char->attributes.cursor = false;
+        guac_terminal_display_set_columns(term->display, term->visible_cursor_row + term->scroll_offset,
+                term->visible_cursor_col, term->visible_cursor_col, guac_char);
+    }
```
mike-jumper: Here and elsewhere, please use the same style as the surrounding code. Where braces are used for `if`, `else`, etc., we have the opening brace on the same line. Instead of:

```c
if (foo)
{
}
else
{
}
```

You'll find:

```c
if (foo) {
}
else {
}
```

src/terminal/terminal.c
```diff
@@ -1240,7 +1253,8 @@ static void __guac_terminal_resize(guac_terminal* term, int width, int height) {
             /* Update buffer top and cursor row based on shift */
             term->buffer->top += shift_amount;
             term->cursor_row  -= shift_amount;
-            term->visible_cursor_row  -= shift_amount;
+            if (term->visible_cursor_row != -1)
```
mike-jumper: Any reason a sentinel value of -1 is being used vs. the `cursor_visible` flag?

If -1 is going to be a value assigned to `visible_cursor_row` and `visible_cursor_col` to indicate that the cursor is not rendered, that should be documented where those properties are defined, however I'm not sure of the utility of a -1 constant when there is a boolean flag also in use and with clear semantics.

If the -1 sentinel is not needed after all, and `cursor_visible` is preferred, it would be good to document that the values of the `visible_cursor_row`, etc. properties have no meaning if `cursor_visible` is `true`.

src/terminal/terminal.c
```diff
@@ -777,31 +778,43 @@ void guac_terminal_commit_cursor(guac_terminal* term) {
 
     guac_terminal_char* guac_char;
 
-    guac_terminal_buffer_row* old_row;
-    guac_terminal_buffer_row* new_row;
+    guac_terminal_buffer_row* row;
 
     /* If no change, done */
     if (term->visible_cursor_row == term->cursor_row && term->visible_cursor_col == term->cursor_col)
```
mike-jumper: Will this potentially result in changes to cursor visibility not taking effect until cursor is moved?

src/terminal/terminal.c
mawww: While most of the time `cursor_visible == false` is equivalent to `visible_cursor_... == -1`, it is not always the case. In particular when we just switched visibility back, we can have `cursor_visible = true` and `visible_cursor_...` still at -1. In that case this condition would change `visibile_curor_row` from `-1` to `-1 - shift_amount` which would be semantically incorrect.

Similarly, on line 788 if we used `cursor_visible` we could trigger a buffer underflow by trying to access row/column -1.

Documenting those properties makes sense.

An alternative would be to introduce an additional `bool visible_cursor_visible` that would track the effective visibility of the cursor (while `cursor_visible` tracks the requested visibility), the proposed naming here is consistent with `visible_cursor_...` tracking the effective state of `cursor_...`.

src/terminal/terminal.c
mawww: I think you are right, we would also need to check `term->cursor_visible == (term->visible_cursor_col == -1)` or `term->cursor_visible == term->visible_cursor_visible` (see the other comment).

src/terminal/terminal.c
mawww: Will do.

mawww:
Added more comments, fixed style issues, and fixed the potential bug where hiding a cursor would only take effect after moving it.

src/terminal/terminal.c
```diff
@@ -777,31 +778,42 @@ void guac_terminal_commit_cursor(guac_terminal* term) {
 
     guac_terminal_char* guac_char;
 
-    guac_terminal_buffer_row* old_row;
-    guac_terminal_buffer_row* new_row;
+    guac_terminal_buffer_row* row;
 
     /* If no change, done */
-    if (term->visible_cursor_row == term->cursor_row && term->visible_cursor_col == term->cursor_col)
+    if (term->cursor_visible && term->visible_cursor_row == term->cursor_row && term->visible_cursor_col == term->cursor_col)
```
mike-jumper: Will this result in the cursor being repeatedly cleared despite already being invisible?

mawww: If `cursor_visible == false` we will clear only if `visible_cursor_col != -1 && visible_cursor_row != -1`, then we will unconditionally set those to `-1` which might be redundent but should not be an issue (if we are counting CPU cycles, I dont expect setting those to -1 vs checking if they already are to early out to be noticeably different).

mike-jumper: Don't worry - no counting of CPU cycles is happening in this case. Orders of magnitude of cycles, sure, but this isn't a performance-critical region of code where individual cycles would have huge multipliers.

PR #238, opened 2019-09-19
GUACAMOLE-625: Included Latin American keymap

dkmstr:
Included keymap file and added it to build process

### DIFF
--- a/src/protocols/rdp/Makefile.am
+++ b/src/protocols/rdp/Makefile.am
@@ -268,7 +268,7 @@ rdp_keymaps =                   \
     $(srcdir)/keymaps/en_gb_qwerty.keymap \
     $(srcdir)/keymaps/en_us_qwerty.keymap \
     $(srcdir)/keymaps/es_es_qwerty.keymap \
-    $(srcdir)/keymaps/es_la_qwerty.keymap \
+    $(srcdir)/keymaps/es_latam_qwerty.keymap \
     $(srcdir)/keymaps/fr_be_azerty.keymap \
     $(srcdir)/keymaps/fr_ch_qwertz.keymap \
     $(srcdir)/keymaps/fr_fr_azerty.keymap \
--- a/src/protocols/rdp/Makefile.am
+++ b/src/protocols/rdp/Makefile.am
@@ -268,6 +268,7 @@ rdp_keymaps =                   \
     $(srcdir)/keymaps/en_gb_qwerty.keymap \
     $(srcdir)/keymaps/en_us_qwerty.keymap \
     $(srcdir)/keymaps/es_es_qwerty.keymap \
+    $(srcdir)/keymaps/latam_qwerty.keymap \
     $(srcdir)/keymaps/fr_fr_azerty.keymap \
     $(srcdir)/keymaps/fr_ch_qwertz.keymap \
     $(srcdir)/keymaps/hu_hu_qwertz.keymap \
--- a/src/protocols/rdp/keymaps/es_latam_qwerty.keymap
+++ b/src/protocols/rdp/keymaps/es_latam_qwerty.keymap
@@ -18,7 +18,7 @@
 #
 
 parent  "base"
-name    "es-la-qwerty"
+name    "es-latam-qwerty"
 freerdp "KBD_LATIN_AMERICAN"
 
 #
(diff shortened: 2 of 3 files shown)


src/protocols/rdp/Makefile.am
```diff
@@ -268,6 +268,7 @@ rdp_keymaps =                   \
     $(srcdir)/keymaps/en_gb_qwerty.keymap \
     $(srcdir)/keymaps/en_us_qwerty.keymap \
     $(srcdir)/keymaps/es_es_qwerty.keymap \
+    $(srcdir)/keymaps/latam_qwerty.keymap \
```
necouchman: A couple of things, here:
- Is "latam" actually the proper identifier for this keyboard layout?  I would suspect something following existing patterns - like "es_la_qwerty" or the like?  I don't know if there's an official standard for this or just what we're calling it in Guacamole, but it should be consistent.
- It should also be organized alphabetically - so if "es_la_qwerty" is the correct name for the layout, this is the correct spot - if "latam_qwerty" is correct, then it should be down a little further in the file.

dkmstr: https://askubuntu.com/questions/8797/what-is-the-two-character-language-code-for-the-latin-american-keyboard-layout

necouchman: Thanks.  I'm good with this as long as @mike-jumper or anyone else doesn't have any objections?

necouchman:
Okay, @dkmstr, looks fine to me as long as I can get some sort of general consensus that we're okay with that keyboard layout name.  Once that's confirmed I'll be happy to merge it.

src/protocols/rdp/Makefile.am
mike-jumper: I agree that we should prefer to remain consistent with the existing convention (`[language code]-[country code]-[variant]`) or, if that isn't working any longer in practice, investigate whether we should be using a different convention.

Following current convention, if "latam" is the proper identifier for all countries associated with this keyboard layout, it sounds like the layout identifier that follows our convention would be `es-latam-qwerty`.

**BUT**, from the [forum thread linked above](https://askubuntu.com/questions/8797/what-is-the-two-character-language-code-for-the-latin-american-keyboard-layout), we may want to look into a better system:

> ... That aside, I would recommend against using language codes to indicate keyboard layouts. There is a separate system already in place for abbreviating the names of the layouts themselves. See `shortDescription` in [xkb-data](http://packages.ubuntu.com/maverick/xkb-data)'s `base.xml`.

necouchman:
@dkmstr: Based on another commit that went in for another keyboard layout, this will need to be updated.

@mike-jumper: <Nudge> Do we want to move forward with this keyboard layout name, for now, or use the opportunity to define a new standard??

mike-jumper:
> Do we want to move forward with this keyboard layout name, for now, or use the opportunity to define a new standard??

I think it's best to move forward with these changes, either with the `es-` prefix or without (depending on what is truly correct).

Though the contributed layout has resulted in our established naming convention being called into question, it's still solid within the scope of the contribution itself (a keyboard layout) and can be made to comply with the current naming convention if it doesn't already comply. A new standard can be sorted out independently.

necouchman:
@mike-jumper Thanks for the input.  I'll proceed with the merging of this after @dkmstr has resolved the merge conflict.  Based on past discussion, I believe the way it is currently implemented - without the "es" prefix - is correct for this keyboard layout.

dkmstr:
Finally, based on suggestions, i renamed to keep currently naming convetion to "es_la_qwerty", fixed files & conflicts.

mike-jumper:
Unfortunately, "la" is the two-letter code for Lao People's Democratic Republic, not Latin America (see: https://en.m.wikipedia.org/wiki/ISO_3166-1_alpha-2#LA).

necouchman:
@dkmstr: Also, your last commit doesn't have the JIRA issue prepended to the commit message.

dkmstr:
Sorry, didn't even remember about it.
It's right now?

dkmstr:
did my commit solved the issue?

mike-jumper:
@dkmstr - not quite. It looks like there's still a commit in the middle which is missing the issue number:

guacamole-server-pr-238

necouchman:
@dkmstr opened #253 to cover this, closing this one.

PR #241, opened 2019-11-14
GUACAMOLE-901: Belgian French keymap

jukeks:


src/protocols/rdp/Makefile.am
```diff
@@ -269,6 +269,7 @@ rdp_keymaps =                   \
     $(srcdir)/keymaps/en_us_qwerty.keymap \
     $(srcdir)/keymaps/es_es_qwerty.keymap \
     $(srcdir)/keymaps/fr_fr_azerty.keymap \
+    $(srcdir)/keymaps/fr_be_azerty.keymap \
```
necouchman: Nitpick - this should probably be *above* the `fr_fr` keymap, as should the `fr_ch` one.

jukeks: They are now alphabetically sorted.

jukeks:
Elaborated the commit message a bit.

necouchman:
@jukeks: Thank you, again, for the contribution.  Please note that the change will also need to be made on the guacamole-client side, as well, to add it as an option for use when configuring connections.

PR #242, opened 2019-12-26
GUACAMOLE-474: Allow file uploads and downloads to be disabled

necouchman:
This change adds options for the protocols that support various file transfer mechanisms to disable either uploads or downloads (or both) if so desired.

I took the approach of creating new boolean options for this, which seems the simplest approach; however, I did think of one possible alternative approach which would be to change the current "Enable Drive" (RDP) and "Enable SFTP" (RDP, SSH, and VNC) options such that, instead of boolean (checkbox), they are multi-option (select) with enable, upload-only, download-only, or disabled.  Let me know if that seems a better approach.

### DIFF
--- a/src/common-ssh/sftp.c
+++ b/src/common-ssh/sftp.c
@@ -790,7 +790,8 @@ static int guac_common_ssh_sftp_get_handler(guac_user* user,
         /* If downloads are disabled, log and return. */
         if (filesystem->disable_download) {
             guac_user_log(user, GUAC_LOG_INFO,
-                    "File downloads have been disabled.");
+                    "Unable to download file \"%s\", "
+                    "file downloads have been disabled.", fullpath);
             return 0;
         }
         
--- a/src/common-ssh/sftp.c
+++ b/src/common-ssh/sftp.c
@@ -429,7 +429,7 @@ int guac_common_ssh_sftp_handle_file_stream(
 
 /**
  * Handler for ack messages which continue an outbound SFTP data transfer
- * (download), signalling the current status and requesting additional data.
+ * (download), signaling the current status and requesting additional data.
  * The data associated with the given stream is expected to be a pointer to an
  * open LIBSSH2_SFTP_HANDLE for the file from which the data is to be read.
  *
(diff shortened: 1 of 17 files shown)


ibuler:
Can dispatch a event when upload or download ?

mike-jumper:
Looks like this will need some rebasing due to #243.

necouchman:
@mike-jumper: Rebase completed.

src/common-ssh/sftp.c
```diff
@@ -787,6 +787,13 @@ static int guac_common_ssh_sftp_get_handler(guac_user* user,
     /* Otherwise, send file contents */
     else {
 
+        /* If downloads are disabled, log and return. */
+        if (filesystem->disable_download) {
+            guac_user_log(user, GUAC_LOG_INFO,
+                    "File downloads have been disabled.");
```
mike-jumper, line 793: In line with the other download failure messages (`Unable to read file "..."`, `Unable to read directory "...": (reason)`), I suggest phrasing this message as a failed/refused operation and including the name of the file.

The current message (`File downloads have been disabled.`) is not immediately clear that it indicates that a request isn't being serviced, and could be interpreted as a simple acknowledgement of the download setting.

src/protocols/rdp/channels/rdpdr/rdpdr-fs-messages.c
```diff
@@ -98,9 +98,13 @@ void guac_rdpdr_fs_process_create(guac_rdp_common_svc* svc,
         /* Create \Download if it doesn't exist */
         file = guac_rdp_fs_get_file((guac_rdp_fs*) device->data, file_id);
         if (file != NULL && strcmp(file->absolute_path, "\\") == 0) {
-            int download_id =
-                guac_rdp_fs_open((guac_rdp_fs*) device->data, "\\Download",
-                    GENERIC_READ, 0, FILE_OPEN_IF, FILE_DIRECTORY_FILE);
+            
+            /* Only create Download folder if downloads are enabled. */
+            if(!((guac_rdp_fs*) devices->data)->disable_download) {
```
mike-jumper, line 103: Whoops - need that space after `if`.

src/protocols/rdp/fs.c
```diff
@@ -77,11 +79,15 @@ void guac_rdp_fs_free(guac_rdp_fs* fs) {
...
 guac_object* guac_rdp_fs_alloc_object(guac_rdp_fs* fs, guac_user* user) {
-
+    
     /* Init filesystem */
     guac_object* fs_object = guac_user_alloc_object(user);
     fs_object->get_handler = guac_rdp_download_get_handler;
-    fs_object->put_handler = guac_rdp_upload_put_handler;
+    
+    /* Assign handler only if uploads are not disabled. */
+    if(!fs->disable_upload)
```
mike-jumper, line 88: `if(` -> `if (`

;)

src/protocols/rdp/channels/rdpdr/rdpdr-fs-messages.c
necouchman: Agh!  I'm slipping into old habits...

PR #243, opened 2020-01-04
GUACAMOLE-249: Migrate to FreeRDP 2.x

mike-jumper:
This change refactors the entirety of Guacamole's RDP support to use FreeRDP 2.x, removing any remaining compatibility with older FreeRDP releases. These changes also:

* Reorganize the source tree in the hopes of making things more maintainable going forward
* Add missing documentation
* Add support for the new "extended NLA" mode added by FreeRDP 2.x (I'm not entirely sure what this is, however) and switch over to "any" as the default security method (rather than "rdp").
* Remove most plugins, replacing guacsvc/guacsnd/guacdr with a single common SVC plugin. The RDP support leverages that common plugin to provide a single API which the arbitrary SVC, RDPSND, and RDPDR support then use without having to make plugins for each.

There are going to be conflicts with master for the following non-1.1.0 changes:

* [GUACAMOLE-296](https://issues.apache.org/jira/browse/GUACAMOLE-296) - Migrating to FreeRDP 2.x removes the need for this, as libwinpr is now automatically linked using pkgconfig.
* [GUACAMOLE-381](https://issues.apache.org/jira/browse/GUACAMOLE-381) - Support for clipboard copy/paste disable flags within CLIPRDR need to be manually re-added due to the number of changes made to CLIPRDR (see below). Much of this applies cleanly without conflict.
* [GUACAMOLE-764](https://issues.apache.org/jira/browse/GUACAMOLE-764) - Most applied cleanly without conflict, however one `int` -> `uint64_t` change has to be reapplied manually (see below).
* [GUACAMOLE-847](https://issues.apache.org/jira/browse/GUACAMOLE-847) - The RDPSND memory leak is now inherently resolved via the common SVC plugin and changes to channel memory handling in FreeRDP 2.x.

To hopefully ease this, I've performed a test merge and manually resolved the above conflicts. I recommend resolving conflicts such that things match: https://github.com/mike-jumper/guacamole-server/tree/freerdp-migrate-2.x-resolve-merge-master

... Good luck.


mike-jumper:
Forgot to mention: I've also leveraged FreeRDP 2.x's support for overriding the default logging to send all FreeRDP log messages to the debug level of Guacamole's logger, so there shouldn't be as much log spam, nor any need to run guacd in foreground to see debug messages from FreeRDP.

mike-jumper:
After some self review, I've added a few more commits:

* The color conversion function behaved incorrectly due to `FreeRDPConvertColor()` expecting the input color to be coming from `ReadColor()`. `ReadColor()` converts a color which is independent of native endianness to an intermediate format which is dependent on native endianness, and `FreeRDPConvertColor()` will only work correctly with that intermediate format. This was causing an issue where black text was occasionally rendering as bright blue, with the bright blue coming from the 0xFF value for the alpha channel being stuck in the wrong place. (See commit a54a123)
* I had previously modified the common surface implementation to avoid automatically flushing the surface when a rectangle is drawn (typically used for initializing the surface) when that surface hasn't yet been sent to the client (common for RDP, where we sometimes receive instructions to create an off-screen surface and then destroy it without ever actually using it). The logic wasn't entirely correct, and accidentally removed some consideration of the dirty flag when determining whether to combine updates. (See commit 0677a9a)
* I ran the Include What You Use tool before opening this PR to try to clean up the headers included over the ages, as many no longer need to be included in those locations due to changes made. That tool added empty `struct` forward declarations which were unnecessary. I have removed these. (See commit 1f24c47)

I didn't redo the test merge on https://github.com/mike-jumper/guacamole-server/tree/freerdp-migrate-2.x-resolve-merge-master, but I did at least cherry pick the above commits onto that branch so the contents will match the expected merge resolution.

mike-jumper:
OK - the only remaining things I managed to find are:

* Pointer `SetNull` and `SetDefault` implementations were stubs. Don't know how long this has been the case, but since it's essentially one line to add these, I have done so.
* I had manually set the `ReceiveChannelData` handler to `freerdp_channels_data()`, but this is the default and need not be manually set. I've now removed that line.
* There was a potential segfault if the connection ends early enough that any of the various SVCs didn't actually connect before their termination handlers were invoked, as a couple of those implementations did not check whether they had actually allocated their internal structures before attempting to free them. Fixed this with commit 67c5bdf.
* Security negotiation would fail apparently due to having the "extended mode" of NLA enabled by default. Checking the FreeRDP source, they do not do this, and disabling "extended mode" by default fixes things.
* The new default of "any" for security mode is more convenient given that NLA is the new normal for RDP, but this breaks past usage where users may be relying on the lack of username/password to result in a traditional Windows login prompt. I've modified "any" such that it will not attempt to negotiate NLA if doing so would be impossible due to the lack of username/password (at least until parameter prompting is in there).

configure.ac
```diff
@@ -541,507 +540,17 @@ AC_ARG_WITH([rdp],
...
-        AC_MSG_WARN([
-  --------------------------------------------
-   Unknown event interface
-   RDP will be disabled.
-  --------------------------------------------])
-        have_freerdp=no
-    fi
-fi
-
-AM_CONDITIONAL([LEGACY_FREERDP_EXTENSIONS], [test "x${legacy_freerdp_extensions}" = "xyes"])
```
jmuehlner: It's pretty awesome that all this stuff is going away!

src/protocols/rdp/channels/audio-input.h
```diff
@@ -296,18 +292,13 @@ guac_user_end_handler guac_rdp_audio_end_handler;
 /**
  * Adds Guacamole's "guacai" plugin to the list of dynamic virtual channel
  * plugins to be loaded by FreeRDP's "drdynvc" plugin. The plugin will only
- * be loaded once guac_rdp_load_drdynvc() is invoked with the guac_rdp_dvc_list
- * passed to this function. The "guacai" plugin ultimately adds support for the
- * "AUDIO_INPUT"  dynamic virtual channel.
+ * be loaded once the "drdynvc" plugin is loaded. The "guacai" plugin
+ * ultimately adds support for the "AUDIO_INPUT"  dynamic virtual channel.
```
jmuehlner: Double space in this comment should be removed.

src/protocols/rdp/channels/pipe-svc.h
```diff
@@ -0,0 +1,188 @@
...
+ * Handler for "blob" instructions which automatically writes received data to
+ * the associated SVC using guac_rdp_pipe_svc_write().
+ */
+guac_user_blob_handler guac_rdp_pipe_svc_blob_handler;
+
+/**
+ * Handler for "pipe" instructions which automatically prepares received pipe
+ * streams to automatically write received blobs to the SVC having the same
+ * name as the pipe stream. Received pipe streams are associated with the
+ * relevant guac_rdp_pipe_svc instance and the SVC-specific "blob" instructino
```
jmuehlner: Typo: "instructino"

src/protocols/rdp/channels/rail.c
```diff
@@ -0,0 +1,197 @@
...
+ * server. No communication for RemoteApp may occur until the Handshake PDU
+ * (or, alternatively, the HandshakeEx PDU) is received. See:
+ *
+ * https://docs.microsoft.com/en-us/openspecs/windows_protocols/ms-rdperp/cec4eb83-b304-43c9-8378-b5b8f5e7082a
+ *
+ * @param rail
+ *     The RailClientContext structure used by FreeRDP to handle the RAIL
+ *     channel for the current RDP session.
+ *
+ * @param handshake
```
jmuehlner: This isn't used for anything?

src/protocols/rdp/channels/rdpdr/rdpdr.h
```diff
@@ -0,0 +1,250 @@
...
+ * @return
+ *     A new wStream containing an I/O completion header, followed by the
+ *     requested additional free space.
+ */
+wStream* guac_rdpdr_new_io_completion(guac_rdpdr_device* device,
+        int completion_id, int status, int size);
+
+/**
+ * Initializes device redirection support (file transfer, printing, etc.) for
+ * RDP and handling of the RDPDR channel.  If failures occur, messages noting
```
jmuehlner: Another double space here

src/protocols/rdp/doc/svc-example/README.md
```diff
@@ -0,0 +1,169 @@
...
+ * [`WTSVirtualChannelWrite()`](https://docs.microsoft.com/en-us/windows/win32/api/wtsapi32/nf-wtsapi32-wtsvirtualchannelwrite)
+ * [`WTSVirtualChannelClose()`](https://docs.microsoft.com/en-us/windows/win32/api/wtsapi32/nf-wtsapi32-wtsvirtualchannelclose)
+
+On the Guacamole side, bidirectional communication is established using:
+
+ * The `static-channels` connection parameter (in the case of the example, this should be set to `EXAMPLE`).
+ * An [`onpipe`](http://guacamole.apache.org/doc/guacamole-common-js/Guacamole.Client.html#event:onpipe)
+   handler which handles inbound (server-to-client) pipe streams named identically
+   to the SVC. The inbound pipe stream will be received upon establishing the RDP
+   connection and is used to transmit any data sent along the SVC **from** within
```
jmuehlner: Why is just the word `from` emphasized here?

src/protocols/rdp/download.c
```diff
@@ -0,0 +1,240 @@
...
+    int file_id = guac_rdp_fs_open(filesystem, path,
+            FILE_READ_DATA, 0, FILE_OPEN, 0);
+
+    /* If file opened successfully, start stream */
+    if (file_id >= 0) {
+
+        guac_rdp_download_status* download_status;
+        const char* basename;
+
+        int i;
```
jmuehlner: Why not declare these variables down where they're actually used?

src/protocols/rdp/plugins/channels.c
```diff
@@ -0,0 +1,129 @@
...
+
+    return wrapper;
+
+}
+
+PVIRTUALCHANNELENTRY guac_rdp_plugin_wrap_entry(PVIRTUALCHANNELENTRY entry) {
+
+    /* Do not wrap if there is insufficient space to store the wrapped
+     * function */
+    if (guac_rdp_wrapped_entry_count == GUAC_RDP_MAX_CHANNELS)
```
jmuehlner: No warning or anything if wrapping fails? What are the consequences if this happens?

src/protocols/rdp/channels/rail.c
mike-jumper: Not currently, no. We ignore the contents of the handshake, simply using it to signal when client system parameters should be sent and execution of the RemoteApp should be requested.

This is actually in line with the behavior of xfreerdp, which also ignores the contents of the handshake PDU received from the server:

https://github.com/FreeRDP/FreeRDP/blob/02614cce49ce6ff120a1306962c5d3ff6b05ea75/client/X11/xf_rail.c#L892-L901

src/protocols/rdp/download.c
mike-jumper: It was the coding style at the time that this function was originally written. The body of this function was actually moved from what used to be `rdpdr_service.c` and was originally called `guac_rdpdr_download_to_owner`:

https://github.com/apache/guacamole-server/blob/f5b5ac718396b24352290d2fac976260f4b3829e/src/protocols/rdp/guac_rdpdr/rdpdr_service.c#L237-L303

I can clean this up a bit now in the locations I can find such code.

src/protocols/rdp/upload.c
```diff
@@ -0,0 +1,236 @@
...
+
+}
+
+int guac_rdp_upload_blob_handler(guac_user* user, guac_stream* stream,
+        void* data, int length) {
+
+    int bytes_written;
+    guac_rdp_upload_status* upload_status = (guac_rdp_upload_status*) stream->data;
+
+    /* Get filesystem, return error if no filesystem 0*/
```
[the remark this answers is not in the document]
mike-jumper: Arg, it's a vim typo. I must have been moving the cursor to the beginning of the line.

src/protocols/rdp/channels/rail.c
jmuehlner: Huh, ok.

src/protocols/rdp/plugins/channels.c
mike-jumper: > No warning or anything if wrapping fails?

To log a message using the normal logging subsystem exposed by libguac, we would need access to the `guac_client`, however the `guac_client` is unavailable at this level as the calling function (`guac_freerdp_channels_load_plugin()`) was meant to be a drop-in replacement for part of the FreeRDP API (`freerdp_channels_load_plugin()`) and none of the parameters received by these functions have a path for reaching the `guac_client`.

We could work around this by:

* *Not* having this function be a true drop-in replacement. There is arguably little utility to having the function be a drop-in replacement now that the code has been migrated. Passing different arguments (perhaps `rdpContext`) would allow access to the associated `guac_client`.
* Using FreeRDP's logging subsystem (part of libwinpr). That subsystem is rerouted early on in the connection process such that all messages logged from FreeRDP actually get logged to the `guac_client` at the debug level.

> What are the consequences if this happens?

Generally, it shouldn't - there is an upper bound on the maximum number of static channels that can be loaded. This upper bound is set by RDP itself and by FreeRDP. The `GUAC_RDP_MAX_CHANNELS` constant is set to the maximum number of static channels that FreeRDP can load, and while we can't verify this at runtime, there is a compile-time check that will fail the build if the constant is set incorrectly:

https://github.com/apache/guacamole-server/blob/55959b54564cd7ce8800225ed8d3741345eb90a2/src/protocols/rdp/plugins/channels.h#L39-L42

The plugin would fail to load due to there not being a slot available to hold the plugin.

Setting that aside, though, let's assume something strange is going on and the plugin does successfully load after the limit is reached:

The first attempt to load a plugin X would succeed, as prior instances of X had their entry points wrapped and FreeRDP sees this load attempt as unique. After this point, further attempts to load copies of X would fail due to its entry point already being present, and FreeRDP would refuse to load the plugin.

src/protocols/rdp/channels/rdpdr/rdpdr-messages.c
```diff
@@ -0,0 +1,349 @@
...
+    guac_client_log(svc->client, GUAC_LOG_DEBUG, "Client ID confirmed");
+}
+
+void guac_rdpdr_process_device_reply(guac_rdp_common_svc* svc,
+        wStream* input_stream) {
+
+    guac_rdpdr* rdpdr = (guac_rdpdr*) svc->data;
+
+    unsigned int device_id, ntstatus;
+    int severity, c, n, facility, code;
```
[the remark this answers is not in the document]
mike-jumper: Weirdly enough, they are standard names. The standard is not set by FreeRDP but by Microsoft. Each of these variables refers to a portion of a 32-bit "NTSTATUS" code:

https://docs.microsoft.com/en-us/openspecs/windows_protocols/ms-erref/87fba13e-bf06-450e-83b1-9241dc81e781

As I'm calling the "sev" field "severity" here, it might make sense to change `c` to `customer`, however I'm not sure anyone knows what `n` is:

> **N (1 bit)**: Reserved. MUST be set to 0 so that it is possible to map an NTSTATUS value to an equivalent HRESULT value, as specified in section 2.1, by setting this bit.

configure.ac
mike-jumper: Yup!

src/protocols/rdp/download.c
mike-jumper: OK - I've cleaned this up, moved the basename logic to its own function within `fs.h`, and added a unit test for said function.

src/protocols/rdp/channels/rdpdr/rdpdr.h
mike-jumper: Should now be fixed - I did an overall search for double spaces occurring within comments and corrected all occurrences.

src/protocols/rdp/channels/pipe-svc.h
mike-jumper: Fixedo.

src/protocols/rdp/channels/audio-input.h
mike-jumper: Should now be fixed.

src/protocols/rdp/upload.c
```diff
@@ -0,0 +1,236 @@
...
+/**
+ * Writes the given filename to the given upload path, sanitizing the filename
+ * and translating the filename to the root directory.
+ *
+ * @param filename
+ *     The filename to sanitize and move to the root directory.
+ *
+ * @param path
+ *     A pointer to a buffer which should receive the sanitized path. The
+ *     buffer must hav at least GUAC_RDP_FS_MAX_PATH bytes available.
```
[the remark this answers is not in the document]
mike-jumper: Fixed.

src/protocols/rdp/upload.c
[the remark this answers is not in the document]
mike-jumper: Fixed.

src/protocols/rdp/plugins/channels.c
mike-jumper: OK - I went ahead and refactored `guac_freerdp_channels_load_plugin()` such that it accepts a `rdpContext` instead, allowing warnings to be logged if plugins may fail to load due to limits being reached. It's no longer a true drop-in replacement, but it really doesn't need to be.

mike-jumper:
OK, I believe I've addressed everything with the exception of the `n` and `c` naming, which I've commented on above. Let me know if you still think that needs fixing and I'll look around for possibilities.

necouchman:
I’m working through a review, as well - will try to buckle down and get it
submitted tomorrow...

mike-jumper:
Thanks, guys.

src/protocols/rdp/upload.c
```diff
@@ -0,0 +1,236 @@
...
+    char file_path[GUAC_RDP_FS_MAX_PATH];
+
+    /* Get filesystem, return error if no filesystem */
+    guac_rdp_fs* fs = rdp_client->filesystem;
+    if (fs == NULL) {
+        guac_protocol_send_ack(user->socket, stream, "FAIL (NO FS)",
+                GUAC_PROTOCOL_STATUS_SERVER_ERROR);
+        guac_socket_flush(user->socket);
+        return 0;
+    }
```
necouchman: Are we really returning an error if we're returning 0, which also seems to be the success code?

src/protocols/rdp/channels/audio-input.h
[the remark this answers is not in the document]
mike-jumper: This should be fixed, I think. @jmuehlner found a few of these, as well, and I fixed them in commit 2139d40 after doing a search of all files beneath `src/protocols/rdp/` for any other occurrences of extra spaces in comments.

src/protocols/rdp/channels/rdpdr/rdpdr-fs-messages-dir-info.h
```diff
@@ -17,58 +17,76 @@
...
  * documentation, this is "defined as all the basic information, plus extended
  * attribute size."
  */
-void guac_rdpdr_fs_process_query_full_directory_info(guac_rdpdr_device* device,
-        const char* entry_name, int file_id, int completion_id);
+guac_rdpdr_directory_query_handler guac_rdpdr_fs_process_query_full_directory_info;
 
 /**
  * Processes a query request for FileBothDirectoryInformation. From the
  * documentation, this absurdly-named request is "basic information plus
```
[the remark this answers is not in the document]
mike-jumper: Heh. :)

src/protocols/rdp/channels/pipe-svc.h
```diff
@@ -0,0 +1,188 @@
...
+ * handler (guac_rdp_pipe_svc_blob_handler).
```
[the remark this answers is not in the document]
mike-jumper: @jmuehlner beat you to it - this should already be fixed via commit feb376e. ;)

src/protocols/rdp/channels/pipe-svc.h
```diff
@@ -0,0 +1,188 @@
...
+
+/**
+ * Handler for "blob" instructions which automatically writes received data to
+ * the associated SVC using guac_rdp_pipe_svc_write().
+ */
+guac_user_blob_handler guac_rdp_pipe_svc_blob_handler;
+
+/**
+ * Handler for "pipe" instructions which automatically prepares received pipe
+ * streams to automatically write received blobs to the SVC having the same
```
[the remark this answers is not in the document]
mike-jumper: Fair point.

src/protocols/rdp/channels/cliprdr.c
```diff
@@ -0,0 +1,601 @@
...
+     * callable, until after the relevant guac_rdp_clipboard structure is
+     * allocated and associated with the CliprdrClientContext */
+    guac_rdp_clipboard* clipboard = (guac_rdp_clipboard*) cliprdr->custom;
+    assert(clipboard != NULL);
+
+    guac_client_log(clipboard->client, GUAC_LOG_TRACE, "CLIPRDR: Received "
+            "monitor ready.");
+
+    /* Respond with capabilities and supported format list */
+    guac_rdp_cliprdr_send_capabilities(cliprdr);
```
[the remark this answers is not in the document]
mike-jumper: My original thinking was that if `guac_rdp_cliprdr_send_capabilities()` fails, the following call to `guac_rdp_cliprdr_send_format_list()` should also fail. I'll add a check to handle both.

src/protocols/rdp/bitmap.h
```diff
@@ -66,8 +61,11 @@ typedef struct guac_rdp_bitmap {
  *
  * @param bitmap
  *     The bitmap to cache.
+ *
+ * @return
+ *     TRUE if successful, FALSE otherwise.
```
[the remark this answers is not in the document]
mike-jumper: In this case, no, I don't believe so. I was probably just on a roll with `BOOL`-ing everything.

I'll recheck the body of the function and switch things back to `void` if `BOOL` is not needed. The fewer things need changing, the better.

src/protocols/rdp/channels/pipe-svc.c
```diff
@@ -0,0 +1,230 @@
...
+    guac_rdp_pipe_svc* pipe_svc = guac_rdp_pipe_svc_get(user->client, name);
+
+    /* Fail if no such SVC */
+    if (pipe_svc == NULL) {
+        guac_user_log(user, GUAC_LOG_WARNING, "User requested non-existent "
+                "pipe (no such SVC configured): \"%s\"", name);
+        guac_protocol_send_ack(user->socket, stream, "FAIL (NO SUCH PIPE)",
+                GUAC_PROTOCOL_STATUS_CLIENT_BAD_REQUEST);
+        guac_socket_flush(user->socket);
+        return 0;
```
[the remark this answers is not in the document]
mike-jumper: Yes. Returning an error code from within an instruction handler aborts the user's connection:

https://github.com/apache/guacamole-server/blob/381ff1a4214e06803f0f587f1ed97cf7e54853e7/src/libguac/user-handshake.c#L164-L177

In this case, the instruction is successfully handled in that the fault has been communicated to the user with an `ack`.

src/protocols/rdp/upload.c
mike-jumper: Yep - the error is being returned to the user via the `ack`. The received instruction has otherwise been successfully handled and the rest of the user's connection can continue.

src/protocols/rdp/channels/cliprdr.c
```diff
@@ -0,0 +1,601 @@
...
+         * not supported here, then something has been implemented incorrectly.
+         * Either incorrect values are (somehow) being stored, or support for
+         * the format indicated by that value is incomplete and must be added
+         * here. The values which may be stored within requested_format are
+         * completely within our control. */
+        default:
+            guac_client_log(clipboard->client, GUAC_LOG_DEBUG, "Requested "
+                    "clipboard data in unsupported format (0x%X).",
+                    clipboard->requested_format);
+            return CHANNEL_RC_OK;
```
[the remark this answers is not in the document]
mike-jumper: Yes. FreeRDP handles error return codes similarly to the way we do, with such errors being treated as fatal (or at least near-fatal). In this case, I'm not sure whether the entire RDP connection would terminate, but returning a non-zero code here would result in the message processing thread for CLIPRDR being shut down.

The root of the relevant code flow is the thread body of FreeRDP's CLIPRDR handling:

https://github.com/FreeRDP/FreeRDP/blob/32bf768652a1a6d6872453ff5c84ab97bedaaa00/channels/cliprdr/client/cliprdr_main.c#L891-L935

In a nutshell:

1. Loading FreeRDP's "cliprdr" plugin results in the above thread starting. It handles messages received along the CLIPRDR channel, delegating processing of those messages to `cliprdr_order_recv()`, an internal function.
2. In the case of a Format Data Response PDU, `cliprdr_order_recv()` further delegates to `cliprdr_process_format_data_response()` which finally delegates to the function assigned as the `ServerFormatDataResponse` handler, which is `guac_rdp_cliprdr_format_data_response()` in our case.
3. If the `ServerFormatDataResponse` handler returns anything but `CHANNEL_RC_OK`, that error code propogates its way back up the chain and results in the thread being killed.

So, we could return an error code here, but it would result in no further CLIPRDR messages being processed. Receiving data in an unexpected format from the RDP server isn't great, and would be a violation of CLIPRDR, but maybe is not so bad that CLIPRDR should be killed for the rest of the session. Returning success here (but logging that things are odd) allows things to continue, assuming copy/paste is otherwise working despite the server misbehaving.

src/protocols/rdp/channels/cliprdr.c
```diff
@@ -0,0 +1,601 @@
...
+        case CF_UNICODETEXT:
+            reader = GUAC_READ_UTF16;
+            break;
+
+        /* If the format ID stored within the guac_rdp_clipboard structure is actually
+         * not supported here, then something has been implemented incorrectly.
+         * Either incorrect values are (somehow) being stored, or support for
+         * the format indicated by that value is incomplete and must be added
+         * here. The values which may be stored within requested_format are
+         * completely within our control. */
```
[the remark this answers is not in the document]
mike-jumper: I believe so (see my other response regarding FreeRDP's CLIPRDR thread and its error handling).

src/protocols/rdp/bitmap.h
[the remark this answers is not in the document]
mike-jumper: Regarding the other cases, yes - the other functions are written to comply with the API of the FreeRDP library, which does require a `BOOL` return to signal failures, even if our implementation doesn't have a failure condition in most (all?) cases.

mike-jumper:
OK - I think I have addressed everything noted thus far with the exception of:

1. The naming of the `n` and `c` variables representing components of an NTSTATUS code.
2. The concerns with CLIPRDR handler return codes.

I believe the above are correct as they stand (see comments), but please let me know if you believe they still need to be addressed.

src/protocols/rdp/channels/pipe-svc.h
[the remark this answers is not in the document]
necouchman: Sorry about that - I even reloaded the review page a couple of times thinking that I'd see those items get updated.

src/protocols/rdp/channels/cliprdr.c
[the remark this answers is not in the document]
necouchman: Okay, makes sense.  I'm good with it.

necouchman:
Overall I'm fine with it.  Going to do one more quick(er) review pass and then do a test compile and run it and make sure everything looks okay.

This is great work, Mike - so nice to have this one behind us and look at actually releasing 1.1.0 - and then another version pretty quickly after that, hopefully, and start our more frequent release process.  This is so cool!

src/protocols/rdp/channels/cliprdr.c
```diff
@@ -0,0 +1,605 @@
...
+     * allocated and associated with the CliprdrClientContext */
+    guac_rdp_clipboard* clipboard = (guac_rdp_clipboard*) cliprdr->custom;
+    assert(clipboard != NULL);
+
+    guac_client_log(clipboard->client, GUAC_LOG_TRACE, "CLIPRDR: Received "
+            "monitor ready.");
+
+    /* Respond with capabilities ... */
+    int status = guac_rdp_cliprdr_send_capabilities(cliprdr);
+    if (status != CHANNEL_RC_OK)
```
necouchman: `if (!guac_rdp_cliprdr_send_capabilities(cliprdr))`

;-)

src/protocols/rdp/channels/cliprdr.c
```diff
@@ -0,0 +1,605 @@
...
+ *     (non-zero) otherwise.
+ */
+static UINT guac_rdp_cliprdr_format_list(CliprdrClientContext* cliprdr,
+        const CLIPRDR_FORMAT_LIST* format_list) {
+
+    /* FreeRDP-specific handlers for CLIPRDR are not assigned, and thus not
+     * callable, until after the relevant guac_rdp_clipboard structure is
+     * allocated and associated with the CliprdrClientContext */
+    guac_rdp_clipboard* clipboard = (guac_rdp_clipboard*) cliprdr->custom;
+    assert(clipboard != NULL);
```
necouchman: This is more a curiosity than anything, but is there any special reason to using `assert` here instead of `if (clipboard == NULL); return`?  I'm familiar with `assert` in the Java side of the code, but it seems pretty rare in Guacamole's C code-based, so I'm just curious why it goes here and how it changes behavior?

necouchman:
FWIW - I just applied this change to the current `staging/1.1.0` branch and I get some compile failures.  Going to clean up and try, again, but not sure if it's mine or something missing from the commits here:
```
channels/cliprdr.c: In function 'guac_rdp_cliprdr_channel_connected':
channels/cliprdr.c:477:27: error: assignment from incompatible pointer type [-Werror]
     cliprdr->MonitorReady = guac_rdp_cliprdr_monitor_ready;
                           ^
channels/cliprdr.c:478:31: error: assignment from incompatible pointer type [-Werror]
     cliprdr->ServerFormatList = guac_rdp_cliprdr_format_list;
                               ^
channels/cliprdr.c:479:38: error: assignment from incompatible pointer type [-Werror]
     cliprdr->ServerFormatDataRequest = guac_rdp_cliprdr_format_data_request;
                                      ^
channels/cliprdr.c:480:39: error: assignment from incompatible pointer type [-Werror]
     cliprdr->ServerFormatDataResponse = guac_rdp_cliprdr_format_data_response;
                                       ^
cc1: all warnings being treated as errors
```

mike-jumper:
I'm not seeing those errors on my end. Which Linux distribution and version of FreeRDP are you building against?

necouchman:
> I'm not seeing those errors on my end. Which Linux distribution and version of FreeRDP are you building against?

Just tried a fresh copy straight from Apache repo and same results.  I'm using CentOS7, freerdp-devel version is 2.0.0-1.rc4.el7 (latest, I think).  My process is:
- Clone Apache repo
- Checkout `staging/1.1.0`
- Checkout pull/243/head to merge-243
- Merge --no-ff merge-243
- autoreconf -fi
- ./configure --prefix=/opt/guac-test
- make

mike-jumper:
Hm. I will recheck. I really hope this doesn't mean we need to add new configure tests for API differences already...

mike-jumper:
Sigh ... sure enough: https://github.com/FreeRDP/FreeRDP/commit/65812bdbc00df2247f51147e78b25076d16ab1d8

I'll add some checks...

necouchman:
Ugh.  It would be nice if the API would stay constant enough to be able to use it...

src/protocols/rdp/channels/cliprdr.c
mike-jumper: `assert()` in C will cause SIGABRT to be sent to the process if the condition fails. If we were expecting `clipboard` to occasionally be `NULL`, then adding an `if` would be best here. In this case, it is reasoned that `clipboard` absolutely cannot be `NULL` under any circumstance where this code can be reached, and we declare that assumption with an `assert()`. Smart compilers, static checkers, etc. will take that `assert()` into account, and it allows the system to bail out if a required assumption ends up being false.

mike-jumper:
Actually ... as the calling convention is the same between the two variants, and the only difference is that the version we currently code against explicitly declares that it will not modify the contents of that structure, we may have lucked out here. I think we can safely typecast to the type expected.

src/protocols/rdp/channels/cliprdr.c
necouchman: Thx.

mike-jumper:
OK - I think I've fixed the issue, which seems to be the only build issue resulting from changes to the FreeRDP after CentOS produced their package. I encountered a different issue after that was addressed due to the way that I was checking platform endianness, but I think I've fixed that, too.

necouchman:
Okay - I'm trying to finish up testing here tonight and might be able to get it merged tonight or tomorrow.  @jmuehlner Any other comments or objections prior to proceeding with the merge?

necouchman:
Okay, testing looks good for me - I'll try to kick the tires on it some more over the next few days to be sure, but initial stuff looks fine.

@jmuehlner Any last objections before the merge?

mike-jumper:
I'm currently seeing the following after building on Debian:

    undefined symbol: guac_freerdp_dynamic_channel_collection_add	(/usr/local/lib/freerdp2/libguacai-client.so)

Not sure why this isn't an issue under other distros, but I think I can resolve by moving the common parts of audio input to a different unit. The function referenced above is never invoked within the plugin and doesn't actually need to be present.

mike-jumper:
OK, the above should be addressed by commit 4282da6. Checking against various Ubuntus, I'm now seeing the following on Bionic:

```
  CC       channels/libguac_client_rdp_la-rail.lo
channels/rail.c: In function 'guac_rdp_rail_channel_connected':
channels/rail.c:167:27: error: assignment from incompatible pointer type [-Werror=incompatible-pointer-types]
     rail->ServerHandshake = guac_rdp_rail_handshake;
                           ^
channels/rail.c:168:29: error: assignment from incompatible pointer type [-Werror=incompatible-pointer-types]
     rail->ServerHandshakeEx = guac_rdp_rail_handshake_ex;
                             ^
cc1: all warnings being treated as errors
```

Will see whether the solution here is as simple as before.

mike-jumper:
As an aside, I'd like to say it's definitely irksome that "FreeRDP 2.0.0" does not mean the same thing across the various distros, with variances in the API despite otherwise being apparently the same version.

Looks like we should be able to make things flexible enough to build across the board, though.

mike-jumper:
After fixing the RAIL issue, I encountered the following:

```
...
  CC       libguac_client_rdp_la-color.lo
color.c: In function 'guac_rdp_convert_color':
color.c:62:16: error: implicit declaration of function 'FreeRDPConvertColor'; did you mean 'ConvertColor'? [-Werror=implicit-function-declaration]
     intermed = FreeRDPConvertColor(intermed, src_format, dst_format,
                ^~~~~~~~~~~~~~~~~~~
                ConvertColor
cc1: all warnings being treated as errors
```

Apparently, the `FreeRDPConverColor()` function was formerly known as `ConvertColor()`. It was renamed along with similarly-named functions due to Mac applications using `GetColor()` failing to build because that function clashes with a system function of the same name (see FreeRDP/FreeRDP#3810).

Adding tests to determine the proper function doesn't appear to be an option at the moment, as:

* The necessary include path for FreeRDP is only known via pkg-config and I'm unaware of any way to get autoconf tests to find headers that can only be found when using the `CFLAGS` returned by pkg-config.
* We can't simply test version numbers because all of these FreeRDP variants have the same version.

I tried addressing this by switching to `ConvertColor()`, which should work due to compatibility macros present in the FreeRDP headers (commit a6b7e07). *But*, I then encountered issues with the glyph callbacks:

```
  CC       libguac_client_rdp_la-rdp.lo
rdp.c: In function 'rdp_freerdp_pre_connect':
rdp.c:155:16: error: assignment from incompatible pointer type [-Werror=incompatible-pointer-types]
     glyph.Draw = guac_rdp_glyph_draw;
                ^
rdp.c:156:21: error: assignment from incompatible pointer type [-Werror=incompatible-pointer-types]
     glyph.BeginDraw = guac_rdp_glyph_begindraw;
                     ^
rdp.c:157:19: error: assignment from incompatible pointer type [-Werror=incompatible-pointer-types]
     glyph.EndDraw = guac_rdp_glyph_enddraw;
                   ^
cc1: all warnings being treated as errors
```

In this case, the incompatibility is due to switching from `UINT32` to `INT32` for width and height. At this point, I'm thinking we simply shouldn't support the version of FreeRDP on Bionic. The version of the API in use appears to be unique to Bionic and from before any 2.0.0 release candidate was made.

I've also created some Jenkins jobs to verify the build, so we should be able to better ensure build compatibility going forward:

* https://builds.apache.org/view/E-G/view/Guacamole/job/guacamole-server-centos/
* https://builds.apache.org/view/E-G/view/Guacamole/job/guacamole-server-fedora/
* https://builds.apache.org/view/E-G/view/Guacamole/job/guacamole-server-debian/
* https://builds.apache.org/view/E-G/view/Guacamole/job/guacamole-server-ubuntu/

mike-jumper:
Rerunning builds for all FreeRDP tags, it looks like the current code will build only against 2.0.0-rc4. Earlier tags suffer from the issues noted above for Bionic.

necouchman:
> Rerunning builds for all FreeRDP tags, it looks like the current code will build only against 2.0.0-rc4. Earlier tags suffer from the issues noted above for Bionic.

Lovely...so we're tied to a current RC of FreeRDP 2.  That feels safe.

mike-jumper:
> Lovely...so we're tied to a current RC of FreeRDP 2. That feels safe.

Well, 2.0.0-rc4 and master, but yeah - I know what you mean.

I'm not necessarily against trying to be compatible with rc0 through rc3, but I'm not sure how we would distinguish them. It doesn't seem possible to write configure tests because of pkg-config, and the version numbers are identical, even within FreeRDP's version.h.

### 2.0.0-rc0

```c
#define FREERDP_VERSION_MAJOR 2
#define FREERDP_VERSION_MINOR 0
#define FREERDP_VERSION_REVISION 0
#define FREERDP_VERSION_SUFFIX "dev5"
#define FREERDP_API_VERSION "2"
#define FREERDP_VERSION "2.0.0"
#define FREERDP_VERSION_FULL "2.0.0-dev5"
#define GIT_REVISION "326a44895"
```

### 2.0.0-rc4

```c
#define FREERDP_VERSION_MAJOR 2
#define FREERDP_VERSION_MINOR 0
#define FREERDP_VERSION_REVISION 0
#define FREERDP_VERSION_SUFFIX "dev5"
#define FREERDP_API_VERSION "2"
#define FREERDP_VERSION "2.0.0"
#define FREERDP_VERSION_FULL "2.0.0-dev5"
#define GIT_REVISION "326a44895"
```

mike-jumper:
I'll try nested autoconf. Perhaps adding a configure script which is dedicated to RDP and which is invoked by the top-level configure script will allow us to perform tests which inherit flags from pkg-config.

mike-jumper:
aHA! Manually setting `CPPFLAGS` within seems to allow tests to work, as it adds the needed `-I` flags globally:

```
    PKG_CHECK_MODULES([RDP], [freerdp2 freerdp-client2 winpr2],
                      [CPPFLAGS="${RDP_CFLAGS} $CPPFLAGS"],
                      [AC_MSG_WARN([
```

The downside is that it adds the flags globally. It shouldn't hurt the build, as part of the reason this is so cumbersome is that FreeRDP 2.x is double-namespaced, with headers being within `freerdp2/freerdp/`.

mike-jumper:
OK, I'm having success with configure tests and am rewriting the previous typecast solutions for CLIPRDR, RAIL, etc. to instead use tests. This should be more future-proof in the sense that we'll get obvious build failures if API compatibility breaks, and it should allow us to support FreeRDP: Bionic Edition as well as 2.0.0-rc0 onward.

necouchman:
Everything looks good with CentOS7 with 2.0.0-rc4

mike-jumper:
OK, I've added configure tests and things seem to build across the board now: git master, 2.0.0-rc0 through 2.0.0-rc4, as well as Bionic Edition.

The ASF Jenkins instance is too backed up at the moment to be used to verify the builds, but I'm doing so manually on my own dev machine via Docker.

mike-jumper:
I've manually confirmed on:

* CentOS 7
* Fedora 29, 30, and 31
* Debian stable and testing
* Ubuntu Bionic, Disco, and Eoan
* Builds of FreeRDP from source using 2.0.0-rc0, 2.0.0-rc1, 2.0.0-rc2, 2.0.0-rc3, 2.0.0-rc4, and master

I *think* we're good to go.

necouchman:
Sounds good.  Giving @jmuehlner one last chance to throw in any input - otherwise I'll plan to merge it later today.

You mentioned needing to deal with conflicts in the master - anything special that we need to worry about for that?

mike-jumper:
> You mentioned needing to deal with conflicts in the master - anything special that we need to worry about for that?

Only that these conflicts are likely to be more difficult to resolve than the usual conflicts. I've tried to summarize the specific issues that conflicted and how to resolve in the issue description. AFAIK, there's no real way to provide a patch which addresses merge conflicts, but I went ahead and performed the merge myself as a test, manually resolving the conflicts, and have been cherry-picking any new commits to that branch since then:

https://github.com/mike-jumper/guacamole-server/tree/freerdp-migrate-2.x-resolve-merge-master

My hope is that the above branch might help in verifying the result of the merge and/or in guiding the resolution of the conflicts.

necouchman:
Sounds good...I think it's go time...

:shipit:

necouchman:
Okay, merge to `staging/1.1.0` is done.  I have to step out for a little while, but will work through the merge to master here, shortly.

necouchman:
Merge to `master` completed.  I had to work around a couple of settings that got added into master, but I think I've managed to do it without clobbering any of the updates that have happened to `master` in the meantime.  There's certainly an off-chance we'll have to go back and fix something that I accidentally undid, but I tried to go back and verify a few times that nothing substantial was removed.

mike-jumper:
Looks real solid. I just manually tried merging the previous `HEAD` of `master` onto my test merge resolution branch, resolved the conflicts manually, and the diff against current `master` showed no differences at all ... except for a new CLIPRDR-related log message.

I think that log message should probably be removed, and I will open a PR to that effect in a moment.

PR #245, opened 2020-01-16
GUACAMOLE-249: Automatically determine correct location for FreeRDP plugins.

mike-jumper:
This change leverages pkg-config to determine the actual, correct location used by FreeRDP for its plugins, logging the discovered directory in the summary at the end of `configure` and using that directory for the install location of the FreeRDP plugins (regardless of any prefix, libdir, etc. that might otherwise have been given as options to the `configure` script):

```
[mjumper@dev-mjumper guacamole-server]$ ./configure
...
config.status: creating config.h
config.status: executing depfiles commands
config.status: executing libtool commands

------------------------------------------------
guacamole-server version 1.1.0
------------------------------------------------

   Library status:

     freerdp2 ............ yes
     pango ............... yes
     libavcodec .......... yes
     libavutil ........... yes
     libssh2 ............. yes
     libssl .............. yes
     libswscale .......... yes
     libtelnet ........... yes
     libVNCServer ........ yes
     libvorbis ........... yes
     libpulse ............ yes
     libwebsockets ....... yes
     libwebp ............. yes
     wsock32 ............. no

   Protocol support:

      Kubernetes .... yes
      RDP ........... yes
      SSH ........... yes
      Telnet ........ yes
      VNC ........... yes

   Services / tools:

      guacd ...... yes
      guacenc .... yes
      guaclog .... yes

   FreeRDP plugins: /usr/lib64/freerdp2
   Init scripts: no
   Systemd units: no

Type "make" to compile guacamole-server.

[mjumper@dev-mjumper guacamole-server]$ 
```

I figured this might be worth doing as part of GUACAMOLE-249, given that the FreeRDP plugins being installed are changing and symbolic links from installs of Apache Guacamole 1.0.0 and older will not work. If others think this should be a separate issue, I'm not extremely against letting this wait, but I think this will save everyone some pain.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -532,11 +532,19 @@ fi
 #
 
 have_freerdp2=disabled
+FREERDP2_PLUGIN_DIR=
+
 AC_ARG_WITH([rdp],
             [AS_HELP_STRING([--with-rdp],
                             [support RDP @<:@default=check@:>@])],
             [],
             [with_rdp=check])
+
+# FreeRDP plugin directory
+AC_ARG_WITH(freerdp_plugin_dir,
+            [AS_HELP_STRING([--with-freerdp-plugin-dir=<path>],
+                            [install FreeRDP plugins to the given directory @<:@default=check@:>@])
+            ],FREERDP2_PLUGIN_DIR=$withval)
 
 # Preserve CPPFLAGS so it can be restored later, following the addition of
 # options specific to FreeRDP tests
@@ -547,7 +555,8 @@ then
     have_freerdp2=yes
     PKG_CHECK_MODULES([RDP], [freerdp2 freerdp-client2 winpr2],
                       [CPPFLAGS="${RDP_CFLAGS} -Werror $CPPFLAGS"]
-                      [FREERDP2_PLUGIN_DIR="`$PKG_CONFIG --variable=libdir freerdp2`/freerdp2"],
+                      [AS_IF([test "x${FREERDP2_PLUGIN_DIR}" = "x"],
+                             [FREERDP2_PLUGIN_DIR="`$PKG_CONFIG --variable=libdir freerdp2`/freerdp2"])],
                       [AC_MSG_WARN([
   --------------------------------------------
    Unable to find FreeRDP (libfreerdp2 / libfreerdp-client2 / libwinpr2)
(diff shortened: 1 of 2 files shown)


necouchman:
I'm fine adding this to -249 - it makes sense to me.

configure.ac
```diff
@@ -994,6 +996,12 @@ AM_COND_IF([ENABLE_INIT], [build_init="${init_dir}"], [build_init=no])
 
 AM_COND_IF([ENABLE_SYSTEMD], [build_systemd="${systemd_dir}"], [build_systemd=no])
 
+#
+# FreeRDP plugins
+#
+
+AM_COND_IF([ENABLE_RDP], [build_rdp_plugins="${FREERDP2_PLUGIN_DIR}"], [build_rdp_plugins=no])
```
necouchman: What happens if this ends up no or undefined - because FreeRDP2, for whatever reason, lacks the plugins directory?

mike-jumper: It depends on which variable you're referring to:

The `build_rdp_plugins` variable is only used to represent the status of the build visually within the summary. If it ends up being "no" (due to FreeRDP being absent or due to RDP being explicitly disabled via command line options passed to `configure`), then the summary ends up looking like:

```
...
   Services / tools:

      guacd ...... yes
      guacenc .... yes
      guaclog .... yes

   FreeRDP plugins: no
   Init scripts: no
   Systemd units: no
...
```

`FREERDP2_PLUGIN_DIR` cannot be "no". It could potentially be a strange value if the pkg-config file installed by FreeRDP is broken, in which case the RDP part of the build would be broken.

I'll add a `--with-freerdp-plugin-dir` option to override the auto-detected value, in case the detected value ends up being incorrect.

mike-jumper: OK, I've added a `--with-freerdp-plugin-dir` option. From `./configure --help`:

```
...
  --with-terminal         support text-based protocols [default=check]
  --with-vnc              support VNC [default=check]
  --with-rdp              support RDP [default=check]
  --with-freerdp-plugin-dir=<path>
                          install FreeRDP plugins to the given directory
                          [default=check]
...
```

If given, the value will be used instead of attempting to derive the value from the package lib directory exposed by pkg-config.

PR #246, opened 2020-01-16
GUACAMOLE-928: Keep all build artifacts of Docker image within /usr/local/guacamole, including FreeRDP plugins.

mike-jumper:
The change introduced by #245 actually causes issues with the Docker image build due to the way a separate builder image is used. The image build process expects that absolutely all build artifacts will be installed beneath `/usr/local/guacamole`, but this is no longer the case by default due to the automatic detection of FreeRDP plugin location.

This change overrides the FreeRDP plugin location detection that is part of the guacamole-server build, keeping things together for the sake of the Docker image build and relying instead on the rest of the build to add any symbolic links needed for FreeRDP to find the plugins.

### DIFF
--- a/src/guacd-docker/bin/build-guacd.sh
+++ b/src/guacd-docker/bin/build-guacd.sh
@@ -43,7 +43,7 @@ PREFIX_DIR="$2"
 
 cd "$BUILD_DIR"
 autoreconf -fi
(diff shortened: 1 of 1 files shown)


mike-jumper:
Created a separate JIRA issue for this regression (and any others), rather than continue using GUACAMOLE-249.

PR #247, opened 2020-01-17
GUACAMOLE-927: Automatically set $HOME for sake of FreeRDP initialization process

mike-jumper:
The `freerdp_settings_new()` function, part of the FreeRDP initialization process, assumes that the `HOME` environment variable will be set and aborts initialization if it isn't, ultimately resulting in a segfault on the Guacamole side.

This change automatically sets the `HOME` variable based on the home directory retrieved from the system's user database. In case initialization still unexpectedly fails, checking has also been added around `freerdp_context_new()` which will abort the connection cleanly rather than segfault.

### DIFF
--- a/src/protocols/rdp/client.c
+++ b/src/protocols/rdp/client.c
@@ -38,10 +38,44 @@
 #include <guacamole/audio.h>
 #include <guacamole/client.h>
 
+#include <errno.h>
 #include <pthread.h>
+#include <pwd.h>
 #include <stdlib.h>
+#include <string.h>
+#include <sys/types.h>
 
 int guac_client_init(guac_client* client, int argc, char** argv) {
+
+    /* Automatically set HOME environment variable if unset (FreeRDP's
+     * initialization process will fail within freerdp_settings_new() if this
+     * is unset) */
+    const char* current_home = getenv("HOME");
+    if (current_home == NULL) {
+
+        /* Warn if the correct home directory cannot be determined */
+        struct passwd* passwd = getpwuid(getuid());
+        if (passwd == NULL)
+            guac_client_log(client, GUAC_LOG_WARNING, "FreeRDP initialization "
+                    "may fail: The \"HOME\" environment variable is unset and "
+                    "its correct value could not be automatically determined: "
+                    "%s", strerror(errno));
+
+        /* Warn if the correct home directory could be determined but can't be
+         * assigned */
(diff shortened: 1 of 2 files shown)


mike-jumper:
I don't think it would make sense as a fallback, but I agree it would make tons of sense to _always_ leverage `GUACAMOLE_HOME/freerdp/` or similar, applying the same standard logic for determining `GUACAMOLE_HOME` behind that. It would match the webapp and is essentially what we already do for SSH known hosts storage.

**But:**

FreeRDP is going to (unfortunately) attempt to create its config directory structure if it doesn't exist, even if that structure will not be used. We _can_ override the config directory location to something like `/etc/guacamole/freerdp/`, but FreeRDP will fail and abort initialization if `/etc/guacamole` does not exist or if `/etc/guacamole` cannot be written, both of which are expected configurations in our case.

The only way to prevent FreeRDP from attempting to create this structure (and, incidentally, to also bypass the `HOME` variable check and corresponding abort) is to set the `ServerMode` flag in the settings. This sounds great at first, but `ServerMode` is actually a special-case flag which causes the library to assume it's operating as an RDP server rather than an RDP client. There's no way to indicate to the library that automatic creation of config directories, checking `HOME`, etc. is inappropriate while still having a functional RDP client.

Unless a future version avoids imposing its own config directory location and structure, instead leaving that as an optional utility for downstream users that wish to use it (similar to what we do with the `LocalEnvironment` class), I think we're stuck with providing FreeRDP with a directory that it can expect to both exist and be writable.

necouchman:
So, is the suggestion that we proceed with this pull request as you have written it - to determine the `HOME` directory of the user running guacd?  Or are you going to rework it to use `GUACAMOLE_HOME`?  I'm good either way, just wanted to throw out `GUACAMOLE_HOME` for discussion...

mike-jumper:
I believe the only way to proceed is as written - determining the home directory of the user running guacd and setting `HOME` to that if it isn't already set.

I don't think it's possible to rework things to use `GUACAMOLE_HOME` in this case because of how the FreeRDP library assumes `HOME` will be set, the way it assumes the config directory will be based off `HOME`, and the way it automatically creates that structure. There would need to be changes in the FreeRDP library for us to use `GUACAMOLE_HOME` the way we would like.

mike-jumper:
Correction: for us to _safely_ use `GUACAMOLE_HOME` the way we would like.

necouchman:
Got it.  Let's merge.

PR #250, opened 2020-01-21
GUACAMOLE-934: Add support for the Play Sound PDU.

mike-jumper:
This change adds support for the [Play Sound PDU](https://docs.microsoft.com/en-us/openspecs/windows_protocols/ms-rdpbcgr/e9fb1ecb-191c-45ca-84a0-b7db6366e467), which requests that the client beep at a given frequency and for a given duration.

As written, this support limits the maximum duration of each requested beep to half a second, despite RDP technically supporting beeps up to 50 days long.

### DIFF
--- a/src/libguac/audio.c
+++ b/src/libguac/audio.c
@@ -117,6 +117,12 @@ guac_audio_stream* guac_audio_stream_alloc(guac_client* client,
     audio->client = client;
     audio->stream = guac_client_alloc_stream(client);
 
+    /* Abort allocation if underlying stream cannot be allocated */
+    if (audio->stream == NULL) {
+        free(audio);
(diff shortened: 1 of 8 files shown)


mike-jumper:
Hold on, there seems to be something off in my square wave math.

mike-jumper:
OK - my calculations were indeed wrong, suffering from rounding error in the width of the cycle of the square wave. Correcting for this, tones were correct but suffered from severe aliasing distortion at all but high sample rates, so I've switched to a triangle wave.

Things should now be correct and distortion should be minimal.

jmuehlner:
I think any users will appreciate that we don't play 50 day long beeps.

PR #251, opened 2020-01-22
GUACAMOLE-935: Free internals of rdpBitmap only when required.

mike-jumper:
This change adds an test to the `configure` script which determines whether we are building against FreeRDP 2.0.0-rc0, which had a `Bitmap_Free()` which freed bitmaps. For all newer versions of FreeRDP 2.0.0, `Bitmap_Free()` doesn't - it simply wraps the implementation-specific free handler, and it's on the implementation to know which internal parts of `rdpBitmap` were allocated by `Bitmap_Alloc()` and must now be freed.

The latter was handled in commit 8dda26a when the resulting memory leak was discovered during valgrind testing, but this now must be made conditional as this would result in a double-free against 2.0.0-rc0.

Unfortunately, the test itself is awkward:

* It is not possible to test `Bitmap_Free()` behavior in itself. We must test version numbers.
* While we can test numeric version macros, the only numeric version macros available always indicate that the version is "2.0.0".
* Version macros exist which indicate the RC number and inter-RC number, but they are strings. The C preprocessor cannot test string values.

Because of these limitations, the test can only be performed by either compiling a test program and running it (which can only be done when not cross-compiling) or by checking preprocessor output. This change ultimately goes with the latter approach.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -589,23 +589,16 @@ then
     # manually free all data associated with the rdpBitmap, even data which
     # was not allocated by the implementation.
     AC_MSG_CHECKING([whether Bitmap_Free() frees the rdpBitmap and its image data])
-    AC_RUN_IFELSE([AC_LANG_SOURCE([[
+    AC_EGREP_CPP([\"2.0.0-dev\"], [
 
         #include <freerdp/version.h>
+        FREERDP_VERSION_FULL
 
-        #include <string.h>
-
-        int main() {
-            /* Return successfully if version is 2.0.0-rc0 */
-            return strcmp(FREERDP_VERSION_FULL, "2.0.0-dev");
-        }
-
-    ]])],
+    ],
     [AC_MSG_RESULT([yes])]
     [AC_DEFINE([FREERDP_BITMAP_FREE_FREES_BITMAP],,
                [Whether Bitmap_Free() frees the rdpBitmap and its image data])],
-    [AC_MSG_RESULT([no])],
-    [AC_MSG_RESULT([assuming no (cross-compiling)])])
+    [AC_MSG_RESULT([no])])
 
 fi
 
--- a/configure.ac
+++ b/configure.ac
@@ -578,6 +578,37 @@ then
 
 fi
 
+# Variation in memory internal allocation/free behavior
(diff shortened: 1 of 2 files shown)


configure.ac
```diff
@@ -578,6 +578,30 @@ then
...
+    # associated data member within Bitmap_Free(), relying on the
+    # implementation-specific free handler to free only implementation-specific
+    # data. This changed in commit 2cf10cc, and implementations must now
+    # manually free all data associated with the rdpBitmap, even data which
+    # was not allocated by the implementation.
+    AC_MSG_CHECKING([whether Bitmap_Free() frees the rdpBitmap and its image data])
+    AC_EGREP_CPP([\"2.0.0-dev\"], [
+
+        #include <freerdp/version.h>
+        FREERDP_VERSION_FULL
```
jmuehlner: What specifically are you looking for in the header file to indicate that we're on rc1+?

Is it this `FREERDP_VERSION_FULL` string?

jmuehlner: I mean, I get that it's the whole section included here, I was just wondering why you chose this in particular.

mike-jumper: This is looking for the specific text `"2.0.0-dev"` to detect that we're on rc0.

`AC_EGREP_CPP` passes the given source through the C preprocessor (CPP) and then egrep. The C preprocessor will expand that macro to its defined value, which will match the expression `"2.0.0-dev"` if the version is 2.0.0-rc0.

mike-jumper: I guess I actually need to escape the periods in that regex. This will currently match `"200.0-dev"`, too...

mike-jumper: There.

jmuehlner: Ah, that makes sense.

jmuehlner:
I'm not sure I understand the title of this PR. From the description, it sounds like you want to free the internals for 2.0.0-rc0 and later, whereas the title says earlier. Am I missing something?

mike-jumper:
Nope, you're right - I have the title backwards.

mike-jumper:
...and one of the commit messages. I'll correct that one, too.

mike-jumper:
OK, I think that should be accurate.

PR #253, opened 2020-01-23
GUACAMOLE-625: Included Latin American keymap

dkmstr:
I have recreated the patch with a new clean guacamole-server

Hope this solves everything now.

Best Regards,


necouchman:
Ah, actually @dkmstr, one minor thing.  You titled both the PR and the commit with "GUACAMOLE 625" and not "GUACAMOLE-625" (dash instead of a space).  This seems like a minor thing, but it helps us keep commits coming in from different places consistent, so please correct both of those.

mike-jumper:
A minor inconsistency in that JIRA issue prefix has definitely bitten me pretty hard before when grepping the git logs. It's not too bad when reading commit messages individually as a human, but it makes things tricky for anything automated.

mike-jumper:
And, yes, looks good to me (except for that missing dash). ;)

dkmstr:
Hope this is fine now, sorry for the mistake :)

PR #255, opened 2020-02-06
GUACAMOLE-947: Fix the initialization issue of CLIPRDR_FORMAT_LIST object

myjimmy:
I have fixed GUACAMOLE-947 issue registered by me.

### DIFF
--- a/src/protocols/rdp/channels/cliprdr.c
+++ b/src/protocols/rdp/channels/cliprdr.c
@@ -78,6 +78,7 @@ static UINT guac_rdp_cliprdr_send_format_list(CliprdrClientContext* cliprdr) {
 
     /* We support CP-1252 and UTF-16 text */
     CLIPRDR_FORMAT_LIST format_list = {
+        .msgType = CB_FORMAT_LIST,
         .formats = (CLIPRDR_FORMAT[]) {
             { .formatId = CF_TEXT },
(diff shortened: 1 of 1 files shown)


necouchman:
@myjimmy: Very nice - thank you for the contribution!  One question from me - have you verified that this change is backward-compatible with other Windows RDP versions/servers, such that adding it does not break compatibility?

mike-jumper:
The `msgType` field of the Format List PDU _does_ need to be set to `CLIPRDR_FORMAT_LIST` according to Microsoft's documentation for CLIPRDR; the relevant message can't be identified otherwise.

The reason this wasn't done here was that the `msgType` member of FreeRDP's `CLIPRDR_FORMAT_LIST` structure was ignored by FreeRDP, with their internal handling always sending `CLIPRDR_FORMAT_LIST` regardless of what `msgType` was set to. Our implementation followed the pattern used by xfreerdp and omitted setting that property, relying instead on the known behavior of FreeRDP's functions for sending this PDU.

It looks like the above has changed within FreeRDP's git master. This warning will not occur against 2.0.0-rc4 and earlier. It will occur for the xfreerdp client, as well, which appears to not have been updated to set `msgType`.

Since it should be safe to set this property for older versions of FreeRDP (it will just be overwritten), I believe this change should be fine, though it's annoying that library function behavior would change like this.

PR #256, opened 2020-02-10
GUACAMOLE-951: Add ability to set KeyboardType for FreeRDP

mather:
KeyboardType, KeyboardSubtype, KeyboardFunctionKey is required for Japanese keyboard setting actually,
but the patch has been applied to overwrite values only for Japanese keyboards.
see: https://github.com/FreeRDP/FreeRDP/pull/707

I think it should be specified from Guacamole as FreeRDP user side.

### DIFF
--- a/src/protocols/rdp/keymap.h
+++ b/src/protocols/rdp/keymap.h
@@ -103,6 +103,24 @@ struct guac_rdp_keymap {
      */
     const UINT32 freerdp_keyboard_layout;
 
+    /**
+     * FreeRDP keyboard type definition.
+     * See https://docs.microsoft.com/ja-jp/windows/win32/api/winuser/nf-winuser-getkeyboardtype
+     */
+    const UINT32 freerdp_keyboard_type;
+
+    /**
+     * FreeRDP keyboard subtype definition.
+     * See https://docs.microsoft.com/ja-jp/windows/win32/api/winuser/nf-winuser-getkeyboardtype
+     */
+    const UINT32 freerdp_keyboard_subtype;
+
+    /**
+     * FreeRDP keyboard number of function keys.
+     * See https://docs.microsoft.com/ja-jp/windows/win32/api/winuser/nf-winuser-getkeyboardtype#remarks
+     */
+    const UINT32 freerdp_keyboard_function_key;
+
 };
 
 /**
--- a/src/protocols/rdp/keymaps/ja_jp_qwerty.keymap
+++ b/src/protocols/rdp/keymaps/ja_jp_qwerty.keymap
@@ -20,6 +20,9 @@
 parent  "base"
 name    "ja-jp-qwerty"
 freerdp "KBD_JAPANESE"
+freerdp_keyboard_type "KBD_TYPE_JAPANESE"
+freerdp_keyboard_subtype "2"
+freerdp_keyboard_function_key "12"
 
 map -shift      0x02..0x0D 0x7D ~ "1234567890-^\"
 map -shift      0x10..0x1B      ~ "qwertyuiop@["
--- a/src/protocols/rdp/settings.c
+++ b/src/protocols/rdp/settings.c
@@ -1189,6 +1189,9 @@ void guac_rdp_push_settings(guac_client* client,
     rdp_settings->DesktopHeight = guac_settings->height;
     rdp_settings->AlternateShell = guac_rdp_strdup(guac_settings->initial_program);
     rdp_settings->KeyboardLayout = guac_settings->server_layout->freerdp_keyboard_layout;
+    rdp_settings->KeyboardType = guac_settings->server_layout->freerdp_keyboard_type;
+    rdp_settings->KeyboardSubType = guac_settings->server_layout->freerdp_keyboard_subtype;
+    rdp_settings->KeyboardFunctionKey = guac_settings->server_layout->freerdp_keyboard_function_key;
 
     /* Performance flags */
     /* Explicitly set flag value */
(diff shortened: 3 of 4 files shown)


src/protocols/rdp/keymap.h
```diff
@@ -103,6 +103,24 @@ struct guac_rdp_keymap {
      */
     const UINT32 freerdp_keyboard_layout;
 
+    /**
+     * FreeRDP keyboard type definition.
+     * See https://docs.microsoft.com/ja-jp/windows/win32/api/winuser/nf-winuser-getkeyboardtype
```
mike-jumper: Please point to the English documentation link rather than the Japanese version.

src/protocols/rdp/keymaps/ja_jp_qwerty.keymap
```diff
@@ -20,6 +20,9 @@
 parent  "base"
 name    "ja-jp-qwerty"
 freerdp "KBD_JAPANESE"
+freerdp_keyboard_type "KBD_TYPE_JAPANESE"
```
mike-jumper: Is this something that should be set for the other layouts?

src/protocols/rdp/keymap.h
```diff
@@ -103,6 +103,24 @@ struct guac_rdp_keymap {
      */
     const UINT32 freerdp_keyboard_layout;
 
+    /**
+     * FreeRDP keyboard type definition.
```
mike-jumper: If the keyboard layout, type, subtype, etc. are Windows/RDP standard values and not specific to FreeRDP, perhaps we should take this opportunity to move away from `freerdp_keyboard_layout`, `freerdp_keyboard_type`, etc. here (and `freerdp` within the keymaps) and migrate to naming that isn't tied to the underlying RDP implementation.

src/protocols/rdp/settings.c
```diff
@@ -1189,6 +1189,9 @@ void guac_rdp_push_settings(guac_client* client,
     rdp_settings->DesktopHeight = guac_settings->height;
     rdp_settings->AlternateShell = guac_rdp_strdup(guac_settings->initial_program);
     rdp_settings->KeyboardLayout = guac_settings->server_layout->freerdp_keyboard_layout;
+    rdp_settings->KeyboardType = guac_settings->server_layout->freerdp_keyboard_type;
+    rdp_settings->KeyboardSubType = guac_settings->server_layout->freerdp_keyboard_subtype;
+    rdp_settings->KeyboardFunctionKey = guac_settings->server_layout->freerdp_keyboard_function_key;
```
mike-jumper: Will this cause problems for the keymap files that don't have these values defined?

src/protocols/rdp/keymaps/ja_jp_qwerty.keymap
mather: FreeRDP has default values.
https://github.com/FreeRDP/FreeRDP/blob/master/libfreerdp/core/settings.c#L345-L347
So, no need to specify if there is no problem for other layouts.

src/protocols/rdp/keymap.h
mather: It's good to migrate naming.
Should I replace these name in this PR?

src/protocols/rdp/keymap.h
mather: I will replace to English page URL.

src/protocols/rdp/settings.c
mather: Sorry, it should not be set if no value.

src/protocols/rdp/keymap.h
mike-jumper: Yes, I would say so. With three new layout-related members being added here, if `freerdp_` is not an appropriate prefix, then it shouldn't be used for the new members, and the older, existing member should be corrected to follow the new convention. The member in question is not part of a public API, and avoiding the rename would amount to technical debt.

src/protocols/rdp/settings.c
```diff
@@ -1189,9 +1189,15 @@ void guac_rdp_push_settings(guac_client* client,
     rdp_settings->DesktopHeight = guac_settings->height;
     rdp_settings->AlternateShell = guac_rdp_strdup(guac_settings->initial_program);
     rdp_settings->KeyboardLayout = guac_settings->server_layout->freerdp_keyboard_layout;
-    rdp_settings->KeyboardType = guac_settings->server_layout->freerdp_keyboard_type;
-    rdp_settings->KeyboardSubType = guac_settings->server_layout->freerdp_keyboard_subtype;
-    rdp_settings->KeyboardFunctionKey = guac_settings->server_layout->freerdp_keyboard_function_key;
+    if (guac_settings->server_layout->freerdp_keyboard_type > 0) {
```
mike-jumper: Seeing as the value being compared is a `UINT32`, testing that the value is non-zero would be more appropriate. The value can't be negative.

nwf00:
Hi, 
I´m new with Guacamole. I´m having problems with ctrl key and I use es-latam keyboard. I know there is already an es-latam leyout but I can't finde where to add this layout to my Guacamole. I use RDP.

for example I can't find this path " guacamole-server/src/protocols/rdp/keymaps/"

if someone can help me it would be great

Thanks

mike-jumper:
@nwf00, please don't post support questions here. This is a pull request (a contribution of code), not a help forum.

The place to post general support questions is the user@guacamole.apache.org mailing list:

http://guacamole.apache.org/support/#mailing-lists

necouchman:
@mather Unfortunately the last two commits added need to be cleaned up:
* The commit renaming the variables needs to be tagged with the JIRA issue.
* There should not be any merge commits - instead, the branch should be rebased on the current master.

mather:
@necouchman So, should I push rebased (with updated commit message) branch in my branch with `push -f` ?

mike-jumper:
@mather - yep, that's correct.

necouchman:
@mike-jumper Further thoughts on this one?

PR #258, opened 2020-02-22
Added Norwegian keyboard map.

oxcrag:


oxcrag:
I fixed the entry order, renamed the pull request, and the commit message for the new version of the files. Is there a way of changing the previous commit message after the fact, or will that do bad things to the pull request/version management?

necouchman:
> Is there a way of changing the previous commit message after the fact

Yes, you can do an interactive rebase, choose the option to modify the commit message, and then fix the messages.  After you're done do a force push and it should be good.

necouchman:
Also note that the correct format for both PR (which you've done properly) and commit messages is "GUACAMOLE-968: ".

oxcrag:
I'm (slowly) getting the hang of this; thanks again! I think the requested changes are in place now.

src/protocols/rdp/Makefile.am
```diff
@@ -224,6 +224,7 @@ BUILT_SOURCES =                         \
 rdp_keymaps =                                \
     $(srcdir)/keymaps/base.keymap            \
     $(srcdir)/keymaps/failsafe.keymap        \
+    $(srcdir)/keymaps/da_dk_qwerty.keymap    \
```
mike-jumper: Is this from something else?

oxcrag: No, I just figured I'd order the rest of the keymaps in the code block while I put the Norwegian one in its correct place. On the contrary I was unsure about whether to move `failsafe.keymap` to its alphabetical position and decided against it, since it could be argued to be a special case that perhaps deserves special visibility. As you see, I just moved the Danish keymap from directly below the Swedish to its more alphabetically correct position above the German one.

mike-jumper: Ah, OK. I didn't see the deletion below. I thought this was adding a keymap that didn't yet exist.

necouchman:
This also may need to be updated, now, to take into account changes from #285 and #287 as part of GUACAMOLE-518.

mike-jumper:
Yep. Caps Lock will not behave correctly without the keymap dictating Caps Lock's behavior.

necouchman:
@oxcrag This still needs to be updated to take Caps Lock changes into account.

PR #259, opened 2020-02-25
GUACAMOLE-962: Allow misbehaving RDP servers to use OpaqueRect and PatBlt.

mike-jumper:
This change requests relaxed validation of server orders if supported by FreeRDP, warning the user during the build process if their version of FreeRDP lacks this support.

Fallback support for OpaqueRect and PatBlt orders is restored through reverting commit 9855d875c794e9517567e89ad13acccd7e7e03d0, allowing RDP servers that (incorrectly) assume support for those orders to generally function without introducing visual artifacts.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -722,6 +722,21 @@ then
 
 fi
 
+# Support for receiving unannounced orders from the RDP server
+if test "x${have_freerdp2}" = "xyes"
+then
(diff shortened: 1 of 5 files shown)


necouchman:
I'll get the merge back to master done, soon, too...

PR #260, opened 2020-02-25
GUACAMOLE-818: Break SFTP directory JSON at blob boundaries. Do not skip entries.

mike-jumper:
The intent of the previous version of the SFTP directory listing code was to break the JSON transfer at blob boundaries, waiting for an ack before sending the next blob, however the ordering of the `blob_written` and directory read checks could result in a directory entry being skipped at the boundary of each blob.

The proper order would be to check the `blob_written` flag first, however the `blob_written` flag is unnecessary. It's simpler and more correct to just break out of the loop once the desired blob has been flushed.


necouchman:
Will get this into master with the other one.

PR #266, opened 2020-03-09
GUACAMOLE-982: Fix the error log message when the creating thread of RDP client fails

myjimmy:
I have fixed the GUACAMOLE-982 issue.


necouchman:
@mike-jumper: Thoughts on including this in 1.2.0?

mike-jumper:
My thoughts are that we should never allow the release scope to creep after it's been settled, except for the special-case exceptions we decided on for 1.2.0 due to its timing (regressions related to the massive FreeRDP 2.0.0 / Apache Directory API migrations of 1.1.0).

It's is a simple change, but scope creep is scope creep. This particular copypasta has been there since 0.9.10-incubating (2016) and can easily wait until 1.3.0.

necouchman:
Sounds good.

PR #267, opened 2020-03-18
GUACAMOLE-513: Implement Wake-on-LAN support in libguac

necouchman:
Okay, rough start, here, I'm sure, but hopefully not too bad :-).  I used a couple of examples I found online to guide me in writing my code, so hopefully it makes sense.  Let me know what needs to be fixed.

### DIFF
--- a/src/libguac/Makefile.am
+++ b/src/libguac/Makefile.am
@@ -69,8 +69,8 @@ libguacinc_HEADERS =                  \
     guacamole/user.h                  \
     guacamole/user-constants.h        \
     guacamole/user-fntypes.h          \
-    guacamole/user-types.h	      \
-    guacamole/wol.h		      \
+    guacamole/user-types.h            \
+    guacamole/wol.h                   \
     guacamole/wol-constants.h
 
 noinst_HEADERS =      \
--- a/src/libguac/Makefile.am
+++ b/src/libguac/Makefile.am
@@ -69,7 +69,9 @@ libguacinc_HEADERS =                  \
     guacamole/user.h                  \
     guacamole/user-constants.h        \
     guacamole/user-fntypes.h          \
-    guacamole/user-types.h
+    guacamole/user-types.h	      \
+    guacamole/wol.h		      \
+    guacamole/wol-constants.h
 
 noinst_HEADERS =      \
     id.h              \
@@ -104,7 +106,8 @@ libguac_la_SOURCES =   \
     user.c             \
     user-handlers.c    \
     user-handshake.c   \
-    wait-fd.c
+    wait-fd.c	       \
+    wol.c
 
 # Compile WebP support if available
 if ENABLE_WEBP
--- a/src/libguac/guacamole/wol-constants.h
+++ b/src/libguac/guacamole/wol-constants.h
@@ -40,5 +40,12 @@
  */
 #define GUAC_WOL_DEFAULT_WAIT_TIME 60
 
+/**
+ * The size of the magic Wake-on-LAN packet to send to wake a remote host.  This
+ * consists of 6 bytes of 0xFF, and then the MAC address repeated 16 times.
+ * https://en.wikipedia.org/wiki/Wake-on-LAN#Magic_packet
+ */
+#define GUAC_WOL_PACKET_SIZE 102
+
 #endif /* GUAC_WOL_CONSTANTS_H */
 
--- a/src/libguac/guacamole/wol-constants.h
+++ b/src/libguac/guacamole/wol-constants.h
@@ -33,12 +33,6 @@
 #define GUAC_WOL_LOCAL_IPV4_BROADCAST "255.255.255.255"
 
 /**
- * The number of seconds to wait after sending the Wake-on-LAN packet
- * for the destination host to start responding.
- */
-#define GUAC_WOL_BOOT_WAIT_TIME 60
-
-/**
  * The size of the magic Wake-on-LAN packet to send to wake a remote host.  This
  * consists of 6 bytes of 0xFF, and then the MAC address repeated 16 times.
  * https://en.wikipedia.org/wiki/Wake-on-LAN#Magic_packet
--- a/src/libguac/guacamole/wol-constants.h
+++ b/src/libguac/guacamole/wol-constants.h
@@ -17,8 +17,8 @@
  * under the License.
  */
 
-#ifndef __GUAC_WOL_CONSTANTS_H
-#define __GUAC_WOL_CONSTANTS_H
+#ifndef GUAC_WOL_CONSTANTS_H
+#define GUAC_WOL_CONSTANTS_H
 
 /**
  * Header file that provides constants and defaults related to libguac
@@ -29,7 +29,8 @@
 
 /**
  * The default broadcast address to which Wake-on-LAN packets will be sent
- * if one is not specified.
+ * if one is not specified, which is the special value for the IPv4 local
+ * network broadcast.
  */
 #define GUAC_WOL_DEFAULT_BROADCAST "255.255.255.255"
 
@@ -39,5 +40,5 @@
  */
 #define GUAC_WOL_DEFAULT_WAIT_TIME 60
 
-#endif /* __GUAC_WOL_CONSTANTS_H */
+#endif /* GUAC_WOL_CONSTANTS_H */
 
--- a/src/libguac/guacamole/wol-constants.h
+++ b/src/libguac/guacamole/wol-constants.h
@@ -47,5 +47,13 @@
  */
 #define GUAC_WOL_PACKET_SIZE 102
 
+/**
+ * The port number that the magic packet should contain as the destination.  In
+ * reality this doesn't matter all that much, since the packet is not usually
+ * processed by a full IP stack, but it is considered a pretty standard
+ * practice.
+ */
+#define GUAC_WOL_PORT 9
+
 #endif /* GUAC_WOL_CONSTANTS_H */
 
--- a/src/libguac/guacamole/wol-constants.h
+++ b/src/libguac/guacamole/wol-constants.h
@@ -1,0 +1,43 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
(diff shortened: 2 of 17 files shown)


src/libguac/guacamole/wol-constants.h
```diff
@@ -0,0 +1,43 @@
...
+ * Wake-on-LAN support.
+ * 
+ * @file wol-constants.h
+ */
+
+/**
+ * The default broadcast address to which Wake-on-LAN packets will be sent
+ * if one is not specified.
+ */
+#define GUAC_WOL_DEFAULT_BROADCAST "255.255.255.255"
```
mike-jumper: Unless there is logic within libguac that uses this value automatically, this doesn't feel like a "default" here (though implementations may well use it as such). According to [RFC 919](https://tools.ietf.org/html/rfc919#section-7), 255.255.255.255 is a special broadcast address for strictly the local network. Perhaps `GUAC_WOL_LOCAL_BROADCAST`, `GUAC_WOL_LOCAL_IPV4_BROADCAST`, or something similar would be better?

Alternatively, providing default logic and using this value if `NULL` is given might make sense. If so, and this remains `GUAC_WOL_DEFAULT_BROADCAST`, I recommend documenting the fact that this is "255.255.255.255", the IPv4 broadcast address for the local network.

src/libguac/guacamole/wol-constants.h
```diff
@@ -0,0 +1,43 @@
...
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+#ifndef __GUAC_WOL_CONSTANTS_H
+#define __GUAC_WOL_CONSTANTS_H
```
mike-jumper: Though we (incorrectly) have used a double-underscore prefix for this in the past, doing so is technically not good and intrudes on a reserved namespace.

This should just be `GUAC_WOL_CONSTANTS_H`.

src/libguac/wol.c
```diff
@@ -0,0 +1,128 @@
...
+    
+    /* Attempt to set broadcast; exit with error if this fails. */
+    if (setsockopt(wol_socket, SOL_SOCKET, SO_BROADCAST, &wol_bcast, sizeof(wol_bcast)) < 0) {
+        return 0;
+    }
+    
+    /* Set server endpoint to the broadcast address. */
+    wol_dest.sin_family = AF_INET;
+    wol_dest.sin_port = htons(9);
+    wol_dest.sin_addr.s_addr = inet_addr(broadcast_addr);
```
mike-jumper: What if the address is invalid? Will this work with IPv6?

necouchman:
Okay, identified one issue, already - if the wait time is set for too large amount of time, the connection closes down.  Not sure what I need to do to really make it wait for the amount of time and not terminate the connection?  Also, might be good to send a custom message back to the client that it can display during this time, if that's possible?

src/libguac/wol.c
```diff
@@ -0,0 +1,128 @@
...
+    
+    /* Generate the magic packet. */
+    __guac_wol_create_magic_packet(wol_packet, dest_mac);
+    
+    /* Send the packet and record bytes sent. */
+    int bytes_sent = __guac_wol_send_packet(broadcast_addr, wol_packet);
+    
+    /* Sleep for the specified amount of time to wait for the remote
+     host to boot. */
+    sleep(wait_time);
```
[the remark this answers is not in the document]
necouchman: I'm happy to move the sleep to the caller at the protocol level - I have no argument with that.  i was trying to avoid as much code duplication as possible, but that's fine with me.  I'll move it out to the individual protocols and use the function you mentioned, which will probably also take care of the other issue I posted above with the connection timing out while sleeping.  Thanks!

src/libguac/wol.c
```diff
@@ -0,0 +1,128 @@
...
+    /* Send the packet and return number of bytes sent. */
+    return sendto(wol_socket, &packet, sizeof(unsigned char) * 102, 0,
+            (struct sockaddr*) &wol_dest, sizeof(wol_dest));
+ 
+}
+
+int guac_wol_wake(const char* mac_addr, const char* broadcast_addr,
+        const int wait_time) {
+    
+    unsigned char wol_packet[102];
```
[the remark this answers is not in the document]
necouchman: Probably should be a constant, eh?  This is the size of the "magic packet" - https://en.wikipedia.org/wiki/Wake-on-LAN#Magic_packet.  Six bytes of 255, followed by the MAC address repeated 16 times.  Which, come to think of it, I may not have done correctly -  I'll need to double-check that...

src/libguac/guacamole/wol-constants.h
necouchman: The logic that uses this by default is in the protocol-specific parts where the setting is pulled in - if there's nothing specified for the broadcast address parameter, this is used.  I think it should remain the default, though I'm happy to name it whatever.  I will update the documentation to reflect that it is the local broadcast address.

src/libguac/guacamole/wol-constants.h
necouchman: ok

src/libguac/wol.c
necouchman: No, it will not work with IPv6 as it is written.  I should probably put some logic in to detect what is provided and choose the correct `AF_INET`/`AF_INET6` option.

If the address is invalid then the subsequent call to `sendto()` will fail.  But worth some more robust error checking...I'll beef it up.

necouchman:
Thanks for the quick review, @mike-jumper  - I'll try to get some of these items addressed quickly.

src/libguac/wol.c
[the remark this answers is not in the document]
necouchman: Okay, size is now a constant.

src/libguac/guacamole/wol-constants.h
necouchman: Fixed.

src/libguac/wol.c
necouchman: Okay, this may take me a little longer to figure out how to do - the `inet_addr()` function appears to be a little outdated and only supports IPv4 (according to the Linux manpage, anyway), and several of the other ones appear to be ones that are not recommended for us, so I've got to figure out a good way to detect IPv4 vs. IPv6 and parse out into the arguments that the `socket` stuff expects.

Also, for what it's worth, IPv6 doesn't support broadcast, only link-local multicast, so I feel like we're going to end up with a couple of different functions for this - one for IPv4 and one for IPv6.  Definitely a little more work to do, here...

necouchman: Okay, taken a stab at IPv4 and IPv6 support concurrently, so see if that looks okay.

necouchman:
Okay, I think I've cleaned up all of those initial items.  What else?!

src/libguac/guacamole/wol-constants.h
mike-jumper: The reason I have issue with "default" here is that we're essentially _recommending_ that behavior be implemented as a default, not actually implementing a default. It feels like libguac should not concern itself with what protocol implementations may choose to be their defaults.

If some aspect of this WoL implementation implemented its own defaults in the case that specific values are omitted (if `NULL` is passed for the address, a negative value for the timeout, etc.), then I completely agree with documenting these as defaults within libguac since that's exactly what they would be - libguac's defaults. Outside of that, I think the aspect of these values being defaults (or not) would be a private concern of the implementation using libguac.

If the main goal with these defaults is to avoid duplicated code within the bundled protocol support, perhaps some sort of private `defaults.h` within `src/common/` would be better?

src/libguac/Makefile.am
```diff
@@ -69,7 +69,9 @@ libguacinc_HEADERS =                  \
     guacamole/user.h                  \
     guacamole/user-constants.h        \
     guacamole/user-fntypes.h          \
-    guacamole/user-types.h
+    guacamole/user-types.h	      \
+    guacamole/wol.h		      \
+    guacamole/wol-constants.h
```
mike-jumper, line 74: I recommend aligning the `\` with the lines above.

src/libguac/guacamole/wol-constants.h
necouchman: Okay, I see your point.  I've renamed them, at this point, such that they are not defaults, anymore.  The boot time one is still a little bit of a gray area - I can move that constant definition out to the individual protocols, if you prefer?

src/libguac/Makefile.am
necouchman: Ahem.  Yes.

src/libguac/wol.c
```diff
@@ -0,0 +1,144 @@
...
+        int hops = 1;
+        
+        if (setsockopt(wol_socket, IPPROTO_IPV6, IPV6_MULTICAST_HOPS, &hops,
+                sizeof(hops)))
+            return 0;
+    }
+    
+    /* Send the packet and return number of bytes sent. */
+    return sendto(wol_socket, packet, GUAC_WOL_PACKET_SIZE, 0,
+            (struct sockaddr*) &wol_dest, sizeof(wol_dest));
```
mike-jumper, line 117: Whether successful or not, does `wol_socket` need to be closed before it goes out of scope to avoid leaking a descriptor?

src/libguac/guacamole/wol-constants.h
```diff
@@ -0,0 +1,57 @@
...
+/**
+ * The value for the local IPv4 broadcast address.
+ */
+#define GUAC_WOL_LOCAL_IPV4_BROADCAST "255.255.255.255"
+
+/**
+ * The number of seconds to wait after sending the Wake-on-LAN packet
+ * for the destination host to start responding.
+ */
+#define GUAC_WOL_BOOT_WAIT_TIME 60
```
mike-jumper, line 39: I'm still not sure about this.

The other constants make total sense; `GUAC_WOL_LOCAL_IPV4_BROADCAST` is exactly that, as are `GUAC_WOL_PACKET_SIZE` and `GUAC_WOL_PORT`. Each is a standard value which any WoL implementation should reasonably provide as a reusable constant.

`GUAC_WOL_BOOT_WAIT_TIME` isn't a standard value nor a standard part of WoL. My understanding from reading the code thus far is that this is just a constant which we are using internally to avoid having to define such a constant within each protocol. I agree that we should aim to be DRY and that defining such a constant somewhere makes sense ... but I don't think the public API of libguac is the place for it. I think it's outside the scope of concerns for the core, public library of guacamole-server.

There is a place where we have been placing internally-reusable code that isn't part of the public API (`src/common/`). If it goes anywhere, I think that would be the place for it, even it it's just something like `src/common/common/defaults.h`. Such a header might even serve as a place to receive other default values that have otherwise been duplicated across each protocol implementation but _could_ be common.

If you think this constant does belong in the public libguac API, I think I can be swayed, but I'd like to read through the reasoning. As it stands, I currently can't see it as something within scope.

src/libguac/wol.c
necouchman: Okay, this should be good, now.

src/libguac/guacamole/wol-constants.h
necouchman: Makes sense.  I've created the `common/defaults.h` file and added it, there, instead.

PR #271, opened 2020-04-23
GUACAMOLE-1047: Notify connecting client on unrecognized connection ID

holograph:
The can easily be tested by sending a well-formed `select` command using a `$`-prefixed, but fake, connection ID.

Before patch:
```
tomer@druuge:~/dev/guacamole-server$ echo '6.select,10.$123456789;' | nc localhost 4822
tomer@druuge:~/dev/guacamole-server$ 
```

Note complete lack of response above, the server simply drops the connection. Contrast with the same command after the patch:

```
tomer@druuge:~/dev/guacamole-server$ echo '6.select,10.$123456789;' | nc localhost 4822
5.error,38.Connection "$123456789" does not exist,3.768;
tomer@druuge:~/dev/guacamole-server$ 
```

### DIFF
--- a/src/guacd/connection.c
+++ b/src/guacd/connection.c
@@ -278,10 +278,18 @@ static int guacd_route_connection(guacd_proc_map* map, guac_socket* socket) {
         proc = guacd_proc_map_retrieve(map, identifier);
         new_process = 0;
 
-        /* Warn if requested connection does not exist */
-        if (proc == NULL)
-            guacd_log(GUAC_LOG_INFO, "Connection \"%s\" does not exist.",
-                    identifier);
+        /* Warn and ward off client if requested connection does not exist */
+        if (proc == NULL) {
+            char message[2048];
+
+            snprintf(message, sizeof(message),
+                    "Connection \"%s\" does not exist", identifier);
+
+            guacd_log(GUAC_LOG_INFO, message);
+            guac_protocol_send_error(socket, message,
+                    GUAC_PROTOCOL_STATUS_CLIENT_BAD_REQUEST);
+        }
+
         else
             guacd_log(GUAC_LOG_INFO, "Joining existing connection \"%s\"",
                     identifier);
--- a/src/guacd/connection.c
+++ b/src/guacd/connection.c
@@ -287,7 +287,7 @@ static int guacd_route_connection(guacd_proc_map* map, guac_socket* socket) {
 
             guacd_log(GUAC_LOG_INFO, message);
             guac_protocol_send_error(socket, message,
-                    GUAC_PROTOCOL_STATUS_CLIENT_BAD_REQUEST);
+                    GUAC_PROTOCOL_STATUS_RESOURCE_NOT_FOUND);
         }
 
         else


src/guacd/connection.c
```diff
@@ -278,10 +278,18 @@ static int guacd_route_connection(guacd_proc_map* map, guac_socket* socket) {
...
+        /* Warn and ward off client if requested connection does not exist */
+        if (proc == NULL) {
+            char message[2048];
+
+            snprintf(message, sizeof(message),
+                    "Connection \"%s\" does not exist", identifier);
+
+            guacd_log(GUAC_LOG_INFO, message);
+            guac_protocol_send_error(socket, message,
+                    GUAC_PROTOCOL_STATUS_CLIENT_BAD_REQUEST);
```
necouchman, line 290: In general I think this is probably a good thing to do, but, have you tested the mainstream Guacamole Client with these changes?  Presumably you're developing a custom application that interacts directly with guacd, but we need to make sure these changes are compatible with the upstream client.

Also, I haven't looked, yet, but I'm guessing there may need to be some matching changes done within the guacamole-common-js code on the guacamole-client side of things to make sure errors are handled when send back by guacd?  Or maybe this is already in place?

necouchman: Ah, I see that you just put in a JIRA issue for the guacamole-common side :-)

holograph: Yes, and yes:

* The current guacd with the current mainstream (Java) client will simply error out due to socket read error;
* Once this is merged, the existing Java client will error out due to "unexpected opcode", which amounts to the same thing;
* Once the fix for GUACAMOLE-1048 is hopefully merged, the client will actually provide an honest-to-god error message :-)

src/guacd/connection.c
```diff
@@ -278,10 +278,18 @@ static int guacd_route_connection(guacd_proc_map* map, guac_socket* socket) {
...
-            guacd_log(GUAC_LOG_INFO, "Connection \"%s\" does not exist.",
-                    identifier);
+        /* Warn and ward off client if requested connection does not exist */
+        if (proc == NULL) {
+            char message[2048];
+
+            snprintf(message, sizeof(message),
+                    "Connection \"%s\" does not exist", identifier);
+
+            guacd_log(GUAC_LOG_INFO, message);
```
mike-jumper, line 288: `guacd_log()` will interpret `message` here as a printf-style format string:

https://github.com/apache/guacamole-server/blob/ca1fbd5e980392dc91b52ddff473698035d8f357/src/guacd/log.h#L45-L49

To do this safely, the call would need to be:

```c
guacd_log(GUAC_LOG_INFO, "%s", message);
```

However, given that the connection is failing, I think `GUAC_LOG_ERROR` would be more appropriate.

src/guacd/connection.c
mike-jumper, line 290: I think `GUAC_PROTOCOL_STATUS_RESOURCE_NOT_FOUND` would be a better choice.

src/guacd/connection.c
```diff
@@ -278,10 +278,18 @@ static int guacd_route_connection(guacd_proc_map* map, guac_socket* socket) {
         proc = guacd_proc_map_retrieve(map, identifier);
         new_process = 0;
 
-        /* Warn if requested connection does not exist */
-        if (proc == NULL)
-            guacd_log(GUAC_LOG_INFO, "Connection \"%s\" does not exist.",
-                    identifier);
+        /* Warn and ward off client if requested connection does not exist */
```
mike-jumper, line 281: What about the case where a particular protocol is requested (for a new connection) and that protocol is not defined?

mike-jumper (changes requested):

src/guacd/connection.c
holograph: Oh, I missed that -- so I can basically safely log a format string and skip constructing the message entirely? i.e.
```guacd_log(GUAC_LOG_INFO, "Connection \"%s\" does not exist", identifier);```

As for the level, this isn't a runtime error so much as a client error though, akin to HTTP 400 "bad request" or 404 "not found", which to me sounds more like an informative message if any; If you still think it's appropriate to log this an error, however, please ping me and I'll adjust the PR.

src/guacd/connection.c
holograph: This isn't handled here - this code path is specifically for existing connections (i.e. selecting $<connection_id>)

src/guacd/connection.c
mike-jumper: > Oh, I missed that -- so I can basically safely log a format string and skip constructing the message entirely? i.e. `guacd_log(GUAC_LOG_INFO, "Connection \"%s\" does not exist", identifier);`

Yep!

> As for the level, this isn't a runtime error so much as a client error though, akin to HTTP 400 "bad request" or 404 "not found", which to me sounds more like an informative message if any; ...

Good point. I agree it makes sense for this to remain at the info level, with any further handling of the client error naturally up to the client.

holograph: Woah, it's been a while. I am no longer with my previous employer and have no real-world way of testing this (nor even a toolchain to build it with...). Hopefully it's still somewhat useful as-is.

mike-jumper: Ah - right. I now remember something about that from your other PR...

I'll build off what you've done here and get this closed out.

mike-jumper:
Alrighty - superseded by #355.

PR #273, opened 2020-04-30
GUACAMOLE-1053: guacd segfaults if user actively presses keys during RDP disconnect

trengri:
Added a read/write lock to the **guac_rdp_client** structure, and acquire/release the lock in input handlers and RDP client thread.
I was not able to reproduce a segmentation fault with this patch applied.

### DIFF
--- a/src/protocols/rdp/client.c
+++ b/src/protocols/rdp/client.c
@@ -155,6 +155,9 @@ int guac_client_init(guac_client* client, int argc, char** argv) {
     pthread_mutexattr_settype(&(rdp_client->attributes),
             PTHREAD_MUTEX_RECURSIVE);
 
+    /* Initalize the lock */
+    pthread_rwlock_init(&(rdp_client->lock), NULL);
+
     /* Set handlers */
     client->join_handler = guac_rdp_user_join_handler;
     client->free_handler = guac_rdp_client_free_handler;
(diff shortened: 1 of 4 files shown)


mike-jumper:
Thanks, @trengri.

Since [GUACAMOLE-1053](https://issues.apache.org/jira/browse/GUACAMOLE-1053) has been pulled into scope of 1.2.0, I've updated the merge base of this PR to `staging/1.2.0`, but doing so has naturally pulled in a bunch of commits from `master`.

When you can, please rebase on `staging/1.2.0`.

trengri:
Rebased to staging/1.2.0

trengri:
Rebased to resolve conflicts after #267

src/protocols/rdp/client.c
```diff
@@ -155,6 +157,12 @@ int guac_client_init(guac_client* client, int argc, char** argv) {
     pthread_mutexattr_settype(&(rdp_client->attributes),
             PTHREAD_MUTEX_RECURSIVE);
 
+    /* Initalize the lock */
+    pthread_rwlockattr_init(&lock_attributes);
+    pthread_rwlockattr_setpshared(&lock_attributes, PTHREAD_PROCESS_SHARED);
```
mike-jumper, line 162: Does this lock need `PTHREAD_PROCESS_SHARED`? I don't believe instances of this lock will be shared across different processes, just among threads of the same process.

trengri: You are right. Removed lock attributes altogether.

PR #274, opened 2020-05-04
GUACAMOLE-1059: Use FreeRDP function for verifying Stream length

necouchman:
This implements the use of the `Stream_GetRemainingLength()` function prior to reading RDP Streams throughout the code.

There is one place that I need a little help with - will make a note at the correct place in the pull request...

### DIFF
--- a/src/protocols/rdp/channels/rdpdr/rdpdr-fs-messages-file-info.c
+++ b/src/protocols/rdp/channels/rdpdr/rdpdr-fs-messages-file-info.c
@@ -135,6 +135,10 @@ void guac_rdpdr_fs_process_set_rename_info(guac_rdp_common_svc* svc,
     wStream* output_stream;
     char destination_path[GUAC_RDP_FS_MAX_PATH];
 
+    /* Check stream size prior to reading. */
+    if (Stream_GetRemainingLength(input_stream) < 6)
+        return;
+    
     /* Read structure */
     Stream_Seek_UINT8(input_stream); /* ReplaceIfExists */
     Stream_Seek_UINT8(input_stream); /* RootDirectory */
--- a/src/protocols/rdp/channels/rdpdr/rdpdr-fs-messages-file-info.c
+++ b/src/protocols/rdp/channels/rdpdr/rdpdr-fs-messages-file-info.c
@@ -136,9 +136,13 @@ void guac_rdpdr_fs_process_set_rename_info(guac_rdp_common_svc* svc,
     char destination_path[GUAC_RDP_FS_MAX_PATH];
 
     /* Check stream size prior to reading. */
-    if (Stream_GetRemainingLength(input_stream) < 6)
+    if (Stream_GetRemainingLength(input_stream) < 6) {
+        guac_client_log(svc->client, GUAC_LOG_WARNING, "File Stream does not "
+                "contain the required number of bytes.  File sharing may not "
+                "work as expected.");
         return;
-    
+    }
+        
     /* Read structure */
     Stream_Seek_UINT8(input_stream); /* ReplaceIfExists */
     Stream_Seek_UINT8(input_stream); /* RootDirectory */
--- a/src/protocols/rdp/channels/rdpdr/rdpdr-fs-messages-file-info.c
+++ b/src/protocols/rdp/channels/rdpdr/rdpdr-fs-messages-file-info.c
@@ -137,9 +137,10 @@ void guac_rdpdr_fs_process_set_rename_info(guac_rdp_common_svc* svc,
 
     /* Check stream size prior to reading. */
     if (Stream_GetRemainingLength(input_stream) < 6) {
-        guac_client_log(svc->client, GUAC_LOG_WARNING, "File Stream does not "
-                "contain the required number of bytes.  File sharing may not "
-                "work as expected.");
+        guac_client_log(svc->client, GUAC_LOG_WARNING, "Server Drive Set "
+                "Information Request (FileRenameInformation) PDU does not "
+                "contain the expected number of bytes.  File redirection "
+                "may not work as expected.");
         return;
     }
         
@@ -148,6 +149,14 @@ void guac_rdpdr_fs_process_set_rename_info(guac_rdp_common_svc* svc,
     Stream_Seek_UINT8(input_stream); /* RootDirectory */
     Stream_Read_UINT32(input_stream, filename_length); /* FileNameLength */
 
+    if (Stream_GetRemainingLength(input_stream) < filename_length) {
+        guac_client_log(svc->client, GUAC_LOG_WARNING, "Server Drive Set "
+                "Information Request (FileRenameInformation) PDU does not "
+                "contain the expected number of bytes.  File redirection "
+                "may not work as expected.");
+        return;
+    }
+    
     /* Convert name to UTF-8 */
     guac_rdp_utf16_to_utf8(Stream_Pointer(input_stream), filename_length/2,
             destination_path, sizeof(destination_path));
@@ -200,6 +209,15 @@ void guac_rdpdr_fs_process_set_allocation_info(guac_rdp_common_svc* svc,
     UINT64 size;
     wStream* output_stream;
 
+    /* Check to make sure the stream has at least 8 bytes (UINT64) */
+    if (Stream_GetRemainingLength(input_stream) < 8) {
+        guac_client_log(svc->client, GUAC_LOG_WARNING, "Server Drive Set "
+                "Information Request (FileAllocationInformation) PDU does not "
+                "contain the expected number of bytes.  File redirection "
+                "may not work as expected.");
+        return;
+    }
+    
     /* Read new size */
     Stream_Read_UINT64(input_stream, size); /* AllocationSize */
 
@@ -252,6 +270,15 @@ void guac_rdpdr_fs_process_set_end_of_file_info(guac_rdp_common_svc* svc,
     UINT64 size;
     wStream* output_stream;
 
+    /* Check to make sure stream contains at least 8 bytes (UINT64) */
+    if (Stream_GetRemainingLength(input_stream) < 8) {
+        guac_client_log(svc->client, GUAC_LOG_WARNING, "Server Drive Set "
+                "Information Request (FileEndOfFileInformation) PDU does not "
+                "contain the expected number of bytes.  File redirection "
+                "may not work as expected.");
+        return;
+    }
+    
     /* Read new size */
     Stream_Read_UINT64(input_stream, size); /* AllocationSize */
 
--- a/src/protocols/rdp/channels/rdpdr/rdpdr-fs-messages.c
+++ b/src/protocols/rdp/channels/rdpdr/rdpdr-fs-messages.c
@@ -49,8 +49,12 @@ void guac_rdpdr_fs_process_create(guac_rdp_common_svc* svc,
     char path[GUAC_RDP_FS_MAX_PATH];
 
     /* Check remaining stream data prior to reading. */
-    if (Stream_GetRemainingLength(input_stream) < 32)
+    if (Stream_GetRemainingLength(input_stream) < 32) {
+        guac_client_log(svc->client, GUAC_LOG_WARNING, "File stream does not "
+                "contain the expected number of bytes. File sharing may not "
+                "work as expected.");
         return;
+    }
     
     /* Read "create" information */
     Stream_Read_UINT32(input_stream, desired_access);
@@ -128,8 +132,12 @@ void guac_rdpdr_fs_process_read(guac_rdp_common_svc* svc,
     wStream* output_stream;
 
     /* Check remaining bytes before reading stream. */
-    if (Stream_GetRemainingLength(input_stream) < 12)
+    if (Stream_GetRemainingLength(input_stream) < 12) {
+        guac_client_log(svc->client, GUAC_LOG_WARNING, "File Stream does not "
+                "contain the expected number of bytes. File sharing may not "
+                "work as expected.");
         return;
+    }
     
     /* Read packet */
     Stream_Read_UINT32(input_stream, length);
@@ -181,8 +189,12 @@ void guac_rdpdr_fs_process_write(guac_rdp_common_svc* svc,
     wStream* output_stream;
 
     /* Check remaining length. */
-    if (Stream_GetRemainingLength(input_stream) < 32)
+    if (Stream_GetRemainingLength(input_stream) < 32) {
+        guac_client_log(svc->client, GUAC_LOG_WARNING, "File Stream does not "
+                "contain the expected number of bytes. File sharing may not "
+                "work as expected.");
         return;
+    }
     
     /* Read packet */
     Stream_Read_UINT32(input_stream, length);
@@ -257,8 +269,12 @@ void guac_rdpdr_fs_process_volume_info(guac_rdp_common_svc* svc,
     int fs_information_class;
 
     /* Check remaining length */
-    if (Stream_GetRemainingLength(input_stream) < 4)
+    if (Stream_GetRemainingLength(input_stream) < 4) {
+        guac_client_log(svc->client, GUAC_LOG_WARNING, "File Stream does not "
+                "contain the expected number of bytes. File sharing may not "
+                "work as expected.");
         return;
+    }
     
     Stream_Read_UINT32(input_stream, fs_information_class);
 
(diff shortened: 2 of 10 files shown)


src/protocols/rdp/channels/rdpsnd/rdpsnd.c
```diff
@@ -35,11 +35,23 @@ void guac_rdpsnd_process_receive(guac_rdp_common_svc* svc,
...
     Stream_Read_UINT16(input_stream, header.body_size);
-
+    
+    if (Stream_GetRemainingLength(input_stream) < header.body_size) {
+        guac_client_log(svc->client, GUAC_LOG_DEBUG, "Not enough bytes in stream."
+                "  Remaining: %d, Body size: %d",
+                Stream_GetRemainingLength(input_stream),
+                header.body_size);
+        return;
+    }
```
necouchman, line 53: This area here is causing me a little trouble - for whatever reason, if I do this check, sound doesn't work.  Maybe I'm doing some pointer access incorrectly or something??

trengri: I'm not a RDP expert. It's just a guess, based on reading the comments in the source code.
It looks like some PDUs like do not have headers, so there is little sense in inspecting header.body_size for them...
Maybe you need to move this check a few lines lower (i.e. put it after the following piece of code)?
https://github.com/apache/guacamole-server/blob/9c37fc56171d55b8fa838949d7fd40db85506857/src/protocols/rdp/channels/rdpsnd/rdpsnd.c#L43-L50

necouchman: Thanks for the hint @trengri - I will take a look!

necouchman: Actually, looking further at the code, it looks like each of the functions that are called in the switch statement already do the Stream size verification prior to reading, so maybe there's no need to do it, here??
* `guac_rdpsnd_wave_handler()` is doing Stream write operations, not reads.
* `guac_rdpsnd_formats_handler`, `guac_rdpsnd_training_handler`, and `guac_rdpsnd_wave_info_handler` already do the check.
* `guac_rdpsnd_close_handler` doesn't do any reads (or anything else).

So, seems like checking within this function may not be necessary?

necouchman: So, went ahead and moved it down below that first check and it seems to be working fine.  Still not sure it's strictly necessary since we're checking it in the other functions, but works for now!

src/protocols/rdp/channels/rdpdr/rdpdr-fs-messages-file-info.c
```diff
@@ -135,6 +135,10 @@ void guac_rdpdr_fs_process_set_rename_info(guac_rdp_common_svc* svc,
     wStream* output_stream;
     char destination_path[GUAC_RDP_FS_MAX_PATH];
 
+    /* Check stream size prior to reading. */
+    if (Stream_GetRemainingLength(input_stream) < 6)
+        return;
```
mike-jumper, line 140: Is it worth logging a warning when a received PDU is unexpectedly small? Should make it easier to trace down the cause of unexpected behavior due to RDP protocol violations.

src/protocols/rdp/channels/rdpdr/rdpdr-fs-messages.c
```diff
@@ -406,6 +430,9 @@ void guac_rdpdr_fs_process_query_directory(guac_rdp_common_svc* svc,
     if (file == NULL)
         return;
 
+    if (Stream_GetRemainingLength(input_stream) < 9)
```
mike-jumper, line 433: I recommend adding a comment noting the reason for doing this (that you're verifying the length of the PDU for future reads, and that the main header is 9 bytes). The check is otherwise initially mysterious and magic-numbery until the mental gears turn further. A comment noting that this number must line up with future reads would also serve to call attention to this section if any changes are made to those reads.

src/protocols/rdp/channels/rdpdr/rdpdr-fs-messages-file-info.c
necouchman: Yeah, I can add these in.

src/protocols/rdp/channels/rdpdr/rdpdr-fs-messages.c
necouchman: Yeah, that makes sense.

src/protocols/rdp/channels/rdpsnd/rdpsnd.c
```diff
@@ -48,6 +52,10 @@ void guac_rdpsnd_process_receive(guac_rdp_common_svc* svc,
         guac_rdpsnd_wave_handler(svc, input_stream, &header);
         return;
     }
+    
+    /* Check body size */
+    if (Stream_GetRemainingLength(input_stream) < header.body_size)
```
mike-jumper, line 57: RDPSND is a bit convoluted in the way it handles each segment of sound data. I think the reason you're seeing this fail is that audio data is split across two PDUs:

1. "WaveInfo" (with type `SNDC_WAVE`).
2. "Wave" (the PDU immediately following a "WaveInfo" PDU)

In the case of "WaveInfo", the body size is not the size of that PDU, but the size of that PDU _plus_ the next:

> The BodySize field of the RDPSND PDU Header is the size of the WaveInfo PDU plus the size of the data field of the Wave PDU that immediately follows this packet minus the size of the Header.

mike-jumper: Given the above, my thinking would be that it's not correct to check against `header.body_size` here, but that each logical attempt to read/seek through a block of data within `input_stream` should take it upon itself to first verify that the necessary number of bytes are available before attempting that read/seek.

necouchman: Okay - moving it down to where I currently have it _seems_ to work, but if there's something different I should do, here, I'm happy to change it up.

mike-jumper: Well, the [documentation for RDPSND regarding the WaveInfo PDU](https://docs.microsoft.com/en-us/openspecs/windows_protocols/ms-rdpea/c53cd81c-0d7f-4e68-8b95-1c1da68dbaac
) pretty specifically states that the body size does not necessarily indicate the size of the remainder of the PDU:

> **Header (4 bytes):** An [RDPSND PDU Header (section 2.2.1)](https://docs.microsoft.com/en-us/openspecs/windows_protocols/ms-rdpea/863d154b-dbe0-4781-979d-fa48daed721e). The **msgType** field of the RDPSND PDU Header MUST be set to SNDC_WAVE (0x02). The **BodySize** field of the RDPSND PDU Header is the size of the WaveInfo PDU plus the size of the data field of the [Wave PDU](https://docs.microsoft.com/en-us/openspecs/windows_protocols/ms-rdpea/81841793-f5d3-4305-aa22-ddcbd81a96b5) that immediately follows this packet minus the size of the Header.

Given this, I don't believe it is correct to check `body_size` in this way, and I'm not sure why it would be working. The above indicates to me that `body_size` would always be larger than `Stream_GetRemainingLength()` for a WaveInfo PDU .

src/protocols/rdp/channels/rdpdr/rdpdr-fs-messages-file-info.c
necouchman: Okay, warning logging has been added in - let me know if I missed anything or if the warning messages need more detail, etc.

src/protocols/rdp/channels/rdpdr/rdpdr-fs-messages.c
necouchman: Added comments in that hopefully clarify this - let me know if they need any more detail.

src/protocols/rdp/channels/rdpsnd/rdpsnd.c
necouchman: So, sounds like I should pull it out of there?  Particularly since the Streams get checked elsewhere?  And it doesn't seem like it's possible to accurately check this size given that we have to also check the next packet?  Or should it be checked against some other value than `body_size`?

mike-jumper:
I think there's a check missing here:

https://github.com/apache/guacamole-server/blob/9c37fc56171d55b8fa838949d7fd40db85506857/src/protocols/rdp/channels/rdpsnd/rdpsnd-messages.c#L273-L277

src/protocols/rdp/channels/rdpsnd/rdpsnd.c
mike-jumper: Yeah, I think so. The value of `body_size` is needed, but as it doesn't (necessarily) correspond to the size of the PDU, it can't really be used as a sanity check for PDU size.

I think the critical thing is simply that a check of remaining buffer size is made before any attempt to read/seek through a block of the input stream. In the case of the code above, this would be both within that function (making sure there is enough space for the header that the function reads) and within the PDU-specific handlers invoked by that function (making sure there is enough space for whatever data those handlers read, which only those handlers should be aware of).

necouchman:
> I think there's a check missing here

Okay, added that one using `Stream_Length()` (and not `Stream_GetRemainingLength()`).

Let me know if anything else needs tweaking.

src/protocols/rdp/plugins/guacai/guacai-messages.c
```diff
@@ -48,8 +52,9 @@ static void guac_rdp_ai_read_format(wStream* stream,
     Stream_Read_UINT16(stream, format->bps); /* wBitsPerSample */
     Stream_Read_UINT16(stream, format->data_size); /* cbSize */
 
-    /* Read arbitrary data block (if applicable) */
-    if (format->data_size != 0) {
+    /* Read arbitrary data block (if applicable) and data is available. */
+    if (format->data_size != 0
+            && Stream_GetRemainingLength(stream) >= format->data_size) {
```
mike-jumper, line 57: This may produce some odd behavior. The check is necessary, but the failure status may need to be carried outward to the caller of `guac_rdp_ai_read_format()`.

This function is invoked by `guac_rdp_ai_process_formats()` in a loop, using each call to process one of the received audio input formats. If this check fails for one of the formats, and other formats remain to be read, then their headers will start to be read within what theoretically is the data portion of the preceding format.

src/protocols/rdp/channels/rdpsnd/rdpsnd-messages.c
```diff
@@ -266,6 +307,14 @@ void guac_rdpsnd_wave_handler(guac_rdp_common_svc* svc,
 
     guac_rdp_client* rdp_client = (guac_rdp_client*) client->data;
     guac_audio_stream* audio = rdp_client->audio;
+    
+    /* Verify we have at least 4 bytes, which is manually copied below. */
+    if (Stream_Length(input_stream) < 4) {
```
mike-jumper, line 312: That 4 bytes is actually copied from `rdpsnd->initial_wave_data`, not the `input_stream`. For `input_stream`, the read happens when `guac_audio_stream_write_pcm()` is called, and reads `rdpsnd->incoming_wave_size + 4` bytes:

https://github.com/apache/guacamole-server/blob/9c37fc56171d55b8fa838949d7fd40db85506857/src/protocols/rdp/channels/rdpsnd/rdpsnd-messages.c#L281-L282

src/protocols/rdp/plugins/guacai/guacai-messages.c
necouchman: Okay, so maybe change this from a `static void` to `static int` and return something non-zero if this happens?

src/protocols/rdp/channels/rdpsnd/rdpsnd-messages.c
necouchman: So, remove this entirely, or we should be checking `rdpsnd->initial_wave_data`?

src/protocols/rdp/plugins/guacai/guacai-messages.c
necouchman: Okay, took a stab at changing this such that any errors in `guac_rdp_ai_read_format()` results in `guac_rdp_ai_process_formats()` bailing out without continuing trying to process formats.  Let me know if that needs any work.

necouchman:
Okay, tried to fix up the comments with the correct PDU names where applicable and add the missing checks.  Just a couple of questions still open above on the previous changes.

src/protocols/rdp/plugins/guacai/guacai-messages.c
necouchman: Well, my initial try didn't work so well - it just bails out and doesn't process any sound input :-).  Apparently I need to try something else...

necouchman: Okay, I think I got it - have to deal with the availability of the arbitrary data block separately from checking the Stream for the amount of data.  Once I split that up it works as expected.

src/protocols/rdp/channels/rdpsnd/rdpsnd-messages.c
mike-jumper: We should be ensuring that `Stream_Length(input_stream)` is at least `rdpsnd->incoming_wave_size + 4`.

necouchman: Got it.  Done.

src/protocols/rdp/channels/rdpdr/rdpdr-messages.c
```diff
@@ -212,6 +212,10 @@ void guac_rdpdr_process_server_announce(guac_rdp_common_svc* svc,
 
     unsigned int major, minor, client_id;
 
+    /* Stream should contain at least 8 bytes (UINT16 + UINT16 + UINT32) */
+    if (Stream_GetRemainingLength(input_stream) < 8)
+        return;
```
mike-jumper, line 217: Missed one `guac_client_log()` here, I think.

src/protocols/rdp/channels/rdpdr/rdpdr.c
```diff
@@ -38,6 +38,17 @@ void guac_rdpdr_process_receive(guac_rdp_common_svc* svc,
...
     int packet_id;
 
+    /* 
+     * Check that device redirection stream contains at least 4 bytes
+     * (UINT16 + UINT16).
+     */
+    if (Stream_GetRemainingLength(input_stream) < 4) {
+        guac_client_log(svc->client, GUAC_LOG_WARNING, "Device redirection "
+                "channel receive Stream does not contain the expected number "
+                "of bytes. Device redirection may not function as expected.");
```
mike-jumper, line 48: > Device redirection channel receive Stream ...

As phrased, this does not make sense to me. Looking at the context, this is guarding the read operation around the header common to all RDPDR PDUs, so I would suggest rephrasing this in that respect ("Device redirection channel PDU header" or "RDPDR channel PDU header" or similar).

src/protocols/rdp/channels/rdpsnd/rdpsnd-messages.c
```diff
@@ -50,6 +50,14 @@ void guac_rdpsnd_formats_handler(guac_rdp_common_svc* svc,
     /* Reset own format count */
     rdpsnd->format_count = 0;
 
+    /* Check to make sure the stream has at least 20 bytes, which */
```
mike-jumper, line 53: Which what?

src/protocols/rdp/channels/rdpsnd/rdpsnd.c
```diff
@@ -35,11 +35,19 @@ void guac_rdpsnd_process_receive(guac_rdp_common_svc* svc,
     guac_rdpsnd* rdpsnd = (guac_rdpsnd*) svc->data;
     guac_rdpsnd_pdu_header header;
 
+    /* Check that we at least the 4 byte header (UINT8 + UINT8 + UINT16) */
```
mike-jumper, line 38: Did you mean "we have"?

src/protocols/rdp/plugins/guacai/guacai.c
```diff
@@ -52,10 +52,18 @@
 static void guac_rdp_ai_handle_data(guac_client* client,
         IWTSVirtualChannel* channel, wStream* stream) {
 
+    /* Verify we have at least 1 byte in the stream (UINT8) */
+    if (Stream_GetRemainingLength(stream) < 1) {
+        guac_client_log(client, GUAC_LOG_WARNING, "Audio input stream does not "
```
mike-jumper, line 57: This should probably be "PDU" or "PDU header" rather than "stream".

src/protocols/rdp/channels/rdpsnd/rdpsnd-messages.c
necouchman: It's a cliff-hanger.  Just channeling my inner novelist.

PR #275, opened 2020-05-17
GUACAMOLE-1059: Check array boundary for sound formats

necouchman:


src/protocols/rdp/channels/rdpsnd/rdpsnd-messages.c
```diff
@@ -296,7 +296,7 @@ void guac_rdpsnd_wave_info_handler(guac_rdp_common_svc* svc,
     rdpsnd->next_pdu_is_wave = TRUE;
 
     /* Reset audio stream if format has changed */
-    if (audio != NULL)
+    if (audio != NULL && format < sizeof(rdpsnd->formats))
```
mike-jumper, line 299: Looks good - very straightforward. The only nit I'd like to pick is logging: rather than silently trudge forward, I think we should log that something isn't right, just as in #274. If we encounter a faulty RDP server and this occurs, it would be good to know that it occurred, both for debugging and for reporting upstream.

Lacking that, this sort of failure would no doubt result in JIRA issues / mailing list threads regarding mysterious lack of sound, and it would take some involved digging before things could be established to not be a guac issue.

necouchman: Okay, took a stab at a log message - probably needs some tweaking, so let me know what needs to be changed.

src/protocols/rdp/channels/rdpsnd/rdpsnd-messages.c
```diff
@@ -296,11 +296,16 @@ void guac_rdpsnd_wave_info_handler(guac_rdp_common_svc* svc,
...
         guac_audio_stream_reset(audio, NULL,
                 rdpsnd->formats[format].rate,
                 rdpsnd->formats[format].channels,
                 rdpsnd->formats[format].bps);
+    
+    else
+        guac_client_log(svc->client, GUAC_LOG_WARNING, "RDP server attempted "
+                "to specify an invalid audio format.  Sound may not work as "
+                "expected.");
```
mike-jumper, line 308: This `else` will also be triggered if `audio == NULL`.

PR #276, opened 2020-05-17
VNC, RDP or SSH Client Connect and Session Recording

pheomis:
Hi,
Is there a feature where we can log in with the client (openssh, tighvnc, mstsc etc) that we use daily and session recording?

Sample;
terminal
ssh username:password:serverip@guacamole-server.com

### DIFF
--- a/bin/guacctl
+++ b/bin/guacctl
@@ -117,7 +117,7 @@ error() {
 ##
 usage() {
     cat >&2 <<END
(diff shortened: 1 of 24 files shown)


necouchman:
@pheomis: Please subscribe to the mailing list and ask your question, there.  This is not a support forum.

http://guacamole.apache.org/support/

PR #277, opened 2020-05-24
GUACAMOLE-343: Add setting for explicitly disabling recording.

necouchman:
I decided to tackle this real quick - if it's not something we're actually interested in doing let me know and I'll close this PR and the JIRA issue, but it was easy enough to knock out.


necouchman:
Closing for now per the JIRA discussion: https://issues.apache.org/jira/browse/GUACAMOLE-343

PR #282, opened 2020-06-09
GUACAMOLE-513: Several updates to Wake-on-LAN functionality

necouchman:
In addition to the issues identified by Coverity for socket open failures, this pull request enhances error logging and also changes the default boot time to zero in the recently-merged Wake-on-LAN functionality.

### DIFF
--- a/src/libguac/wol.c
+++ b/src/libguac/wol.c
@@ -19,6 +19,9 @@
 
 #include "config.h"
 
+#include "guacamole/error.h"
+#include "guacamole/wol.h"
+
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
@@ -26,8 +29,6 @@
 #include <arpa/inet.h>
 #include <netinet/in.h>
 #include <sys/socket.h>
-
-#include "guacamole/wol.h"
 
 /**
  * Generate the magic Wake-on-LAN (WoL) packet for the specified MAC address
@@ -85,8 +86,11 @@ static ssize_t __guac_wol_send_packet(const char* broadcast_addr,
     wol_dest.sin_family = AF_INET;
     if (inet_pton(AF_INET, broadcast_addr, &(wol_dest.sin_addr)) == 0) {
         wol_dest.sin_family = AF_INET6;
-        if (inet_pton(AF_INET6, broadcast_addr, &(wol_dest.sin_addr)) == 0)
+        if (inet_pton(AF_INET6, broadcast_addr, &(wol_dest.sin_addr)) == 0) {
+            guac_error = GUAC_STATUS_SEE_ERRNO;
+            guac_error_message = "Wake-on-LAN address does not appear to be IPv4 or IPv6";
             return 0;
+        }
     }
     
     /* Set up socket for IPv4 broadcast. */
@@ -95,13 +99,18 @@ static ssize_t __guac_wol_send_packet(const char* broadcast_addr,
         wol_socket = socket(AF_INET, SOCK_DGRAM, 0);
         
         /* If opening a socket fails, bail out. */
-        if (wol_socket < 0)
+        if (wol_socket < 0) {
+            guac_error = GUAC_STATUS_SEE_ERRNO;
+            guac_error_message = "Failed to open IPv4 Wake-on-LAN socket";
             return 0;
+        }
 
         /* Attempt to set broadcast; exit with error if this fails. */
         if (setsockopt(wol_socket, SOL_SOCKET, SO_BROADCAST, &wol_bcast,
                 sizeof(wol_bcast)) < 0) {
             close(wol_socket);
+            guac_error = GUAC_STATUS_SEE_ERRNO;
+            guac_error_message = "Failed to set IPv4 broadcast for Wake-on-LAN socket";
             return 0;
         }
     }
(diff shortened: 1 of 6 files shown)


src/libguac/wol.c
```diff
@@ -93,6 +93,12 @@ static ssize_t __guac_wol_send_packet(const char* broadcast_addr,
     if (wol_dest.sin_family == AF_INET) {
         int wol_bcast = 1;
         wol_socket = socket(AF_INET, SOCK_DGRAM, 0);
+        
+        /* If opening a socket fails, bail out. */
+        if (wol_socket < 0) {
+            close(wol_socket);
```
mike-jumper, line 99: There's nothing to `close()` if `wol_socket` is negative.

necouchman:
> Before merging, any thoughts on using guac_error and guac_error_message to communicate the nature of any failures within the WoL functions?

Yeah, I'm happy to take a look at those and see about passing through whatever error messages are available/applicable.

necouchman:
Okay, stab has been taken at using the `guac_error` framework to communicate error messages.  Let me know if the messages or status codes need any tweaks.

src/libguac/wol.c
```diff
@@ -85,19 +86,31 @@ static ssize_t __guac_wol_send_packet(const char* broadcast_addr,
     wol_dest.sin_family = AF_INET;
     if (inet_pton(AF_INET, broadcast_addr, &(wol_dest.sin_addr)) == 0) {
         wol_dest.sin_family = AF_INET6;
-        if (inet_pton(AF_INET6, broadcast_addr, &(wol_dest.sin_addr)) == 0)
+        if (inet_pton(AF_INET6, broadcast_addr, &(wol_dest.sin_addr)) == 0) {
+            guac_error = GUAC_STATUS_SEE_ERRNO;
+            guac_error_message = "Wake-on-LAN address does not appear to be IPv4 or IPv6";
             return 0;
+        }
```
mike-jumper, line 93: From the man page for `inet_pton()`:

> inet_pton() returns 1 on success (network address was successfully converted).  0 is returned if src does not contain a character string representing a valid network address in the specified address family. If af does not contain a valid address family, -1 is returned and errno is set to EAFNOSUPPORT.

[The relevant part of POSIX](https://pubs.opengroup.org/onlinepubs/9699919799/functions/inet_ntop.html) states similarly:

> The inet_pton() function shall return 1 if the conversion succeeds, with the address pointed to by dst in network byte order. It shall return 0 if the input is not a valid IPv4 dotted-decimal string or a valid IPv6 address string, or -1 with errno set to [EAFNOSUPPORT] if the af argument is unknown.

It doesn't look like `errno` is set for the case that `inet_pton()` returns 0. It _is_ set for -1, but that condition isn't handled here.

So, overall:

* `GUAC_STATUS_SEE_ERRNO` is probably incorrect for the case that `inet_pton()` returns 0
* We should consider handling the case that `inet_pton()` returns -1. A return value of -1 would be interpreted as success here, resulting in who-knows-what being used for the broadcast address.

src/libguac/wol.c
```diff
@@ -85,19 +86,31 @@ static ssize_t __guac_wol_send_packet(const char* broadcast_addr,
...
     }
     
     /* Set up socket for IPv4 broadcast. */
     if (wol_dest.sin_family == AF_INET) {
         int wol_bcast = 1;
         wol_socket = socket(AF_INET, SOCK_DGRAM, 0);
```
mike-jumper, line 99: Didn't notice this until now, but why `0` for IPv4 and `IPPROTO_UDP` for IPv6? It seems that's the only difference between the `socket()` calls which would prevent using `socket(wol_dest.sin_family, ...)`.

src/libguac/wol.c
necouchman: Okay, I'll refactor this a bit to handle both of those cases.

src/libguac/wol.c
necouchman: Looks like we can probably just specify `0` for both of them and collapse the socket piece into a single declaration, so I'll work on that.  According to `socket()` man page, the socket type (`SOCK_DGRAM`) usually determines the protocol and you can just put `0` for the protocol.  Guessing that will work equally well for IPv6 as IPv4 in this case.

src/libguac/wol.c
necouchman: Okay, this should be addressed, though it feels like there should be a slightly more elegant solution than this.  Played around with a `switch()` statement to see if that would work, but nothing quite as simple as `if...else`.

src/libguac/wol.c
necouchman: This should be a little more efficient, now.

src/libguac/wol.c
```diff
@@ -83,35 +84,68 @@ static ssize_t __guac_wol_send_packet(const char* broadcast_addr,
...
+    }
+    
+    /* If return value is zero, address doesn't match the IPv4, so try IPv6. */
+    else if (retval == 0) {
         wol_dest.sin_family = AF_INET6;
-        if (inet_pton(AF_INET6, broadcast_addr, &(wol_dest.sin_addr)) == 0)
+        retval = inet_pton(wol_dest.sin_family, broadcast_addr, &(wol_dest.sin_addr));
+        
+        /* IPv6 didn't work, either, so bail out. */
+        if (retval == 0) {
```
mike-jumper, line 102: Does `retval < 0` need to be checked for this call to `inet_pton()`, as well?

necouchman: Ah, I was mis-reading what the man page said about the `-1` return value of `inet_pton()`.  You're right, let me fix that real quick...

PR #284, opened 2020-06-20
GUACAMOLE-1076: Correct RemoteApp launch behavior.

mike-jumper:
The handshake procedure implemented by Apache Guacamole 1.1.0 for RemoteApp is partially incorrect, omitting the Client Information PDU. Without this PDU, RemoteApp behaves unexpectedly, launching multiple copies of the requested application.

In investigating this, additional differences from the 1.0.0 release were identified and have also been addressed here:

* The `RAIL_EXEC_FLAG_EXPAND_ARGUMENTS` flag was missing from the Client Execute PDU.
* Several system parameters were not specified, having previously been included by default via FreeRDP's internal implementation of RAIL. These are now explicitly specified.
* System Parameters Update PDUs were not being sent at all because the `params` flags were unset. These flags are used within FreeRDP to denote which system parameters are being updated. Each set flag results in a separate System Parameters Update PDU being sent for that portion of the provided `RAIL_SYSPARAM_ORDER` structure.

See: https://issues.apache.org/jira/browse/GUACAMOLE-1076?focusedCommentId=17141211&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-17141211

### DIFF
--- a/src/protocols/rdp/channels/rail.c
+++ b/src/protocols/rdp/channels/rail.c
@@ -100,13 +100,36 @@ static UINT guac_rdp_rail_complete_handshake(RailClientContext* rail) {
         return status;
 
     RAIL_SYSPARAM_ORDER sysparam = {
+
+        .dragFullWindows = FALSE,
+
+        .highContrast = {
+            .flags = 0x7E,
+            .colorScheme = {
+                .string = NULL,
+                .length = 0
+            }
+        },
+
+        .keyboardCues = FALSE,
+        .keyboardPref = FALSE,
+        .mouseButtonSwap = FALSE,
+
         .workArea = {
             .left   = 0,
             .top    = 0,
             .right  = rdp_client->settings->width,
             .bottom = rdp_client->settings->height
         },
-        .dragFullWindows = FALSE
+
+        .params =
+              SPI_MASK_SET_DRAG_FULL_WINDOWS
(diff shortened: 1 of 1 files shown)


src/protocols/rdp/channels/rail.c
```diff
@@ -88,14 +90,46 @@ static UINT guac_rdp_rail_complete_handshake(RailClientContext* rail) {
...
+
+        .dragFullWindows = FALSE,
+
+        .highContrast = {
+            .flags = 0x7E,
+            .colorScheme = {
+                .string = NULL,
+                .length = 0
+            }
+        },
```
necouchman, line 112: High contrast has to be enabled?

mike-jumper: My understanding is that this value (`0x7E`) is not _enabling_ high contrast mode, but declaring that the RDP server should allow the user to select high contrast mode through standard means. I've added commit 7598f5a to split this value up into its corresponding flags, which should make things clearer.

The value `0x7E` is the standard value passed when connecting via RemoteApp. It is shown within the example PDU in Microsoft's RemoteApp spec, is the default within FreeRDP, and was the value that Guacamole sent prior to the migration to FreeRDP 2.0.0.

Omitting this doesn't appear to break anything, but it would be an uncomfortable inconsistency with known-good (and possibly expected) behavior.

PR #289, opened 2020-06-24
GUACAMOLE-465: Correct usage of printf-like guacenc_log().

mike-jumper:
A few usages of `guacenc_log()` within guacenc are incorrect:

* The newline character is automatically added. Specifying it manually as part of the string will result in two newline characters being present.
* The `%d` format specifier was given for a string parameter, when `%s` was likely intended ("s" is right next to "d").

These changes remove the extra `\n` and switch the incorrect usage of `%d` to `%s`.

### DIFF
--- a/src/guacenc/video.c
+++ b/src/guacenc/video.c
@@ -483,7 +483,7 @@ int guacenc_video_free(guacenc_video* video) {
     /* write trailer, if needed */
     if (video->container_format_context != NULL &&
             video->output_stream != NULL) {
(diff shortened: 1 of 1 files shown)


necouchman:
Ah, crud, sorry @mike-jumper - merged #290 prior to this one and now there's a conflict.  If you can fix that up I'll get this merged and then get them both merged back to master.

mike-jumper:
No worries - they were doomed to conflict. I'll rebase.

mike-jumper:
Should be OK now.

PR #292, opened 2020-06-25
GUACAMOLE-1110: Size and security improvements for Docker images

IT-Experte:
Dockage image size will reduced by install only package dependencies (without recommendations). The guacd process runs as a non-privileged user and will be checked by netcat.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -113,8 +113,12 @@ RUN ${PREFIX_DIR}/bin/link-freerdp-plugins.sh \
 # Checks the operating status every 5 minutes with a timeout of 5 seconds
 HEALTHCHECK --interval=5m --timeout=5s CMD nc -z 127.0.0.1 4822 || exit 1
 
-# Run with user nobody
-USER nobody
+# Create a new user guacd
+ARG UID=1000
+RUN useradd --system --create-home --shell /usr/sbin/nologin --uid $UID --no-user-group guacd
+
+# Run with user guacd
+USER guacd
 
 # Expose the default listener port
 EXPOSE 4822
--- a/Dockerfile
+++ b/Dockerfile
@@ -90,6 +90,7 @@ ENV LD_LIBRARY_PATH=${PREFIX_DIR}/lib
 ENV GUACD_LOG_LEVEL=info
 
 ARG RUNTIME_DEPENDENCIES="            \
+        netcat-openbsd                \
         ca-certificates               \
         ghostscript                   \
         fonts-liberation              \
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -100,15 +101,21 @@ ARG RUNTIME_DEPENDENCIES="            \
...
+    apt-get install -y --no-install-recommends $RUNTIME_DEPENDENCIES                && \
+    apt-get install -y --no-install-recommends $(cat "${PREFIX_DIR}"/DEPENDENCIES)  && \
     rm -rf /var/lib/apt/lists/*
 
 # Link FreeRDP plugins into proper path
 RUN ${PREFIX_DIR}/bin/link-freerdp-plugins.sh \
         ${PREFIX_DIR}/lib/freerdp2/libguac*.so
 
+# Checks the operating status every 5 minutes with a timeout of 5 seconds
+HEALTHCHECK --interval=5m --timeout=5s CMD nc -z 127.0.0.1 4822 || exit 1
```
necouchman, line 114: My Docker knowledge is fairly limited, so forgive me if this is an ignorant question, but is this a generally acceptable method for doing healthchecks?  Is there some advantage to this kind of healthcheck over something like:
```
ss -tnlp|grep -q 4822 || exit 1
```
?

Dockerfile
```diff
@@ -100,15 +101,21 @@ ARG RUNTIME_DEPENDENCIES="            \
...
+
+# Run with user nobody
+USER nobody
```
necouchman, line 117: I'm curious if this will work with FreeRDP 2.x in this particular Docker environment, particularly given the issues we've seen with FreeRDP requiring a writable home directory for the user (including the user running guacd)?

Dockerfile
IT-Experte: Yes it works too, but you have to install package iproute2 for program ss. This package need 4204 kB additional disk space. Package netcat-openbsd only consume 114 kB space.

Dockerfile
IT-Experte: I will check this out and switch to Debian branche testing with include FreeRDP 2.0. (ARG DEBIAN_VERSION=testing)

Dockerfile
necouchman: Okay, fair enough.

Dockerfile
necouchman: The current Guacamole image uses a FreeRDP 2.x build, so it's really just a matter of verifying that the `nobody` account within the Docker image has a writable home directory, and that FreeRDP connections work as expected after this change.  If you've verified that both of those are the case, then that's fine - just want to be sure.

IT-Experte: I tested Debian 10 (stable) with (libfreerdp2.0.0~git20190204) and Debian 11 (testing) with (libfreerdp2.1.1 dfsg1-1 b1).

Guacd 1.2 and libfreerdp2.0 is working well. Guacd print out a warning message, but Filetransfer and clipboard are working. Guacd 1.2 and libfreerdp2.1.1 is not working.

Should I change the user "Nobody" to a new user "guacd" with a writable home directory?

----
WARNING:	FreeRDP initialization may fail: Writability of the current user's home directory ("/nonexistent") could not be determined: No such file or directory

mike-jumper: > Should I change the user "Nobody" to a new user "guacd" with a writable home directory?

Yes. That warning is important and things will likely unexpectedly fail. This shouldn't be the default behavior of the image.

IT-Experte: At the time of image creation, a new user is created to run the process. File Transfer works without any warning into logs (in both directions).

benrubson:
Please see https://github.com/apache/guacamole-server/pull/314 which completes this, creating a proper group for the `guacd` service user. Thx 👍

PR #293, opened 2020-06-25
GUACAMOLE-474: Correctly disable RDP uploads and downloads in get/put handlers

necouchman:


mike-jumper:
I've opened #294, now realizing that you opened a pull request for similar. I think some checks may be missing here, though, in particular some low-level catch-alls in case we miss something.

necouchman:
Closing in favor of #294

PR #297, opened 2020-07-02
GUACAMOLE-221: Add convenience API for automatically handling received "argv" streams.

mike-jumper:
This change adds a new header to libguac, `argv.h`, which provides a series of convenience functions for automatically handling "argv" streams:

* `guac_argv_register()` - Registers a callback which should be invoked when the full, updated value of an argument has been received via an "argv" stream. Option flags can be specified to declare that a particular callback should be invoked only once, or that received values should be echoed back to all users of the connection.
* `guac_argv_received()` - Hands over control of a received "argv" stream such that registered callbacks will automatically be invoked.
* `guac_argv_handler` - Convenient implementation of the "argv" instruction handler which automatically handles sending of "ack" instructions and invokes `guac_argv_received()` appropriately.

In general, `guac_argv_register()` and `guac_argv_handler` should be sufficient to replace all current usage of the "argv" stream. For handling of "argv" streams that must block the connection, two additional functions are provided:

* `guac_argv_await()` - Blocks until each of the provided NULL-terminated list of arguments has been received.
* `guac_argv_stop()` - Safely unblocks any `guac_argv_await()` that are currently blocked, and prevents any future calls from blocking.

### DIFF
--- a/src/libguac/argv.c
+++ b/src/libguac/argv.c
@@ -1,0 +1,340 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+#include "config.h"
+
+#include "guacamole/argv.h"
+#include "guacamole/protocol.h"
+#include "guacamole/socket.h"
+#include "guacamole/stream.h"
+#include "guacamole/string.h"
+#include "guacamole/user.h"
+
+#include <pthread.h>
+#include <stdlib.h>
+#include <string.h>
+
+/**
+ * The state of an argument that will be automatically processed. Note that
+ * this is distinct from the state of an argument value that is currently being
+ * processed. Argument value states are dynamically-allocated and scoped by the
+ * associated guac_stream.
+ */
+typedef struct guac_argv_state {
+
+    /**
+     * The name of the argument.
+     */
+    char name[GUAC_ARGV_MAX_NAME_LENGTH];
+
+    /**
+     * Whether at least one value for this argument has been received since it
+     * was registered.
+     */
+    int received;
+
+    /**
+     * Bitwise OR of all option flags that should affect processing of this
+     * argument. The only current flag is GUAC_ARGV_OPTION_ONCE.
+     */
+    int options;
+
+    /**
+     * The callback that should be invoked when a new value for the associated
+     * argument has been received. If the GUAC_ARGV_OPTION_ONCE flag is set,
+     * the callback will be invoked at most once.
+     */
+    guac_argv_callback* callback;
+
+    /**
+     * The arbitrary data that should be passed to the callback.
+     */
+    void* data;
+
+} guac_argv_state;
+
+/**
+ * The current state of automatic processing of "argv" streams.
+ */
+typedef struct guac_argv_await_state {
+
+    /**
+     * Whether automatic argument processing has been stopped via a call to
+     * guac_argv_stop().
+     */
+    int stopped;
+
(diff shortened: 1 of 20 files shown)


src/libguac/guacamole/argv-constants.h
```diff
@@ -0,0 +1,71 @@
...
+ * Additional streams will be rejected.
+ */
+#define GUAC_ARGV_OPTION_ONCE 1
+
+/**
+ * Option flag which declares to guac_argv_register() that the values received
+ * and accepted for the associated argument should be echoed to all connected
+ * users via outbound "argv" streams.
+ */
+#define GUAC_ARGV_OPTION_ECHO 2
```
necouchman, line 42: With this option, here and implemented in the new `libguac/argv.c` file, is there still a need for the following functions:
`guac_telnet_send_current_argv()`
`guac_ssh_send_current_argv()`
`guac_kubernetes_send_current_argv()`

?  It seems like the echo here and the `guac_*_send_current_argv()` have the same purpose - sending the argv value back to the client.  Is there a reason they are both required?

mike-jumper: Yep - the current value of the parameters that users may change needs to be sent to the user when the connection starts, as well as to new users joining a shared connection. This is distinct from the case where an update to a parameter is broadcast to all current users of the connection (`GUAC_ARGV_OPTION_ECHO`).

src/libguac/argv.c
```diff
@@ -0,0 +1,350 @@
...
+
+    }
+
+    pthread_mutex_unlock(&await_state.lock);
+}
+
+int guac_argv_handler(guac_user* user, guac_stream* stream,
+        char* mimetype, char* name) {
+
+    /* Refuse stream if not argument is not registered */
```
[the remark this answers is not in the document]
mike-jumper: You're not not not wrong.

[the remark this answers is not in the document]
mike-jumper: OK - fixed up the relevant commit and rebased.

src/libguac/argv.c
```diff
@@ -0,0 +1,350 @@
...
+
+} guac_argv;
+
+/**
+ * Statically-allocated, shared state of the guac_argv_*() family of functions.
+ */
+static guac_argv_await_state await_state = {
+    .lock = PTHREAD_MUTEX_INITIALIZER,
+    .changed = PTHREAD_COND_INITIALIZER
+};
```
necouchman, line 144: I'm guessing statically allocated `mutex`/`cond` don't have to be `free`d?  Or is it just assumed that it'll be freed when the process is shut down?

src/protocols/telnet/user.c
```diff
@@ -92,7 +92,7 @@ int guac_telnet_user_join_handler(guac_user* user, int argc, char** argv) {
         user->pipe_handler = guac_telnet_pipe_handler;
 
         /* Updates to connection parameters */
-        user->argv_handler = guac_telnet_argv_handler;
+        user->argv_handler = guac_argv_handler;
```
necouchman, line 95: Looks like this is consistently changed to the `guac_argv_handler` for all of the protocols - does it make sense to continue to allow this to be configurable, or is it something that should be changed in the `guac_user` code such that `guac_argv_handler` is always what is called?

I'm good with it either way, just curious - I'm all for leaving things flexible/configurable, just wasn't sure if the new `argv` code is intended to be the new flexible/configurable framework that is always referred to, or if there are foreseeable situations where you'd still want to change that?

src/libguac/argv.c
```diff
@@ -0,0 +1,350 @@
...
+    /**
+     * Whether automatic argument processing has been stopped via a call to
+     * guac_argv_stop().
+     */
+    int stopped;
+
+    /**
+     * The total number of arguments registered.
+     */
+    unsigned int num_registered;
```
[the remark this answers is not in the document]
mike-jumper: > Or does that happen automatically during the static allocation below?

Exactly. The C standard specifies that a statically-initialized array or structure is initialized to the values specified, with all other values being initialized to zero.

src/libguac/argv.c
mike-jumper: Correct. There does not need to be a corresponding call to `pthread_potato_destroy()` so long as it's a statically-allocated potato.

src/protocols/telnet/user.c
mike-jumper: I definitely think that `argv_handler` needs to stay there:

The pattern of the API provided by libguac is one of gradually increasing levels of abstraction, and I think it's important that we maintain that pattern consistently. We provide higher-level convenience functions, default implementations, etc. as needed, but aim to always allow implementations to take control if those conveniences aren't well-suited in their case.

Regarding this specific implementation, it provides its convenience at the cost of limited buffer space for received argument values. Should an implementation need greater space, or to process arguments of any size as true streams, they will need to be able to handle the raw stream themselves.

PR #301, opened 2020-09-10
GUACAMOLE-1174: Added exec call implementation for kubernetes protocol

yaroslavNqualisystems:
Added support for using `exec` for attaching to kubernetes pods.
Modified generation of Kubernetes API endpoint.

Added connection argument  _exec-command_.
 
**exec_command:**
 The command used by exec call. If omitted, attach call will be used.
 Used to run a new process and attach to it, instead of the main container process.

### DIFF
--- a/src/protocols/kubernetes/settings.c
+++ b/src/protocols/kubernetes/settings.c
@@ -31,6 +31,8 @@ const char* GUAC_KUBERNETES_CLIENT_ARGS[] = {
     "namespace",
     "pod",
     "container",
+    "use-exec",
+    "exec-command",
     "use-ssl",
     "client-cert",
     "client-key",
@@ -85,6 +87,16 @@ enum KUBERNETES_ARGS_IDX {
      * in the pod will be used.
      */
     IDX_CONTAINER,
+
+    /**
+     * Whether exec call should be used. If omitted, attach call will be used.
+     */
+    IDX_USE_EXEC,
+
+    /**
+     * The command used by exec call.
+     */
+    IDX_EXEC_COMMAND,
 
     /**
      * Whether SSL/TLS should be used. If omitted, SSL/TLS will not be used.
@@ -274,6 +286,16 @@ guac_kubernetes_settings* guac_kubernetes_parse_args(guac_user* user,
     settings->kubernetes_container =
         guac_user_parse_args_string(user, GUAC_KUBERNETES_CLIENT_ARGS, argv,
                 IDX_CONTAINER, NULL);
+
+    /* Parse whether exec call should be used */
+    settings->use_exec =
+        guac_user_parse_args_boolean(user, GUAC_KUBERNETES_CLIENT_ARGS, argv,
+                IDX_USE_EXEC, false);
+
+    /* Read exec command (optional) */
+    settings->exec_command =
+        guac_user_parse_args_string(user, GUAC_KUBERNETES_CLIENT_ARGS, argv,
+                IDX_EXEC_COMMAND, GUAC_KUBERNETES_DEFAULT_EXEC_COMMAND);
 
     /* Parse whether SSL should be used */
     settings->use_ssl =
--- a/src/protocols/kubernetes/settings.c
+++ b/src/protocols/kubernetes/settings.c
@@ -31,7 +31,6 @@ const char* GUAC_KUBERNETES_CLIENT_ARGS[] = {
     "namespace",
     "pod",
     "container",
-    "use-exec",
     "exec-command",
     "use-ssl",
     "client-cert",
@@ -89,12 +88,7 @@ enum KUBERNETES_ARGS_IDX {
     IDX_CONTAINER,
 
     /**
-     * Whether exec call should be used. If omitted, attach call will be used.
-     */
-    IDX_USE_EXEC,
-
-    /**
-     * The command used by exec call.
+     * The command used by exec call. If omitted, attach call will be used.
      */
     IDX_EXEC_COMMAND,
 
@@ -287,15 +281,10 @@ guac_kubernetes_settings* guac_kubernetes_parse_args(guac_user* user,
         guac_user_parse_args_string(user, GUAC_KUBERNETES_CLIENT_ARGS, argv,
                 IDX_CONTAINER, NULL);
 
-    /* Parse whether exec call should be used */
-    settings->use_exec =
-        guac_user_parse_args_boolean(user, GUAC_KUBERNETES_CLIENT_ARGS, argv,
-                IDX_USE_EXEC, false);
-
     /* Read exec command (optional) */
     settings->exec_command =
         guac_user_parse_args_string(user, GUAC_KUBERNETES_CLIENT_ARGS, argv,
-                IDX_EXEC_COMMAND, GUAC_KUBERNETES_DEFAULT_EXEC_COMMAND);
+                IDX_EXEC_COMMAND, NULL);
 
     /* Parse whether SSL should be used */
     settings->use_ssl =
--- a/src/protocols/kubernetes/settings.c
+++ b/src/protocols/kubernetes/settings.c
@@ -417,6 +417,9 @@ void guac_kubernetes_settings_free(guac_kubernetes_settings* settings) {
     free(settings->kubernetes_pod);
     free(settings->kubernetes_container);
 
+    /* Free Kubernetes exec command */
+    free(settings->exec_command);
+
     /* Free SSL/TLS details */
     free(settings->client_cert);
     free(settings->client_key);
--- a/src/protocols/kubernetes/url.c
+++ b/src/protocols/kubernetes/url.c
@@ -89,15 +89,16 @@ int guac_kubernetes_escape_url_component(char* output, int length,
 
 }
 
-int guac_kubernetes_endpoint_attach(char* buffer, int length,
+int guac_kubernetes_endpoint_uri(char* buffer, int length,
         const char* kubernetes_namespace, const char* kubernetes_pod,
-        const char* kubernetes_container) {
+        const char* kubernetes_container, int use_exec, const char* exec_command) {
 
     int written;
 
     char escaped_namespace[GUAC_KUBERNETES_MAX_ENDPOINT_LENGTH];
     char escaped_pod[GUAC_KUBERNETES_MAX_ENDPOINT_LENGTH];
     char escaped_container[GUAC_KUBERNETES_MAX_ENDPOINT_LENGTH];
+    char escaped_exec_command[GUAC_KUBERNETES_MAX_ENDPOINT_LENGTH];
 
     /* Escape Kubernetes namespace */
     if (guac_kubernetes_escape_url_component(escaped_namespace,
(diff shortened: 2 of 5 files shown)


src/protocols/kubernetes/settings.c
```diff
@@ -31,6 +31,8 @@ const char* GUAC_KUBERNETES_CLIENT_ARGS[] = {
     "namespace",
     "pod",
     "container",
+    "use-exec",
+    "exec-command",
```
mike-jumper, line 35: Why not just `exec-command`, defaulting to "attach" if no command is provided for execution?

src/protocols/kubernetes/url.c
```diff
@@ -109,26 +110,42 @@ int guac_kubernetes_endpoint_attach(char* buffer, int length,
...
-                "?container=%s&stdin=true&stdout=true&tty=true",
-                escaped_namespace, escaped_pod, escaped_container);
-    }
-    else {
-        written = snprintf(buffer, length,
-                "/api/v1/namespaces/%s/pods/%s/attach"
-                "?stdin=true&stdout=true&tty=true",
-                escaped_namespace, escaped_pod);
+        snprintf(endpoint_params+param_length, GUAC_KUBERNETES_MAX_ENDPOINT_LENGTH-param_length,
+                "container=%s&", escaped_container);
```
mike-jumper, line 143: The return value of `snprintf()` shouldn't be discarded. This function needs to know whether it has successfully assembled the entire endpoint URI.

src/protocols/kubernetes/url.c
```diff
@@ -109,26 +110,42 @@ int guac_kubernetes_endpoint_attach(char* buffer, int length,
...
+
+    char endpoint_path[GUAC_KUBERNETES_MAX_ENDPOINT_LENGTH];
+    snprintf(endpoint_path, GUAC_KUBERNETES_MAX_ENDPOINT_LENGTH*3, 
+    "/api/v1/namespaces/%s/pods/%s/%s", escaped_namespace, escaped_pod, call);
+
+    /* Generate endpoint params */
+    char endpoint_params[GUAC_KUBERNETES_MAX_ENDPOINT_LENGTH]="";
+    int param_length=0;
+
+    if(use_exec){
```
mike-jumper, line 126: Please follow surrounding style - there should be a space after the `if` and before the opening `{`.

src/protocols/kubernetes/url.c
```diff
@@ -109,26 +110,42 @@ int guac_kubernetes_endpoint_attach(char* buffer, int length,
...
-    /* Generate attachment endpoint URL */
-    if (kubernetes_container != NULL) {
+    /* Generate endpoint path depending on the call type */
+    char* call="attach";
+    if (use_exec)
+        call = "exec";
+
+    char endpoint_path[GUAC_KUBERNETES_MAX_ENDPOINT_LENGTH];
+    snprintf(endpoint_path, GUAC_KUBERNETES_MAX_ENDPOINT_LENGTH*3, 
+    "/api/v1/namespaces/%s/pods/%s/%s", escaped_namespace, escaped_pod, call);
```
mike-jumper, line 120: As with the other call below, the return value of `snprintf()` should not be discarded.

mike-jumper, line 120: When breaking a statement across multiple lines like this, the continuation of the statement should be indented to make that clear. As is, this is a bit hard to read.

mike-jumper, line 120: Why is the length passed to `snprintf()` here three times the actual length of the buffer?

src/protocols/kubernetes/url.c
```diff
@@ -109,26 +110,42 @@ int guac_kubernetes_endpoint_attach(char* buffer, int length,
...
+    int param_length=0;
+
+    if(use_exec){
+        /* Escape exec command */
+        if (guac_kubernetes_escape_url_component(escaped_exec_command,
+                    sizeof(escaped_exec_command), exec_command))
+            return 1;
 
+        param_length += snprintf(endpoint_params, GUAC_KUBERNETES_MAX_ENDPOINT_LENGTH, 
+        "command=%s&", escaped_exec_command);
```
mike-jumper, line 133: If this fails because there is insufficient space for the command within the buffer, that failure needs to be handled. As written, if the command somehow exceeds buffer length, subsequent calls to `snprintf()` will attempt to write off the end of the buffer.

mike-jumper, line 133: As above, when breaking a statement across multiple lines, please indent subsequent lines such that this is clear.

mike-jumper (changes requested):

mike-jumper:
@yaroslavNqualisystems, before I jump back in and re-review:

GUACAMOLE-1175: Fixes

Please update your commit message such that it reflects the high-level nature of your changes. If that forces the change to be split across multiple, logical commits, so much the better. The commit logs should serve as a useful reference for anyone that wants to discover the nature of changes made, what motivated those changes, etc.

A good (and amusing) reference on this: https://www.codelord.net/2015/03/16/bad-commit-messages-hall-of-shame/

yaroslavNqualisystems:
> @yaroslavNqualisystems, before I jump back in and re-review:
> 
> GUACAMOLE-1175: Fixes
> 
> Please update your commit message such that it reflects the high-level nature of your changes. If that forces the change to be split across multiple, logical commits, so much the better. The commit logs should serve as a useful reference for anyone that wants to discover the nature of changes made, what motivated those changes, etc.
> 
> A good (and amusing) reference on this: https://www.codelord.net/2015/03/16/bad-commit-messages-hall-of-shame/

I have updated the commit message, I guess you can proceed with review.

src/protocols/kubernetes/settings.c
```diff
@@ -275,6 +281,11 @@ guac_kubernetes_settings* guac_kubernetes_parse_args(guac_user* user,
         guac_user_parse_args_string(user, GUAC_KUBERNETES_CLIENT_ARGS, argv,
                 IDX_CONTAINER, NULL);
 
+    /* Read exec command (optional) */
+    settings->exec_command =
+        guac_user_parse_args_string(user, GUAC_KUBERNETES_CLIENT_ARGS, argv,
+                IDX_EXEC_COMMAND, NULL);
```
mike-jumper, line 287: The string returned by `guac_user_parse_args_string()` is dynamically-allocated and will need to eventually be freed. See:

https://github.com/apache/guacamole-server/blob/2aa0218cc761b0810d20d92ba3deccc57c993a63/src/protocols/kubernetes/settings.c#L399-L429

src/protocols/kubernetes/url.c
```diff
@@ -109,26 +110,52 @@ int guac_kubernetes_endpoint_attach(char* buffer, int length,
...
-                "?container=%s&stdin=true&stdout=true&tty=true",
-                escaped_namespace, escaped_pod, escaped_container);
-    }
-    else {
-        written = snprintf(buffer, length,
-                "/api/v1/namespaces/%s/pods/%s/attach"
-                "?stdin=true&stdout=true&tty=true",
-                escaped_namespace, escaped_pod);
+        written = snprintf(endpoint_params, sizeof(endpoint_params),
+                "container=%s&", escaped_container);
```
mike-jumper, line 149: Won't this overwrite the string previously written to `endpoint_params` if `exec_command` is set?

src/protocols/kubernetes/url.c
```diff
@@ -109,26 +110,52 @@ int guac_kubernetes_endpoint_attach(char* buffer, int length,
...
+        if (guac_kubernetes_escape_url_component(escaped_exec_command,
+                    sizeof(escaped_exec_command), exec_command))
+            return 1;
 
+        written = snprintf(endpoint_params, sizeof(endpoint_params), 
+            "command=%s&", escaped_exec_command);
+
+        if (written < 0 || written >= sizeof(endpoint_params))
+            return 1;
+    }
```
mike-jumper, line 140: Perhaps there is a way to clean up these repeated, similar calls to `snprintf()` intended to append parameters to a URI? Seems the sort of thing that would benefit greatly from being moved into its own simple, testable function.

mike-jumper (changes requested):

src/protocols/kubernetes/settings.c
yaroslavNqualisystems: Done

src/protocols/kubernetes/url.c
yaroslavNqualisystems: Fixed

src/protocols/kubernetes/url.c
yaroslavNqualisystems: Fixed

src/protocols/kubernetes/url.c
```diff
@@ -89,15 +89,33 @@ int guac_kubernetes_escape_url_component(char* output, int length,
...
+    char escaped_param_value[GUAC_KUBERNETES_MAX_ENDPOINT_LENGTH];
+
+    /* Escape value */
+    if (guac_kubernetes_escape_url_component(escaped_param_value,
+                    sizeof(escaped_param_value), param_value))
+            return 1;
+    
+    int written;
+    written = snprintf(buffer + strlen(buffer), length - strlen(buffer), 
+            "%s=%s&", param_name, escaped_param_value);
```
mike-jumper, line 104: Rather than iterate through the string twice (by calling `strlen()` twice), and rather than assuming that further parameters will be appended and always including an "&", I suggest killing two birds with one stone and determining the length while determining whether the "?" is already present (and thus whether a "?" or "&" prefix is needed).

This function would then be usable in any context, perhaps appending directly to the path being built rather than the awkward:

```c
    written = snprintf(buffer, length, "%s?%sstdin=true&stdout=true&tty=true", 
        endpoint_path, endpoint_params);
```

mike-jumper (changes requested):

src/protocols/kubernetes/url.c
mike-jumper: Beware also that calculating the length passed to `snprintf()` as `length - strlen(buffer)` has the potential to write into memory off the end of the buffer:

Without `strlen()` being bounded by the length of the buffer, it may well return a value larger than `length` if `buffer` is not null-terminated. That would result in a negative value that is ultimately interpreted as a very large positive value when converted into the `size_t` used by `snprintf()`.

src/protocols/kubernetes/url.c
```diff
@@ -89,15 +89,33 @@ int guac_kubernetes_escape_url_component(char* output, int length,
 
 }
 
-int guac_kubernetes_endpoint_attach(char* buffer, int length,
+int guac_kubernetes_append_endpoint_param(char* buffer, int length, 
+        const char* param_name, const char* param_value) {
```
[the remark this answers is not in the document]
yaroslavNqualisystems: Added prototype 004f57e19a42ddf5b14f355835302b8042a16afb

src/protocols/kubernetes/url.c
yaroslavNqualisystems: 7a1ba51baef5a6e8e8d85f6a5b68c022d6ba0fa1 I have added algorithm which determine the delimiter and calculate the buffer string length also added verification the buffer null terminated, not sure 100% it is correct.
7809447c3ffd7c0fddef807231e994f52a6c15ed Added usage of modified function.

mike-jumper (approved):

PR #303, opened 2020-10-01
GUACAMOLE-1190: Create AF_INET6 socket if no IPv4 addresses are available

lyind:
`guacd` cannot currently bind to IPv6 addresses (ie. the IPv6 anyaddress `::`), due to `AF_INET` being explicitly specified when creating the socket using `socket(AF_INET, ...)`.

This PR changes that in a convervative way and enables `guacd` to switch to the IPv6 address family in case no IPv4 address is returned by `getaddrinfo()`. No creating a `AF_INET6` at this point would always cause the later call to `bind()` to fail.

The added code may later be further enhanced to try to more aggressively bind to IPv6 on all supported operating systems. Unfortunately IPv6 socket behavior differs a bit between platforms and also depends on OS settings, thus the conservative approach.

For now this is mostly useful to support explicit IPv6-only deployments and has been tested to reliably bind to IPv6 addresses in such an environment.

### DIFF
--- a/src/guacd/daemon.c
+++ b/src/guacd/daemon.c
@@ -304,20 +304,6 @@ int main(int argc, char* argv[]) {
 
     }
 
-    /* Get socket */
-    socket_fd = socket(AF_INET, SOCK_STREAM, 0);
-    if (socket_fd < 0) {
-        guacd_log(GUAC_LOG_ERROR, "Error opening socket: %s", strerror(errno));
-        exit(EXIT_FAILURE);
-    }
-
-    /* Allow socket reuse */
-    if (setsockopt(socket_fd, SOL_SOCKET, SO_REUSEADDR,
-                (void*) &opt_on, sizeof(opt_on))) {
-        guacd_log(GUAC_LOG_WARNING, "Unable to set socket options for reuse: %s",
-                strerror(errno));
-    }
-
     /* Attempt binding of each address until success */
     current_address = addresses;
     while (current_address != NULL) {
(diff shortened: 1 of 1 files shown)


necouchman:
@lyind: Thanks for this contribution! Before we can go further with reviewing this code, though, the contribution guidelines need to be followed, which first involves creating a JIRA issue for this change, and then tagging the PR and the commit messages with that JIRA issue.  JIRA is located here:

https://issues.apache.org/jira/projects/GUACAMOLE

Full contribution guidelines can be found, here:

http://guacamole.apache.org/open-source/

lyind:
Fixes GUACAMOLE-1190

src/guacd/daemon.c
```diff
@@ -304,8 +304,23 @@ int main(int argc, char* argv[]) {
...
 
+    /* create socket for AF_INET6 to allow IPv6 bind in case there is no IPv4 address available */
+    sa_family_t family = AF_INET6;
+    current_address = addresses;
+    while (current_address != NULL) {
+
+        if (current_address->ai_family == AF_INET)
+        {
+            family = AF_INET;
+        }
```
mike-jumper, line 315: Please follow the same style as established elsewhere (opening brace on same line). See surrounding code and: http://guacamole.apache.org/guac-style/

src/guacd/daemon.c
```diff
@@ -304,8 +304,23 @@ int main(int argc, char* argv[]) {
...
+
+        current_address = current_address->ai_next;
+    }
```
mike-jumper, line 318: I don't think it is strictly correct to have two loops through the possible addresses like this - one to choose the family and then another to choose the address (presumably failing repeatedly for if that family was pulled from a different address). Shouldn't this all be done in the same loop?

mike-jumper (changes requested):
Please also follow the established convention for formatting of commit messages ("GUACAMOLE-XYZ: High-level description of the changes made.")

The current commit ("[GUACAMOLE-1190] Create AF_INET6 socket if no IPv4 addresses are available") happily does mention the JIRA issue, but does not follow the convention used within our git.

src/guacd/daemon.c
lyind: I had a similar thought, but initially didn't feel like changing the proven flow of control much.

It would probably be OK to re-create the socket a few times (we set `SO_REUSEADDR`). I re-factored the code accordingly.

Please review my updated (force-pushed) commit.

Thank you!

src/guacd/daemon.c
lyind: Sorry for missing this, fixed.

src/guacd/daemon.c
```diff
@@ -333,27 +319,46 @@ int main(int argc, char* argv[]) {
             guacd_log(GUAC_LOG_ERROR, "Unable to resolve host: %s",
                     gai_strerror(retval));
 
+        /* Get socket */
+        socket_fd = socket(current_address->ai_family, SOCK_STREAM, 0);
+        if (socket_fd < 0) {
+            guacd_log(GUAC_LOG_ERROR, "Error opening socket: %s", strerror(errno));
+
+            /** Unable to get a socket for the resolved address family, try next */
```
mike-jumper, line 327: Looks like an extra asterisk snuck in here. I assume this isn't intended to be a Doxygen-style comment. ;)

mike-jumper (changes requested):

src/guacd/daemon.c
lyind: Thanks for the review.

- Comment style fixed
- Also log socket protocol family on failure to bind (may help in fixing the configuration error)

mike-jumper (approved):

PR #307, opened 2020-11-02
GUACAMOLE-1205: Update Guacamole Server version numbers for 1.3.0 release

necouchman:
Updates `guacamole-server` versions for `1.3.0` version release, including `libguac` due to changes from GUACAMOLE-221. I think I did the update to `libguac` correctly, but I'm a little uncertain on the `age` field for this release, so let me know if that needs to be tweaked.

### DIFF
--- a/src/libguac/Makefile.am
+++ b/src/libguac/Makefile.am
@@ -135,7 +135,7 @@ libguac_la_CFLAGS = \
     -Werror -Wall -pedantic
 
 libguac_la_LDFLAGS =     \
-    -version-info 18:0:1 \
+    -version-info 19:0:0 \
     -no-undefined        \
     @CAIRO_LIBS@         \
     @DL_LIBS@            \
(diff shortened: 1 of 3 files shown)


src/libguac/Makefile.am
```diff
@@ -135,7 +135,7 @@ libguac_la_CFLAGS = \
     -Werror -Wall -pedantic
 
 libguac_la_LDFLAGS =     \
-    -version-info 18:0:1 \
+    -version-info 19:0:2 \
```
mike-jumper, line 138: I believe `age` needs to be reset to 0 as interfaces were changed (a new `protocol_version` member was added to the non-opaque `guac_user_info` structure). Code expecting the newer version of `guac_user_info` may not work (segfault) if it is given an older version and it attempts to access `protocol_version`.

If `guac_user_info` were opaque (with its members exposed only through getter functions), this would be correct.

mike-jumper (changes requested):

src/libguac/Makefile.am
necouchman: Okay, I have reset it to 0.

mike-jumper (approved):

PR #308, opened 2020-11-02
GUACAMOLE-221: Rely on FreeRDP error code if no RDP disconnect reason is available.

mike-jumper:
Currently, if the user enters incorrect credentials when prompted, they are advised that the remote desktop server is unavailable. This is incorrect and results from the RDP support not handling the error code(s) associated within invalid login attempts, which may not be exposed within the disconnect reason.

These changes handle the additional error codes supported by FreeRDP that expose whether the connection has failed due to incorrect credentials.

### DIFF
--- a/src/protocols/rdp/error.c
+++ b/src/protocols/rdp/error.c
@@ -18,14 +18,90 @@
  */
 
 #include "error.h"
-#include "rdp.h"
 
 #include <freerdp/freerdp.h>
 #include <guacamole/client.h>
 #include <guacamole/protocol.h>
 #include <guacamole/socket.h>
+#include <winpr/wtypes.h>
 
-void guac_rdp_client_abort(guac_client* client) {
+/**
+ * Translates the error code returned by freerdp_get_last_error() for the given
+ * RDP instance into a Guacamole status code and human-readable message. If no
+ * error was reported, a successful error code and message will be assigned.
+ *
+ * @param rdp_inst
+ *     The FreeRDP client instance handling the RDP connection that failed.
+ *
+ * @param status
+ *     Pointer to the variable that should receive the guac_protocol_status
+ *     value equivalent to the error returned by freerdp_get_last_error().
+ *
+ * @param message
+ *     Pointer to the variable that should receive a static human-readable
+ *     message generally describing the error returned by
+ *     freerdp_get_last_error().
+ */
+static void guac_rdp_translate_last_error(freerdp* rdp_inst,
+        guac_protocol_status* status, const char** message) {
+
+    UINT32 last_error = freerdp_get_last_error(rdp_inst->context);
+    switch (last_error) {
+
+        /* Normal disconnect */
+        case FREERDP_ERROR_NONE:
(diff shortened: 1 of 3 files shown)


src/protocols/rdp/error.c
```diff
@@ -18,14 +18,90 @@
...
+        case FREERDP_ERROR_CONNECT_PASSWORD_MUST_CHANGE:
+        case FREERDP_ERROR_CONNECT_WRONG_PASSWORD:
+        case FREERDP_ERROR_INSUFFICIENT_PRIVILEGES:
+        case FREERDP_ERROR_SECURITY_NEGO_CONNECT_FAILED:
+        case FREERDP_ERROR_SERVER_DENIED_CONNECTION:
+        case FREERDP_ERROR_SERVER_INSUFFICIENT_PRIVILEGES:
+        case FREERDP_ERROR_SERVER_FRESH_CREDENTIALS_REQUIRED:
+            *status = GUAC_PROTOCOL_STATUS_CLIENT_UNAUTHORIZED;
+            *message = "Authentication failure.";
+            break;
```
necouchman, line 80: I guess my only question, here, is if it's worth exposing something more useful to the user about the nature of the failure through Guacamole rather than bundling all of the credential failures into a single status message? I'm thinking of the kind of errors that users would receive if they attempted to connect to a RDP server with something like Microsoft Terminal Services Client (mstsc) versus Guacamole - I think mstsc exposes more specific information about these failures to the users - e.g. if your account is locked out, your told that your account is locked out.

mike-jumper: There are currently only two authentication-related status codes:

* `GUAC_PROTOCOL_STATUS_CLIENT_UNAUTHORIZED` - The credentials are wrong or the user has not yet been authorized (has not logged in).
* `GUAC_PROTOCOL_STATUS_CLIENT_FORBIDDEN` - The operation is simply denied (regardless of whether the credentials are correct).

If we define additional status codes to cover things like expired credentials, perhaps something like `GUAC_PROTOCOL_STATUS_CLIENT_EXPIRED` and `GUAC_PROTOCOL_STATUS_CLIENT_LOCKED`, we can provide more user-visible feedback. Before we do that, it'd be good to understand why there are different codes within RDP / FreeRDP for "password expired", "password _certainly_ expired", "password must change", "account expired", etc. as those statuses seem awfully ... identical.

As for administrator-visible feedback, I definitely think it's reasonable to add more logging around this.

necouchman: Okay - I'm fine to leave this off of here, now, and come back later and fix it up, later, if you think it would be better scoped elsewhere.

mike-jumper: New status codes, yes, I think it'd be best to save that for a different change.

Logging, I agree we should do this. I'll add that real quick.

mike-jumper: OK - I've split things apart more and added more verbose messages.

necouchman (approved):

PR #311, opened 2020-11-03
GUACAMOLE-221: Handle FREERDP_ERROR_* constants only when defined.

mike-jumper:
The changes from #308 broke the build against FreeRDP 2.0.0-rc2 and earlier. This is because earlier versions of FreeRDP do not support all the constants used within `error.c`.

This change wraps affected constants in `#ifdef` / `#endif`, handling that specific error constant only if it's actually defined.


necouchman (approved):
That's a lot of `ifdef`s.

PR #312, opened 2020-11-06
GUACAMOLE-1201: Fix audio input issue on mme

myjimmy:
I have fixed the audio input issues (noise & sped up) on MME.

### DIFF
--- a/src/protocols/rdp/channels/audio-input/audio-buffer.c
+++ b/src/protocols/rdp/channels/audio-input/audio-buffer.c
@@ -19,6 +19,7 @@
 
 #include "channels/audio-input/audio-buffer.h"
 #include "rdp.h"
+#include "client.h"
 
 #include <guacamole/client.h>
 #include <guacamole/protocol.h>
@@ -120,6 +121,11 @@ void guac_rdp_audio_buffer_begin(guac_rdp_audio_buffer* audio_buffer,
         void* data) {
 
     pthread_mutex_lock(&(audio_buffer->lock));
+    
+    /* Create a thread */
+    audio_buffer->stop_thread = 0;
+    pthread_create(&audio_buffer->send_thread, NULL,
+                   guac_rdp_audio_send_thread, audio_buffer);
 
     /* Reset buffer state to provided values */
     audio_buffer->bytes_written = 0;
@@ -130,6 +136,19 @@ void guac_rdp_audio_buffer_begin(guac_rdp_audio_buffer* audio_buffer,
     audio_buffer->packet_size = packet_frames
                               * audio_buffer->out_format.channels
                               * audio_buffer->out_format.bps;
+
+    /* Calculate the interval to send each packet in ms.
+     * We need to set half of the time duration designated
+     * by packet frames according to Nyquist's theorem.
+     */
+    audio_buffer->packet_interval = packet_frames
+                                  * 1000
+                                  / audio_buffer->out_format.rate
+                                  / 2;
+    if (audio_buffer->packet_interval == 0) {
+        audio_buffer->packet_interval =
+            GUAC_RDP_AUDIO_INPUT_SEND_DEFAULT_INTERVAL;
+    }
 
     /* Allocate new buffer */
     free(audio_buffer->packet);
@@ -233,6 +252,97 @@ static int guac_rdp_audio_buffer_read_sample(
 void guac_rdp_audio_buffer_write(guac_rdp_audio_buffer* audio_buffer,
         char* buffer, int length) {
 
+    pthread_mutex_lock(&(audio_buffer->lock));
+
+    /* Ignore packet if there is no buffer */
+    if (audio_buffer->packet_size == 0 || audio_buffer->packet == NULL) {
+        pthread_mutex_unlock(&(audio_buffer->lock));
+        return;
+    }
+
+    /* Allocate the new buffer. */
+    audio_stream_list *new_stream = (audio_stream_list*)malloc(sizeof(audio_stream_list));
+    new_stream->stream_data = malloc(length);
+
+    /* Write the stream data to the new buffer. */
+    new_stream->length = length;
+    new_stream->next_stream = NULL;
+    memcpy(new_stream->stream_data, buffer, new_stream->length);
+    
+    /* Add the new buffer into the list. */
+    if (audio_buffer->first_stream_list == NULL) {
+        audio_buffer->first_stream_list = new_stream;
+        audio_buffer->last_stream_list = new_stream;
+    }
+    else {
+        audio_buffer->last_stream_list->next_stream = new_stream;
+        audio_buffer->last_stream_list =
+            (audio_stream_list*)audio_buffer->last_stream_list->next_stream;
+    }
+
+    pthread_mutex_unlock(&(audio_buffer->lock));
+
+}
+
+void guac_rdp_audio_buffer_end(guac_rdp_audio_buffer* audio_buffer) {
+
+    /* Set the flag to stop the thread */
+    pthread_mutex_lock(&(audio_buffer->lock));
+    audio_buffer->stop_thread = 1;
+    pthread_mutex_unlock(&(audio_buffer->lock));
+
+    /* Stop the thread */
+    pthread_join(audio_buffer->send_thread, NULL);
+
+    pthread_mutex_lock(&(audio_buffer->lock));
+
+    /* The stream is now closed */
+    guac_rdp_audio_buffer_ack(audio_buffer,
+            "CLOSED", GUAC_PROTOCOL_STATUS_RESOURCE_CLOSED);
+
+    /* Unset user and stream */
+    audio_buffer->user = NULL;
+    audio_buffer->stream = NULL;
+
+    /* Reset buffer state */
+    audio_buffer->bytes_written = 0;
+    audio_buffer->packet_size = 0;
+    audio_buffer->flush_handler = NULL;
+
+    /* Reset I/O counters */
+    audio_buffer->total_bytes_sent = 0;
+    audio_buffer->total_bytes_received = 0;
+
+    /* Free packet (if any) */
+    free(audio_buffer->packet);
+    audio_buffer->packet = NULL;
+
+    /* Free stream list */
+    audio_stream_list* current = audio_buffer->first_stream_list;
+    while (current != NULL) {
+
+        audio_buffer->first_stream_list =
+            audio_buffer->first_stream_list->next_stream;
+
+        free(current->stream_data);
+        free(current);
+
+        current = audio_buffer->first_stream_list;
+    }
+
+    pthread_mutex_unlock(&(audio_buffer->lock));
+
+}
+
+void guac_rdp_audio_buffer_free(guac_rdp_audio_buffer* audio_buffer) {
+    pthread_mutex_destroy(&(audio_buffer->lock));
+    free(audio_buffer->packet);
+    free(audio_buffer);
+}
+
+void guac_rdp_audio_buffer_send(guac_rdp_audio_buffer* audio_buffer,
+        char* buffer, int length) {
+
     int16_t sample;
 
     pthread_mutex_lock(&(audio_buffer->lock));
@@ -242,6 +352,7 @@ void guac_rdp_audio_buffer_write(guac_rdp_audio_buffer* audio_buffer,
         pthread_mutex_unlock(&(audio_buffer->lock));
         return;
     }
+    pthread_mutex_unlock(&(audio_buffer->lock));
 
     int out_bps = audio_buffer->out_format.bps;
 
@@ -276,6 +387,9 @@ void guac_rdp_audio_buffer_write(guac_rdp_audio_buffer* audio_buffer,
             /* Reset buffer in all cases */
             audio_buffer->bytes_written = 0;
 
+            /* Wait in a specified interval. */
+            guac_timestamp_msleep(audio_buffer->packet_interval);
+
         }
 
     } /* end packet write loop */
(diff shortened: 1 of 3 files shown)


necouchman:
@myjimmy: Please tag the pull request with the JIRA issue, as well.

myjimmy:
> @myjimmy: Please tag the pull request with the JIRA issue, as well.

@necouchman, what does it mean? Sorry, I didn't understand.

myjimmy:
Sorry, I understand. I have just modified the title of the pull request.

mike-jumper (changes requested):
Please look through these changes and try to document things with an eye toward what would be helpful for someone to see when they read through your code.

For example:

```c
/**
 * A structure of list to take the audio buffers.
 */
typedef struct audio_stream_list {
    /**
     * The data of audio buffer
     */
    char*  data;
...
```

The documentation of the structure and its members should clarify things at a high level. You say "the data of the audio buffer", but what _is_ that? What data? "A structure of list to take the audio buffers" ... but why? How? What do you mean by take? Reading through the usage, it seems clear that it's a linked list of received audio packets, but certainly it would be better to describe it as such, and to clarify what that data is rather than just call it "the data".

Similarly:

```c
            /* Wait in a specified interval. */
            guac_timestamp_msleep(GUAC_RDP_AUDIO_INPUT_SEND_INTERVAL);
```

Yes, that's what `guac_timestamp_msleep()` does, but there must be something that you can capture within the documentation that adds information and context beyond restating the code in English form.

Consider the following semi-contrived example:

```c
/* Add 2 to pointer */
ptr += 2;
```

Technically correct, sure. Useful, nope. No one reading the above will be able to tell why this was done or whether it really is the correct way to do what it's doing.

Compare that to:

```c
/* Advance to next element */
current += 2;
```

Now the reader knows what the intent of that line is, _even if it's wrong_. If someone reads through this later and notices "hey, that's not the correct element size", then they can correct it knowing what that code is _supposed_ to do.

The same concept goes in principle with variable naming:

```c
    audio_stream_list *tmp = (audio_stream_list*)malloc(sizeof(audio_stream_list));
```

Presumably, this variable has a concrete purpose and there is a name which would capture that purpose better than `tmp`.

mike-jumper:
Regarding the above, I'd like to also point out the commit messages themselves:

Screenshot of commit messages

Just like the comments, these should be aimed at capturing what someone reviewing the code later will need to understand about the context of your changes.

Something like "Fixed the audio issues on MME" doesn't capture what is being fixed and why. If someone goes through this in a `git bisect` later, finds that this is the change that broke something, they are going to have quite a bit of further research ahead of them before they find out what issue that is and why these changes were made.

A messages like "Added some comments" is only negligibly better that a message like "Changed some code". What were you doing and why? If the change doesn't feel significantly separate your earlier commits, it can just be squashed into the most relevant commit. If it does, then there should be some context that can be provided. Presumably those comments were to document something specific?

myjimmy:
@mike-jumper, Thanks for your reply.
In addition to the comments, the coding style, and the commit message that you pointed out, aren't there technically any other problems in my code?

mike-jumper:
> In addition to the comments, the coding style, and the commit message that you pointed out, aren't there technically any other problems in my code?

I'll check, but ease of reviewing is part of the reason I'd like that addressed.

src/protocols/rdp/channels/audio-input/audio-buffer.c
```diff
@@ -242,6 +338,7 @@ void guac_rdp_audio_buffer_write(guac_rdp_audio_buffer* audio_buffer,
         pthread_mutex_unlock(&(audio_buffer->lock));
         return;
     }
+    pthread_mutex_unlock(&(audio_buffer->lock));
```
mike-jumper, line 341: This mutex is documented as guarding all access to `audio_buffer`. Unlocking things here and then operating on the `audio_buffer` without holding the mutex looks like a threadsafety issue.

If you are certain this is safe, can you describe how that has been verified?

src/protocols/rdp/channels/audio-input/audio-buffer.c
```diff
@@ -233,6 +239,96 @@ static int guac_rdp_audio_buffer_read_sample(
...
+    /* Add the new buffer into the list. */
+    if (audio_buffer->first_stream_list == NULL) {
+        audio_buffer->first_stream_list = tmp;
+        audio_buffer->last_stream_list = tmp;
+    }
+    else {
+        audio_buffer->last_stream_list->next = (void*)tmp;
+        audio_buffer->last_stream_list =
+            (audio_stream_list*)audio_buffer->last_stream_list->next;
+    }
```
mike-jumper, line 268: If new memory will be allocated for every received `blob` of audio data, there needs to be some reasonable upper bound on how much pending audio data will be stored.

src/protocols/rdp/channels/audio-input/audio-buffer.h
```diff
@@ -66,6 +66,27 @@ typedef struct guac_rdp_audio_format {
...
+
+    /**
+     * The length of audio buffer
+     */
+    int    length;
+
+    /**
+     * The pointer of the next audio buffer
+     */
+    void*  next;
```
mike-jumper, line 86: This should be a pointer to a concrete type, presumably `audio_stream_list` (which would then need a forward declaration). We should only be using `void*` for pointers that truly are intended to point to arbitrary data.

src/protocols/rdp/channels/audio-input/audio-buffer.h
```diff
@@ -273,5 +310,10 @@ void guac_rdp_audio_buffer_end(guac_rdp_audio_buffer* audio_buffer);
  */
 void guac_rdp_audio_buffer_free(guac_rdp_audio_buffer* audio_buffer);
 
+/**
+ * This thread handler sends the audio input stream data to the remote server.
+ */
+void* guac_rdp_audio_send_thread(void* data);
```
mike-jumper, line 316: The parameter and return value need to be documented here.

Also, as mentioned in the earlier review regarding code comments, the documentation describing this function should also capture the high-level use. In this case, the fact that the thread will flush data at a regular, constant rate, and that this rate is intended to prevent the remote audio buffer from running out of space seems like it would be good to note here. Lacking that information, it is unclear why a timed and threaded approach is needed, and future changes may well revert this in an attempt to make things more efficient without realizing the consequences.

myjimmy:
@mike-jumper, Please let me know how to modify the code and the comment and the commit message in the pull request.

src/protocols/rdp/client.h
```diff
@@ -87,6 +87,11 @@
  */
 #define GUAC_RDP_AUDIO_BPS 16
 
+/**
+ * Interval to send the audio input packet to a remote server, in ms.
+ */
+#define GUAC_RDP_AUDIO_INPUT_SEND_INTERVAL 5
```
mike-jumper, line 93: Rather than specify an arbitrary static value, perhaps the interval should be derived from the number of audio frames that the RDP server dictates should be sent in each PDU?

myjimmy:
@mike-jumper, I have just modified the code, comments, and commits.
Could you check the pull request again?

src/protocols/rdp/channels/audio-input/audio-buffer.c
```diff
@@ -276,49 +387,54 @@ void guac_rdp_audio_buffer_write(guac_rdp_audio_buffer* audio_buffer,
             /* Reset buffer in all cases */
             audio_buffer->bytes_written = 0;
 
+            /* Wait in a specified interval. */
+            guac_timestamp_msleep(audio_buffer->packet_interval);
```
mike-jumper, line 391: As mentioned in the previous review:

> Yes, that's what `guac_timestamp_msleep()` does, but there must be something that you can capture within the documentation that adds information and context beyond restating the code in English form.

src/protocols/rdp/channels/audio-input/audio-buffer.c
```diff
@@ -120,6 +121,11 @@ void guac_rdp_audio_buffer_begin(guac_rdp_audio_buffer* audio_buffer,
         void* data) {
 
     pthread_mutex_lock(&(audio_buffer->lock));
+    
+    /* Create a thread */
+    audio_buffer->stop_thread = 0;
+    pthread_create(&audio_buffer->send_thread, NULL,
+                   guac_rdp_audio_send_thread, audio_buffer);
```
mike-jumper, line 128: Please make sure that your comments provide more context than simply restating the code itself in English.

src/protocols/rdp/channels/audio-input/audio-buffer.h
```diff
@@ -66,6 +66,29 @@ typedef struct guac_rdp_audio_format {
...
+/**
+ * A structure of linked list to store the audio stream periodically
+ * sent by the client.
+ */
+typedef struct audio_stream_list audio_stream_list;
+struct audio_stream_list {
+    /**
+     * The pointer of audio stream
+     */
+    char* stream_data;
```
mike-jumper, line 78: Previously, this read:

```c
    /**
     * The data of audio buffer
     */
    char*  data;
```

From the previous review:

> The documentation of the structure and its members should clarify things at a high level. You say "the data of the audio buffer", but what _is_ that? What data?

I think the above still applies here. It is unclear what this structure member actually is. The documentation should clarify things at a high level.

src/protocols/rdp/client.h
```diff
@@ -87,6 +87,11 @@
  */
 #define GUAC_RDP_AUDIO_BPS 16
 
+/**
+ * The default interval to send the audio input packet to a remote server, in ms.
+ */
+#define GUAC_RDP_AUDIO_INPUT_SEND_DEFAULT_INTERVAL 5
```
mike-jumper, line 93: I don't think this should be needed if the outbound audio data rate is correctly throttled based on the server-requested frames per packet.

src/protocols/rdp/channels/audio-input/audio-buffer.c
```diff
@@ -233,6 +252,97 @@ static int guac_rdp_audio_buffer_read_sample(
...
+    /* Add the new buffer into the list. */
+    if (audio_buffer->first_stream_list == NULL) {
+        audio_buffer->first_stream_list = new_stream;
+        audio_buffer->last_stream_list = new_stream;
+    }
+    else {
+        audio_buffer->last_stream_list->next_stream = new_stream;
+        audio_buffer->last_stream_list =
+            (audio_stream_list*)audio_buffer->last_stream_list->next_stream;
+    }
```
mike-jumper, line 281: As mentioned in the previous review:

> If new memory will be allocated for every received `blob` of audio data, there needs to be some reasonable upper bound on how much pending audio data will be stored.

mike-jumper (changes requested):
Overall:

* Please make sure that your comments provide more context than simply restating the code itself in English. I believe I've commented on most of these occurrences, but there may be others.
* I don't think we should use an approach which involves polling and busy-looping while waiting for inbound audio data, nor which potentially uses a static default send interval that is not determined by the required packet size. If the problem is that outbound audio data may instantaneously exceed the implicit maximum rate, we should be able to throttle the outbound data rate in a consistent and correct way.
* This approach does not impose a bound on the inbound data from the client. It should, both to avoid excessive resource use and to avoid too great a time difference between local and remote audio streams.

src/protocols/rdp/channels/audio-input/audio-buffer.h
```diff
@@ -273,5 +317,22 @@ void guac_rdp_audio_buffer_end(guac_rdp_audio_buffer* audio_buffer);
...
+ * receiving the audio input stream from the client, the missing some audio stream
+ * occurs due to the lack of remote audio buffer.
+ * So, to prevent the remote audio buffer from running out of space seems,
+ * we need to use the thread to send the audio stream in the uniform interval.
+ *
+ * @param data
+ *     The pointer to the guac_rdp_audio_buffer object maintained by RDP session.
+ *
+ * @return
+ *     NULL. This is a normal format of the general thread callback functions.
```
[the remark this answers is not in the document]
myjimmy: When defining the thread callback function and using it, the function generally returns a NULL pointer.
I documented this as the above comment.

src/protocols/rdp/channels/audio-input/audio-buffer.h
```diff
@@ -145,6 +168,27 @@ typedef struct guac_rdp_audio_buffer {
...
+    /**
+     * The flag to stop the thread.
+     */
+    int stop_thread;
+
+    /**
+     * The list to take the audio buffers.
+     */
+    audio_stream_list* first_stream_list;
+    audio_stream_list* last_stream_list;
```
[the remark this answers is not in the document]
myjimmy: I'll modify this.

src/protocols/rdp/channels/audio-input/audio-buffer.c
```diff
@@ -131,6 +137,19 @@ void guac_rdp_audio_buffer_begin(guac_rdp_audio_buffer* audio_buffer,
...
                               * audio_buffer->out_format.bps;
 
+    /* Calculate the interval to send each packet in ms.
+     * We need to set half of the time duration designated
+     * by packet frames according to Nyquist's theorem.
+     */
+    audio_buffer->packet_interval = packet_frames
+                                  * 1000
+                                  / audio_buffer->out_format.rate
+                                  / 2;
```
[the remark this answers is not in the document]
myjimmy: Because the time interval of sending audio packets isn't even and the time interval occasionally exceeds one packet length, the choppy issue occurred. 
So, I think that we need to make sure that the intervals will at least be one half of one packet length.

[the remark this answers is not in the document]
mike-jumper: Sure. That being the case:

* That would be the reasoning to cite here, not Nyquist's theorem.
* This sounds like a workaround for a problem that results from not properly calculating the send interval in the first place. Rather than use a constant interval, falling back to a hard-coded 5ms if the interval is sufficiently small, shouldn't the interval be determined dynamically based on (1) how much time has elapsed since the last time a packet was sent and (2) how much audio data is about to be sent?

mike-jumper:
Reattempted and addressed via #337. Thanks for finding this and the initial stab at the fix, @myjimmy.

myjimmy:
@mike-jumper Thanks for your reply.
And I checked your fix (#337).
This is wonderful. I learned many things from you.

PR #314, opened 2020-11-30
GUACAMOLE-1110: Create a proper dedicated service group

benrubson:
Hi,

This PR is a slight modification to #292 which has been merged in `staging/1.3.0` (thus my PR against this branch).
It creates a dedicated group for the `guacd` service user.
So that it is not assigned to the default `users` group, among other possible users...

Thank you 👍

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -115,7 +115,9 @@ HEALTHCHECK --interval=5m --timeout=5s CMD nc -z 127.0.0.1 4822 || exit 1
 
 # Create a new user guacd
 ARG UID=1000
-RUN useradd --system --create-home --shell /usr/sbin/nologin --uid $UID --no-user-group guacd
+ARG GID=1000
+RUN groupadd --gid $GID guacd
+RUN useradd --system --create-home --shell /usr/sbin/nologin --uid $UID --gid $GID guacd
 
 # Run with user guacd
 USER guacd


benrubson:
@IT-Experte @mike-jumper @necouchman could you please review / merge this, which completes #292 in `staging/1.3.0` ?
Note that same dedicated user/group creation has been applied in https://github.com/apache/guacamole-client/pull/554.
Thank you very much 👍

Dockerfile
```diff
@@ -115,7 +115,8 @@ HEALTHCHECK --interval=5m --timeout=5s CMD nc -z 127.0.0.1 4822 || exit 1
 
 # Create a new user guacd
 ARG UID=1000
-RUN useradd --system --create-home --shell /usr/sbin/nologin --uid $UID --no-user-group guacd
+RUN groupadd --gid $UID guacd
+RUN useradd --system --create-home --shell /usr/sbin/nologin --uid $UID --gid $UID guacd
```
mike-jumper, line 119: I don't think the GID should be set to the UID (or that it should be assumed that doing so is safe). Perhaps this should be its own `ARG`?

mike-jumper (changes requested):

Dockerfile
benrubson: Many thanks for your feedback @mike-jumper, updated accordingly :+1:

benrubson: Note that there's no risk for both `UID` and `GID` to have the same numerical value (`1000` here).

mike-jumper: Well ... there is no risk if they _happen_ to have the same value, or if the same value is intentionally chosen for both the UID and the GID after that value was been confirmed to be available within both UID and GID pools.

The important thing is to not have the image build written to directly use a UID as if it were a GID.

benrubson: OK, both UID and GID with a `1000` value are free for use in the image (the free range around 1000 is rather large).
And both are now identified with their own `ARG`.
Should then be OK, or did I perhaps miss something ?
Thank you again 👍

mike-jumper (approved):

benrubson:
Many thanks @mike-jumper 👍

PR #317, opened 2020-12-18
GUACAMOLE-1186: Maintain the clipboard content of the remote server.

myjimmy:
When users copy any data within the remote server, the guacamole maintains the clipboard content of the remote server unless users copy any text in the client computer.

### DIFF
--- a/src/protocols/rdp/channels/cliprdr.c
+++ b/src/protocols/rdp/channels/cliprdr.c
@@ -283,7 +283,8 @@ static UINT guac_rdp_cliprdr_format_list(CliprdrClientContext* cliprdr,
     if (guac_rdp_cliprdr_format_supported(format_list, CF_TEXT))
         return guac_rdp_cliprdr_send_format_data_request(cliprdr, CF_TEXT);
 
-    /* Ignore any unsupported data */
+    /* Suppose that any unsupported data is raw data (in this case, CF_RAW) */
+    guac_rdp_cliprdr_send_format_data_request(cliprdr, CF_RAW);
     guac_client_log(clipboard->client, GUAC_LOG_DEBUG, "Ignoring unsupported "
             "clipboard data. Only Unicode and text clipboard formats are "
             "currently supported.");
@@ -437,11 +438,15 @@ static UINT guac_rdp_cliprdr_format_data_response(CliprdrClientContext* cliprdr,
          * Either incorrect values are (somehow) being stored, or support for
          * the format indicated by that value is incomplete and must be added
          * here. The values which may be stored within requested_format are
-         * completely within our control. */
+         * completely within our control.
+         * But because some data rather than text was copied, it needs to reset 
+         * the current local clipboard. */
         default:
             guac_client_log(clipboard->client, GUAC_LOG_DEBUG, "Requested "
                     "clipboard data in unsupported format (0x%X).",
                     clipboard->requested_format);
+            guac_common_clipboard_reset(clipboard->clipboard, "text/plain");
+            guac_common_clipboard_send(clipboard->clipboard, clipboard->client);
             return CHANNEL_RC_OK;
 
     }
@@ -612,6 +617,11 @@ int guac_rdp_clipboard_end_handler(guac_user* user, guac_stream* stream) {
      * the data that was received */
     guac_rdp_clipboard* clipboard = rdp_client->clipboard;
     if (clipboard == NULL)
+        return 0;
+
+    /* Don't need to send it to the remote server, if the local clipboard is
+     * empty */
+    if (clipboard->clipboard->length == 0)
         return 0;
 
     /* Terminate clipboard data with NULL */


src/protocols/rdp/channels/cliprdr.c
```diff
@@ -283,7 +283,8 @@ static UINT guac_rdp_cliprdr_format_list(CliprdrClientContext* cliprdr,
     if (guac_rdp_cliprdr_format_supported(format_list, CF_TEXT))
         return guac_rdp_cliprdr_send_format_data_request(cliprdr, CF_TEXT);
 
-    /* Ignore any unsupported data */
+    /* Suppose that any unsupported data is raw data (in this case, CF_RAW) */
+    guac_rdp_cliprdr_send_format_data_request(cliprdr, CF_RAW);
```
mike-jumper, line 287: Why?

myjimmy: guac_rdp_cliprdr_format_list() function is called when copying any data in the remote server.
In this function, we need to tell the remote server whether the remote client supports the clipboard format of the remote clipboard.
Users can copy any kind of data including text in the remote server.
Because I thought that CF_RAW indicates any kind of clipboard data, I selected the clipboard format as CF_RAW.

mike-jumper: If the format of the data in the remote clipboard is not supported (is not handled by one of the `if` statements above), why are we requesting that the RDP server send us that data?

myjimmy: > why are we requesting that the RDP server send us that data?
Because we need to reset the local clipboard. That's, to get the opportunity to reset the local clipboard when users copy any data other than text.

The added new code of `guac_rdp_cliprdr_format_list()` function let the remote server to call `guac_rdp_cliprdr_format_data_response()` function according to **"Remote Desktop Protocol: Clipboard Virtual Channel Extension"**.
If users copy any kind of data other than text, so we can use `guac_rdp_cliprdr_format_data_response()` function to reset the local clipboard.

necouchman: Yes, but why reset the local clipboard? Why is resetting the local clipboard the solution to not erasing what's on the remote clipboard?

myjimmy: In the Apache Guacamole, the process related to the clipboard has one cyclic loop.
If the remote clipboard is changed, the content of the remote clipboard is reflected in the local clipboard.
This produces one regression to reflect the local clipboard to the remote clipboard, again.

So, I thought that we have to reset the local clipboard when copying any data other than text in the remote server.
Then, we can check the length of the local clipboard in `guac_rdp_clipboard_end_handler()` function to decide whether we maintain the clipboard content of the remote server.

myjimmy: Also, you can see the following image for the workflow when copying any data in the remote server:

image

src/protocols/rdp/channels/cliprdr.c
```diff
@@ -614,6 +619,11 @@ int guac_rdp_clipboard_end_handler(guac_user* user, guac_stream* stream) {
     if (clipboard == NULL)
         return 0;
 
+    /* Don't need to send it to the remote server, if the local clipboard is
+     * empty */
+    if (clipboard->clipboard->length == 0)
+        return 0;
```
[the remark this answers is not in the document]
myjimmy: No, it can never happen. I think that we don't need to update the remote clipboard when there is no data in the local clipboard.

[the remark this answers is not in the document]
mike-jumper: It looks like it can:

https://www.windowscentral.com/how-clear-clipboard-data-shortcut-windows-10

Testing the suggestion at the link above (running `echo off | clip` from within the Windows command line), that does indeed clear the clipboard, and Guacamole currently handles that correctly.

mike-jumper:
I'm not against altering things such that things work cleanly even if you attempt to copy data that Guacamole itself does not currently handle, but I don't think this is the correct approach. This will remove the ability to clear the clipboard, and looks like it will result in requesting the RDP server send us data that we will simply drop (it would be better to just not request it).

You mentioned:

> Because we need to reset the local clipboard. That's, to get the opportunity to reset the local clipboard when users copy any data other than text.

There is no need to request the RDP server send us _anything_ in order to do that. If you want to reset the local clipboard, you can just reset it. That said, I don't think we should be resetting the local clipboard as a means of achieving what you're trying to achieve.

I'd think the solution to the difficulty you're encountering with images would be:

* Don't clear the clipboard if unknown data is received (behave as if nothing has changed)
* Don't send the local clipboard to the RDP server unless it's changed locally.

I believe the server is already doing the former, and the client _should_ actually already be doing the latter. If either of those are not the case, then perhaps correcting that would correct this cleanly?

mike-jumper (changes requested):
See comments above. It looks like:

* We shouldn't be clearing the local clipboard as a means to achieve this.
* We shouldn't be requesting the RDP server to send us data unless we actually need that data.
* Server-side changes might not be necessary, and further investigation might be needed to determine why things are not already behaving this way.

In past discussion, I mentioned:

> I don't believe this is a bug. ... When you switch to another tab, application, etc. on the client side, Guacamole will attempt to resync the remote clipboard with the contents of the client clipboard. This is intentional, however since the client clipboard does not contain the previously-copied image, it has the counterintuitive effect of clearing what was copied.

This may not be entirely correct. The client _shouldn't_ be resyncing the local clipboard unless the local clipboard has changed since the last time it was synced.

src/protocols/rdp/channels/cliprdr.c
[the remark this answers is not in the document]
myjimmy: Is this related to this pull request?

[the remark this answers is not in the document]
mike-jumper: Yes. In response to Nick asking noting that this change would remove the ability for a user to clear the remote clipboard, you said:

> No, it can never happen. I think that we don't need to update the remote clipboard when there is no data in the local clipboard.

On the contrary, from the above link:

* It can happen.
* It is desirable to be able to do so for the sake of privacy.

If the local clipboard is intentionally cleared, the remote clipboard needs to be cleared, as well.

PR #321, opened 2020-12-28
GUACAMOLE-1133: initialize GCrypt in VNC protocol prior to client start-up

necouchman:
This small change does some initialization on the GCrypt library used by libvncclient prior to the VNC client start-up. In my testing this looks to resolve the GUACAMOLE-1133 issue, which deals with failures to connect to Apple Remote Desktop systems. Further testing would be greatly appreciated. I'm also not 100% sure what options we should be using to initialize GCrypt - this is a very basic initialization of the library and then some secure memory.

### DIFF
--- a/src/common-ssh/ssh.c
+++ b/src/common-ssh/ssh.c
@@ -140,11 +140,21 @@ static void guac_common_ssh_openssl_free_locks(int count) {
 int guac_common_ssh_init(guac_client* client) {
 
 #ifdef LIBSSH2_USES_GCRYPT
-    /* Init threadsafety in libgcrypt */
-    gcry_control(GCRYCTL_SET_THREAD_CBS, &gcry_threads_pthread);
-    if (!gcry_check_version(GCRYPT_VERSION)) {
-        guac_client_log(client, GUAC_LOG_ERROR, "libgcrypt version mismatch.");
-        return 1;
+    
+    if (!gcry_control(GCRYCTL_INITIALIZATION_FINISHED_P)) {
+    
+        /* Init threadsafety in libgcrypt */
+        gcry_control(GCRYCTL_SET_THREAD_CBS, &gcry_threads_pthread);
+        
+        /* Initialize GCrypt */
+        if (!gcry_check_version(GCRYPT_VERSION)) {
+            guac_client_log(client, GUAC_LOG_ERROR, "libgcrypt version mismatch.");
+            return 1;
+        }
+
+        /* Mark initialization as completed. */
+        gcry_control(GCRYCTL_INITIALIZATION_FINISHED, 0);
+    
     }
 #endif
 
--- a/src/protocols/vnc/vnc.c
+++ b/src/protocols/vnc/vnc.c
@@ -48,11 +48,21 @@
 #include <guacamole/timestamp.h>
 #include <guacamole/wol.h>
 #include <rfb/rfbclient.h>
+#include <rfb/rfbconfig.h>
 #include <rfb/rfbproto.h>
+
+#ifdef LIBVNCSERVER_WITH_CLIENT_GCRYPT
+#include <errno.h>
+#include <gcrypt.h>
+#endif
 
 #include <stdlib.h>
 #include <string.h>
 #include <time.h>
+
+#ifdef LIBVNCSERVER_WITH_CLIENT_GCRYPT
+GCRY_THREAD_OPTION_PTHREAD_IMPL;
+#endif
 
 char* GUAC_VNC_CLIENT_KEY = "GUAC_VNC";
 
@@ -133,6 +143,27 @@ rfbClient* guac_vnc_get_client(guac_client* client) {
     /* TLS Locking and Unlocking */
     rfb_client->LockWriteToTLS = guac_vnc_lock_write_to_tls;
     rfb_client->UnlockWriteToTLS = guac_vnc_unlock_write_to_tls;
+#endif
+    
+#ifdef LIBVNCSERVER_WITH_CLIENT_GCRYPT
+    
+    /* Check if GCrypt is initialized, do it if not. */
+    if (!gcry_control(GCRYCTL_INITIALIZATION_FINISHED_P)) {
+    
+        guac_client_log(client, GUAC_LOG_DEBUG, "GCrypt initialization started.");
+
+        /* Initialize thread control. */
+        gcry_control(GCRYCTL_SET_THREAD_CBS, &gcry_threads_pthread);
+
+        /* Basic GCrypt library initialization. */
+        gcry_check_version(NULL);
+
+        /* Mark initialization as completed. */
+        gcry_control(GCRYCTL_INITIALIZATION_FINISHED, 0);
+        guac_client_log(client, GUAC_LOG_DEBUG, "GCrypt initialization completed.");
+    
+    }
+    
 #endif
 
     /* Do not handle clipboard and local cursor if read-only */


src/protocols/vnc/vnc.c
```diff
@@ -49,11 +49,17 @@
...
+#include <gcrypt.h>
+#endif
 
 #include <stdlib.h>
 #include <string.h>
 #include <time.h>
 
+/**
+ * The client key to use to identify this VNC client.
+ */
```
mike-jumper, line 62: The documentation for `GUAC_VNC_CLIENT_KEY` exists in `vnc.h`:

https://github.com/apache/guacamole-server/blob/68c5dd1730804409abf96bda43f11e0dafde4224/src/protocols/vnc/vnc.h#L167-L171

src/protocols/vnc/vnc.c
```diff
@@ -133,6 +139,18 @@ rfbClient* guac_vnc_get_client(guac_client* client) {
...
+    
+#ifdef LIBVNCSERVER_WITH_CLIENT_GCRYPT
+    
+    guac_client_log(client, GUAC_LOG_DEBUG, "GCrypt initialization started.");
+    gcry_check_version(NULL);
+    gcry_control(GCRYCTL_SUSPEND_SECMEM_WARN);
+    gcry_control(GCRYCTL_INIT_SECMEM, 16384, 0);
+    gcry_control(GCRYCTL_RESUME_SECMEM_WARN);
+    gcry_control(GCRYCTL_INITIALIZATION_FINISHED, 0);
+    guac_client_log(client, GUAC_LOG_DEBUG, "GCrypt initialization completed.");
```
mike-jumper, line 152: `guac_common_ssh_init()` does similar GCrypt initialization if libssh has been built against GCrypt:

https://github.com/apache/guacamole-server/blob/68c5dd1730804409abf96bda43f11e0dafde4224/src/common-ssh/ssh.c#L142-L149

Which brings a few questions:

* Why `gcry_check_version(NULL)` vs. `gcry_check_version(SOMETHING_SPECIFIC)`?
* There's a good deal of initialization here that we aren't doing in the common SSH code. Is that a problem?
* What happens if GCrypt is initialized twice due to libvncclient and libssh2 having both been built against it?

mike-jumper (changes requested):

src/protocols/vnc/vnc.c
necouchman: Ah, oops.

src/protocols/vnc/vnc.c
necouchman: * At this point I'm not sure if there's a reason to go with `SOMETHING_SPECIFIC` versus `NULL` - unfortunately the documentation on LibVNC integration with GCrypt seems to be sparse, and I'm not sure if there's a specific version of GCrypt that is required for LibVNC to function correctly. The scant examples I can find (vino from Gnome, for example) just use `NULL`. I'm happy to just stick with `GCRYPT_VERSION` for consistency, if that is preferred. However, I'm not sure that necessarily accomplishes anything useful, as it just checks to make sure that the version of GCrypt that you built against is identical to the version you're running with. Maybe that's worthwhile, maybe not? My impression is that specifying a version is more for the purpose of making sure that GCrypt supports the features/algorithms you need for a particular application?
* The GCrypt documentation on initializing the library (https://www.gnupg.org/documentation/manuals/gcrypt/Initializing-the-library.html) talks about initializing secure memory to protect things if the memory gets swapped out to disk. I don't know how much of a concern this is - for either VNC or SSH - it was in the documentation, so I went with it.
* Not sure if there are adverse effects of initializing the library twice, but there is a method for checking on status of initialization, so it's probably worth just doing that to make sure it hasn't already been initialized. That said, I'm not sure what the implications are of initializing it in one place with the thread-safe options (SSH) and not doing so in another place (VNC). So, I might need some guidance on how that should be handled, since one initialization is in the common-ssh code and the other in the VNC protocol code. My familiarity with GCrypt is very, very limited...

necouchman: I went ahead and put the checks in place for initialization, and synced up the init code between VNC and SSH. It's still worth figuring out what the best options are for initializing the library.

src/protocols/vnc/vnc.c
```diff
@@ -49,11 +49,19 @@
 #include <guacamole/wol.h>
 #include <rfb/rfbclient.h>
 #include <rfb/rfbproto.h>
+#ifdef LIBVNCSERVER_WITH_CLIENT_GCRYPT
+#include <asm-generic/errno.h>
```
mike-jumper, line 53: Why this header?

necouchman: The macro for the GCrypt pthread implementation requires it. If you remove it, you get this build error:

```
Making all in src/protocols/vnc
make[2]: Entering directory `/guac/working/guacamole-server/src/protocols/vnc'
  CC       libguac_client_vnc_la-vnc.lo
In file included from vnc.c:53:0:
vnc.c: In function 'gcry_pthread_mutex_init':
vnc.c:61:1: error: 'ENOMEM' undeclared (first use in this function)
 GCRY_THREAD_OPTION_PTHREAD_IMPL;
 ^
vnc.c:61:1: note: each undeclared identifier is reported only once for each function it appears in
make[2]: *** [libguac_client_vnc_la-vnc.lo] Error 1
make[2]: Leaving directory `/guac/working/guacamole-server/src/protocols/vnc'
make[1]: *** [all-recursive] Error 1
make[1]: Leaving directory `/guac/working/guacamole-server'
make: *** [all] Error 2
```

mike-jumper: I believe `asm-generic/errno.h` is the internal, platform-specific header, and this should just be `errno.h`.

necouchman: Aha, makes sense. will fix momentarily.

src/common-ssh/ssh.c
```diff
@@ -140,11 +140,26 @@ static void guac_common_ssh_openssl_free_locks(int count) {
...
+        /* Initialize GCrypt */
+        if (!gcry_check_version(GCRYPT_VERSION)) {
+            guac_client_log(client, GUAC_LOG_ERROR, "libgcrypt version mismatch.");
+            return 1;
+        }
+        
+        /* Initialize secure memory space. */
+        gcry_control(GCRYCTL_SUSPEND_SECMEM_WARN);
+        gcry_control(GCRYCTL_INIT_SECMEM, 16384, 0);
+        gcry_control(GCRYCTL_RESUME_SECMEM_WARN);
```
mike-jumper, line 158: I'm somewhat concerned about allocating secure memory unless we expect that it will be used, as it is apparently a finite resource. From my read of the initialization docs for gcrypt, this will not affect usage of memory in general, but rather only explicit attempts to allocate secure memory using gcrypt's functions for doing so. Looking through the libvncclient (and libssh2) code, I'm so far not seeing any usage of secure memory.

necouchman: Ah, okay, sounds good. Removed.

src/protocols/vnc/vnc.c
```diff
@@ -133,6 +141,27 @@ rfbClient* guac_vnc_get_client(guac_client* client) {
     /* TLS Locking and Unlocking */
     rfb_client->LockWriteToTLS = guac_vnc_lock_write_to_tls;
     rfb_client->UnlockWriteToTLS = guac_vnc_unlock_write_to_tls;
+#endif
+    
+#ifdef LIBVNCSERVER_WITH_CLIENT_GCRYPT
```
mike-jumper, line 146: Where is this (conditionally) defined?

necouchman: rfbconfig.h:
```
/* Enable support for libgcrypt in libvncclient */
#ifndef LIBVNCSERVER_WITH_CLIENT_GCRYPT 
#define LIBVNCSERVER_WITH_CLIENT_GCRYPT  1 
#endif
```

which is then included in rfbproto.h.

mike-jumper: Sure - that's a great idea.

I was initially just looking for where the macro was coming from, as there were no changes to `configure.ac` and my quick grep of libvncclient code didn't reveal anything, but this sounds excellent. It avoids any potential for what we saw with the accidental omission of our own `config.h` in handling of FreeRDP behavior.

necouchman: Should be so, now.

mike-jumper (approved):

PR #322, opened 2021-01-06
Force fast WebP compression

andrewsali:
This PR is based on re-evaluating WebP lossless and is meant for testing and opening a discussion. The background is the question and response (by @mike-jumper  and @necouchman ) here: http://mail-archives.apache.org/mod_mbox/guacamole-dev/202008.mbox/browser

Based on our testing, using a very-fast WebP lossless compression significantly improves the responsiveness (e.g. scrolling, moving windows around) of Guacamole compared to the default settings. This was tested using libwebp > 0.6.1. 

In our configuration there is little latency between the Remote Desktop server and the Guacamole gateway, so most of the latency was due to the compression time for PNG. With using fast WebP this is much improved. 

It might be possible that our results differ from the original evaluation of WebP due to network bandwith improving faster than compute speed over the last few years. 

Caveat: It might also be the case that if there is much latency between the Remote Desktop server and the Gateway, then this change doesn't have a beneficial effect.

This PR is of course just a suggestion, feel free to discard, however it might be worthwhile to test in different scenarios as per our testing the UX is much improved.

### DIFF
--- a/src/libguac/encode-webp.c
+++ b/src/libguac/encode-webp.c
@@ -192,10 +192,10 @@ int guac_webp_write(guac_socket* socket, guac_stream* stream,
         return -1;
 
     /* Add additional tuning */
-    config.lossless = lossless;
-    config.quality = quality;
-    config.thread_level = 1; /* Multi threaded */
-    config.method = 2; /* Compression method (0=fast/larger, 6=slow/smaller) */
+    config.lossless = 1; /* force lossless */
+    config.quality = 30.f; /* prefer fast encoding over size */
+    config.thread_level = 0; /* Disable Multi thread */
+    config.method = 0; /* Compression method (0=fast/larger, 6=slow/smaller) */
 
     /* Validate configuration */
     WebPValidateConfig(&config);
--- a/src/common/surface.c
+++ b/src/common/surface.c
@@ -557,14 +557,8 @@ static int __guac_common_surface_should_use_webp(guac_common_surface* surface,
     if (!guac_client_supports_webp(surface->client))
         return 0;
 
-    /* Calculate the average framerate for the given rect */
-    int framerate = __guac_common_surface_calculate_framerate(surface, rect);
-
-    /* WebP is preferred if:
-     * - frame rate is high enough
-     * - PNG is not more optimal based on image contents */
-    return framerate >= GUAC_COMMON_SURFACE_JPEG_FRAMERATE
-        && __guac_common_surface_png_optimality(surface, rect) < 0;
+    /* Prefer always WebP */
+    return 1;
 
 }
 


src/libguac/encode-webp.c
```diff
@@ -192,10 +192,10 @@ int guac_webp_write(guac_socket* socket, guac_stream* stream,
         return -1;
 
     /* Add additional tuning */
-    config.lossless = lossless;
-    config.quality = quality;
-    config.thread_level = 1; /* Multi threaded */
-    config.method = 2; /* Compression method (0=fast/larger, 6=slow/smaller) */
+    config.lossless = 1; /* force lossless */
```
mike-jumper, line 195: The `guac_webp_write()` function should definitely not ignore its `lossless` parameter entirely, as that's essentially crippling a library function. If lossless WebP should be used across the board wherever WebP would be used, it should be the surface implementation making that decision, not the internals of the generic WebP writer.

src/libguac/encode-webp.c
```diff
@@ -192,10 +192,10 @@ int guac_webp_write(guac_socket* socket, guac_stream* stream,
...
+    config.quality = 30.f; /* prefer fast encoding over size */
+    config.thread_level = 0; /* Disable Multi thread */
```
mike-jumper, line 197: Why?

src/libguac/encode-webp.c
```diff
@@ -192,10 +192,10 @@ int guac_webp_write(guac_socket* socket, guac_stream* stream,
         return -1;
 
     /* Add additional tuning */
-    config.lossless = lossless;
-    config.quality = quality;
-    config.thread_level = 1; /* Multi threaded */
-    config.method = 2; /* Compression method (0=fast/larger, 6=slow/smaller) */
+    config.lossless = 1; /* force lossless */
+    config.quality = 30.f; /* prefer fast encoding over size */
```
mike-jumper, line 196: The `quality` parameter should not be ignored, either. If `30.f` is the desired value here, shouldn't that be what's passed in?

mike-jumper (changes requested):
Please also be sure to open a JIRA issue covering the intent of these changes, tag your commits with the issue (see established commit history), etc.

mike-jumper:
Interesting. It definitely sounds like this is worth reevaluating, though I am skeptical that we should force lossless across the board vs. force WebP across the board where possible.

How did bandwidth consumption compare with PNG vs. forced WebP?

How did performance compare when allowing _both_ lossless and lossy WebP, with lossless being used only where PNG would have been used?

What did you observe in terms of performance with `config.method` set to `0` vs. the previous value of `2`?

andrewsali:
Sorry, this PR is not meant to be a definitive code-base change at all - it is just a suggestion for testing and if others can reproduce the improved speed, then formalize the aspects requested in the code-review.

For encoding speed / bandwidth, what we tested was to take a screenshot PNG (Full HD screen) and see how much time it is to convert to webp using the cwebp utility.

Using m=0, q=30, converting 50 times

```
real    0m2.549s
user    0m2.151s
sys     0m0.404s
```

Output file size: 220K

Using m=2, q=30, converting 50 times:
```
real    0m13.037s
user    0m12.575s
sys     0m0.466s
```

Output file size: 112K

The original PNG file size was also 220K.

Using m=0, q=90, converting 50 times:

```
real    0m3.201s
user    0m2.945s
sys     0m0.261s
```

Output file size: 186K.

Using m=2, q=90, converting 50 times:

```
real    0m15.867s
user    0m15.531s
sys     0m0.333s
```

Output file size: 103K.

So compared to the m=2 setting, the bandwidth requirement seems about double, but encode speeds are about x5.

Given the improvements in bandwidth, it might be worthwhile to go for faster compression. In our subjective experience this has helped substantially. But it would be good for some tests in other deployments to see if it is really a good way to go.

andrewsali:
Just for some back-of-the-envelope calculations:

With WebP lossless q=0, the average compression time is 0.05s, whereas with q=2 it is around 0.25s, however increases bandwidth by around 100KB. So if the bandwidth is larger than 100KB / (0.25 - 0.05)s ~ 500KB/s (4 Megabits / sec), then faster compression should be beneficial.

Given that 4 Megabits / sec bandwidth is nowadays considered fairly slow (usually it's at least an order of magnitude faster), in most cases the faster compression should result in better UX.

necouchman:
@andrewsali: 

> Given that 4 Megabits / sec bandwidth is nowadays considered fairly slow (usually it's at least an order of magnitude faster), in most cases the faster compression should result in better UX.

My only comment about the bandwidth would be that, while 4 Mb/s is on the low end of individual internet connections, if you're aggregating a bunch of those together, you have the potential to hit limits. For example, if 200 users have active sessions, that has at least the potential to be 800 Mb/s, which, even for larger offices, may be at or beyond the limit of those connections. And, besides that, Guacamole may not be the only thing that those 200 users are doing - they probably have other websites opened, and, if these theoretical users are anything like the real ones I encounter day-to-day, they also have Pandora or Spotify or YouTube going, so all of the sudden 800Mb/s isn't quite as much as it used to be, and now I start to figure out how to chop that 4 Mb/s connection down some.

Aside from that, it's also useful to consider the other side of the connection - the servers running Tomcat and/or guacd where those connections terminate. Not just for raw bandwidth sake, but also for cost - for example, if I'm running guacd in a public cloud provider, I'm paying for both the CPU time and the bandwidth, and it may be more cost-effective to purchase one over the other.

All that is not to say that I'm opposed to the changes, just that I agree with Mike that I'm not sure hard-coding some of those parameters is a good idea - there should either be some logic and detection of connection performance that impacts them, or they should be configurable, or, perhaps, both.

jackhadrill:
I have tested these changes and have noticed a significant improvement in terms of CPU usage on the server, and in terms of input latency.

I am interested in developing this further, and wish to provide compression settings through the control panel; however, I am not familiar with Guacamole's code base.

Please could someone point me towards some documentation which describes how I can get a setting or configuration value in the web UI/database as far as the two modified files in this commit?

necouchman:
@jackhadrill: As this pull request stands today, it is not "configurable" - the changes are simply static changes to the code and do not include any way to dynamically change these settings. This would require that both the code being changed, here, be made configurable, and that such settings be added to the Guacamole Client side.

cobmojo:
I would really love to see this developed. Would it be appropriate to start a bug bounty?

necouchman:
@jackhadrill If you're looking to configure this, you'll want to make it a connection parameter. There are several places to look in both the client and server code to see examples of this:

https://github.com/apache/guacamole-server/blob/master/src/protocols/ssh/settings.c
https://github.com/apache/guacamole-server/blob/master/src/protocols/ssh/settings.h
https://github.com/apache/guacamole-server/blob/master/src/protocols/vnc/settings.c
https://github.com/apache/guacamole-server/blob/master/src/protocols/vnc/settings.h

(Basically, any of the protocols you can find the `settings.c` and `settings.h` files and look at those, and then look where those settings are used.)

On the client-side, the connection parameters for each protocol are stored in JSON files:

https://github.com/apache/guacamole-client/tree/master/guacamole-ext/src/main/resources/org/apache/guacamole/protocols

ne0YT:
> Apologies, I wasn't particularly clear in my previous response.
> 
> I am looking to develop the changes both here and within the Guacamole client's code base, but was hoping someone could either describe or point me towards an example within this code base where a client-configured setting is used, such that I can develop this extra functionality in a similar manner.
> 
> In the meantime, I'm happy to do some digging and familiarise myself with this code base. :-)
hey there, just wanted to aks for an update :)

ne0YT:
> I would really love to see this developed. Would it be appropriate to start a bug bounty?

hey there, did you create this somewhere? I might be able to alos put some money into this.

ne0YT:
is there an easy way to patch this into an already running docker to check the difference?

neandrake:
@ne0YT you would not patch this into an already running container. Clone the repo and apply this PR then run the dockerfile which will compile the current state of the source code and run the `guacd` service with this change applied.

---

@andrewsali:
> So compared to the m=2 setting, the bandwidth requirement seems about double, but encode speeds are about x5.

That's quite a hike in bandwidth usage. Is the increase in encoding speed all that noticeable? At a certain point the server's encoding speed will have diminishing returns if the client is unable to receive and decode just as fast, right?

ne0YT:
> @ne0YT you would not patch this into an already running container. Clone the repo and apply this PR then run the dockerfile which will compile the current state of the source code and run the `guacd` service with this change applied.
> 
> @andrewsali:
> 
> > So compared to the m=2 setting, the bandwidth requirement seems about double, but encode speeds are about x5.
> 
> That's quite a hike in bandwidth usage. Is the increase in encoding speed all that noticeable? At a certain point the server's encoding speed will have diminishing returns if the client is unable to receive and decode just as fast, right?

how would one do this when using docker-compose.yml?
I know there is the option to use "build: githubrepo", but how to do this when I use this pull request and only for one service?

neandrake:
@ne0YT I'm not sure what you're referring to. I don't see a `docker-compose.yml` in this repository and your setup is likely to be different from others'.

I set up my own Dockerfile based on their great documentation here:
https://guacamole.apache.org/doc/gug/installing-guacamole.html#building-guacamole-server

If you'd like to reference it I attached it here:
https://github.com/apache/guacamole-server/pull/316#issuecomment-758321806

ne0YT:
> @ne0YT I'm not sure what you're referring to. I don't see a `docker-compose.yml` in this repository and your setup is likely to be different from others'.
> 
> I set up my own Dockerfile based on their great documentation here: https://guacamole.apache.org/doc/gug/installing-guacamole.html#building-guacamole-server
> 
> If you'd like to reference it I attached it here: [#316 (comment)](https://github.com/apache/guacamole-server/pull/316#issuecomment-758321806)

teh idea is to only test this change while using th elatets version of the autobuild while also beaing able to change the parameters myself a bit, so I could do some more tests:
https://hub.docker.com/r/guacamole/guacamole

so it's not possible to patch this manually into a runnign build?

necouchman:
@ne0YT : These changes are to the source code of guacd, which means, in order to test them, guacd has to be re-compiled from source. So, no, you cannot just "patch this manually into a running build" - you have to re-build. You can check out this code, either from this pull request or from the forked repo that the pull request is based on, and build the docker image that way. I suppose you also could compile guacd on a system and place it into an existing container, but that seems like a lot of additional work.

ne0YT:
can someone tell me a reliable way to check the performance differences and data usage?

ne0YT:
@necouchman 
@mike-jumper 
@andrewsali 
Hey there
I can just not find anyone who can do this.
maybe one of you can help me out?

necouchman:
@ne0YT I cannot help you with determining performance differences, here - my utilization of Guacamole is pretty routine administrative access to servers via RDP and SSH, and does not drive performance to any extremes, nor do I have any issues with connections lagging, scrolling issues, etc.

ne0YT:
> @ne0YT I cannot help you with determining performance differences, here - my utilization of Guacamole is pretty routine administrative access to servers via RDP and SSH, and does not drive performance to any extremes, nor do I have any issues with connections lagging, scrolling issues, etc.

I don't have any connection issues.
But I do have performance issues dragging windows (even without them being rendered) and general performance issues on weak thinclients.
As smh KASM Workspace is able to perform better for me using Guacamole as a backend. I hoped there's a possibility to make rdp sessions smoother in official guacamole too.

so I'm generally asking for someone to implement this as an option.
Network speed is not a problem in my scenario.

But I'm really not able to find a developer who is able to do this change.

ne0YT:
> @ne0YT I cannot help you with determining performance differences, here - my utilization of Guacamole is pretty routine administrative access to servers via RDP and SSH, and does not drive performance to any extremes, nor do I have any issues with connections lagging, scrolling issues, etc.

is there any way you can think of, I could change my setup to always use fast WebP for all rdp connections without having to build the software myself so I can just use my docker setup?

necouchman:
@ne0YT: No, without either incorporating the changes from this pull request or some other changes to the source code, and rebuilding the software, there isn't any way you can force a particular image/compression algorithm. Guacamole (specifically guacd) evaluates the performance characteristics of the link between the client and guacd, along with the types of screen updates being done, and automatically selects the best image format based on that and the negotiated/supported image formats. You'd have to either force negotiation down to only WebP (which would require JavaScript changes in the guacamole-client code), or do something like this pull request and force it in guacd. Either one requires source code changes and building the code.

necouchman:
@ne0YT Even if you go this route, the only reliable way to do it is to check out the source code, make your source code modifications, and then re-build the client and client container with the updated source files. If you try to change out the JS files within the web application, while it may work temporarily, you will almost certainly lose it at some point due to a re-deploy of the application.

PR #326, opened 2021-01-15
GUACAMOLE-1263: Mark freed memory as freed prior to calling rfbClientCleanup().

mike-jumper:
Older versions of libvncclient did not free all memory within `rfbClientCleanup()`, but this has been corrected as of their 0.9.12
release. As guacamole-server may well be built against older versions of libvncclient, we can't simply remove the manual `free()` calls, but we should be sure to set any memory that we free ourselves to `NULL` so that `rfbClientCleanup()` does not attempt to free it again.


necouchman (approved):

mike-jumper:
Wow that was a fast merge.

necouchman:
image

PR #327, opened 2021-01-19
GUACAMOLE-1267: Add VNC setting 'disable-server-input'

stefan-esq:
Add a new VNC setting 'disable-remote-input' which attempts to turn the remote input off when set to true.

### DIFF
--- a/src/protocols/vnc/settings.c
+++ b/src/protocols/vnc/settings.c
@@ -91,6 +91,7 @@ const char* GUAC_VNC_CLIENT_ARGS[] = {
     "wol-broadcast-addr",
     "wol-udp-port",
     "wol-wait-time",
+    "disable-server-input",
     NULL
 };
 
@@ -373,6 +374,11 @@ enum VNC_ARGS_IDX {
      */
     IDX_WOL_WAIT_TIME,
 
+    /*
+     * Whether or not to disable the input on the server side. The default is not to disable the input.
+     */
+    IDX_DISABLE_SERVER_INPUT,
+
     VNC_ARGS_COUNT
 };
 
@@ -426,6 +432,11 @@ guac_vnc_settings* guac_vnc_parse_args(guac_user* user,
     settings->read_only =
         guac_user_parse_args_boolean(user, GUAC_VNC_CLIENT_ARGS, argv,
                 IDX_READ_ONLY, false);
+
+    /* Disable server input */
+    settings->disable_server_input =
+            guac_user_parse_args_boolean(user, GUAC_VNC_CLIENT_ARGS, argv,
+                                         IDX_DISABLE_SERVER_INPUT, false);
 
     /* Parse color depth */
     settings->color_depth =
--- a/src/protocols/vnc/settings.h
+++ b/src/protocols/vnc/settings.h
@@ -300,6 +300,11 @@ typedef struct guac_vnc_settings {
      */
     int wol_wait_time;
 
+    /**
+     * Whether or not to disable the input on the server side.
+     */
+    bool disable_server_input;
+
 } guac_vnc_settings;
 
 /**
--- a/src/protocols/vnc/vnc.c
+++ b/src/protocols/vnc/vnc.c
@@ -385,6 +385,17 @@ void* guac_vnc_client_thread(void* data) {
     }
 #endif
 
+    /* Disable remote console (Server input) */
+    if (settings->disable_server_input) {
+        rfbSetServerInputMsg msg;
+        msg.type = rfbSetServerInput;
+        msg.status = 1;
+        msg.pad = 0;
+
+        rfbBool success = WriteToRFBServer(rfb_client, (char*)&msg, sz_rfbSetServerInputMsg);
+        guac_client_log(client, GUAC_LOG_DEBUG, "%s send request to disable server input", success ? "Successfully" : "Failed to");
+    }
+
     /* Set remaining client data */
     vnc_client->rfb_client = rfb_client;
 


src/protocols/vnc/vnc.c
```diff
@@ -385,6 +385,17 @@ void* guac_vnc_client_thread(void* data) {
     }
 #endif
 
+    /* Disable remote console */
+    if (settings->disable_remote_input) {
+        rfbSetServerInputMsg no_remote_console = {
+                .type = rfbSetServerInput, .status = 1, .pad = 0
+        };
+        rfbBool success = WriteToRFBServer(rfb_client,
+                                           (char *) &no_remote_console, sz_rfbSetServerInputMsg);
```
mike-jumper, line 394: Why not leverage `PermitServerInput()`?

src/protocols/vnc/settings.h
```diff
@@ -300,6 +300,11 @@ typedef struct guac_vnc_settings {
      */
     int wol_wait_time;
 
+    /**
+    * Whether or not to write the disable remote input message to the remote host
+    */
+    bool disable_remote_input;
```
mike-jumper, line 306: I think this should be described at more of a high level. Rather than describe the low-level impact on which VNC protocol messages are sent, this should describe the high-level effect of the setting on session/server behavior.

src/protocols/vnc/settings.c
```diff
@@ -373,6 +374,8 @@ enum VNC_ARGS_IDX {
      */
     IDX_WOL_WAIT_TIME,
 
+    IDX_DISABLE_REMOTE_INPUT,
```
mike-jumper, line 377: Please document.

mike-jumper (changes requested):

src/protocols/vnc/settings.h
```diff
@@ -300,6 +300,11 @@ typedef struct guac_vnc_settings {
      */
     int wol_wait_time;
 
+    /**
+    * Whether or not to write the disable remote input message to the remote host
+    */
```
mike-jumper, line 305: I should also point out that something happened here to misalign the block comment.

src/protocols/vnc/vnc.c
stefan-esq: I have no clue, this is based on custom code written by another developer inside our company, I have rewritten most of it so it is configurable and does not always disable the input. As requested I have changed this to PermitServerInput.

src/protocols/vnc/settings.h
stefan-esq: I hope the following message is ok, "Whether or not to disable the input on the server side."

src/protocols/vnc/settings.c
stefan-esq: Added documentation: "Whether or not to disable the input on the server side.".

src/protocols/vnc/settings.h
stefan-esq: Should hopefully be fixed.

src/protocols/vnc/vnc.c
```diff
@@ -385,6 +385,12 @@ void* guac_vnc_client_thread(void* data) {
     }
 #endif
 
+    /* Disable remote console (Server input) */
+    if (settings->disable_server_input) {
+        rfbBool success = PermitServerInput(rfb_client, 1);
+        guac_client_log(client, GUAC_LOG_INFO, "Disabling server input %s.", success ? "succeeded" : "failed");
```
mike-jumper, line 391: There is admittedly not much documentation on the return value of `PermitServerInput()`, but reading through the libvncclient code, the value doesn't appear to indicate whether disabling input succeeded:

```c
/*
 * UltraVNC Server Input Disable
 * Apparently, the remote client can *prevent* the local user from interacting with the display
 * I would think this is extremely helpful when used in a HelpDesk situation
 */
rfbBool PermitServerInput(rfbClient* client, int enabled)
{
    rfbSetServerInputMsg msg;

    if (!SupportsClient2Server(client, rfbSetServerInput)) return TRUE;
    /* enabled==1, then server input from local keyboard is disabled */
    msg.type = rfbSetServerInput;
    msg.status = (enabled ? 1 : 0);
    msg.pad = 0;
    return  (WriteToRFBServer(client, (char *)&msg, sz_rfbSetServerInputMsg) ? TRUE : FALSE);
}
```

(https://github.com/LibVNC/libvncserver/blob/c82b3abb369983cd737190331b897814a580a980/libvncclient/rfbproto.c#L1502-L1517)

Given the above, it looks like:

* If the VNC server supports the relevant message, the return value indicates only if the message was successfully sent, not whether the server honored the message.
* If the VNC server _doesn't_ support the relevant message, `PermitServerInput()` pretends to succeed.

The silent success of `PermitServerInput()` makes this log message less useful. If you want to keep the message, I suggest moving it to the `GUAC_LOG_DEBUG` level and rephrasing to capture only what we know.

stefan-esq: As stated in your comment it appears that the status is only related to if the sending of the message succeeds. I have adjusted the message for this:

`guac_client_log(client, GUAC_LOG_DEBUG, "%s send request to disable server input", success ? "Successfully" : "Failed to");`

In addition to this, after retesting the code locally PermitServerInput does not seem to work with some vnc clients even though the vnc clients do support the functionality.
After going through the libvncserver code I found the following:

```	/** Which messages are supported by the server
	 * This is a *guess* for most servers.
	 * (If we can even detect the type of server)
	 *
	 * If the server supports the "rfbEncodingSupportedMessages"
	 * then this will be updated when the encoding is received to
	 * accurately reflect the servers capabilities.
	 */
	rfbSupportedMessages supportedMessages;
```

It would appear that libvncserver guesses if the vnc server supports disabling the remote input and if it guesses the server does not, then PermitServerInput  will not send the request even if the client supports it.

For this reason I have changed the code back to use WriteToRFBServer, in my opinion it is better to always send the request.

src/protocols/vnc/vnc.c
```diff
@@ -385,6 +385,17 @@ void* guac_vnc_client_thread(void* data) {
...
 
+    /* Disable remote console (Server input) */
+    if (settings->disable_server_input) {
+        rfbSetServerInputMsg msg;
+        msg.type = rfbSetServerInput;
+        msg.status = 1;
+        msg.pad = 0;
+
+        rfbBool success = WriteToRFBServer(rfb_client, (char*)&msg, sz_rfbSetServerInputMsg);
+        guac_client_log(client, GUAC_LOG_DEBUG, "%s send request to disable server input", success ? "Successfully" : "Failed to");
```
mike-jumper, line 396: This reads well in the case of failure ("Failed to send request ..."), but the grammar gets wonky if things are sent successfully ("Successfully send request ...").

Perhaps it would be better to simply have two separate `guac_client_log()` calls? That would also allow for the failure case to be logged at a higher log level like `GUAC_LOG_WARNING` without getting too cluttered.

src/protocols/vnc/vnc.c
```diff
@@ -385,6 +385,17 @@ void* guac_vnc_client_thread(void* data) {
     }
 #endif
 
+    /* Disable remote console (Server input) */
+    if (settings->disable_server_input) {
+        rfbSetServerInputMsg msg;
+        msg.type = rfbSetServerInput;
+        msg.status = 1;
+        msg.pad = 0;
```
mike-jumper, line 393: FYI - I'm sure this is fine as-is (all structure members are explicitly initialized here), but you can do this quickly and cleanly in one step with:

```c
rfbSetServerInputMsg msg = {
    .type = rfbSetServerInput,
    .status = 1
};
```

with `pad` implicitly zeroed.

I'm OK with things as you have them, but just wanted to point the above out. We have additional syntax at our disposal here thanks to the codebase being written for C99.

mike-jumper (changes requested):

necouchman:
@stefan-esq : Any chance you could finish this up and we can get this merged?

PR #328, opened 2021-01-19
GUACAMOLE-1268 rename recordings after the session has finished

stefan-esq:
Apply the suffix ".guac" to the recording files so that scripts can pick up finished session recordings.


necouchman:
Closing this as the JIRA issue has been closed as "Won't Do"

PR #329, opened 2021-01-24
GUACAMOLE-1133: Add build check for headers when libvncserver includes gcrypt support

necouchman:
Copied the check from libssh2 for gcrypt support and adapted it for libvncserver. Should fix the current build issues, and I also added Dockerfile update to include `libgcrypt-dev` in the build requirements.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -57,6 +57,7 @@ ARG BUILD_DEPENDENCIES="              \
         freerdp2-dev                  \
         gcc                           \
         libcairo2-dev                 \
+        libgcrypt-dev                 \
         libjpeg62-turbo-dev           \
         libossp-uuid-dev              \
         libpango1.0-dev               \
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -57,6 +57,7 @@ ARG BUILD_DEPENDENCIES="              \
         freerdp2-dev                  \
         gcc                           \
         libcairo2-dev                 \
+        libgcrypt-dev                 \
```
mike-jumper, line 60: Is this the correct package name? I don't see it in the Debian package search.

necouchman: This is what I was going off of:
https://packages.debian.org/buster/libgcrypt-dev

Looks like it is a "virtual package" for libgcrypt20-dev?

mike-jumper:
Ah, OK.

mike-jumper (approved):

PR #331, opened 2021-02-07
GUACAMOLE-1204: Add server-side support for pass-through of multi-touch events.

mike-jumper:
This change adds RDP support for processing of received multi-touch events leveraging the RDPEI channel. If enabled on the connection and supported by the RDP server, this allows direct touch interaction with applications running within the remote desktop that support touch.

This new support relies on:

* A new Guacamole instruction, `touch`, which communicates the details of each touch contact from client to server.
* A new Guacamole layer parameter, `multi-touch`, which allows the server to report to the client that touch events are supported (to allow the client to decide whether to rely on mouse emulation vs. touch pass-through).

Each touch event consists of:

* An arbitrary ID (to distinguish one contact from another)
* X/Y coordinates
* An approximation of the contact area (X/Y radius of an ellipse, as well as the angle of that ellipse)
* The relative force applied (where 1 is the maximum force represented by the device and 0 is a touch that has been released)

Unlike the JavaScript touch events, and for the sake of simplicity, I have leveraged the force value to denote whether a touch has ended. There is no separate instruction for deleting a touch - the existing touch is simply updated with a force of 0.

### DIFF
--- a/src/libguac/guacamole/user-fntypes.h
+++ b/src/libguac/guacamole/user-fntypes.h
@@ -96,6 +96,51 @@ typedef int guac_user_mouse_handler(guac_user* user, int x, int y,
         int button_mask);
 
 /**
+ * Handler for Guacamole touch events, invoked when a "touch" instruction has
+ * been received from a user.
+ *
+ * @param user
+ *     The user that sent the touch event.
+ *
+ * @param id
+ *     An arbitrary integer ID which uniquely identifies this contact relative
+ *     to other active contacts.
+ *
+ * @param x
+ *     The X coordinate of the center of the touch contact within the display
+ *     when the event occurred, in pixels. This value is not guaranteed to be
+ *     within the bounds of the display area.
+ *
+ * @param y
+ *     The Y coordinate of the center of the touch contact within the display
+ *     when the event occurred, in pixels. This value is not guaranteed to be
+ *     within the bounds of the display area.
+ *
+ * @param x_radius
+ *     The X radius of the ellipse covering the general area of the touch
+ *     contact, in pixels.
+ *
+ * @param y_radius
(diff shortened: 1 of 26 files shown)


src/libguac/guacamole/user-fntypes.h
```diff
@@ -95,6 +95,51 @@ typedef void* guac_user_callback(guac_user* user, void* data);
...
+ *     The rough angle of clockwise rotation of the general area of the touch
+ *     contact, in degrees.
+ *
+ * @param force
+ *     The relative force exerted by the touch contact, where 0 is no force
+ *     (the touch has been lifted) and 1 is maximum force (the maximum amount
+ *     of force representable by the device).
+ *
+ * @return
+ *     Zero if the mouse event was handled successfully, or non-zero if an
```
jmuehlner, line 137: mouse?

src/libguac/guacamole/user.h
```diff
@@ -509,6 +509,27 @@ struct guac_user {
      */
     guac_user_argv_handler* argv_handler;
 
+    /**
+     * Handler for mouse events sent by the Gaucamole web-client.
```
jmuehlner, line 513: mouse?

src/libguac/guacamole/user-fntypes.h
mike-jumper: Whoops. Corrected via rebase.

🐭 🧀

src/libguac/guacamole/user.h
mike-jumper: Whoops. Fixed this one via rebase, too.

🧀 🐁

jmuehlner:
Well, this all looks pretty reasonable to me.

PR #333, opened 2021-03-02
GUACAMOLE-1302: Add support for forcing lossless compression on VNC/RDP connections.

mike-jumper:
This change adds a new `force-lossless` parameter to the VNC and RDP support, allowing the default lossy compression heuristics to be overridden. The default "use lossy compression if it looks appropriate" behavior of connections remains the same unless overridden with this new parameter.

### DIFF
--- a/src/common/common/display.h
+++ b/src/common/common/display.h
@@ -100,6 +100,13 @@ typedef struct guac_common_display {
     guac_common_display_layer* buffers;
 
     /**
+     * Non-zero if all graphical updates for this display should use lossless
+     * compression, 0 otherwise. By default, newly-created displays will use
+     * lossy compression when heuristics determine it is appropriate.
+     */
+    int lossless;
+
+    /**
      * Mutex which is locked internally when access to the display must be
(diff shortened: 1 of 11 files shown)


src/common/common/display.h
```diff
@@ -228,5 +235,27 @@ void guac_common_display_free_layer(guac_common_display* display,
...
+ * for graphical updates when heuristics determine that doing so is
+ * appropriate. Specifying a non-zero value here will force all graphical
+ * updates to always use lossless compression, whereas specifying zero will
+ * restore the default policy.
+ *
+ * Note that this can also be adjusted on a per-layer / per-buffer basis with
+ * guac_common_surface_set_lossless().
+ *
+ * @param display
+ *     The display to modify
```
jmuehlner, line 251: Looks like there's a period missing here on this param doc - same as `guac_common_surface_set_lossless`.

mike-jumper: Oops. Fixed and rebased.

PR #335, opened 2021-03-10
GUACAMOLE-1307: Use VerifyCertificateEx callback if supported.

mike-jumper:


configure.ac
```diff
@@ -897,6 +897,14 @@ then
...
 fi
 
+# Updated certificate verification callback (introduced with 2.0.0, not present
+# in 2.0.0-rc4 or earlier)
+if test "x${have_freerdp2}" = "xyes"
+then
+    AC_CHECK_MEMBERS([freerdp.VerifyCertificateEx],,,
+                     [[#include <freerdp/freerdp.h>]])
+fi
+
```
necouchman, line 907: Is `HAVE_FREERDP_VERIFYCERTIFICATEEX` defined in the `freerdp/freerdp.h` file? Or does `AC_CHECK_MEMBERS` do some sort of automatic definition? I was looking at the other instances of `AC_CHECK_MEMBERS` in `configure.ac` and they seem to at least have an `AC_DEFINE` call, if not a block of code that works through it.

necouchman (changes requested):

configure.ac
mike-jumper: `AC_CHECK_MEMBERS` will automatically define its own macro based on the name of the structure and member provided. For example:

https://github.com/apache/guacamole-server/blob/e2a136f41e729a80f14ef479321ab8a5d753e6d2/configure.ac#L888-L897

The above call results in the following within `config.h` when the check passes:

```c
/* Define to 1 if `AllowUnanouncedOrdersFromServer' is a member of
   `rdpSettings'. */
#define HAVE_RDPSETTINGS_ALLOWUNANOUNCEDORDERSFROMSERVER 1
```

From https://www.gnu.org/software/autoconf/manual/autoconf-2.63/html_node/Generic-Structures.html:

> ... When _member_ belongs to _aggregate_, define HAVE_aggregate_member (in all capitals, with spaces and dots replaced by underscores). ...

necouchman: Ah, okay - the documentation I was reading must not have included that bit of detail.

necouchman (approved):

PR #336, opened 2021-04-08
GUACAMOLE-1283: Add synchronization around absolutely all outbound RDP messages.

mike-jumper:
The FreeRDP library is intended to be threadsafe, but is not reliably so with respect to legacy RDP encryption and outbound messages. When outbound messages are sent by multiple threads, the encryption key used for legacy RDP encryption may not be updated correctly, resulting in a fatal connection error like:

> "ERRINFO_DECRYPT_FAILED (0x00001192):(a) Decryption using Standard RDP Security mechanisms (section 5.3.6) failed. (b) Session key creation using Standard RDP Security mechanisms (section 5.3.5) failed."

These changes add synchronization around absolutely all parts of the RDP support that may send RDP messages, including `freerdp_check_event_handles()`.

### DIFF
--- a/src/protocols/rdp/channels/audio-input/audio-buffer.h
+++ b/src/protocols/rdp/channels/audio-input/audio-buffer.h
@@ -25,8 +25,19 @@
 #include <pthread.h>
 
 /**
+ * A buffer of arbitrary audio data. Received audio data can be written to this
+ * buffer, and will automatically be flushed via a given handler once the
+ * internal buffer reaches capacity.
+ */
+typedef struct guac_rdp_audio_buffer guac_rdp_audio_buffer;
+
+/**
  * Handler which is invoked when a guac_rdp_audio_buffer's internal packet
  * buffer has reached capacity and must be flushed.
+ *
+ * @param audio_buffer
+ *     The guac_rdp_audio_buffer that has reached capacity and is being
+ *     flushed.
  *
  * @param buffer
  *     The buffer which needs to be flushed as an audio packet.
@@ -40,8 +51,8 @@
  *     The arbitrary data pointer provided when the audio buffer was
  *     initialized.
  */
-typedef void guac_rdp_audio_buffer_flush_handler(char* buffer, int length,
-        void* data);
+typedef void guac_rdp_audio_buffer_flush_handler(guac_rdp_audio_buffer* audio_buffer,
+        char* buffer, int length, void* data);
 
 /**
  * A description of an arbitrary PCM audio format.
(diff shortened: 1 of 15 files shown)


src/protocols/rdp/channels/audio-input/audio-buffer.h
```diff
@@ -40,8 +51,8 @@
  *     The arbitrary data pointer provided when the audio buffer was
  *     initialized.
  */
-typedef void guac_rdp_audio_buffer_flush_handler(char* buffer, int length,
-        void* data);
+typedef void guac_rdp_audio_buffer_flush_handler(guac_rdp_audio_buffer* audio_buffer,
+        char* buffer, int length, void* data);
```
necouchman, line 55: Is there reason to expect that there are cases when `buffer` may be outside of the `audio_buffer`? Passing pointers around isn't that expensive, but seems like if we have the entire `audio_buffer` structure we could just retrieve `buffer` from there?

necouchman (commented):
I'm guessing that the `PubSub_SubscribeChannelConnected()` calls do not need to be wrapped in the mutex lock/unlock, as they're just internal to the client and don't send any messages, correct? For example:

https://github.com/apache/guacamole-server/blob/b2ae2fdf003a6854ac42877ce0fce8e88ceb038a/src/protocols/rdp/channels/cliprdr.c#L541-L543

Just wanted to make sure. Other than that just one comment...

src/protocols/rdp/channels/audio-input/audio-buffer.h
mike-jumper: We could, sure. The same could go for `data`, which is also part of `guac_rdp_audio_buffer`. Would you prefer something like:

```c
typedef void guac_rdp_audio_buffer_flush_handler(guac_rdp_audio_buffer* audio_buffer, int length);
```

?

mike-jumper:
> I'm guessing that the PubSub_SubscribeChannelConnected() calls do not need to be wrapped in the mutex lock/unlock, as they're just internal to the client and don't send any messages, correct?

Right.

src/protocols/rdp/channels/audio-input/audio-buffer.h
necouchman: Yeah, that's fine with me - I wasn't sure if there were any situations where the function would be used to process a `buffer` and `data` that wasn't already part of `audio_buffer`? If not, I'm totally good with the above - if so, we can leave it as is. No strong preference...

mike-jumper: OK - I've removed the redundant parameters.

necouchman (approved):

PR #340, opened 2021-05-22
set client rcMask to NULL after freeing

iljavs:
libvncclient does not currently free rcMask, but if it ever does, the current code would trigger a double free. Set client->rcMask to NULL after freeing to prevent this scenario.


necouchman:
@iljavs Thank you for the contribution. You'll need to create a JIRA issue for this and tag the pull request and the commit messages with the JIRA issue. See other existing pull requests, particularly merged ones, for examples of this.

http://guacamole.apache.org/open-source/

PR #341, opened 2021-05-23
GUACAMOLE-1350: Add code to join leave_handler when connecting

myjimmy:
When the shared users enters by using Share feature in rdp, `guac_rdp_parse_args()` is called.
But, when they leaves, `guac_rdp_settings_free()` isn't called.
We can `guac_rdp_user_leave_handler()` to fix this issue.

### DIFF
--- a/src/protocols/kubernetes/client.c
+++ b/src/protocols/kubernetes/client.c
@@ -101,6 +101,7 @@ int guac_client_init(guac_client* client) {
     /* Set handlers */
     client->join_handler = guac_kubernetes_user_join_handler;
     client->free_handler = guac_kubernetes_client_free_handler;
+    client->leave_handler = guac_kubernetes_user_leave_handler;
 
     /* Register handlers for argument values that may be sent after the handshake */
     guac_argv_register(GUAC_KUBERNETES_ARGV_COLOR_SCHEME, guac_kubernetes_argv_callback, NULL, GUAC_ARGV_OPTION_ECHO);
--- a/src/protocols/rdp/client.c
+++ b/src/protocols/rdp/client.c
@@ -165,6 +165,7 @@ int guac_client_init(guac_client* client, int argc, char** argv) {
     /* Set handlers */
     client->join_handler = guac_rdp_user_join_handler;
     client->free_handler = guac_rdp_client_free_handler;
+    client->leave_handler = guac_rdp_user_leave_handler;
 
 #ifdef ENABLE_COMMON_SSH
     guac_common_ssh_init(client);
--- a/src/protocols/ssh/client.c
+++ b/src/protocols/ssh/client.c
@@ -51,6 +51,7 @@ int guac_client_init(guac_client* client) {
     /* Set handlers */
     client->join_handler = guac_ssh_user_join_handler;
     client->free_handler = guac_ssh_client_free_handler;
+    client->leave_handler = guac_ssh_user_leave_handler;
 
     /* Register handlers for argument values that may be sent after the handshake */
     guac_argv_register(GUAC_SSH_ARGV_COLOR_SCHEME, guac_ssh_argv_callback, NULL, GUAC_ARGV_OPTION_ECHO);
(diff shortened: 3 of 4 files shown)


src/protocols/rdp/user.c
```diff
@@ -121,13 +121,16 @@ int guac_rdp_user_join_handler(guac_user* user, int argc, char** argv) {
...
         user->pipe_handler = guac_rdp_pipe_svc_pipe_handler;
-        
+
         /* If we own it, register handler for updating parameters during connection. */
         if (user->owner)
             user->argv_handler = guac_argv_handler;
 
     }
 
+    /* Process some resources when users leaving */
```
mike-jumper, line 131: What do you mean by "process some resources"?

src/protocols/rdp/user.c
```diff
@@ -121,13 +121,16 @@ int guac_rdp_user_join_handler(guac_user* user, int argc, char** argv) {
...
+    user->leave_handler = guac_rdp_user_leave_handler;
```
mike-jumper, line 132: If all users will always have the same handler, this can be easier/simpler when set at the `client` level. For example, in the VNC support:

https://github.com/apache/guacamole-server/blob/44145f681aa3cf08e9cf01e99a68096c77bcab39/src/protocols/vnc/client.c#L62

mike-jumper:
> When the shared users enters by using Share feature in rdp, `guac_rdp_parse_args()` is called.

This will happen for all users, regardless of whether they leverage the connection sharing feature.

mike-jumper:
What about the other protocols? Are there other cases where a `leave_handler` is defined but not actually ever assigned?

src/protocols/rdp/user.c
myjimmy: This indicates the tasks done by `guac_rdp_user_leave_handler()` function.

src/protocols/rdp/user.c
myjimmy: `guac_rdp_user_leave_handler()` performs the removing of common cursor and the free of `guac_rdp_settings` allocated for each shared user.
But the free to `guac_rdp_settings` of owner is performed through guac_rdp_client_free_handler().
Please refer [this link](https://issues.apache.org/jira/browse/GUACAMOLE-1350?focusedCommentId=17350114&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-17350114) for more information.
So, I think that we can't set this handler at `client` level.

myjimmy:
> > When the shared users enters by using Share feature in rdp, `guac_rdp_parse_args()` is called.
> 
> This will happen for all users, regardless of whether they leverage the connection sharing feature.

Yes, you're right.
I tried to emphasize that the guad doesn't free the `guac_rdp_settings` of shared users when using *share feature*.

myjimmy:
> What about the other protocols? Are there other cases where a `leave_handler` is defined but not actually ever assigned?
There are other cases where a leave_handler is defined for other protocols.
But, I checked that handler isn't actually ever assigned.

myjimmy:
@mike-jumper I modified the code.
I set `guac_rdp_user_leave_handler()` at `client` level.

mike-jumper:
> > What about the other protocols? Are there other cases where a `leave_handler` is defined but not actually ever assigned?
> 
> There are other cases where a `leave_handler` is defined for other protocols.
> But, I checked that handler isn't actually ever assigned.

Soooooo ... shall we fix those, as well?

myjimmy:
> > > What about the other protocols? Are there other cases where a `leave_handler` is defined but not actually ever assigned?
> > 
> > 
> > There are other cases where a `leave_handler` is defined for other protocols.
> > But, I checked that handler isn't actually ever assigned.
> 
> Soooooo ... shall we fix those, as well?

OK. We'll fix those, as well.

myjimmy:
@mike-jumper I set the `leave_handler` at `client` level for other protocols (kubernetes, ssh, telnet) in addition to rdp.

mike-jumper:
Thanks, @myjimmy!

mike-jumper (approved):

PR #342, opened 2021-05-28
GUACAMOLE-1064: add Norwegian keyboard

oharboe:
@jmoehlma Best guess... How do I test this?

### DIFF
--- a/src/protocols/rdp/Makefile.am
+++ b/src/protocols/rdp/Makefile.am
@@ -238,6 +238,7 @@ rdp_keymaps =                                \
     $(srcdir)/keymaps/hu_hu_qwertz.keymap    \
     $(srcdir)/keymaps/it_it_qwerty.keymap    \
     $(srcdir)/keymaps/ja_jp_qwerty.keymap    \
+    $(srcdir)/keymaps/no_no_qwerty.keymap    \
     $(srcdir)/keymaps/pt_br_qwerty.keymap    \
     $(srcdir)/keymaps/sv_se_qwerty.keymap    \
     $(srcdir)/keymaps/da_dk_qwerty.keymap    \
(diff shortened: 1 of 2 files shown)


necouchman:
@oharboe: Looks fine without having an actual keyboard to test it with. Unfortunately I don't have a way to test, either :smile:

oharboe:
@necouchman Is there a way to test this beyond manual integration testing, i.e. set up a couple of machines w/Guacamole?

mike-jumper:
@oharboe, no, there is no automated test mechanism for keyboard layout behavior. The only way to test that a new keyboard layout behaves correctly is to deploy and directly test against an RDP connection using that layout.

oharboe:
@mike-jumper Tested. Seems to work :-)

`sudo showkey` was useful to find scancodes.

src/protocols/rdp/Makefile.am
```diff
@@ -241,7 +241,8 @@ rdp_keymaps =                                \
     $(srcdir)/keymaps/pt_br_qwerty.keymap    \
     $(srcdir)/keymaps/sv_se_qwerty.keymap    \
     $(srcdir)/keymaps/da_dk_qwerty.keymap    \
-    $(srcdir)/keymaps/tr_tr_qwerty.keymap
+    $(srcdir)/keymaps/tr_tr_qwerty.keymap    \
+    $(srcdir)/keymaps/no_no_qwerty.keymap
```
mike-jumper, line 245: With the exception of "base" and "failsafe" (which are special cases) and "da_dk_qwerty" (which somehow slipped in here unsorted), the files listed here are being maintained in sorted order. Please move your new keymap up in the list such that sorted order is maintained.

mike-jumper (changes requested):
Except for the ordering of the keymaps (see comment), this LGTM.

mike-jumper (approved):

PR #345, opened 2021-07-29
GUACAMOLE-1388: Ensure RDP-specific resources are cleaned up after channel disconnect.

mike-jumper:
Without these changes, RDP-specific resources like the CLIPRDR and RDPEI channels may remain from past connections if the RDP connection is dynamically reconnected via the "Reconnect" display resize method, resulting in assertion failures or memory errors if those stale resources are reused after reconnect is completed.

### DIFF
--- a/src/protocols/rdp/rdp.c
+++ b/src/protocols/rdp/rdp.c
@@ -471,18 +471,6 @@ static int guac_rdp_handle_connection(guac_client* client) {
 
     pthread_rwlock_wrlock(&(rdp_client->lock));
 
-    /* Set up screen recording, if requested */
-    if (settings->recording_path != NULL) {
-        rdp_client->recording = guac_common_recording_create(client,
-                settings->recording_path,
-                settings->recording_name,
-                settings->create_recording_path,
-                !settings->recording_exclude_output,
-                !settings->recording_exclude_mouse,
-                !settings->recording_exclude_touch,
-                settings->recording_include_keys);
-    }
-
     /* Create display */
     rdp_client->display = guac_common_display_alloc(client,
             rdp_client->settings->width,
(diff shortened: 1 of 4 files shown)


src/protocols/rdp/rdp.c
```diff
@@ -816,6 +804,18 @@ void* guac_rdp_client_thread(void* data) {
...
+        rdp_client->recording = guac_common_recording_create(client,
+                settings->recording_path,
+                settings->recording_name,
+                settings->create_recording_path,
+                !settings->recording_exclude_output,
+                !settings->recording_exclude_mouse,
+                !settings->recording_exclude_touch,
+                settings->recording_include_keys);
+    }
+
```
necouchman, line 818: Some reason you're wanting to include these changes?

necouchman (changes requested):
Just wondering about that last file change that doesn't seem to be related...

src/protocols/rdp/rdp.c
mike-jumper: As I walked through the set of resources that are allocated/freed for each (re)connection attempt, I noticed that this particular resource is created and destroyed asymmetrically. It was allocated for each reconnect, but freed only at the absolute end of the connection, thus each reconnect would erroneously start a new recording and fail to properly free resources.

It struck me as related (mishandled resources on reconnect), but you're right that it's not related to channel disconnect and not related to specifically the clipboard issue noted in GUACAMOLE-1388.

Should I separate this into a different commit and broaden the issue+PR description?

necouchman: Ah, okay. I'm fine rolling them into this one.

necouchman (approved):

PR #346, opened 2021-09-10
GUACAMOLE-1416: Fix unreleased the lock in the ssh_client_thread

ycaibb:
GUACAMOLE-1416: Fix unreleased the lock ssh_client->term_channel_lock in the ssh_client_thread.

### DIFF
--- a/src/protocols/ssh/ssh.c
+++ b/src/protocols/ssh/ssh.c
@@ -438,8 +438,10 @@ void* ssh_client_thread(void* data) {
         /* Send keepalive at configured interval */
         if (settings->server_alive_interval > 0) {
             timeout = 0;
-            if (libssh2_keepalive_send(ssh_client->session->session, &timeout) > 0)
+            if (libssh2_keepalive_send(ssh_client->session->session, &timeout) > 0) {
+                pthread_mutex_unlock(&(ssh_client->term_channel_lock));
                 break;
+            }
             timeout *= 1000;
         }
         /* If keepalive is not configured, sleep for the default of 1 second */


src/protocols/ssh/ssh.c
```diff
@@ -439,6 +439,7 @@ void* ssh_client_thread(void* data) {
         if (settings->server_alive_interval > 0) {
             timeout = 0;
             if (libssh2_keepalive_send(ssh_client->session->session, &timeout) > 0)
+                pthread_mutex_unlock(&(ssh_client->term_channel_lock));
                 break;
```
necouchman, line 443: You'll need to add the braces `{..}` for this `if` statement. By just adding this line the code will then always execute the `break` if the `settings->server_alive_interval` is greater than `0`, which is probably not what we want.

necouchman (changes requested):

src/protocols/ssh/ssh.c
ycaibb: Thanks, please check the new one

src/protocols/ssh/ssh.c
```diff
@@ -438,8 +438,10 @@ void* ssh_client_thread(void* data) {
         /* Send keepalive at configured interval */
         if (settings->server_alive_interval > 0) {
             timeout = 0;
-            if (libssh2_keepalive_send(ssh_client->session->session, &timeout) > 0)
+            if (libssh2_keepalive_send(ssh_client->session->session, &timeout) > 0) {
+                pthread_mutex_unlock(&(ssh_client->term_channel_lock));
                 break;
+	    }
```
necouchman, line 444: This is almost lined up, but not quite. Sorry for the #nitpick, but we'd like the style consistent :-).

necouchman (changes requested):

src/protocols/ssh/ssh.c
ycaibb: Please check. Sorry for the inconvenience : )

necouchman (approved):

PR #348, opened 2021-10-18
GUACAMOLE-1447: bump docker base image to Debian Bullseye

xens:
Bumping Docker base image to Debian latest stable `Debian Bullseye`.

* https://issues.apache.org/jira/browse/GUACAMOLE-1447
* https://www.debian.org/releases/

What I tested on Guacamole version `1.3.0`:

:heavy_check_mark: Connection to a RDP host
:heavy_check_mark: Connection to a SSH target + verifying the remote-server key (known host key)
:heavy_check_mark: Connection to a SSH target with user/pass authentication
:heavy_check_mark: Connection to a SSH target with pubkey authentication (with an un-encrypted private key)
:heavy_check_mark: Connection to a SSH target with pubkey authentication (with an un-encrypted private key)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -148,7 +148,8 @@ RUN apt-get update                                                              
 
 # Link FreeRDP plugins into proper path
 RUN ${PREFIX_DIR}/bin/link-freerdp-plugins.sh \
-        ${PREFIX_DIR}/lib/freerdp2/libguac*.so
+        ${PREFIX_DIR}                         \
+        /lib/freerdp2/libguac*.so
 
 # Checks the operating status every 5 minutes with a timeout of 5 seconds
 HEALTHCHECK --interval=5m --timeout=5s CMD nc -z 127.0.0.1 4822 || exit 1
@@ -158,8 +159,6 @@ ARG UID=1000
 ARG GID=1000
 RUN groupadd --gid $GID guacd
 RUN useradd --system --create-home --shell /usr/sbin/nologin --uid $UID --gid $GID guacd
-RUN cp -r /usr/local/guacamole/lib/* /usr/lib/x86_64-linux-gnu/freerdp2/
-RUN cp -r /usr/local/guacamole/lib/freerdp2/* /usr/lib/x86_64-linux-gnu/freerdp2/
 
 # Run with user guacd
 USER guacd
--- a/Dockerfile
+++ b/Dockerfile
@@ -158,6 +158,8 @@ ARG UID=1000
 ARG GID=1000
 RUN groupadd --gid $GID guacd
 RUN useradd --system --create-home --shell /usr/sbin/nologin --uid $UID --gid $GID guacd
+RUN cp -r /usr/local/guacamole/lib/* /usr/lib/x86_64-linux-gnu/freerdp2/
+RUN cp -r /usr/local/guacamole/lib/freerdp2/* /usr/lib/x86_64-linux-gnu/freerdp2/
 
 # Run with user guacd
 USER guacd
--- a/Dockerfile
+++ b/Dockerfile
@@ -22,7 +22,7 @@
 #
 
 # The Debian image that should be used as the basis for the guacd image
-ARG DEBIAN_BASE_IMAGE=buster-slim
+ARG DEBIAN_BASE_IMAGE=bullseye-slim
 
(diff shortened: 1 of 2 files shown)


src/guacd-docker/bin/link-freerdp-plugins.sh
```diff
@@ -58,26 +57,25 @@ where_is_freerdp() {
...
 
-while [ -n "$1" ]; do
+GUAC_PLUGIN="$1$2"
+FREERDP_DIR="$(where_is_freerdp "$1")"
+FREERDP_PLUGIN_DIR="${FREERDP_DIR}/freerdp2"
 
-    # Determine correct install location for FreeRDP plugins
-    FREERDP_DIR="$(where_is_freerdp "$1")"
-    FREERDP_PLUGIN_DIR="${FREERDP_DIR}/freerdp2"
+for value in $GUAC_PLUGIN; do
```
mike-jumper, line 71: This seems a bit wonky. With `GUAC_PLUGIN` being assigned `"$1$2"`, won't `for value in $GUAC_PLUGIN` always just loop through a single value except in the case where `GUAC_PLUGIN` happens to contain whitespace, in which case it would (presumably undesirably) split that value up?

Dockerfile
```diff
@@ -148,7 +148,8 @@ RUN apt-get update
 
 # Link FreeRDP plugins into proper path
 RUN ${PREFIX_DIR}/bin/link-freerdp-plugins.sh \
-        ${PREFIX_DIR}/lib/freerdp2/libguac*.so
+        ${PREFIX_DIR}                         \
+        /lib/freerdp2/libguac*.so
```
mike-jumper, line 152: The guacamole-server build automatically locates the proper directory for FreeRDP plugins and installs its plugins there. Perhaps it would be better to do away with `link-freerdp-plugins.sh` entirely and avoid this problem going forward?

Dockerfile
```diff
@@ -22,7 +22,7 @@
 #
 
 # The Debian image that should be used as the basis for the guacd image
-ARG DEBIAN_BASE_IMAGE=buster-slim
+ARG DEBIAN_BASE_IMAGE=bullseye-slim
```
mike-jumper, line 25: Rather than referencing bullseye directly for `DEBIAN_BASE_IMAGE` and `DEBIAN_RELEASE`, would it be better to reference Debian stable? I'd think that would allow the nightly rebuilds of the image to pick up on distro updates even if there has not been a corresponding Guacamole update.

mike-jumper (commented):
Beware that the latest two commits don't follow the convention we use for our commit messages and will need to be updated or squashed into that first commit.

xens:
> Beware that the latest two commits don't follow the convention we use for our commit messages and will need to be updated or squashed into that first commit.

no worries, I converted this PR as DRAFT, I'll rebase + reword when it'll be ready to be merged

Dockerfile
xens: My initial understanding was that setting `LD_LIBRARY_PATH` to `/usr/local/guacamole/lib` would be sufficient as everything was available under that path, including the FreeRDP plugins, but if we skip the `link-freerdp-plugins.sh`, `guacd` complains that it can't load `"guac-common-svc" from /usr/lib/x86_64-linux-gnu/freerdp2/` so that's why I tried to adapt the `link-freerdp-plugins.sh`, but I'm happy to help if we can get rid of it and simplify this process.

> The guacamole-server build automatically locates the proper directory for FreeRDP plugins and installs its plugins there

I'm not sure to understand if you mean that it must already work as-is without `link-freerdp-plugins.sh` (which doesn't seems to be the case), or if it means that we need to adapt something based on what has been done during the build process (if that's the case could you give me pointers?).

Dockerfile
xens: Good idea, will do.

mike-jumper:
As of #388 and #389, the guacd Docker image:

* Uses the latest Alpine Linux.
* Automatically builds the latest versions of all protocol libraries at the time the image is built.

So we _should_ be good now, and hopefully will no longer need to bounce back and forth between different versions of Debian and Ubuntu in an effort to get the latest libraries.

Thanks for taking a shot at this, @xens!

PR #349, opened 2021-10-18
GUACAMOLE-745: Support OpenSSH private keys & ED25519

roysjosh:
Let libssh2 parse PEM and ssh-native keys.


necouchman:
A couple of comments, here:
* The pull request needs to be limited to a specific JIRA issue, or split out into two different pull requests. I'd recommend just picking one and going with it, and we can close out the other JIRA issue as "resolved by XXX".
* Does this introduce any minimum version requirements for libssh2 that might impact people currently using it? I'm thinking, most specifically, of the stock libssh2 included with things like CentOS/RHEL7, which is probably going to be a bit outdated. We should, at the very least, know if it introduces a version requirement, and, if it does, discuss if we're good having the code move to that minimum version requirement (and deal with it in configure, Makefile, etc.), or support both.

roysjosh:
Oh, I think this PR will break the agent. I forgot about that. Please consider this PR as a draft until I figure that out.

roysjosh:
It looks like the agent hasn't been touched in a long time and probably has been unable to compile since at least 0fcea2738bc2d58a930c930c6a7bab72a914c87e in 2015. It also looks like the agent was relying on an out-of-tree patch for libssh2 since it was conceived back in 2013. This PR will break it more. Should I try to avoid that or are you okay with proceeding @necouchman ?

necouchman:
@roysjosh Interesting on the SSH agent bit - I don't think this PR need necessarily address that. We should probably spin up a separate JIRA issue for it.

Other than that, if nothing else significant is broken, I think we're probably okay - making 1.6.0 the minimum from 2015 doesn't seem terribly unreasonable to me.

roysjosh:
I modified the libssh2 configure check to look for the `libssh2_userauth_publickey_frommemory` function. Let me know if you want to take another route.

routerino:
Out of curiosity, would this merge be a breaking change? IE: if this pull request is merged and guacamole is deployed with an older copy of libssh, is that gracefully handled? If so there shouldn't be any blocking reason to merge the pull request.

mike-jumper:
@routerino It is gracefully handled here in the sense that SSH support will simply not be built if that core required function is absent. I _think_ it should be OK to start requiring at least libssh2 1.6.0+.

mike-jumper (commented):
With these changes, is it possible to now remove those `common-ssh/dsa-compat.h` and `common-ssh/rsa-compat.h` headers? (And possibly other direct OpenSSL includes)?

If those are no longer needed as of these changes, then those could be removed and these changes would possibly have the additional, happy effect of fixing the Ubuntu devel build (which is currently failing due to the newer OpenSSL):

https://ci-builds.apache.org/job/Guacamole/job/guacamole-server-master-ubuntu/532/JENKINS_LABEL_EXPRESSION=ubuntu,UBUNTU_RELEASE=devel/console

roysjosh:
@mike-jumper compile tested & pushed

mike-jumper:
Thanks, @roysjosh!

mike-jumper (changes requested):

src/common-ssh/key.c
```diff
@@ -33,119 +31,53 @@
...
+    /* OpenSSH v1 private keys are PEM-wrapped base64-encoded blobs. The encoded data begins with:
+     *   "openssh-key-v1\0"
+     * and continues with the following data for a plaintext key:
+     *   pascal string: cipher name ("none")
+     *   pascal string: kdf name ("none")
+     *   pascal string: kdf params (NULL)
+     *   32-bit int: number of keys (1)
+     */
+    const char ssh_v1key_prefix[] = "-----BEGIN OPENSSH PRIVATE KEY-----\nb3BlbnNzaC1rZXktdjEA";
+    const char ssh_v1key_pt_key_prefix[] = "AAAABG5vbmUAAAAEbm9uZQAAAAAAAAAB";
```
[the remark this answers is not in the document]
roysjosh: Done

src/common-ssh/key.c
```diff
@@ -33,119 +31,53 @@
...
+
+    bool is_ssh_priv_key = length > sizeof(ssh_v1key_prefix) && !strncmp(data, ssh_v1key_prefix,
+            sizeof(ssh_v1key_prefix) - 1);
+    bool ssh_pt_key_marker = length > sizeof(ssh_v1key_prefix) + sizeof(ssh_v1key_pt_key_prefix) && !strncmp(data +
+            sizeof(ssh_v1key_prefix) - 1, ssh_v1key_pt_key_prefix, sizeof(ssh_v1key_pt_key_prefix) - 1);
+
+    bool encrypted = pkcs_encrypted_marker || (is_ssh_priv_key && !ssh_pt_key_marker);
+
+    if (encrypted && (passphrase == NULL || *passphrase == '\0'))
```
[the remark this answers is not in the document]
roysjosh: Refactored and added some comments. Extracted the code to test:
```
     dsa-no-pass: is passphrase needed? false
        dsa-pass: is passphrase needed? true
   ecdsa-no-pass: is passphrase needed? false
      ecdsa-pass: is passphrase needed? true
 ed25519-no-pass: is passphrase needed? false
    ed25519-pass: is passphrase needed? true
     rsa-no-pass: is passphrase needed? false
        rsa-pass: is passphrase needed? true
```
The DSA, RSA, and ECDSA keys are the old format while the ed25519 key is the new OpenSSH v1 format.

src/common-ssh/key.c
```diff
@@ -33,119 +31,81 @@
...
-
-        /* Send public key formatted for SSH */
-        guac_common_ssh_buffer_write_string(&pos, "ssh-rsa", sizeof("ssh-rsa")-1);
-        guac_common_ssh_buffer_write_bignum(&pos, key_e);
-        guac_common_ssh_buffer_write_bignum(&pos, key_n);
-
-        /* Save public key to structure */
-        key->public_key = public_key;
-        key->public_key_length = pos - public_key;
+/* Check for a PKCS#1/PKCS#8 ENCRYPTED marker. */
```
mike-jumper, line 39: Please document functions, their parameters, return values, etc. using Doxygen-style comments. If this function is strictly local to this file, it should also be declared `static`. For example:

https://github.com/apache/guacamole-server/blob/90322cd4d3797edb01706b2d327c7d85a26cd31f/src/common-ssh/sftp.c#L104-L118

src/common-ssh/key.c
```diff
@@ -33,119 +31,81 @@
...
+bool is_pkcs_encrypted_key(char* data, int length) {
+    return strstr(data, "ENCRYPTED") != NULL;
+}
```
mike-jumper, line 42: The `length` value is ignored here, but probably shouldn't be. In particular, the `data` value is ultimately received from the input of `guac_common_ssh_key_alloc()`, which is not documented as being null-terminated:

https://github.com/apache/guacamole-server/blob/b2ae2fdf003a6854ac42877ce0fce8e88ceb038a/src/common-ssh/common-ssh/key.h#L109-L117

It looks like all current usage of `guac_common_ssh_key_alloc()` does pass a null-terminated string here, but the structure of the function does not currently require this and will overrun within this new test if there isn't a null byte somewhere despite the explicit `length`.

Possibilities that occur to me here are:

* Modifying the search used by `is_pkcs_encrypted_key()` such that `length` is honored.
* Modifying `guac_common_ssh_key_alloc()` such that null-terminated strings are required.

src/common-ssh/common-ssh/key.h
```diff
@@ -25,75 +25,23 @@
...
-     */
-    SSH_KEY_RSA,
-
-    /**
-     * DSA key.
-     */
-    SSH_KEY_DSA
-
-} guac_common_ssh_key_type;
+#define OPENSSH_V1_UNENCRYPTED_KEY "AAAABG5vbmUAAAAEbm9uZQAAAAAAAAAB"
```
mike-jumper, line 38: The contextual comment at above is helpful, but please document each of these constants separately such that they are individually clear. For example:

https://github.com/apache/guacamole-server/blob/90322cd4d3797edb01706b2d327c7d85a26cd31f/src/common-ssh/common-ssh/key.h#L30-L33

and

https://github.com/apache/guacamole-server/blob/90322cd4d3797edb01706b2d327c7d85a26cd31f/src/common-ssh/common-ssh/key.h#L35-L38

mike-jumper (changes requested):
This is definitely looking _much_ better.

src/common-ssh/key.c
roysjosh: Done

src/common-ssh/key.c
roysjosh: I originally used `strnstr` but that isn't available on Debian without hacks that would break other platforms. Autotools could help but it's been too long since I've played with it. I could possibly do `memmem` instead.

src/common-ssh/common-ssh/key.h
roysjosh: Done

src/common-ssh/key.c
mike-jumper: The same problem will occur with `memmem()`, I expect, as that's not a standard function, either. In either case, there would need to be a fallback implementation for platforms that lack the function.

Implementing a `guac_strnstr()` within libguac could be a possibility. We do similar with `guac_strlcpy()`, which provides its own internal implementation of `strlcpy()` if the platform lacks the function:

https://github.com/apache/guacamole-server/blob/084ddaf81f5c29d6cf8871fffe4e8353b10cfca8/src/libguac/string.c#L45-L71

If we _know_ that the input will always be null-terminated, the simplest solution is probably to modify `guac_common_ssh_key_alloc()` to require null-terminated inputs, remove the `length` parameter from there, and thus allow the call to `strstr()` to be presumed safe.

roysjosh:
@mike-jumper here's a WIP guac_strnstr. I can add some comments to the header file if the approach looks good to you + any other changes needed.

src/libguac/tests/string/strnstr.c
```diff
@@ -0,0 +1,73 @@
...
+#include <guacamole/string.h>
+
+#include <stdlib.h>
+#include <string.h>
+
+/**
+ * Verify guac_strnstr() behaviors:
+ */
+void test_string__strnstr() {
+    char haystack[8] = {'a', 'h', 'i', ' ', 't', 'u', 'n', 'a'};
```
mike-jumper, line 30: 🐟

mike-jumper:
> @mike-jumper here's a WIP guac_strnstr. I can add some comments to the header file if the approach looks good to you + any other changes needed.

@roysjosh Looks great (including the unit test)! I'm happy with the approach and see no need for any changes beyond adding documentation in the header.

src/libguac/tests/string/strnstr.c
roysjosh: 🐬

src/libguac/guacamole/string.h
```diff
@@ -109,6 +109,28 @@ size_t guac_strlcpy(char* restrict dest, const char* restrict src, size_t n);
...
+ *     The string to look for. It must be null-terminated.
+ *
+ * @param len
+ *     The maximum number of bytes to examine in haystack.
+ *
+ * @return
+ *     If needle is the empty string, return haystack. If needle exists in
+ *     haystack, return a pointer to the first instance in haystack. If needle
+ *     does not exist in haystack, return NULL.
+ *
```
mike-jumper, line 130: The convention for `@return` has been to describe the return value as if the "Returns ..." starting part of the sentence is already written, with the remaining documentation written as if that were already there. For example:

https://github.com/apache/guacamole-server/blob/78cbaf27ba09ea5fee6f7869edb9e7513d382690/src/libguac/guacamole/string.h#L187-L191

I think we should do the same here. My suggestion would be something like:

```c
/**
 *...
 * @return
 *     A pointer to the first instance of needle within haystack, or NULL if
 *     needle does not exist in haystack. If needle is the empty string,
 *     haystack is returned.
 */
```

Besides consistency, I think the order is particularly important here, as the edge case of the empty `needle` is noteworthy but probably not the most critical consideration for a developer looking to invoke `guac_strnstr()`.

mike-jumper (changes requested):
The unit test, the new function, and the new logic leveraging that new function all look good to me here. Except for this documentation nitpick, I think we're good to go.

src/libguac/guacamole/string.h
roysjosh: Done

mike-jumper (approved):

roysjosh:
Thanks :)

PR #350, opened 2021-11-04
GUACAMOLE-943 Fixed incorrect sequence for SSH and telnet when shift key is held.

Noor-glyp:


src/terminal/terminal.c
```diff
@@ -1590,6 +1590,23 @@ static int __guac_terminal_send_key(guac_terminal* term, int keysym, int pressed
                 if (keysym == 0xFF54 || keysym == 0xFF99) return guac_terminal_send_string(term, "\x1B[B"); /* Down */
             }
 
+            if (term->mod_shift) 
+            {
```
mike-jumper, line 1594: Please use the same style as the established code with respect to the `if` and the `{`. For example:

https://github.com/apache/guacamole-server/blob/491be8382acfc682a356284701a8535e4e73dd47/src/terminal/terminal.c#L1580

See: https://guacamole.apache.org/guac-style/#braces

mike-jumper (changes requested):
Thanks, @Noor-glyp! In addition to the comment regarding code style, a couple other points:

1. Please correct the commit message to follow the same "ISSUE: message" format as the other messages (see the git history). Consistency regarding this formatting is actually pretty important to ensure people looking to find the commits relevant to a particular issue know exactly what to grep for.
2. What about if F1, F2, etc. are pressed _without_ Shift? Or with different combinations of modifiers? Presumably F1 on its own, Shift+F1, Ctrl+F1, Ctrl+Alt+F1, or even Ctrl+Alt+Shift+F1 all produce different escape sequences?

src/terminal/terminal.c
Noor-glyp: 1. I resolved the formatting issue. And rewrote the commit message. 
2. Well if function key is pressed without Shift and with Shift. Both generate different key events. For the Ctrl+Function or Ctrl+Alt function or Ctrl+Alt+Shift +Function (This opens the menu and refocuses it to menu so function key hold with it, I am not sure the function key event will pass through.)

Noor-glyp: Will add different functionality for the Alt and Ctrl hold and add to this PR branch.

mike-jumper:
For something like Ctrl+Alt+Shift where the mainline webapp reserves that
key combination for itself, you will likely need to either use Guacamole's
on-screen keyboard or send the key events programmatically with guac's JS
API.

Noor-glyp:
Okay. Got it let me implement it, and test it out.

src/terminal/terminal.c
Noor-glyp: Added the different key combination support, in recent commit.

src/terminal/terminal.c
```diff
@@ -1590,17 +1590,129 @@ static int __guac_terminal_send_key(guac_terminal* term, int keysym, int pressed
...
+                if (keysym == 0xFFC4) return guac_terminal_send_string(term, "\x1B[18;3~"); /* F7  */
+                if (keysym == 0xFFC5) return guac_terminal_send_string(term, "\x1B[19;3~"); /* F8  */
+                if (keysym == 0xFFC6) return guac_terminal_send_string(term, "\x1B[20;3~"); /* F9  */
+                if (keysym == 0xFFC7) return guac_terminal_send_string(term, "\x1B[21;3~"); /* F10 */
+                if (keysym == 0xFFC8) return guac_terminal_send_string(term, "\x1B[23;3~"); /* F11 */
+                if (keysym == 0xFFC9) return guac_terminal_send_string(term, "\x1B[24;3~"); /* F12 */
+            }
+
+            /* Shift + Alt + Function keys. */
+            if(term->mod_shift && term->mod_alt){
```
mike-jumper, line 1626: Please follow established code style here - there should be spaces between the `if` and `(` and the `)` and `{` (see elsewhere in this file).

mike-jumper, line 1626: This `if` will never actually be reached, as either of the previous `if (term->mod_shift)` and `if (term->mod_alt)` will exit the function first.

src/terminal/terminal.c
```diff
@@ -1590,17 +1590,129 @@ static int __guac_terminal_send_key(guac_terminal* term, int keysym, int pressed
...
+                if (keysym == 0xFFC4) return guac_terminal_send_string(term, "\x1B[18;5~"); /* F7  */
+                if (keysym == 0xFFC5) return guac_terminal_send_string(term, "\x1B[19;5~"); /* F8  */
+                if (keysym == 0xFFC6) return guac_terminal_send_string(term, "\x1B[20;5~"); /* F9  */
+                if (keysym == 0xFFC7) return guac_terminal_send_string(term, "\x1B[21;5~"); /* F10 */
+                if (keysym == 0xFFC8) return guac_terminal_send_string(term, "\x1B[23;5~"); /* F11 */
+                if (keysym == 0xFFC9) return guac_terminal_send_string(term, "\x1B[24;5~"); /* F12 */
+            }
+
+            /* Shift + Ctrl + Function keys. */
+            if(term->mod_shift && term->mod_ctrl){
```
mike-jumper, line 1658: Same here:

* Please follow established style.
* This will never actually be reached as earlier `if` tests will have already been satisfied and the function will already have exited.

mike-jumper (changes requested):
There are a few issues I'm seeing repeated in this latest revision:

1. Please follow established code style (see the use of spaces between parenthesis and `if` or the opening brace).
2. The conditions tested by the various `if` blocks here conflict, in several cases after simpler tests would have already passed and exited the function. In the cases of the latter and more complicated tests, the body of the `if` is effectively dead code.

I'm also wary of the sheer level of complexity of these tests. I would expect that there would be some sort of system behind the specific sequence generated and the modifiers held - some part that varies based on Ctrl, another based on Shift, etc. If that is the case, it would be much _much_ better to follow that pattern than to exhaustively test every possible combination of modifiers.

Do you know whether there is such a pattern?

src/terminal/terminal.c
Noor-glyp: Sorry about that, I had done that while testing at my machine but forgot to commit. Fixed the formatting issues to. Like added `if (term->mod_alt && term->mod_ctrl) { ` also the functions sequence are
`Ctrl+Alt+Shift -> Fn key
Shift+Alt -> Fn key
Shift+Ctrl -> Fn key
Alt+Ctrl -> Fn key
Shift -> Fn key
Alt -> Fn key
Ctrl -> Fn key` 
Also before proceeding with this approach I searched for a possible case scenario for this but couldn't what we can do is, we have this sequence 
'F1' to 'F4' P to S. Then 15 to 21 for 'F5' to 'F10' and 23 24 for 'F11' and 'F12'. And for different combinations [..;2 ,..., ..;8 is added. We can generate these sequence/strings and send them.

mike-jumper: I really think there has to be a better way. For example, looking at the numeric values in the escape sequences, each combination of modifiers appears to have a corresponding integer value. That value appears to be 1 plus the bitwise OR of the following flag values:

Modifier | Value
-------- | -----
Shift    | 1
Alt      | 2
Ctrl     | 4

For example:

Ctrl+Alt = `(4 | 2) + 1` = 7, hence each escape sequence for Ctrl+Alt+FN will use a value of 7 to represent the modifiers applicable.

This appears to be documented here:

https://www.xfree86.org/current/ctlseqs.html#PC-Style%20Function%20Keys

There will certainly be some special cases for the various function keys, but not to the extent that we need to have the same checks duplicated 8 times.

mike-jumper:
Please also recheck your commit messages - I see a whole bunch of variations:

* `Guacamole 943: ...message...`
* `Guacamole-943: ...message...`
* `GUACAMOLE943: ...message...`
* `GUACAMOLE-943 ...message...`

The correct formatting is actually none of the above:

`GUACAMOLE-943: ...message`

The established commit history is a good example of what to look for. Regarding the messages themselves, we generally point to the following document as a good guide:

https://www.codelord.net/2015/03/16/bad-commit-messages-hall-of-shame/

src/terminal/terminal.c
```diff
@@ -1590,17 +1590,120 @@ static int __guac_terminal_send_key(guac_terminal* term, int keysym, int pressed
...
+
+                if (term->mod_alt) {
+                    keyseq = keyseq | 2; 
+                }
+                
+                if (term->mod_shift) {
+                    keyseq = keyseq | 1;
+                }
+
+                    keyseq = keyseq + 1;
```
mike-jumper, line 1609: What's up with the indentation here?

src/terminal/terminal.c
```diff
@@ -1590,17 +1590,120 @@ static int __guac_terminal_send_key(guac_terminal* term, int keysym, int pressed
...
+                
+                /* F1  */
+                if (keysym == 0xFFBE || keysym == 0xFF91) {
+                    char data[10];
+                    sprintf(data,"\x1B[1;%dP",keyseq);
```
mike-jumper, line 1614: Please follow established code style (see formatting of parameters with respect to commas and spaces). The lack of spaces after the commas separating parameters here (and elsewhere) produces a rather difficult-to-read line.

src/terminal/terminal.c
```diff
@@ -1590,17 +1590,120 @@ static int __guac_terminal_send_key(guac_terminal* term, int keysym, int pressed
...
+                    char data[10];
+                    sprintf(data,"\x1B[23;%d~",keyseq);
+                    return guac_terminal_send_string(term, data); 
+                } 
+
+                /* F12 */
+                if (keysym == 0xFFC9) {
+                    char data[10];
+                    sprintf(data,"\x1B[24;%d~",keyseq);
+                    return guac_terminal_send_string(term, data);
```
mike-jumper, line 1692: With this section consisting of repeated `char data[10]` and `return guac_terminal_send_tring(...)` lines, I feel there must be a more DRY way to approach this.

src/terminal/terminal.c
```diff
@@ -1590,17 +1590,120 @@ static int __guac_terminal_send_key(guac_terminal* term, int keysym, int pressed
...
+
+                /* F11 */
+                if (keysym == 0xFFC8) {
+                    char data[10];
+                    sprintf(data,"\x1B[23;%d~",keyseq);
+                    return guac_terminal_send_string(term, data); 
+                } 
+
+                /* F12 */
+                if (keysym == 0xFFC9) {
```
mike-jumper, line 1689: Is there no way that these various keysyms could also be collapsed into fewer cases? Even if there are special cases for certain groups of function keys, the number of special cases seems like it should be well below 12.

mike-jumper (changes requested):
Please also see the previous feedback regarding the formatting of the previous commit messages.

src/terminal/terminal.c
Noor-glyp: Okay so let me fix that
`if (term->mod_alt)        keyseq = keyseq | 2; `
It should be like this right, like in a single line.

src/terminal/terminal.c
Noor-glyp: In this I think, I can do something like 
`int keystorke;            //It will be the number which keystroke is pressed like from 15 to 24 i.e. F5 to F10` Then pass it to the sprintf like the modifier key combinations in the same statement. This way we can reduce I think 12 function keys if to like 7 as F1 to F4 will remain like this. And F11, F12.

src/terminal/terminal.c
Noor-glyp: It will remain like this as for the function keys we have to pass the escape characters with some data so without the sprintf and this approch, I can't think of a better solution. Please elaborate on this maybe. What you may think be a better way.

Noor-glyp:
On the commit message. I am totally confused like my last commit message is 
GUACAMOLE-943: Reduced the repetition of If statements by using bit wise OR operator.
And is following the format GUACAMOLE-943: ...message. 
So, I am totally confused here. I have read the article you shared and the commits in the recent PR's as well.

mike-jumper:
Right - your latest commit looks absolutely fine. The remaining issue with the commit messages is with the previous commits present in this PR. You'll need to amend those commits to correct their messages, as well.

src/terminal/terminal.c
```diff
@@ -1590,17 +1590,112 @@ static int __guac_terminal_send_key(guac_terminal* term, int keysym, int pressed
...
+                if (term->mod_alt)      keyseq = keyseq | 2; 
+                if (term->mod_shift)    keyseq = keyseq | 1;
+
+                keyseq = keyseq + 1;
+                
+                /* F1  */
+                if (keysym == 0xFFBE || keysym == 0xFF91) {
+                    char data[10];
+                    sprintf(data, "\x1B[1;%dP", keyseq);
+                    return guac_terminal_send_string(term, data);
```
mike-jumper, line 1607: You can avoid this by using `guac_terminal_sendf()`:

https://github.com/apache/guacamole-server/blob/73ac4fcdbe5116a0eb5c5899af1c13ccb7be11d8/src/terminal/terminal/terminal.h#L974-L995

mike-jumper (changes requested):
Please also take another look over your commit messages with the good old https://www.codelord.net/2015/03/16/bad-commit-messages-hall-of-shame/ in mind. The latest two commits are:

* `GUACAMOLE-943: Fix formatting issues.`
* `GUACAMOLE-943: Fixed Formatting Issue.`

The above are problematic because:

1. They don't really add any information to the git history.
2. They say the exact same thing (they are indistinguishable from each other except by their hash).

Besides rewording such that the message(s) are useful, other options at your disposal include squashing the commits together or squashing them into the commit(s) that introduced the code they are correcting.

src/terminal/terminal.c
Noor-glyp: Like you are suggesting like this maybe. Like this
```
if(any mmodifier pressed)
//keyseq calculated
guac_terminal_sendf(term, "\x1B[1;"); 
char charValue = keyseq+'0';
guac_terminal_sendf(term, charValue);
//Till here can be removed outside each if like repetitions can be removed
if(keypress == 0xFFBE) //
return guac_terminal_sendf(term, "P");  //Like for F1
```

mike-jumper: No. From the documentation above, `guac_terminal_sendf()` accepts a printf-style format string and list of arbitrary arguments. Where you have:

```c
char data[10];
sprintf(data, "\x1B[1;%dP", keyseq);
return guac_terminal_send_string(term, data);
```

you can instead:

```c
return guac_terminal_sendf(term, "\x1B[1;%dP", keyseq);
```

Noor-glyp: Done.

necouchman:
@Noor-glyp Sorry, this has languished a bit. If you're still interested in getting these changes merged, I think the main thing to do is clean up the commits - make sure the commit messages are cleaned up, per Mike's request, and you've also got a merge commit that has snuck in there - this likely just needs to be re-based against the `apache:main` branch.

PR #353, opened 2021-11-23
GUACAMOLE-1444: After VNC server network interruption disconnect at client side.

Noor-glyp:
In this I tried with LIBVNC internal `Readtimeout` but even after blocking the main function for some time it still doesn't respond back with a error message. So, applied a ping approach. If the Client is idle and no packets are transferred through VNC. The Guacamole server sends a ping/reconnect request to see if the server is on or not, every 10s. 
We can modify this timeout `GUAC_VNC_IDLE_TIMEOUT`.


src/protocols/vnc/vnc.c
```diff
@@ -512,6 +516,15 @@ void* guac_vnc_client_thread(void* data) {
...
             guac_client_abort(client, GUAC_PROTOCOL_STATUS_UPSTREAM_ERROR, "Connection closed.");
 
+        if (server_runing_start - server_runing_end > GUAC_VNC_IDLE_TIMEOUT) {
+         rfb_client = guac_vnc_get_client(client);
+         server_runing_end = guac_timestamp_current();
+            if(!rfb_client){
+                guac_client_abort(client, GUAC_PROTOCOL_STATUS_UPSTREAM_ERROR, "Connection closed.");
+                break;
+            }
+        }
```
necouchman, line 526: There's some odd indentation going on, here. Please correct to follow existing style - 4-tab spaces.

necouchman (changes requested):
One style issue, and it looks like you have some extra commits in this pull request from your previous work on GUACAMOLE-943. This PR should only contain commits related to this issue.

src/protocols/vnc/vnc.c
Noor-glyp: Done.

src/protocols/vnc/vnc.c
```diff
@@ -512,6 +516,14 @@ void* guac_vnc_client_thread(void* data) {
         if (wait_result < 0)
             guac_client_abort(client, GUAC_PROTOCOL_STATUS_UPSTREAM_ERROR, "Connection closed.");
 
+        if (server_runing_start - server_runing_end > GUAC_VNC_IDLE_TIMEOUT) {
+         rfb_client = guac_vnc_get_client(client);
+         server_runing_end = guac_timestamp_current();
+            if (!rfb_client) {
```
mike-jumper, line 522: Is this just intended to test whether the connection has already closed?

src/protocols/vnc/vnc.c
```diff
@@ -512,6 +516,14 @@ void* guac_vnc_client_thread(void* data) {
...
+                guac_client_abort(client, GUAC_PROTOCOL_STATUS_UPSTREAM_ERROR, "Connection closed.");
```
mike-jumper, line 523: This log message will not be distinguishable from other types of connection closure. If the connection is being closed because the VNC server is not responding, and we can be aware of that cause, then that should be cited in the logs.

src/protocols/vnc/vnc.c
```diff
@@ -451,17 +451,21 @@ void* guac_vnc_client_thread(void* data) {
     guac_socket_flush(client->socket);
 
     guac_timestamp last_frame_end = guac_timestamp_current();
+    guac_timestamp server_runing_end = guac_timestamp_current();
```
mike-jumper, line 454: running*

src/protocols/vnc/client.h
```diff
@@ -54,6 +54,11 @@
  */
 #define GUAC_VNC_CLIPBOARD_MAX_LENGTH 262144
 
+/**
+ * The amount of time in milli-seconds to wait before sending a ping to VNC server.
```
mike-jumper, line 58: milliseconds*

mike-jumper: > ... before sending a ping to VNC server.

I'm not sure this is the case. Can you point to where this ping occurs?

src/protocols/vnc/vnc.c
mike-jumper: With something like the following:

```c
guac_timestamp last_frame_end = guac_timestamp_current();
guac_timestamp server_runing_end = guac_timestamp_current();
```

1. Repeatedly calling `guac_timestamp_current()` will incur unnecessary overhead. Unless there is an expectation that these values will be significantly different, it would be better to just reuse the value from the earlier call.
2. Can you clarify the difference between the use of these timestamps? Overall, they look very similar to the timestamps that are already being obtained for the frame timing heuristics.

src/protocols/vnc/client.h
```diff
@@ -54,6 +54,11 @@
...
+ */
+#define GUAC_VNC_IDLE_TIMEOUT 10000
```
mike-jumper, line 60: This value should probably be a connection configuration option, not a hard-coded value.

src/protocols/vnc/vnc.c
```diff
@@ -512,6 +516,14 @@ void* guac_vnc_client_thread(void* data) {
         if (wait_result < 0)
             guac_client_abort(client, GUAC_PROTOCOL_STATUS_UPSTREAM_ERROR, "Connection closed.");
 
+        if (server_runing_start - server_runing_end > GUAC_VNC_IDLE_TIMEOUT) {
+         rfb_client = guac_vnc_get_client(client);
+         server_runing_end = guac_timestamp_current();
```
mike-jumper, line 521: As Nick mentioned, please correct the indentation such that it matches established style.

mike-jumper (changes requested):

src/protocols/vnc/vnc.c
Noor-glyp: Well this is how it is working. If the system is idle for like say 10s. After 10s the guacd using `guac_vnc_get_client()`. In that when we call the libvnc API. If the connection is already connected. It will return true i.e. Client connected. If the server doesn't respond it will try to create a new connection if it still doesn't respond it will generate a error message.
First I tried using the socket ping function but in that the issue is and it is a issue reported in LibVNC as well. And is generated from their sub library TCP socket library, as on some platforms it doesn't return a error message. As explained here [TCP connection Error](https://stackoverflow.com/questions/283375/detecting-tcp-client-disconnect). 
So, I tried it by sending multiple write requests and that didn't work like took more than 5 mins before a response was generated and in some cases more than 15 mins.

src/protocols/vnc/vnc.c
Noor-glyp: Updated it in latest commit.

src/protocols/vnc/vnc.c
Noor-glyp: Okay so the difference between them is the frame ending requests are being operated in the do while loop. While this loop isn't gonna be called in idle state like wait message = 0. So, it will be operating in the following lines while waiting on the wait_result in line 462.
https://github.com/Noor-glyp/guacamole-server/blob/a4b98eac54fa168a43eca9b76d9094bf8739d16d/src/protocols/vnc/vnc.c#L457-L463

src/protocols/vnc/client.h
Noor-glyp: Well like the other frame buffer timeout. `GUAC_VNC_CONNECT_INTERVAL` and other values I think it will be a hardcoded value as this is a server timeout wait check. So like rest if the user wants he can edit in the source file. Otherwise this option isn't going to be used by anyone normally.

src/protocols/vnc/vnc.c
Noor-glyp: Corrected.

src/protocols/vnc/client.h
necouchman: I think this also may be mis-named - the name `GUAC_VNC_IDLE_TIMEOUT` implies that this is the time that the connection is idle before it times out. Based on the comment, though, I think maybe something like `GUAC_VNC_PING_INTERVAL` would be more appropriate?

necouchman:
@Noor-glyp You requested a review on this one a while back, but there's still one or two outstanding comments to be resolved.

PR #357, opened 2021-12-25
GUACAMOLE-478: Add clipboard line ending normalization option for RDP.

mike-jumper:
This change adds a new connection parameter for RDP, `normalize-clipboard`, that allows the line endings within clipboard data to be normalized to a particular format, such as Windows (CRLF) or Unix (LF).

By default, no normalization is performed.


necouchman (approved):
:christmas_tree:

mike-jumper:
🎅

PR #358, opened 2021-12-26
GUACAMOLE-1293: Add support for notifying that a user has joined an existing connection.

necouchman:
This pull request adds support for notifying the connection owner that another user has joined their connection. This is broken up into three commits - the first adds the underlying support for the "msg" instruction and updates the protocol accordingly, the second adds support for the "name" handshake instruction, allowing guacd to have a human-readable user name, and the third the actual support for sending the notification.

I'm sure it needs some work and clean-up, and there may be better approaches, but I took a shot :-D.

### DIFF
--- a/src/libguac/client.c
+++ b/src/libguac/client.c
@@ -776,7 +776,7 @@ static void* guac_client_owner_notify_join_callback(guac_user* user, void* data)
     
     /* Send required parameters to owner. */
     const char* args[] = { (const char*)joinName, NULL };
-    return (void*) ((intptr_t) guac_protocol_send_msg(user->socket, GUAC_MSG_CLIENT_JOINED, args));
+    return (void*) ((intptr_t) guac_protocol_send_msg(user->socket, GUAC_MESSAGE_USER_JOINED, args));
 
 }
 
@@ -831,7 +831,7 @@ static void* guac_client_owner_notify_leave_callback(guac_user* user, void* data
     
     /* Send required parameters to owner. */
     const char* args[] = { (const char*)quitterName, NULL };
-    return (void*) ((intptr_t) guac_protocol_send_msg(user->socket, GUAC_MSG_CLIENT_LEFT, args));
+    return (void*) ((intptr_t) guac_protocol_send_msg(user->socket, GUAC_MESSAGE_USER_LEFT, args));
 
 }
 
--- a/src/libguac/client.c
+++ b/src/libguac/client.c
@@ -760,35 +760,23 @@ int guac_client_owner_supports_required(guac_client* client) {
 static void* guac_client_owner_notify_join_callback(guac_user* user, void* data) {
 
     const guac_user* joiner = (const guac_user *) data;
-    int retval = 0;
 
-    char* owner = strdup("owner");
-    if (user->info.username != NULL) {
-        free(owner);
-        owner = strdup(user->info.username);
-    }
+    if (user == NULL)
+        return (void*) ((intptr_t) -1);
 
-    char* joinName = strdup("anonymous");
-    if (joiner->info.username != NULL) {
-        free(joinName);
-        joinName = strdup(joiner->info.username);
-    }
+    char* owner = "owner";
+    if (user->info.username != NULL)
+        owner = (char *) user->info.username;
+
+    char* joinName = "anonymous";
+    if (joiner->info.username != NULL)
+        joinName = (char *) joiner->info.username;
 
     guac_user_log(user, GUAC_LOG_DEBUG, "Notifying owner %s of %s joining.", owner, joinName);
     
     /* Send required parameters to owner. */
-    if (user != NULL) {
-        const char* args[] = { (const char*)joinName, NULL };
-        retval = guac_protocol_send_msg(user->socket, GUAC_MSG_CLIENT_JOINED, args);
-    }
-
-    else
-        retval = -1;
-
-    free(owner);
-    free(joinName);
-
-    return (void*) ((intptr_t) retval);
+    const char* args[] = { (const char*)joinName, NULL };
+    return (void*) ((intptr_t) guac_protocol_send_msg(user->socket, GUAC_MSG_CLIENT_JOINED, args));
 
 }
 
@@ -824,30 +812,22 @@ static void* guac_client_owner_notify_leave_callback(guac_user* user, void* data
 
     const guac_user* quitter = (const guac_user *) data;
 
-    char* ownerName = strdup("owner");
-    if (user->info.username != NULL) {
-        free(ownerName);
-        ownerName = strdup(user->info.username);
-    }
+    if (user == NULL)
+        return (void*) ((intptr_t) -1);
 
-    char* quitterName = strdup("anonymous");
-    if (quitter->info.username != NULL) {
-        free(quitterName);
-        quitterName = strdup(quitter->info.username);
-    }
+    char* ownerName = "owner";
+    if (user->info.username != NULL)
+        ownerName = (char *) user->info.username;
+
+    char* quitterName = "anonymous";
+    if (quitter->info.username != NULL)
+        quitterName = (char *) quitter->info.username;
 
     guac_user_log(user, GUAC_LOG_DEBUG, "Notifying owner %s of %s leaving.", ownerName, quitterName);
     
     /* Send required parameters to owner. */
-    if (user != NULL) {
-        const char* args[] = { (const char*)quitterName, NULL };
-        return (void*) ((intptr_t) guac_protocol_send_msg(user->socket, GUAC_MSG_CLIENT_LEFT, args));
-    }
-    
-    free(ownerName);
-    free(quitterName);
-
-    return (void*) ((intptr_t) -1);
+    const char* args[] = { (const char*)quitterName, NULL };
+    return (void*) ((intptr_t) guac_protocol_send_msg(user->socket, GUAC_MSG_CLIENT_LEFT, args));
 
 }
 
--- a/src/libguac/client.c
+++ b/src/libguac/client.c
@@ -772,7 +772,7 @@ static void* guac_client_owner_notify_join_callback(guac_user* user, void* data)
     if (joiner->info.username != NULL)
         joinName = (char *) joiner->info.username;
 
-    guac_user_log(user, GUAC_LOG_DEBUG, "Notifying owner %s of %s joining.", owner, joinName);
+    guac_user_log(user, GUAC_LOG_DEBUG, "Notifying owner \"%s\" of \"%s\" joining.", owner, joinName);
     
     /* Send required parameters to owner. */
     const char* args[] = { (const char*)joinName, NULL };
@@ -783,8 +783,12 @@ static void* guac_client_owner_notify_join_callback(guac_user* user, void* data)
 int guac_client_owner_notify_join(guac_client* client, guac_user* joiner) {
 
     /* Don't send msg instruction if client does not support it. */
-    if (!guac_client_owner_supports_msg(client))
+    if (!guac_client_owner_supports_msg(client)) {
+        guac_client_log(client, GUAC_LOG_DEBUG,
+                        "Client does not support the \"msg\" instruction and "
+                        "will not be notified of the user joining the connection.");
         return -1;
+    }
 
     return (int) ((intptr_t) guac_client_for_owner(client, guac_client_owner_notify_join_callback, joiner));
 
@@ -823,7 +827,7 @@ static void* guac_client_owner_notify_leave_callback(guac_user* user, void* data
     if (quitter->info.username != NULL)
         quitterName = (char *) quitter->info.username;
 
-    guac_user_log(user, GUAC_LOG_DEBUG, "Notifying owner %s of %s leaving.", ownerName, quitterName);
+    guac_user_log(user, GUAC_LOG_DEBUG, "Notifying owner \"%s\" of \"%s\" leaving.", ownerName, quitterName);
     
     /* Send required parameters to owner. */
     const char* args[] = { (const char*)quitterName, NULL };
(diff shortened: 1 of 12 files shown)


src/libguac/user-handlers.c
```diff
@@ -676,6 +677,21 @@ int __guac_handshake_image_handler(guac_user* user, int argc, char** argv) {
     
 }
 
+int __guac_handshake_name_handler(guac_user* user, int argc, char** argv) {
+
+    /* Free any past value */
+    free((char *) user->info.name);
+
+    if (argc > 0 && strcmp(argv[0], ""))
```
jmuehlner, line 685: I think this would benefit from a comment describing what this is for, and maybe some more descriptive variable names.

necouchman: Okay, I've added some additional comments to that area, and I renamed the `name` member of the `info` struct to `real_name`. I was having trouble coming up with anything that fit it any better - as usual, open to suggestions. I need to update the client-side code, as well.

necouchman: Actually, scratch that, I went with "username". I can do "client_username" if that makes more sense....

src/libguac/client.c
```diff
@@ -701,6 +739,120 @@ int guac_client_owner_supports_required(guac_client* client) {
...
+    /* Send required parameters to owner. */
+    if (user != NULL) {
+        const char* args[] = { (const char*)joinName, NULL };
+        retval = guac_protocol_send_msg(user->socket, GUAC_MSG_CLIENT_JOINED, args);
+    }
+
+    else
+        retval = -1;
+
+    free(owner);
```
jmuehlner, line 784: Be careful here - you may be trying to free string constants here.

src/libguac/client.c
```diff
@@ -701,6 +739,120 @@ int guac_client_owner_supports_required(guac_client* client) {
...
+
+    guac_user_log(user, GUAC_LOG_DEBUG, "Notifying owner %s of %s leaving.", ownerName, quitterName);
+    
+    /* Send required parameters to owner. */
+    if (user != NULL) {
+        const char* args[] = { (const char*)quitterName, NULL };
+        return (void*) ((intptr_t) guac_protocol_send_msg(user->socket, GUAC_MSG_CLIENT_LEFT, args));
+    }
+    
+    free(ownerName);
```
jmuehlner, line 839: Same here - you don't want to free string constants.

src/libguac/client.c
necouchman: Okay, I think I've managed to fix this. Maybe there's a more elegant way....

src/libguac/client.c
necouchman: Fixed this one, as well...

necouchman:
@jmuehlner I think this one is also ready for another review.

jmuehlner:
Awesome! Having a look.

src/libguac/client.c
jmuehlner: Actually, do you even need to free these at all? I mean, you're making copies of fields from `info`, or copies of string constants. Why not just avoid making copies at all?

src/libguac/guacamole/protocol-types.h
```diff
@@ -306,9 +306,40 @@ typedef enum guac_protocol_version {
...
      */
-    GUAC_PROTOCOL_VERSION_1_3_0 = 0x010300
+    GUAC_PROTOCOL_VERSION_1_3_0 = 0x010300,
+
+    /**
+     * Protocol version 1.5.0, which supports the "msg" instruction, allowing
+     * messages to be sent to the client, and adds support for the "username"
+     * handshake instruction.
+     */
+    GUAC_PROTOCOL_VERSION_1_5_0 = 0x010500
```
jmuehlner, line 316: What happened to 1.4.0?

jmuehlner:
As with https://github.com/apache/guacamole-client/pull/668, I have a few minor nitpicks, but this overall looks pretty good to me.

I'm curious if @mike-jumper has any thoughts.

src/libguac/client.c
necouchman: That's a good question - I'm not sure why I made copies in the first place. Probably should just use the originals...

necouchman: Oh, I see why I did it - I was trying to come up with a good way to insert some other text in there in the case where `info.username` isn't available. Is there a better way to go about substituting in a string constant if the `username` field isn't populated, that I'm just not thinking about right now?

jmuehlner: Wouldn't it work if you just removed all the `strdup()` and `free()` calls?

jmuehlner: The string constant is statically allocated anyway, so you don't need to worry about cleaning it up. And the fields on `info` are (I hope) getting cleaned up elsewhere.

src/libguac/client.c
```diff
@@ -701,6 +739,128 @@ int guac_client_owner_supports_required(guac_client* client) {
...
+ *
+ * @return
+ *     Zero if the message is sent successfully to the owner, otherwise
+ *     non-zero, cast as a void*.
+ */
+static void* guac_client_owner_notify_leave_callback(guac_user* user, void* data) {
+
+    const guac_user* quitter = (const guac_user *) data;
+
+    char* ownerName = strdup("owner");
```
jmuehlner, line 827: Same here - shouldn't be any need to allocate any new strings here.

src/libguac/client.c
necouchman: Ah, okay - I think I've got it, now - hopefully that works better...

src/libguac/guacamole/user.h
```diff
@@ -89,6 +89,12 @@ struct guac_user_info {
      */
     int optimal_resolution;
     
+    /**
+     * The human-readable name of the Guacamole user. If the client does not
+     * provide a name then this will be NULL.
+     */
+    const char* username;
```
mike-jumper, line 96: New structure members should be added to the end of the `struct` to preserve binary compatibility of the offsets of existing structure members. If at the end, existing code would need only be rebuilt if it allocates `guac_user_info` itself.

mike-jumper, line 96: With this being an arbitrary, human-readable value, with no guarantees of uniqueness nor requirements for uniqueness, I think:

* `name` would be a better choice than `username`, which otherwise implies uniqueness.
* We should note that no requirement is imposed on the nature of this value, including whether it is unique, except that it will be the value supplied during the handshake.

mike-jumper (changes requested):
Looks preetttty nifty.

src/libguac/client.c
```diff
@@ -701,6 +739,108 @@ int guac_client_owner_supports_required(guac_client* client) {
...
+    const char* args[] = { (const char*)joinName, NULL };
+    return (void*) ((intptr_t) guac_protocol_send_msg(user->socket, GUAC_MSG_CLIENT_JOINED, args));
+
+}
+
+int guac_client_owner_notify_join(guac_client* client, guac_user* joiner) {
+
+    /* Don't send msg instruction if client does not support it. */
+    if (!guac_client_owner_supports_msg(client))
+        return -1;
```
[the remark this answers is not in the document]
necouchman: Added

src/libguac/client.c
```diff
@@ -701,6 +739,108 @@ int guac_client_owner_supports_required(guac_client* client) {
...
+    const char* args[] = { (const char*)quitterName, NULL };
+    return (void*) ((intptr_t) guac_protocol_send_msg(user->socket, GUAC_MSG_CLIENT_LEFT, args));
+
+}
+
+int guac_client_owner_notify_leave(guac_client* client, guac_user* quitter) {
+
+    /* Don't send msg instruction if client does not support it. */
+    if (!guac_client_owner_supports_msg(client))
+        return -1;
```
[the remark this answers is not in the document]
necouchman: Added log message.

src/libguac/guacamole/protocol-types.h
```diff
@@ -306,9 +306,40 @@ typedef enum guac_protocol_version {
...
+    GUAC_MSG_CLIENT_JOINED = 0x0001,
+
+    /**
+     * A message that notifies the owner of a connection that another user has
+     * left their connection. There should be a single argument provided, the
+     * username of the user who has left.
+     */
+    GUAC_MSG_CLIENT_LEFT = 0x0002
+
+} guac_msg_client;
```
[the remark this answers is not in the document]
necouchman: Good with me - has been renamed.

src/libguac/guacamole/user.h
necouchman: Got it - have moved it down.

necouchman: Okay, I've renamed this from `username` to `name` throughout the code, including during the various handshake operations. Just need to update client-side code, now...

src/libguac/client.c
```diff
@@ -701,6 +739,108 @@ int guac_client_owner_supports_required(guac_client* client) {
...
+
+    char* owner = "owner";
+    if (user->info.username != NULL)
+        owner = (char *) user->info.username;
+
+    char* joinName = "anonymous";
+    if (joiner->info.username != NULL)
+        joinName = (char *) joiner->info.username;
+
+    guac_user_log(user, GUAC_LOG_DEBUG, "Notifying owner %s of %s joining.", owner, joinName);
```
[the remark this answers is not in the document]
necouchman: Makes sense - added.

necouchman:
Thanks @mike-jumper - I think I've addressed everything, and have confirmed it compiles. Need to update some of the client-side code, as well.

src/libguac/client.c
```diff
@@ -701,6 +739,116 @@ int guac_client_owner_supports_required(guac_client* client) {
...
+        owner = (char *) user->info.name;
+
+    char* joinName = "anonymous";
+    if (joiner->info.name != NULL)
+        joinName = (char *) joiner->info.name;
+
+    guac_user_log(user, GUAC_LOG_DEBUG, "Notifying owner \"%s\" of \"%s\" joining.", owner, joinName);
+    
+    /* Send required parameters to owner. */
+    const char* args[] = { (const char*)joinName, NULL };
```
mike-jumper, line 778: While I agree it makes sense to log something human-readable in the event this is missing ("anonymous", "owner"), I think the value we send via `msg` for `GUAC_MESSAGE_USER_JOINED` and `GUAC_MESSAGES_USER_LEFT` should be strictly the value we receive, or perhaps empty string if `NULL`. If an implementation were to assign its own meaning to the literal `"anonymous"` and send it voluntarily via `name`, then this automatically-assigned value would collide with that.

mike-jumper, line 778: For `GUAC_MESSAGE_USER_JOINED` and `GUAC_MESSAGES_USER_LEFT`, thoughts on including the unique ID automatically assigned by guacd? For example:

```
const char* args[] = { joiner->user_id, (const char*) joinName, NULL };
```

In the event that an implementation assigns the same human-readable name to two or more users of a connection, including the unique ID would allow the receiving client to reliably differentiate between them.

mike-jumper (changes requested):

src/libguac/client.c
necouchman: Yeah, I like that idea - makes sense to me.

necouchman: One follow-up question, here, @mike-jumper - are you thinking that the UUID would be displayed as part of the message the connection owner, or just logged somewhere? In the past we've avoided exposing some of the internals like UUIDs to the end user - not sure if this is a good place to make an exception?

mike-jumper: Nope - definitely should keep internals internal. My thinking was that client implementations may need to reliably differentiate between distinct users of a session, such as to provide a list of current users, regardless of whether their human-readable names match. Including the UUID would provide the necessary information for implementations to do that.

src/libguac/client.c
```diff
@@ -701,6 +739,116 @@ int guac_client_owner_supports_required(guac_client* client) {
...
+    if (user->info.name != NULL)
+        owner = (char *) user->info.name;
+
+    char* joinName = "anonymous";
+    if (joiner->info.name != NULL)
+        joinName = (char *) joiner->info.name;
+
+    guac_user_log(user, GUAC_LOG_DEBUG, "Notifying owner \"%s\" of \"%s\" joining.", owner, joinName);
+    
+    /* Send required parameters to owner. */
```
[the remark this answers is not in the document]
necouchman: Should be fixed.

src/libguac/client.c
```diff
@@ -701,6 +739,116 @@ int guac_client_owner_supports_required(guac_client* client) {
...
+    if (user->info.name != NULL)
+        ownerName = (char *) user->info.name;
+
+    char* quitterName = "anonymous";
+    if (quitter->info.name != NULL)
+        quitterName = (char *) quitter->info.name;
+
+    guac_user_log(user, GUAC_LOG_DEBUG, "Notifying owner \"%s\" of \"%s\" leaving.", ownerName, quitterName);
+    
+    /* Send required parameters to owner. */
```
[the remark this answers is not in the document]
necouchman: Should be fixed.

src/libguac/client.c
necouchman: Okay, I've fixed this up, though it may not be as elegant as it could be - let me know if there's a better way to do this...

necouchman: Server-side is implemented - should send the `user_id` value on. Just have to update things on the client-side, now, as well.

src/libguac/client.c
```diff
@@ -701,6 +739,122 @@ int guac_client_owner_supports_required(guac_client* client) {
...
+    char* logOwner = "owner";
+    if (user->info.name != NULL)
+        logOwner = (char *) user->info.name;
+
+    char* logJoiner = "anonymous";
+    char* sendJoiner = "";
+    if (joiner->info.name != NULL) {
+        logJoiner = (char *) joiner->info.name;
+        sendJoiner = (char *) joiner->info.name;
+    }
```
mike-jumper, line 776: These should be `lowercase_with_underscores` rather than `headlessCamelCase`.

src/libguac/client.c
```diff
@@ -701,6 +739,122 @@ int guac_client_owner_supports_required(guac_client* client) {
...
+    char* logOwner = "owner";
+    if (user->info.name != NULL)
+        logOwner = (char *) user->info.name;
+
+    char* logQuitter = "anonymous";
+    char* sendQuitter = "";
+    if (quitter->info.name != NULL) {
+        logQuitter = (char *) quitter->info.name;
+        sendQuitter = (char *) quitter->info.name;
+    }
```
mike-jumper, line 834: `same_here`

mike-jumper (changes requested):
LGTM - just needs a style correction to a few of the variables (`lowercase_with_underscores` vs. `headlessCamelCase`).

src/libguac/client.c
necouchman: Got it - should be fixed.

src/libguac/client.c
necouchman: Fixed.

mike-jumper (approved):

PR #360, opened 2022-01-03
GUACAMOLE-663: Set pthread stack size for guacd

necouchman:
Applying the patch provided on the Jira issue.

### DIFF
--- a/src/guacd/daemon.c
+++ b/src/guacd/daemon.c
@@ -37,6 +37,7 @@
 #include <libgen.h>
 #include <netdb.h>
 #include <netinet/in.h>
+#include <pthread.h>
 #include <signal.h>
 #include <stdio.h>
 #include <stdlib.h>
@@ -319,6 +320,14 @@ int main(int argc, char* argv[]) {
 
     /* General */
     int retval;
+
+#ifdef HAVE_DECL_PTHREAD_SETATTR_DEFAULT_NP
+    /* Set default stack size */
+    pthread_attr_t default_pthread_attr;
+    pthread_attr_init(&default_pthread_attr);
+    pthread_attr_setstacksize(&default_pthread_attr, GUACD_THREAD_STACK_SIZE);
+    pthread_setattr_default_np(&default_pthread_attr);
+#endif // HAVE_DECL_PTHREAD_SETATTR_DEFAULT_NP
 
     /* Load configuration */
     guacd_config* config = guacd_conf_load();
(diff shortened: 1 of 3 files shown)


src/guacd/daemon.c
```diff
@@ -17,6 +17,8 @@
  * under the License.
  */
 
+#define _GNU_SOURCE 1
```
mike-jumper, line 20: Which function is this required for?

necouchman: Required for `pthread_setattr_default_np()`. I'm happy to either add a comment to that effect, or use an alternative function?

src/guacd/daemon.c
```diff
@@ -275,6 +282,12 @@ int main(int argc, char* argv[]) {
     /* General */
     int retval;
 
+    /* Set default stack size of 8MB */
+    pthread_attr_t default_pthread_attr;
+    pthread_attr_init(&default_pthread_attr);
+    pthread_attr_setstacksize(&default_pthread_attr, GUACD_THREAD_STACK_SIZE);
+    pthread_setattr_default_np(&default_pthread_attr);
```
mike-jumper, line 289: I don't think we should rely on `_GNU_SOURCE` causing `pthread_setattr_default_np()` to be defined. There's no requirement that a libc implementation behave differently with `_GNU_SOURCE` defined, nor that it provide `pthread_setattr_default_np()` to those defining it.

Instead, we should:

1. Use configure tests to check whether `pthread_setattr_default_np()` is available, setting `_GNU_SOURCE` if necessary.
2. Make these calls only if `pthread_setattr_default_np()` is available.

If it's possible to check whether setting `LDFLAGS` options is sufficient to establish the expected stack size, we might want to do that, as well. Setting `LDFLAGS` is the mechanism we use to achieve this within the current Alpine build:

https://github.com/apache/guacamole-server/blob/47b9360d4678bb8e60f0850090ccade666255069/src/guacd-docker/bin/build-all.sh#L34-L36

necouchman: Okay, this should be addressed via rebase.

src/guacd/daemon.c
```diff
@@ -320,6 +321,14 @@ int main(int argc, char* argv[]) {
     /* General */
     int retval;
 
+#ifdef HAVE_DECL_PTHREAD_SETATTR_DEFAULT_NP
+    /* Set default stack size of 8MB */
```
jmuehlner, line 325: I don't love having the value of the constant specified here in the constant - it's an opportunity for things to get out of sync if somebody later updates the value of `GUACD_THREAD_STACK_SIZE`.

jmuehlner:
This PR looks to contain a few unrelated commits - bad rebase?

src/guacd/daemon.c
necouchman: Makes sense. I've removed that part of the comment.

necouchman:
> This PR looks to contain a few unrelated commits - bad rebase?

Whoops, yeah, had rebased it against `staging/1.6.0` there before we decided to remove it, and it got a few extra commits. Should be good, now.

jmuehlner:
LGTM

PR #363, opened 2022-01-07
GUACAMOLE-377: Add support for RemoteFX.

mike-jumper:
This change adds initial RemoteFX support for RDP, enabled via the `enable-gfx` parameter. As RemoteFX includes explicit frame boundaries, and thus allows us to actually track when frames are received, when Guacamole needs to combine them, etc., these changes also include enhancements to the Guacamole protocol that allow the client to track performance with respect to framerate and frame dropping.

### DIFF
--- a/src/protocols/rdp/channels/rdpgfx.c
+++ b/src/protocols/rdp/channels/rdpgfx.c
@@ -1,0 +1,122 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+#include "channels/rdpgfx.h"
+#include "plugins/channels.h"
+#include "rdp.h"
+#include "settings.h"
+
+#include <freerdp/client/rdpgfx.h>
+#include <freerdp/freerdp.h>
+#include <freerdp/gdi/gfx.h>
+#include <freerdp/event.h>
+#include <guacamole/client.h>
+
+#include <stdlib.h>
+#include <string.h>
+
+/**
+ * Callback which associates handlers specific to Guacamole with the
+ * RdpgfxClientContext instance allocated by FreeRDP to deal with received
+ * RDPGFX (Graphics Pipeline) messages.
+ *
+ * This function is called whenever a channel connects via the PubSub event
+ * system within FreeRDP, but only has any effect if the connected channel is
+ * the RDPGFX channel. This specific callback is registered with the
+ * PubSub system of the relevant rdpContext when guac_rdp_rdpgfx_load_plugin() is
+ * called.
+ *
+ * @param context
+ *     The rdpContext associated with the active RDP session.
+ *
+ * @param e
+ *     Event-specific arguments, mainly the name of the channel, and a
+ *     reference to the associated plugin loaded for that channel by FreeRDP.
+ */
+static void guac_rdp_rdpgfx_channel_connected(rdpContext* context,
+        ChannelConnectedEventArgs* e) {
+
+    guac_client* client = ((rdp_freerdp_context*) context)->client;
+
+    /* Ignore connection event if it's not for the RDPGFX channel */
+    if (strcmp(e->name, RDPGFX_DVC_CHANNEL_NAME) != 0)
+        return;
+
+    /* Init GDI-backed support for the Graphics Pipeline */
+    RdpgfxClientContext* rdpgfx = (RdpgfxClientContext*) e->pInterface;
+    rdpGdi* gdi = context->gdi;
+
+    if (!gdi_graphics_pipeline_init(gdi, rdpgfx))
+        guac_client_log(client, GUAC_LOG_WARNING, "Rendering backend for RDPGFX "
+                "channel could not be loaded. Graphics may not render at all!");
+    else
+        guac_client_log(client, GUAC_LOG_DEBUG, "RDPGFX channel will be used for "
+                "the RDP Graphics Pipeline Extension.");
+
+}
+
+/**
+ * Callback which handles any RDPGFX cleanup specific to Guacamole.
+ *
+ * This function is called whenever a channel disconnects via the PubSub event
+ * system within FreeRDP, but only has any effect if the disconnected channel
+ * is the RDPGFX channel. This specific callback is registered with the PubSub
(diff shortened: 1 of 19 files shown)


src/protocols/rdp/gdi.c
```diff
@@ -371,9 +372,110 @@ BOOL guac_rdp_gdi_set_bounds(rdpContext* context, const rdpBounds* bounds) {
...
-    /* IGNORE */
+
+    guac_client* client = ((rdp_freerdp_context*) context)->client;
+    guac_rdp_client* rdp_client = (guac_rdp_client*) client->data;
+    rdpGdi* gdi = context->gdi;
+
+    /* Leverate EndPaint handler to detect end of frame for RDPGFX channel
+     * (ignore otherwise) */
+    if (!rdp_client->settings->enable_gfx)
+        return TRUE;
```
necouchman, line 446: This comment may be a bit confusing - at least, it was to me initially. It seems like it's more likely describing what's done if the block of code below it is not true/executed, not what that block of code is actually doing?

necouchman (changes requested):
Only one initial thing that I see, but I'll take another pass through it, here, soon.

mike-jumper:
I also went ahead and switched the `enable-gfx` parameter over to `disable-gfx`, as it's very likely that this should be enabled in most cases.

jmuehlner:
> I also went ahead and switched the `enable-gfx` parameter over to `disable-gfx`, as it's very likely that this should be enabled in most cases.

In what cases would you not want to enable it?

mike-jumper:
> In what cases would you not want to enable it?

If you are using a RDP server that does not support it well, or if it turns out we've done something wrong here and the old functionality is needed.

mike-jumper:
I'm currently seeing an issue with display resize if RDPGFX is enabled where the display may be resized _smaller_, but increases to the size of the display end up clipped. I think we must be missing a handler somewhere.

jmuehlner:
Good catch!

> I'm currently seeing an issue with display resize if RDPGFX is enabled where the display may be resized _smaller_, but increases to the size of the display end up clipped. I think we must be missing a handler somewhere.

src/protocols/rdp/channels/rdpgfx.c
```diff
@@ -0,0 +1,122 @@
...
+ *
+ * @param context
+ *     The rdpContext associated with the active RDP session.
+ *
+ * @param e
+ *     Event-specific arguments, mainly the name of the channel, and a
+ *     reference to the associated plugin loaded for that channel by FreeRDP.
+ */
+static void guac_rdp_rdpgfx_channel_disconnected(rdpContext* context,
+        ChannelDisconnectedEventArgs* e) {
```
jmuehlner, line 91: Why the single-character variable name?

src/protocols/rdp/gdi.c
```diff
@@ -371,9 +372,110 @@ BOOL guac_rdp_gdi_set_bounds(rdpContext* context, const rdpBounds* bounds) {
...
+     * channel */
+    if (!rdp_client->settings->enable_gfx)
+        return TRUE;
+
+    /* Ignore paint if GDI output is suppressed */
+    if (gdi->suppressOutput)
+        return TRUE;
+
+    /* Ignore paint if nothing has been done (empty rect) */
+    if (gdi->primary->hdc->hwnd->invalid->null)
```
jmuehlner, line 453: `invalid`?

src/protocols/rdp/settings.c
```diff
@@ -371,6 +372,13 @@ enum RDP_ARGS_IDX {
      */
     IDX_DISABLE_GLYPH_CACHING,
 
+    /**
+     * "true" if the RDP Graphics Pipeline Extension should not be used, and
+     * traditional RDP graphics should be used instead, "false" or blank the
```
jmuehlner, line 377: I think some minor rewording would make this comment easier to read - maybe just adding "if" between "blank" and "the".

src/protocols/rdp/gdi.c
mike-jumper: Yep - `invalid` is the pointer to the rect in FreeRDP's GDI implementation that represents the dirty area of the GDI surface. If `invalid->null` is true, then the rect is empty and there have been no changes to the surface.

src/protocols/rdp/settings.c
mike-jumper: Yep, I definitely a word. I'll fix.

src/protocols/rdp/channels/rdpgfx.c
mike-jumper: Fixed here and in the other channels.

src/protocols/rdp/settings.c
mike-jumper: Fixed via rebase.

mike-jumper:
> @mike-jumper is this resolved now?
> 
> > I'm currently seeing an issue with display resize if RDPGFX is enabled where the display may be resized _smaller_, but increases to the size of the display end up clipped. I think we must be missing a handler somewhere.

Yep. The display now resizes correctly regardless of whether RDPGFX is enabled.

jmuehlner:
LGTM! @necouchman any remaining concerns?

jmuehlner:
Well, I'm gonna merge this now. If there are any additional concerns, we can make @mike-jumper fix them in a new PR 😛

necouchman:
Thanks @jmuehlner, sorry for the delayed reply!

landoncolburn:
Was this change reverted? I haven't been able to find any documentation or information regarding RemoteFX

necouchman:
@landoncolburn No, but the change was committed to the master branch, and we have not released a version of Guacamole that includes this support. This will not make it into the 1.5.0 release, and is tagged for 1.6.0 - release date TBD.

https://issues.apache.org/jira/browse/GUACAMOLE-377

ofiriluz:
Hi,

to my understanding, GFX Channel in RDP acts as a more "video centric" type of graphics channel, which passes surface frames instead of bitmap deltas in a more efficient fashion and thus consume less CPU / network

Now, FreeRDP supports it along with H264 / AVC420 / AVC444, is it something supported in Guacamole in some fashion along with this PR? or i am completely off topic and guacamole still works in the same fashion and its just used for tracking frames at the moment

Thanks :)

Stasmin:
> Hi,
> 
> to my understanding, GFX Channel in RDP acts as a more "video centric" type of graphics channel, which passes surface frames instead of bitmap deltas in a more efficient fashion and thus consume less CPU / network
> 
> Now, FreeRDP supports it along with H264 / AVC420 / AVC444, is it something supported in Guacamole in some fashion along with this PR? or i am completely off topic and guacamole still works in the same fashion and its just used for tracking frames at the moment
> 
> Thanks :)

Hello. I'm also interested in AVC444 support, since it's the fastest and most efficient RDP implementation at the moment. As far as I can see, this setting was commented out in the code and then completely removed.

In FreeRDP, you can enable this protocol using /gfx-h264:AVC444.

zhenggexia:
Hello, thank you for your great efforts. I currently cannot find the related code in version 1.5.5. If I urgently need this feature to support my project, could you provide a mirror of the main branch for me to use this feature? I will provide feedback on my user experience. Thank you.

necouchman:
@zhenggexia: This is not in the 1.5.5 code, it will be released in 1.6.0. If you need it urgently, you have a couple of options:
* Check out the current git code, either the `main` branch or the `staging/1.6.0` branch, and build from that code.
* Pull the patch for this pull request and apply it to the 1.5.5 code (and deal with any conflicts/differences in that patch process).

If you want to check out the code, you're already at the right place - https://github.com/apache/guacamole-server

PR #366, opened 2022-02-08
  ssh - support actual keyboard-interactive

jay28374:
Ask user for input
Dont use stored password from password authentication method as it is not possible to know what the server is asking for


jay28374:
Sorry, I'll try to get someone else at my work to deal with these pulls
I unfortunately are limited on time myself.

PR #367, opened 2022-02-11
GUACAMOLE-1536: return file size and file permission information when uploading files and displaying file directories

bookread-he:
return file size and file permission information when uploading files and displaying file directories


src/common-ssh/sftp.c
```diff
@@ -653,10 +654,19 @@ static int guac_common_ssh_sftp_ls_ack_handler(guac_user* user,
...
+		sprintf(tmpstr, "{\"mime\":\"%s\",\"size\":%llu,\"perm\":%lu}",
+			GUAC_USER_STREAM_INDEX_MIMETYPE, attributes.filesize,
+			attributes.permissions);
+	}
+	else {
+		sprintf(tmpstr, "{\"mime\":\"%s\",\"size\":%llu,\"perm\":%lu}",
+			"application/octet-stream", attributes.filesize,
+			attributes.permissions);
+	}
+	mimetype = tmpstr;
```
necouchman, line 669: I don't think we should just re-assign this string to `mimetype`, as the name `mimetype` implies that the data carried is the MIME type, not a string of information with various attributes about the file.

src/common-ssh/sftp.c
```diff
@@ -653,10 +654,19 @@ static int guac_common_ssh_sftp_ls_ack_handler(guac_user* user,
...
         /* Determine mimetype */
         const char* mimetype;
-        if (LIBSSH2_SFTP_S_ISDIR(attributes.permissions))
-            mimetype = GUAC_USER_STREAM_INDEX_MIMETYPE;
-        else
-            mimetype = "application/octet-stream";
+        //adding file size and permission
+        char tmpstr[150];
+	if (LIBSSH2_SFTP_S_ISDIR(attributes.permissions)) {
+		sprintf(tmpstr, "{\"mime\":\"%s\",\"size\":%llu,\"perm\":%lu}",
```
necouchman, line 660: I'm not sure how JSON data is assembled elsewhere in the code - if `sprintf` is routinely used like this, or if there's a library that handles it? Might be worth checking that out - I'd be a little surprised if all of the other places in the guacd code that assemble JSON do it like this.

src/protocols/rdp/ls.c
```diff
@@ -85,10 +86,17 @@ int guac_rdp_ls_ack_handler(guac_user* user, guac_stream* stream,
...
+        //adding file size
+        if (file->attributes & FILE_ATTRIBUTE_DIRECTORY) {
+            sprintf(tmpstr, "{\"mime\":\"%s\",\"size\":%lu}",
+                GUAC_USER_STREAM_INDEX_MIMETYPE, file->size);
+        }
+        else {
+            sprintf(tmpstr, "{\"mime\":\"%s\",\"size\":%lu}",
+                "application/octet-stream", file->size);
+        }
+        mimetype = tmpstr;
```
necouchman, line 99: As with above - just assigning this `tmpstr` back to `mimetype` doesn't really make sense.

necouchman (changes requested):
@bookread-he: Thank you for the contribution. In addition to the concerns that I have about how the code is implemented (see review comments below), you need to make sure that all of your commit messages are tagged with GUACAMOLE-1536 (looks like only one of them is currently), and, also, you should make sure that your commits are squashed down to a more reasonable number of commits - 6 commits for 2 file changes and 30-ish code lines changed seems a bit excessive, and, without looking at each of the commits, I would image that some of them are changing your previous commits.

Also, just want to make sure it is noted that making changes, here, will almost certainly require changes in the guacamole-client code. Perhaps you're already planning that, but just wanted to mention it.

src/common-ssh/sftp.c
mike-jumper: We use the various internal `guac_common_json_*()` functions like `guac_common_json_write_property()`.

src/common-ssh/sftp.c
mike-jumper: +1

This would also break compatibility with any implementation that expects the value to be a mimetype string, which is also what we have defined for the protocol: https://guacamole.apache.org/doc/gug/protocol-reference.html#get-instruction

I think a better approach would be to leverage mimetype parameters to encode the information in a compatible way within the mimetype itself. This would allow the same information to be encoded for other instructions like `file`, as well. For example:

```
application/octet-stream;size=12345;mode=0644;uid=123;gid=456
```

If there are standard parameters for this, we can use them, or failing that we can define our own.

hemabhavani:
Hey, is there any limitation on file size and file types when uploading/downloading the files?

mike-jumper:
@hemabhavani If you have questions about Guacamole, please use the user@guacamole.apache.org mailing list. See: https://guacamole.apache.org/support/#mailing-lists

This is a pull request, and the conversation here should be specific to the changes proposed by that pull request.

src/common-ssh/sftp.c
bookread-he: Hello Mike,
I hope you're doing well. It's been some time since our last interaction, and I've taken this period to carefully consider and implement the MIME type encoding approach you recommended for the project.

With your guidance in mind, I've crafted a string that incorporates the additional file attributes as MIME type parameters, ensuring that the original distinction between files and directories is maintained. Here's the updated code snippet:

```c
char property_value[256];

snprintf(property_value, sizeof(property_value), 
         "%s;size=%llu;mode=%o;uid=%d;gid=%d", 
         mimetype, 
         (unsigned long long) attributes.filesize, 
         attributes.permissions & 07777,
         attributes.uid, 
         attributes.gid);

guac_common_json_write_property(user, stream, &list_state->json_state, absolute_path, property_value);
```

I would be grateful for your feedback on this revision. Does this change correctly reflect the format you envisioned? Your further input would be incredibly valuable.

Thank you for your guidance and patience.

PR #371, opened 2022-02-18
GUACAMOLE-1540: Docker build using Ubuntu 21.10 as a base

jmuehlner:
See [GUACAMOLE-1540](https://issues.apache.org/jira/browse/GUACAMOLE-1540) - this change is confirmed to fix the SSH key extraction issues I was having with the old buster build.

The only real changes of note here are the replacement of `libjpeg62-turbo-dev` with `libjpeg-dev`, and the modification of the `list-dependencies.sh` script to change the way it identifies dependencies.

When testing out this change, I found that the file paths that were linked against are _NOT_ the same paths that are package-managed. 

For example: `libjpeg-turbo8` packages `/usr/lib/x86_64-linux-gnu/libjpeg.so.8`, but the libraries are actually built against `/lib/x86_64-linux-gnu/libjpeg.so.8.2.2`. These two files are hard-linked, but `dpkg` does not understand this, so it failed to find the `libjpeg-turbo8` dependency. The same is true for many other files. Searching by file name instead of absolute path fixes this issue.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -21,11 +21,11 @@
 # Dockerfile for guacamole-server
 #
 
-# The Debian image that should be used as the basis for the guacd image
-ARG DEBIAN_BASE_IMAGE=buster-slim
+# The Ubuntu image that should be used as the basis for the guacd image
+ARG UBUNTU_BASE_IMAGE=21.10
 
 # Use Debian as base for the build
-FROM debian:${DEBIAN_BASE_IMAGE} AS builder
+FROM ubuntu:${UBUNTU_BASE_IMAGE} AS builder
 
 #
 # The Debian repository that should be preferred for dependencies (this will be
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -58,7 +58,7 @@ ARG BUILD_DEPENDENCIES="              \
         gcc                           \
         libcairo2-dev                 \
         libgcrypt-dev                 \
-        libjpeg62-turbo-dev           \
+        libjpeg-dev                   \
```
mike-jumper, line 61: I believe Ubuntu does have a libjpeg-turbo package, just a newer version with a different name: https://guacamole.apache.org/doc/gug/installing-guacamole.html#required-dependencies

mike-jumper (changes requested):

Dockerfile
jmuehlner: Ok, confirmed `libjpeg-turbo8-dev` works. Switched.

PR #372, opened 2022-02-22
GUACAMOLE-1540: Search for libfreerdp2 installation directly instead of checking links.

jmuehlner:
The changes in https://github.com/apache/guacamole-server/pull/371 broke the installation of the FreeRDP plugins. 
Due to a difference in the way libraries are linked in the new build, there are no direct links from the FreeRDP plugins to the FreeRDP libraries, so the `where_is_freerdp` function would return an empty string, resulting in the plugins being linked into `/freerdp2`.

This change does 2 things:
1. Modify the FreeRDP installation directory search to directly search the filesystem for the expected `.so` files instead of assuming that they'll be linked against the plugins.
2. Ensure that a FreeRDP installation directory is actually found; abort the build if it isn't.

### DIFF
--- a/src/guacd-docker/bin/link-freerdp-plugins.sh
+++ b/src/guacd-docker/bin/link-freerdp-plugins.sh
@@ -30,29 +30,20 @@
 ##
 
 ##
-## Given the full path to a FreeRDP plugin, locates the base directory of the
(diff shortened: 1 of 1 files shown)


mike-jumper (approved):
LGTM 👍

PR #373, opened 2022-02-22
GUACAMOLE-1538: Terminal shared libary follow-on fixes

jmuehlner:
This PR fixes issues pointed out by @mike-jumper in https://github.com/apache/guacamole-server/pull/369.

- Rename the newly shared library to `libguac-terminal ` to match convention.
- Fix the linking errors by properly building against the common lib.
- Remove duplicated clipboard size constants.
- To avoid exposing the internal `guac_common_clipboard`, opaque `guac_terminal`.

Additionally, this change adds a doxygen config to autogenerate docs for the new libguac-terminal library.


src/protocols/kubernetes/argv.c
```diff
@@ -21,6 +21,7 @@
 #include "argv.h"
 #include "kubernetes.h"
 #include "terminal/terminal.h"
+#include "terminal/terminal_priv.h"
```
mike-jumper, line 24: Why is the private internal header included here? I'd think it'd be included and required only by the terminal internals.

src/protocols/kubernetes/client.c
```diff
@@ -96,7 +96,7 @@ int guac_client_init(guac_client* client) {
     client->data = kubernetes_client;
 
     /* Init clipboard */
-    kubernetes_client->clipboard = guac_common_clipboard_alloc(GUAC_KUBERNETES_CLIPBOARD_MAX_LENGTH);
+    kubernetes_client->clipboard = guac_common_clipboard_alloc();
```
mike-jumper, line 99: I think allocation of this should be internal to the terminal. Thoughts?

src/protocols/kubernetes/argv.c
jmuehlner: Because internals from the terminal struct are used here - see line 57. The same goes for all the others that have the header included.

src/protocols/kubernetes/client.c
jmuehlner: Yes, I agree. It'll take a bit more shuffling but I can make that happen.

jmuehlner: Yeah, this wouldn't have worked as is. Oops!

src/protocols/kubernetes/argv.c
jmuehlner: I guess it's not very internal if it's used all over the place like this, but there sure is a lot of code that's reaching into the terminal internals right now...

jmuehlner: Ok, should be better now - nothing from `terminal_priv.h` is used outside of the terminal code itself.

src/terminal/terminal/terminal.h
```diff
@@ -915,6 +509,75 @@ void guac_terminal_scroll_handler(guac_terminal_scrollbar* scrollbar, int value)
...
+ *     The width of the terminal, in characters.
+ */
+int guac_terminal_term_width(guac_terminal* term);
+
+/**
+ * Clears the clipboard contents for a given terminal, and assigns a new
+ * mimetype for future data.
+ *
+ * @param terminal The terminal whose clipboard is being reset.
+ * @param mimetype The mimetype of future data.
```
mike-jumper, line 554: Please migrate to the newer style we use for `@param`, `@return`, etc. For example:

https://github.com/apache/guacamole-server/blob/6dd33a8d90907b93a29bc482e888e4a972a0be4d/src/terminal/terminal/terminal.h#L874-L878

src/terminal/terminal/terminal.h
```diff
@@ -591,15 +172,6 @@ typedef struct guac_terminal_options {
      */
     guac_client* client;
```
mike-jumper, line 173: I didn't notice this in the previous iteration, but I don't think this should be an "option" either. It's the sort of thing that would be provided at terminal creation and then never change.

src/terminal/terminal/terminal_priv.h
```diff
@@ -0,0 +1,446 @@
...
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+
+#ifndef _GUAC_TERMINAL_PRIV_H
+#define _GUAC_TERMINAL_PRIV_H
```
mike-jumper, line 22: You'll see this in a lot of older code, but underscore prefixes are apparently reserved for internal use by the C compiler/preprocessor when it comes to global identifiers like these. It's not uncommon to do this anyway, but knowing this now, we should avoid it. This should instead by `GUAC_TERMINAL_PRIV_H`.

The underscore prefix is apparently technically only allowed for things like structure members, and then only if it's not followed by a capital letter or a second underscore.

mike-jumper (changes requested):

src/protocols/kubernetes/argv.c
```diff
@@ -53,8 +53,9 @@ int guac_kubernetes_argv_callback(guac_user* user, const char* mimetype,
     }
 
     /* Update Kubernetes terminal size */
-    guac_kubernetes_resize(client, terminal->term_height,
-            terminal->term_width);
+    guac_kubernetes_resize(client,
+            guac_terminal_term_height(terminal),
+            guac_terminal_term_width(terminal));
```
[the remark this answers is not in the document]
jmuehlner: Yeah, I don't love these names. It's kinda nice to have them named the same as the property they're accessing, but yeah I'd be happy to change them to something like `guac_terminal_get_rows`. It doesn't really match the style of the file - not a lot of getters with `get` in the name, but if you're ok with it, I'm definitely down.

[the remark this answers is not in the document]
jmuehlner: Renamed!

src/terminal/color_scheme.c
```diff
@@ -19,7 +19,7 @@
 
 #include "config.h"
 
-#include "terminal/color-scheme.h"
+#include "terminal/color_scheme.h"
```
jmuehlner, line 22: I renamed these files to match the underscore convention that most of the rest of this library is following.

src/protocols/kubernetes/argv.c
```diff
@@ -67,20 +68,20 @@ void* guac_kubernetes_send_current_argv(guac_user* user, void* data) {
 
     /* Send current color scheme */
     guac_user_stream_argv(user, user->socket, "text/plain",
-            GUAC_KUBERNETES_ARGV_COLOR_SCHEME, terminal->color_scheme);
+            GUAC_KUBERNETES_ARGV_COLOR_SCHEME,
+            guac_terminal_color_scheme(terminal));
```
mike-jumper, line 72: For consistency with things like `guac_terminal_get_rows()`, I think this (and anything similar) should be renamed to `guac_terminal_get_color_scheme()`.

src/terminal/color_scheme.c
mike-jumper: Hm ... I think these should be renamed back. If the rest of the library is following an underscore convention for filenames, I'd lean toward migrating that to dashes as part of public-ifying the library. The dash convention is the convention already used by libguac for its public headers.

mike-jumper (changes requested):

src/terminal/color_scheme.c
jmuehlner: Oh, sure. can do.

src/protocols/kubernetes/argv.c
jmuehlner: Works for me.

src/terminal/color-scheme.c
```diff
@@ -46,7 +46,7 @@
  * @return
  *     Zero if the two strings are equal and non-zero otherwise.
  */
-static int guac_terminal_color_scheme_compare_token(const char* str_start,
+static int guac_terminal_get_color_scheme_compare_token(const char* str_start,
         const char* str_end, const char* literal) {
```
mike-jumper, line 50: While this does return an `int`, it's not a getter; it's a comparator. I think the old naming was correct.

src/terminal/color-scheme.c
```diff
@@ -70,7 +70,7 @@ static int guac_terminal_color_scheme_compare_token(const char* str_start,
  *     Address of a pointer to the end of the string, after the last character.
  *     On return, the pointer is moved back to before any trailing spaces.
  */
-static void guac_terminal_color_scheme_strip_spaces(const char** str_start,
+static void guac_terminal_get_color_scheme_strip_spaces(const char** str_start,
```
mike-jumper, line 73: Similar here: this isn't a getter, but a function which operates directly on the given values. It's in the "color_scheme" namespace beneath the "guac_terminal" namespace, thus has a `guac_terminal_color_scheme_` prefix, and has the descriptive name `strip_spaces` after that describing the action performed. I think that original naming was correct.

src/terminal/terminal/color-scheme.h
```diff
@@ -17,13 +17,13 @@
  * under the License.
  */
 
-#ifndef GUAC_TERMINAL_COLOR_SCHEME_H
-#define GUAC_TERMINAL_COLOR_SCHEME_H
+#ifndef guac_terminal_get_color_scheme_H
+#define guac_terminal_get_color_scheme_H
```
mike-jumper, line 21: The convention for global macros in C is `UPPERCASE_WITH_UNDERSCORES`. I think this may just be the accidental result of a case-insensitive search-and-replace, but this should be `GUAC_TERMINAL_COLOR_SCHEME_H`.

mike-jumper (changes requested):

src/terminal/color-scheme.c
jmuehlner: Ah, sorry. Over-agressive find/replace here. Fixed.

src/terminal/color-scheme.c
jmuehlner: Yeah, this shouldn't have been renamed. Putting it back.

src/terminal/terminal/color-scheme.h
jmuehlner: Yep, yet another over-agressive search/replace.

src/terminal/terminal/color-scheme.h
```diff
@@ -17,12 +17,18 @@
  * under the License.
  */
 
-#ifndef GUAC_TERMINAL_COLOR_SCHEME_H
-#define GUAC_TERMINAL_COLOR_SCHEME_H
+#ifndef GUAC_TERMINAL_GET_COLOR_SCHEME_H
+#define GUAC_TERMINAL_GET_COLOR_SCHEME_H
```
mike-jumper, line 21: Why is this `GET`?

mike-jumper (changes requested):
I think part of the reason that this is involving such untangling is that a good deal of the terminal internals are exposed here that need not be exposed in the public API, for example:

* Internal functions leveraged by the terminal implementation to convert UTF-8, fit values to a particular range, etc. (`common.h`)
* Internal functions for updating the visible region of selected text when the text highlight is changed by mouse or keyboard (`select.h`)
* Internal functions for rendering the terminal scrollbar (`scrollbar.h`)

Except for the core functions required to:

1. Create/free a terminal.
2. Set terminal options.
3. Retrieve important information about terminal state (width, height, current font, etc.)
4. Notify the terminal of received events

I think almost everything here should be private and internal. If we start with these pieces being public, then any change to those internals would have the potential to break API compatibility.

src/terminal/terminal/color-scheme.h
jmuehlner: Ugh, must have been bad search/replace.

jmuehlner:
> I think almost everything here should be private and internal. If we start with these pieces being public, then any change to those internals would have the potential to break API compatibility.

How does this look?

src/terminal/terminal/terminal.h
```diff
@@ -21,11 +21,16 @@
...
+/**
+ * Constants, structures, and function definitions defining the core
+ * functionality for the terminal library.
+ *
+ * @file terminal.h
+ */
+
 #include "config.h"
 
 #include "buffer.h"
```
mike-jumper, line 33: Are any of these remaining private headers still needed here (`buffer.h`, `display.h`, `scrollbar.h`, etc.)? If internal, they shouldn't be part of the public header, and the headers themselves should not be installed (they should be listed as "noinst" in `Makefile.am`).

mike-jumper (changes requested):

src/terminal/terminal/terminal.h
jmuehlner: Nope! They shouldn't be! Fixed.

src/terminal/terminal/terminal.h
```diff
@@ -915,6 +492,82 @@ void guac_terminal_scroll_handler(guac_terminal_scrollbar* scrollbar, int value)
...
+ * terminal connection.
+ *
+ * @param terminal
+ *      The terminal that the given user is leaving.
+ * @param user
+ *      The user who is disconnecting.
+ */
+void guac_terminal_remove_user(guac_terminal* terminal, guac_user* user);
+
 /* INTERNAL FUNCTIONS */
```
mike-jumper, line 571: > ```c
> /* INTERNAL FUNCTIONS */
> ```

I think this is a good indicator that some of the functions that follow should not be part of the public interface within `terminal.h`.

Same thing for:

> ```c
> /* LOW-LEVEL TERMINAL OPERATIONS */
> ```

mike-jumper (changes requested):

src/terminal/terminal/terminal.h
jmuehlner: Yeah, some of them should be internal only, and some definitely should not. More cleanup incoming.

jmuehlner: Ok, more stuff that should be non-public moved out of `terminal.h`.

mike-jumper (changes requested):

src/terminal/terminal/terminal.h
```diff
@@ -21,15 +21,14 @@
...
+ */
 
-#include "buffer.h"
-#include "common/clipboard.h"
-#include "common/cursor.h"
-#include "display.h"
-#include "scrollbar.h"
-#include "types.h"
-#include "typescript.h"
+#include "config.h"
```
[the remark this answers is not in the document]
jmuehlner: Removed from everywhere that doesn't need it.

src/terminal/terminal/terminal.h
```diff
@@ -915,261 +536,86 @@ void guac_terminal_scroll_handler(guac_terminal_scrollbar* scrollbar, int value)
...
- * @param ...
- *     Any srguments to use when filling the format string.
+ *     The terminal whose width should be determined.
  *
  * @return
- *     The number of bytes written to STDIN, or a negative value if an error
- *     occurs preventing the data from being written. This should always be
- *     the size of the data given unless data is intentionally dropped.
- */
-int guac_terminal_sendf(guac_terminal* term, const char* format, ...);
```
mike-jumper, line 1090: IMHO, this should stay in public as a convenient printf-style way to invoke `guac_terminal_send_string()`, assuming there isn't anything about this function that would mean it has to be considered internal.

mike-jumper (changes requested):
Except for `guac_terminal_sendf()` having been removed from the public `terminal.h`, I think we're good.

src/terminal/terminal/terminal.h
jmuehlner: Sure

jmuehlner: Done.

mike-jumper (approved):

PR #376, opened 2022-03-15
GUACAMOLE-1312: Adds Canadian French keymaps

webtroter:
This is from kavenr.

I hope this is the only thing that needs to be updated.

I didnt find those files in my container, so I could change them to test. Maybe I didnt check at the right places.

I can confirm the keymap is correct.

### DIFF
--- a/src/protocols/rdp/Makefile.am
+++ b/src/protocols/rdp/Makefile.am
@@ -235,6 +235,7 @@ rdp_keymaps =                                \
     $(srcdir)/keymaps/es_es_qwerty.keymap    \
     $(srcdir)/keymaps/es_latam_qwerty.keymap \
     $(srcdir)/keymaps/fr_be_azerty.keymap    \
+    $(srcdir)/keymaps/fr_ca_qwerty.keymap    \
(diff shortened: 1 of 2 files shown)


necouchman (approved):
Overall looks okay to me - @mike-jumper or @jmuehlner any concerns?

@webtroter You will need to add the keyboard map on the guacamole-client so that it is an option within the connection parameters, but it's a relatively simple change - just an update to the `rdp.json` file to add it as an option, and then updating the related translation json files.

webtroter:
Is everything OK ?

Are we only waiting on a second opinion?

I can`t wait to be able to use the correct characters, like the interrogation marks ;)

necouchman:
@webtroter Only thing is that the commit message needs to be updated with the GUACAMOLE-1312: prefix.

webtroter:
@necouchman : Done :)

mike-jumper:
> Overall looks okay to me - @mike-jumper or @jmuehlner any concerns?

Nothing from me. I did not compare against a Canadian keyboard, but the mapping looks sane.

webtroter:
If needed, I can a picture of a canadian French keyboard. I have one next to me sometimes.

PR #377, opened 2022-03-17
GUACAMOLE-1555: Call openlog() after client plugin unload

samiponkanen:
… This fixes segfaults from guacd_log() in cases where the client plugin has called openlog() with a non-NULL ident string.

### DIFF
--- a/src/guacd/proc.c
+++ b/src/guacd/proc.c
@@ -42,6 +42,7 @@
 #include <sys/types.h>
 #include <sys/socket.h>
 #include <sys/wait.h>
+#include <syslog.h>
 
 /**
  * Parameters for the user thread.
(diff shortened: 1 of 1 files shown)


mike-jumper:
In what case does a plugin call `openlog()`? Why does a segfault occur if guacd does not re-call `openlog()`?

mike-jumper:
I'm really on the fence regarding whether this is something that belongs in guacd. If third-party code invokes `openlog()` in such a way that it stomps on guacd's logging, I'd think that:

* If possible, it shouldn't.
* It would break guacd's logging while things are running, prior to the segfault in question.

Can you clarify what external code specifically needs to invoke `openlog()`?

samiponkanen:
I fully understand your concerns.

Another way to look at it: if a plugin does call openlog() and you start seeing seemingly unrelated segfaults in guacd_log(), then it really is not easy to figure out that the root cause is in the plugin code. The fix in this PR will remove the risk of getting those segfaults, even if the plugin code is not written like it should be.

But it's your call.

Br,
Sami

PR #379, opened 2022-03-20
Adapt to const parameters of ffmpeg 5.0

dfandrich:
It seems that a few parameters are now const which guacamole treats as
non-const. Mark some variables and parameters as const to match. This
should be pretty safe and compatible. I've verified only that this now
compiles without error.


necouchman:
@dfandrich Thank you for your contribution. In order for us to merge this and include it with the code, you will need to submit a Jira issue for it and tag both the pull request and commit messages with the Jira issue. In addition to the links below, have a look at the other pull requests to see examples.

https://guacamole.apache.org/open-source/
https://issues.apache.org/jira/projects/GUACAMOLE/issues

PR #382, opened 2022-05-18
GUACAMOLE-1610 FFmpeg: Replace deprecated API (av_init_packet)

riko-teki:
av_init_packet is replaced with av_packet_alloc.

### DIFF
--- a/src/guacenc/ffmpeg-compat.c
+++ b/src/guacenc/ffmpeg-compat.c
@@ -54,11 +54,8 @@ static int guacenc_write_packet(guacenc_video* video, void* data, int size) {
     int ret;
 
 #if LIBAVCODEC_VERSION_INT < AV_VERSION_INT(54,1,0)
-
-    AVPacket pkt;
-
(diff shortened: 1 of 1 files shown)


src/guacenc/ffmpeg-compat.c
```diff
@@ -54,11 +54,8 @@ static int guacenc_write_packet(guacenc_video* video, void* data, int size) {
     int ret;
 
 #if LIBAVCODEC_VERSION_INT < AV_VERSION_INT(54,1,0)
-
-    AVPacket pkt;
-
     /* Have to create a packet around the encoded data we have */
-    av_init_packet(&pkt);
+    AVPacket* pkt = av_packet_alloc();
```
necouchman, line 58: I don't think you want to do this. This goal of this entire file is to provide compatibility for older versions of ffmpeg that *don't* support the `av_packet_alloc()` (for example) functions. I believe the rest of the code has already been migrated to the newer code - see the `#else` below that matches this `#if`. I don't think these changes are necessary at all.

PR #383, opened 2022-06-08
GUACAMOLE-1622: Added margins to ssh sessions.

aleitner:
* Offset default layer by new `GUAC_TERMINAL_OFFSET` constant.
* Take new constant `GUAC_TERMINAL_MARGINS` into account when calculating available height and width of the terminal
* Fixed "Reraw" typo to "Redraw"
* Made `get_available_dimensions` function for duplicated logic


src/terminal/display.c
```diff
@@ -229,6 +229,9 @@ guac_terminal_display* guac_terminal_display_alloc(guac_client* client,
     guac_protocol_send_move(client->socket, display->select_layer,
             display->display_layer, 0, 0, 0);
 
+    guac_protocol_send_move(client->socket, display->display_layer,
```
jmuehlner, line 232: A comment here about what this is for would be pretty chill.

src/terminal/terminal.c
```diff
@@ -511,6 +494,38 @@ guac_terminal* guac_terminal_create(guac_client* client,
 
 }
 
+void get_available_dimensions(guac_terminal* term, int width, int height, int* aw, int* ah, int* columns, int* rows) {
```
jmuehlner, line 497: It's a minor thing - but to match styling, there should be a new line between the function definition and the enclosed block.

Additionally, if possible, please wrap code at 80 characters. We generally use 8 spaces as an indent for line continuations.

src/terminal/terminal/terminal.h
```diff
@@ -279,6 +289,33 @@ guac_terminal_options* guac_terminal_options_create(
  */
 void guac_terminal_free(guac_terminal* term);
 
+/**
+ * Calculate the available space within the terminal
```
jmuehlner, line 293: Sorry to be such a stickler about style, but block comments should end in periods. 

Also, a little bit more detail about how the function works is useful to understand the function at a glance. In this case, it would be good to mention something about the calculation results being stuffed into the provided int* pointers.

src/terminal/terminal.c
```diff
@@ -425,27 +416,19 @@ guac_terminal* guac_terminal_create(guac_client* client,
...
+    term->outer_width = width;
+    term->outer_height = height;
 
     /* Set pixel size */
-    term->width = width;
-    term->height = height;
+    term->width = available_width;
+    term->height = available_height;
 
+    /* set rows and columns size */
```
jmuehlner, line 431: Uppercase please

src/terminal/terminal/terminal.h
```diff
@@ -97,6 +97,16 @@
  */
 #define GUAC_TERMINAL_WHEEL_SCROLL_AMOUNT 3
 
+/**
+ * The size of margins between the console text and the border in pixels.
+ */
+#define GUAC_TERMINAL_MARGINS 16
```
mike-jumper, line 103: This and `GUAC_TERMINAL_OFFSET` will need to be in some pixel-independent units to allow the margins to scale for different display resolutions. Not all screens will have the same DPI, and the terminal font rendering takes this into account.

src/terminal/terminal/display.h
```diff
@@ -123,6 +128,12 @@ typedef struct guac_terminal_display {
      */
     int height;
 
+    /**
+     * The size of margins between the console text and the border in pixels.
+     *
```
jmuehlner, line 133: Looks like there's an empty line here that could be removed.

src/terminal/terminal/display.h
```diff
@@ -369,5 +380,16 @@ void guac_terminal_display_clear_select(guac_terminal_display* display);
...
         const char* font_name, int font_size, int dpi);
 
+/**
+ * Calculate the size of margins around the terminal based on DPI.
+ *
+ * @param dpi
+ *     The resolution of the display in DPI.
+ *
+ * @return
+ *     Calculated size of margin in pixels
```
jmuehlner, line 390: Period please.

src/terminal/terminal/terminal.h
```diff
@@ -279,6 +279,34 @@ guac_terminal_options* guac_terminal_options_create(
  */
 void guac_terminal_free(guac_terminal* term);
 
+/**
+ * Calculate the available space within the terminal 
+ * and store the results in the pointer arguments.
+ *
+ * @param terminal
+ *     The terminal provides character width and height for caluclations.
```
jmuehlner, line 287: Typo: `caluclations`

src/terminal/display.c
```diff
@@ -1020,7 +1035,11 @@ int guac_terminal_display_set_font(guac_terminal_display* display,
     if (new_width != display->width || new_height != display->height)
         guac_terminal_display_resize(display, new_width, new_height);
 
+
     return 0;
 
 }
 
+int get_margin_by_dpi(int dpi) {
+    return dpi * GUAC_TERMINAL_MARGINS / 25.4;
```
necouchman, line 1044: Maybe `25.4` should be defined as a constant somewhere, which would also help me understand why the value is chosen :-).

src/terminal/terminal/display.h
```diff
@@ -369,5 +380,16 @@ void guac_terminal_display_clear_select(guac_terminal_display* display);
...
+ */
+int get_margin_by_dpi(int dpi);        
+
```
necouchman, line 393: I don't see this function used anywhere outside of `display.c` - is there some reason it should be made available outside of the file, or perhaps it should be a `static inc` within `display.c`?

src/terminal/terminal/terminal.h
```diff
@@ -279,6 +279,34 @@ guac_terminal_options* guac_terminal_options_create(
...
+ * 
+ * @param columns
+ *     Set with the available width for text of the terminal, by column count.
+ * 
+ * @param rows
+ *     Set with the available height for text of the terminal, by row count.
+ */
+void get_available_dimensions(guac_terminal* term, int width, int height, 
+        int* aw, int* ah, int* columns, int* rows);
+
```
necouchman, line 309: Looks like this is only used in `terminal.c` - is there a reason to make it available in the header, or should it be a `static` function limited to `terminal.c`?

necouchman (changes requested):
Just a few comments from me.

src/terminal/terminal/display.h
```diff
@@ -44,6 +44,17 @@
  */
 #define GUAC_TERMINAL_MAX_CHAR_WIDTH 2
 
+/**
+ * The size of margins between the console text and the border in mm.
+ */
+#define GUAC_TERMINAL_MARGINS 2
```
jmuehlner, line 50: Why not just have a single constant for the margin in inches? I love the metric system as much as anyone else, but it seems a bit overly complicated to use 2 constants to convert mm to inches every time, rather than just working with inches directly.

jmuehlner: Note: this isn't a demand for a change; I'm just curious. If you really prefer it this way, I'm ok with keeping it.

aleitner: I figured if we wanted to make it adjustable we'd remove the `GUAC_TERMINAL_MARGINS` and add it to the options are something, and it's much easier to think of adjusting in terms of mm rather than 0.0393701 inches

jmuehlner: Makes sense to me.

src/terminal/terminal.c
```diff
@@ -335,6 +335,67 @@ guac_terminal_options* guac_terminal_options_create(
...
+ *     Set with the available height for text of the terminal, in pixels.
+ * 
+ * @param columns
+ *     Set with the available width for text of the terminal, by column count.
+ * 
+ * @param rows
+ *     Set with the available height for text of the terminal, by row count.
+ */
+static void get_available_dimensions(guac_terminal* term, int width, int height,
+    int* aw, int* ah, int* columns, int* rows) {
```
mike-jumper, line 364: Is there any case where this would not be used to calculate available dimensions from `outer_width` and `outer_height`, and is there any case where this is not ultimately used to assign the resulting values back to the `guac_terminal`?

If not, perhaps this would be better as `set_available_dimensions(term, width, height)`?

aleitner: 1) It looks like both locations where `get_available_dimensions` is called we end up setting `outer_width` and `outer_height` to the same `width` and `height` that get passed in to `get_available_dimensions`. I can update the function to just use the `terminal` for those values then.

2) I initially had the function as `set_available_dimensions` but in the `guac_terminal_resize` function we needed access to both the original dimensions and the new dimensions from `set_available_dimensions` when the `__guac_terminal_resize` function is called so to prevent rewriting a significant amount of code inside `__guac_terminal_resize` I made the `set_available_dimensions` function a get rather than a set.

src/terminal/terminal.c
```diff
@@ -335,6 +335,68 @@ guac_terminal_options* guac_terminal_options_create(
...
+ * @param ah
+ *     Set with the available height for text of the terminal, in pixels.
+ * 
+ * @param columns
+ *     Set with the available width for text of the terminal, by column count.
+ * 
+ * @param rows
+ *     Set with the available height for text of the terminal, by row count.
+ */
+static void get_available_dimensions(guac_terminal* term,
```
jmuehlner, line 360: At this point, having the function called `get_...` is pretty misleading - usually you'd expect a getter to return a value, which this doesn't do. 

Maybe `calculate_and_save_...` or similar might be better?

src/terminal/terminal.c
aleitner: good call

src/terminal/terminal.c
jmuehlner: Thanks! That's a lot simpler - though now that the function updates the terminal struct directly, the block comment for the function should actually say that - reading the comment as it exists now, you'd think it only updates the passed-in pointers.

mike-jumper (changes requested):

src/terminal/terminal.c
```diff
@@ -335,6 +335,65 @@ guac_terminal_options* guac_terminal_options_create(
...
+ *     to be set on terminal in order to calculate available dimensions.
+ * 
+ * @param columns
+ *     Set with the available width for text of the terminal, by column count.
+ * 
+ * @param rows
+ *     Set with the available height for text of the terminal, by row count.
+ */
+static void calculate_and_save_available_dimensions(guac_terminal* term,
+    int* columns, int* rows) {
```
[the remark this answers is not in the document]
aleitner: Okay instead of cramming everything into the `calculate_and_save_available_dimensions` function I separated the logic into single responsibility functions. The purpose of each function should be much more straightforward

mike-jumper (approved):
LGTM.

@necouchman / @jmuehlner: LGTY?

necouchman (approved):
:shipit: 

@jmuehlner ?

jmuehlner:
LGTM!

PR #385, opened 2022-07-07
GUACAMOLE-1622: Restructured code to resolve scrollbar resizing bug.

aleitner:


aleitner:
> What's the bug?

The scrollbar was sometimes resized to be cut off at the right side of the terminal and it would also not touch the bottom and top edges of the terminal

src/terminal/terminal.c
```diff
@@ -1458,22 +1428,38 @@ int guac_terminal_resize(guac_terminal* terminal, int width, int height) {
     /* Acquire exclusive access to terminal */
     guac_terminal_lock(terminal);
 
-    /* Calculate available text display area by character size */
-    int rows, columns;
-    calculate_rows_and_columns(terminal, height, width, &rows, &columns);
+   /* Set size of available screen area */
```
mike-jumper, line 1431: Looks like a space was accidentally removed here (indentation is now misaligned).

src/terminal/terminal.c
```diff
@@ -1458,22 +1428,38 @@ int guac_terminal_resize(guac_terminal* terminal, int width, int height) {
...
+        available_height = rows * display->char_height;
+        height = available_height + 2 * margin;
+    }
+
+    /* Keep width within predefined maximum */
+    if (columns > GUAC_TERMINAL_MAX_COLUMNS) {
+        columns = GUAC_TERMINAL_MAX_COLUMNS;
+        available_width = columns * display->char_width;
+        width = available_width + GUAC_TERMINAL_SCROLLBAR_WIDTH + 2 * margin;
+    }
```
mike-jumper, line 1458: This used to be within `calculate_rows_and_columns()`, avoiding duplication and ensuring bounds are always enforced. Why did this need to be removed from `calculate_rows_and_columns()` and re-duplicated?

mike-jumper (changes requested):

src/terminal/terminal.c
aleitner: This was due to me reverting to the code's state when it was previously working properly, and then working from there. I agree that re-duplication is a bad thing so I made sure to remove the duplicated code again.

src/terminal/terminal.c
```diff
@@ -500,26 +528,33 @@ guac_terminal* guac_terminal_create(guac_client* client,
...
+    int adjusted_height = height;
+    int adjusted_width = width;
+    calculate_adjusted_height_and_width(term, rows, columns, 
+        &adjusted_height, &adjusted_width);
 
     /* Set pixel size */
-    term->height = available_height;
-    term->width = available_width;
+    term->width = adjusted_width;
+    term->height = adjusted_height;
```
mike-jumper, line 553: This body of code is now identical to its counterpart in `guac_terminal_resize()`:

https://github.com/apache/guacamole-server/blob/f6fcd357616b517eaf19f41c77538704f5a44e24/src/terminal/terminal.c#L1497-L1519

Can these not be rolled into a single, abstract function?

mike-jumper (changes requested):

src/terminal/terminal.c
aleitner: `outer_height` - The outer height of the terminal, in pixels.
`outer_width` - The outer width of the terminal, in pixels.

There are 3 groups of heights and widths that need to be calculated using the `outer_height` and `outer_width` which are defined below:

`available_height` - The height of the display area available to display text, in pixels.
`available_width` - The width of the display area available to display text, in pixels.

`rows` - The height of the display area available to display text, in characters.
`columns` - The width of the display area available to display text, in characters.

`adjusted_height` - The total height of the display area available, in pixels. This is adjusted if `rows` exceeds `GUAC_TERMINAL_MAX_ROWS`
`adjusted_width` - The total width of the display area available, in pixels. This is adjusted if `columns` exceeds `GUAC_TERMINAL_MAX_COLUMNS`

We need `outer_height` and `outer_width` to calculate `available_height` and `available_width`.
We need `available_height` and `available_width` to calculate rows and columns.
We need rows and columns to calculate `adjusted_height` and `adjusted_width`.

I felt as if separating into the 3 distinct operations was cleaner than combing into a large function and returning the 4 values. (`rows`, `columns`, `adjusted_height`, and `adjusted_width`). However, if you are more aligned with a single function I am happy to put it all together

```
static void calculate_dimensions(int outer_height, int outer_width, 
        int *rows, int *columns, int *adjusted_height, int *adjusted_width
);
```

aleitner: I have pushed a commit that shows a solution that tries to make as few changes as possible to resolve the issue, but I feel that previous commit `f6fcd357616b517eaf19f41c77538704f5a44e24` is a more solid design

mike-jumper: Would you be happy with the changes from f6fcd357616b517eaf19f41c77538704f5a44e24, but with the set of common calls to those functions instead called by some sort of more abstract setter for `outer_height` and `outer_width`?

aleitner: This would work for `guac_terminal_create()` but `guac_terminal_resize()` needs the rows and columns to be returned and used later rather than be set on the terminal struct

mike-jumper: Welp, these changes as they stand now are definitely nice, readable, and minimal. If you can squash that "Propose a fix ..." commit up into the main commit (or rephrase to capture the nature of the change rather than the nature of the PR workflow), I'm happy to merge.

aleitner: Okay, I squashed it into a single commit!

mike-jumper (approved):

PR #387, opened 2022-07-30
GUACAMOLE-1652: Only call SSL init functions when the library version requires it.

jmuehlner:
This change fixes the issues with SFTP + VNC as described here: https://issues.apache.org/jira/browse/GUACAMOLE-1652.

### DIFF
--- a/src/guacd/daemon.c
+++ b/src/guacd/daemon.c
@@ -381,10 +381,15 @@ int main(int argc, char* argv[]) {
         CRYPTO_set_locking_callback(guacd_openssl_locking_callback);
 #endif
 
-        /* Init SSL */
+#if OPENSSL_VERSION_NUMBER < 0x10100000L
+        /* Init OpenSSL for OpenSSL Versions < 1.1.0 */
         SSL_library_init();
         SSL_load_error_strings();
         ssl_context = SSL_CTX_new(SSLv23_server_method());
+#else
+        /* Set up OpenSSL for OpenSSL Versions >= 1.1.0 */
+        ssl_context = SSL_CTX_new(TLS_server_method());
+#endif
 
         /* Load key */
         if (config->key_file != NULL) {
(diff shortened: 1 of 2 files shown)


necouchman (approved):

necouchman:
Should we also update this location, where SSL is initialized for Client -> guacd connections:

https://github.com/apache/guacamole-server/blob/eee3ac092c8d1c2b17404c7b52c8a2b794d1ec1a/src/guacd/daemon.c#L384-L387

jmuehlner:
> Should we also update this location, where SSL is initialized for Client -> guacd connections:
> 
> https://github.com/apache/guacamole-server/blob/eee3ac092c8d1c2b17404c7b52c8a2b794d1ec1a/src/guacd/daemon.c#L384-L387

Yeah, might as well. It didn't break anything like this for me in testing, but those functions are not needed when the OpenSSL version is <= 1.1.0.

src/guacd/daemon.c
```diff
@@ -381,10 +381,15 @@ int main(int argc, char* argv[]) {
...
 
-        /* Init SSL */
+#if OPENSSL_VERSION_NUMBER < 0x10100000L
+        /* Init OpenSSL for OpenSSL Versions < 1.1.0 */
         SSL_library_init();
         SSL_load_error_strings();
         ssl_context = SSL_CTX_new(SSLv23_server_method());
+#else
+        /* Set up OpenSSL for OpenSSL Versions >= 1.1.0 */
+        ssl_context = SSL_CTX_new(TLS_server_method());
```
jmuehlner, line 391: Per https://www.openssl.org/docs/manmaster/man3/SSLv23_server_method.html, 

> SSLv23_method(), SSLv23_server_method(), SSLv23_client_method()
> These functions do not exist anymore, they have been renamed to TLS_method(), TLS_server_method() and TLS_client_method() respectively. Currently, the old function calls are renamed to the corresponding new ones by preprocessor macros, to ensure that existing code which uses the old function names still compiles. However, using the old function names is deprecated and new code should call the new functions instead.

necouchman: Nice catch.

necouchman (approved):

PR #388, opened 2022-08-15
GUACAMOLE-1540: Manual build all core protocol libraries for Docker image using Alpine Linux base.

mike-jumper:
This change overhauls the guacamole-server Docker image build such that all core protocol libraries (FreeRDP, libssh2, etc.) are built from source as part of the build on top of a minimal Alpine Linux base. The latest releases of each library are automatically built based on a tag pattern match (PCRE) which can be overridden with a different pattern/tag/branch/commit at build time.

Need the support provided by a new libssh2 or FreeRDP release? You can now get that without having to wait for packages to be built by the maintainers of the distro whose image we happen to currently use for Docker. Need to test things that haven't been released yet? You can do that, too. You can also disable support for a protocol entirely by setting its relevant build arg to `NO`.

For example:

```console
$ docker build -t guacd . # Include support for everything
$ docker build -t guacd --build-arg WITH_FREERDP=stable-2.0 . # Force the build to use FreeRDP's "stable-2.0" dev branch
$ docker build -t guacd --build-arg WITH_FREERDP=NO . # Don't include FreeRDP at all
```

The switch to Alpine also means that this image has been cut in size by almost half:

```console
$ docker images guacamole/guacd:latest
REPOSITORY        TAG       IMAGE ID       CREATED        SIZE
guacamole/guacd   latest    f72e1cd4a0e3   20 hours ago   273MB
$ docker images guacd
REPOSITORY   TAG       IMAGE ID       CREATED         SIZE
guacd        latest    2fcfbcc92a7f   3 minutes ago   139MB
$ 
```


src/guacd-docker/bin/list-dependencies.sh
```diff
@@ -35,19 +35,12 @@ while [ -n "$1" ]; do
     ldd "$1" | grep -v 'libguac' | awk '/=>/{print $(NF-1)}' \
         | while read LIBRARY; do
 
-        # In some cases, the library that's linked against is a hard link
-        # to the file that's managed by the package, which dpkg doesn't understand.
-        # Searching by */basename ensures the package will be found in these cases.
-        LIBRARY_BASENAME=$(basename "$LIBRARY")
```
jmuehlner, line 41: I'm glad we don't need this anymore.

mike-jumper: Don't worry - that funkiness had to be replaced with new funkiness: the `sed 's/\(.*\)-[0-9]\..*$/\1/'` at the end.

The `apk` package management tool provides a `-W` option to find the package that owns a file, but the names of the packages printed are in a format that `apk` can't consume for installation. For example:

```console
$ docker run --rm -it alpine
/ # apk add libssl1.1
fetch https://dl-cdn.alpinelinux.org/alpine/v3.16/main/x86_64/APKINDEX.tar.gz
fetch https://dl-cdn.alpinelinux.org/alpine/v3.16/community/x86_64/APKINDEX.tar.gz
OK: 6 MiB in 14 packages
/ # apk info -W /lib/libssl.so.1.1 
/lib/libssl.so.1.1 is owned by libssl1.1-1.1.1q-r0
/ # apk add libssl1.1-1.1.1q-r0
ERROR: unable to select packages:
  libssl1.1-1.1.1q-r0 (no such package):
    required by: world[libssl1.1-1.1.1q-r0]
/ # apk add libssl1.1
OK: 6 MiB in 14 packages
/ # 
```

jmuehlner: Oh ok, cool. We don't want _too_ much improvement all at once.

jmuehlner:
So just to be clear, this change is relevant to https://issues.apache.org/jira/projects/GUACAMOLE/issues/GUACAMOLE-1540 because it allows building against a newer version of libssh, right?

jmuehlner:
How exactly does this PR help with https://issues.apache.org/jira/projects/GUACAMOLE/issues/GUACAMOLE-1540?

Is it allowing building against a newer libssh?

src/guacd-docker/bin/list-dependencies.sh
jmuehlner: Although, that is a pretty gnarly sed command. A comment explaining what it's doing would be nice.

mike-jumper: Fair - I'll add that.

mike-jumper:
> How exactly does this PR help with https://issues.apache.org/jira/projects/GUACAMOLE/issues/GUACAMOLE-1540?
> 
> Is it allowing building against a newer libssh?

Yes - it avoids this continuing to be a problem by expanding on the merged solution (updating to a distro that has a newer libssh2) by making us independent of distro with respect to those protocol libraries.

We frequently end up in the situation where support for some feature is dependent on an update in an underlying protocol support library, with no code changes on our part, and end up having to tell users that they need to wait for their distro to update their packages. This should hopefully avoid that going forward.

src/guacd-docker/bin/list-dependencies.sh
mike-jumper: OK - fixed via rebase.

jmuehlner (approved):

myamoto:
Hi,

Finally built this Alpine docker Image (to remove security vulnerabilities detected from the 1.4.0 official build).

I had to add cairo and pango as dependencies on the runtime environment dependencies in the dockerfile.
Right here :
https://github.com/apache/guacamole-server/blob/15f6e9f678a584b0fe374ddd279ec453515008ae/Dockerfile#L170

Else I would get those errors on the pod :
```
> ldd  /opt/guacamole/lib/libguac-client-ssh.so
> Error loading shared library libpangocairo-1.0.so.0: No such file or directory (needed by /opt/guacamole/lib/libguac-terminal.so.0)
> Error loading shared library libpango-1.0.so.0: No such file or directory (needed by /opt/guacamole/lib/libguac-terminal.so.0)
> Error loading shared library libgobject-2.0.so.0: No such file or directory (needed by /opt/guacamole/lib/libguac-terminal.so.0) 
```

mike-jumper:
> Finally built this Alpine docker Image (to remove security vulnerabilities detected from the 1.4.0 official build).

The `latest` tag for the Docker image is already rebuilt nightly to incorporate any patches released by package maintainers and the underlying distro image. These changes improve on that further, but you don't _need_ these changes to have an up-to-date image.

> I had to add cairo and pango as dependencies on the runtime environment dependencies in the dockerfile.

This should happen automatically via:

https://github.com/apache/guacamole-server/blob/15f6e9f678a584b0fe374ddd279ec453515008ae/Dockerfile#L141-L146

and:

https://github.com/apache/guacamole-server/blob/15f6e9f678a584b0fe374ddd279ec453515008ae/Dockerfile#L178

See: https://github.com/apache/guacamole-server/blob/15f6e9f678a584b0fe374ddd279ec453515008ae/src/guacd-docker/bin/list-dependencies.sh

Sounds like there must be something the above script is missing that prevents Pango from getting recognized as a dependency. Cairo is recognized and pulled in:

```console
$ docker run --rm guacd cat /opt/guacamole/DEPENDENCIES 
brotli-libs
cairo
expat
fontconfig
freetype
libbz2
libcrypto1.1
libjpeg-turbo
libpng
libssl1.1
libuuid
libwebp
libx11
libxau
libxcb
libxdmcp
libxext
libxrender
musl
pixman
zlib
$
```

myamoto:
Hi @mike-jumper,

Thanks for your prompt and thorough reply.

> The `latest` tag for the Docker image is already rebuilt nightly to incorporate any patches released by package maintainers and the underlying distro image. These changes improve on that further, but you don't _need_ these changes to have an up-to-date image.

I like freezing the image at some point in time, forking the code at a precise commit to be able to build it again if needed, avoiding possible regressions with latest built.

> This should happen automatically via:
> 
> https://github.com/apache/guacamole-server/blob/15f6e9f678a584b0fe374ddd279ec453515008ae/Dockerfile#L141-L146
> 
> and:
> 
> https://github.com/apache/guacamole-server/blob/15f6e9f678a584b0fe374ddd279ec453515008ae/Dockerfile#L178
> 
> See: https://github.com/apache/guacamole-server/blob/15f6e9f678a584b0fe374ddd279ec453515008ae/src/guacd-docker/bin/list-dependencies.sh
> 
> Sounds like there must be something the above script is missing that prevents Pango from getting recognized as a dependency. Cairo is recognized and pulled in:
> 
> ```
> $ docker run --rm guacd cat /opt/guacamole/DEPENDENCIES 
> brotli-libs
> cairo
> expat
> fontconfig
> freetype
> libbz2
> libcrypto1.1
> libjpeg-turbo
> libpng
> libssl1.1
> libuuid
> libwebp
> libx11
> libxau
> libxcb
> libxdmcp
> libxext
> libxrender
> musl
> pixman
> zlib
> $
> ```

That's great thanks for the clues.

Thank you guys for such reactivity on PR 389. I tested it and it's working  seemlessly.

PR #390, opened 2022-08-23
GUACAMOLE-1627: Enforce TLS1.2 

sivagudivadaz:
This is a one-off issue when communicating with Windows 2019 server that has both TLS1.3 and TLS1.2 enabled. 
Win 2019 has partial/experimental support for TLS1.3 - although it advertises this during version negotiation, it doesn't allow the client (FreeRDP) to make a successful RDP connection. 

A new version of FreeRDP has been released https://github.com/FreeRDP/FreeRDP/releases/tag/2.8.0 that supports TLS Version control for this use case, the changes in this PR utilize this feature to establish a successful connection.


necouchman:
I'm wondering if, rather than forcing a setting given the FreeRDP library version, it makes more sense to make this a configurable option that can be driven from the client-side, perhaps with a default value chosen based on the FreeRDP library and/or OpenSSL library version. This would incorporate some of what you've done, but also allow admins to drive particular versions if they so desire, and they've correctly configured the RDP server.

Also, regarding your commit message:
* The commit message needs to include the GUACAMOLE-1627 prefix.
* The commit message should be a bit more descriptive than that - Enforce TLS1.2 in what situations? Why?

sivagudivadaz:
Regarding commit messages: will reword the commit message and provide better descriptive info for the changes. 
>  it makes more sense to make this a configurable option that can be driven from the client-side, perhaps with a default value chosen based on the FreeRDP library and/or OpenSSL library version.

Will look into this, thank you!

sivagudivadaz:
> make this a configurable option that can be driven from the client-side
Are you referring to the guacamole-client project?

necouchman:
@sivagudivadaz : It would require changes in both guacamole-server and guacamole-client, yes. You'd need to make the changes in this code to allow a connection parameter to be provided that would allow you to override the TLS version. The changes in the guacamole-client code would be required to allow connections to provide that parameter.

sivagudivadaz:
@necouchman  Does this look like a good spot to introduce a parameter to version control TLS ? 
image

I don't think the user will run into TLS1.3 issues that often so my preference would be to introduce a configuration setting somewhere guacd can pick up from instead of making a UI change. 

Thoughts ?

necouchman:
@sivagudivadaz : Yes, that seems to be a reasonable place for it.

sivagudivadaz:
Thank you for confirming, Nick. I will raise a PR soon.

src/protocols/rdp/settings.c
```diff
@@ -1604,6 +1612,17 @@ void guac_rdp_push_settings(guac_client* client,
...
 
+    // FreeRDP allows for TLS Version control starting 2.8.0
+#if (defined FREERDP_VERSION_MAJOR && FREERDP_VERSION_MAJOR >=2 && defined FREERDP_VERSION_MINOR && FREERDP_VERSION_MINOR >=8 && defined FREERDP_VERSION_REVISION && FREERDP_VERSION_REVISION >=0)
+    // Faulty servers with partial support for TLSv1.3, like windows server 2019, 
+    // trick FreeRDP into negotiating TLSv1.3 and then send back a RST response after initial "Client Hello" during handshake.
+    // Setting the min and max versions of TLS version allows us to enforce the TLS version the client(FreeRDP) chooses.
+    // Note that older versions of FreeRDP that relied on older versions of Openssl that didn't have TLS1.3 don't run into 
+    // this issue as the max TLS version supported by those clients is TLS1.2.
+    rdp_settings->TLSMinVersion = 0;
+    rdp_settings->TLSMaxVersion = TLS1_2_VERSION;
```
jclc, line 1623: Rather than hard-coding this for 1.2 only, wouldn't it be better to add configuration options for defining min and max versions? For example, [Redemption](https://github.com/wallix/redemption) uses such configuration options.

necouchman: @jclc: Yes, I mentioned this in my original comment, above, and most of the conversation between @sivagudivadaz and me has been regarding the best place to put such an option or options.

I suppose we can talk about whether it makes more sense to force a single specific version or provide options for minimum and maximum versions. I'm not opposed to a version range, per se, but I wonder how valuable it will be - if you have a connection to a single server, it may not be all that useful to have a range of options, as the server is usually going to consistently support the same level, and you ought to be able to just enforce that supported level.

jclc: What about using the [Guacamole protocol](https://guacamole.apache.org/doc/gug/guacamole-protocol.html) to pass these as optional arguments? Then you could do it on a connection-by-connection basis.

necouchman: Yes, that's exactly what we're already talking about. The conversation and screenshot above deals with making these TLS versions configurable on a per-connection basis - if they are added as connection parameters they would be passed along with the rest of the options (hostname, port, username, password, etc.) within the Guacamole Protocol as part of establishing the connection.

PR #391, opened 2022-08-23
GUACAMOLE-1669: Prefer FIPS compliant ciphers and algorithms when FIPS mode is enabled.

jmuehlner:
As discussed in https://issues.apache.org/jira/browse/GUACAMOLE-1669, this change prefers FIPS-compliant ciphers and algorithms if FIPS mode is enabled on the guacd host.

### DIFF
--- a/src/libguac/fips.c
+++ b/src/libguac/fips.c
@@ -1,0 +1,51 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+#include "config.h"
+#include "guacamole/fips.h"
+
+/* If OpenSSL is available, include header for version numbers */
+#ifdef ENABLE_SSL
+    #include <openssl/opensslv.h>
+
+    /* OpenSSL versions prior to 0.9.7e did not have FIPS support */
+    #if !defined(OPENSSL_VERSION_NUMBER) || (OPENSSL_VERSION_NUMBER < 0x00090705f)
+    #define GUAC_FIPS_ENABLED 0
+
+    /* OpenSSL 3+ uses EVP_default_properties_is_fips_enabled() */
+    #elif defined(OPENSSL_VERSION_MAJOR) && (OPENSSL_VERSION_MAJOR >= 3)
+    #include <openssl/evp.h>
+    #define GUAC_FIPS_ENABLED EVP_default_properties_is_fips_enabled(NULL)
+
+    /* For OpenSSL versions between 0.9.7e and 3.0, use FIPS_mode() */
+    #else
+    #include <openssl/crypto.h>
+    #define GUAC_FIPS_ENABLED FIPS_mode()
+    #endif
+
+/* FIPS support does not exist if OpenSSL is not available. */
+#else
+#define GUAC_FIPS_ENABLED 0
+#endif
+
+int guac_fips_enabled() {
+
+    return GUAC_FIPS_ENABLED;
+
+}
--- a/src/libguac/guacamole/fips.h
+++ b/src/libguac/guacamole/fips.h
@@ -1,0 +1,33 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+#ifndef GUAC_FIPS_H
+#define GUAC_FIPS_H
+
+/**
+ * Returns a non-zero value if FIPS mode is enabled, or zero if FIPS mode
+ * is not enabled.
+ *
+ * @return
+ *      A non-zero value if FIPS mode is enabled, or zero if FIPS mode is
+ *      not enabled.
+ */
+int guac_fips_enabled();
+
+#endif
--- a/src/common-ssh/ssh.c
+++ b/src/common-ssh/ssh.c
@@ -22,6 +22,7 @@
 #include "common-ssh/user.h"
 
 #include <guacamole/client.h>
+#include <guacamole/fips.h>
 #include <libssh2.h>
 
 #ifdef LIBSSH2_USES_GCRYPT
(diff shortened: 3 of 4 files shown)


src/libguac/fips.c
```diff
@@ -0,0 +1,68 @@
...
+	return 0;
+
+}
+
+/* OpenSSL 3+ uses EVP_default_properties_is_fips_enabled() */
+#elif defined(OPENSSL_VERSION_MAJOR) && (OPENSSL_VERSION_MAJOR >= 3)
+
+#include <openssl/evp.h>
+
+int guac_fips_enabled() {;
```
jmuehlner, line 42: I'm repeating the function definition here because these different FIPS checks require different headers to be included, and this seemed like the most straightforward way to do it.

necouchman: Is that syntax correct, though, with the `;` at the end of the line?

jmuehlner: Definitely not! fixed.

jmuehlner: Let me test this again before I un-draft it.

necouchman:
Cool, looks okay to me as long as it tests okay. My only other question would be if it's worth making it configurable or not? I'm guessing not - since i'ts just setting an order of preferred cipher and key algorithms, I would guess if it supports it you want it. Just throwing it out, though.

src/libguac/fips.c
jmuehlner: Ooooops, I somehow lost my Makefile changes. Definitely didn't work without those.

jmuehlner:
Yeah, I kinda feel like it's probably not worth it. In the general case, we don't really want to be telling libssh2 what it should be using at all - it's job is to negotiate a set of algorithms and ciphers that both server and client can use. Unfortunately it doesn't seem to work in the case of FIPS.

Luckily for FIPS, there's a pretty small set of options that are both FIPS-compliant, AND libssh2-supported. I just listed those from biggest key to smallest. I guess you could imagine somebody wanting to prefer smaller key sizes for performance reasons, but I'd guess that nobody would ever use such an option ¯\_(ツ)_/¯

> Cool, looks okay to me as long as it tests okay. My only other question would be if it's worth making it configurable or not? I'm guessing not - since i'ts just setting an order of preferred cipher and key algorithms, I would guess if it supports it you want it. Just throwing it out, though.

mike-jumper:
> ...  In the general case, we don't really want to be telling libssh2 what it should be using at all - it's job is to negotiate a set of algorithms and ciphers that both server and client can use. ...

I would echo that. We might need to do this now for the sake of having thins work in a FIPS environment _now_, but in the general case it's one of the things that should just work within the underlying library.

mike-jumper:
> Luckily for FIPS, there's a pretty small set of options that are both FIPS-compliant, AND libssh2-supported. I just listed those from biggest key sizes to smallest. I guess you could imagine somebody wanting to prefer smaller key sizes for performance reasons, but I'd guess that nobody would ever use such an option ¯_(ツ)_/¯

Maybe there's a way we can check what OpenSSH considers its order of priority in a FIPS environment?

Googling around, I found [this page](https://techhub.hpe.com/eginfolib/networking/docs/switches/5130ei/5200-3946_security_cg/content/485048549.htm) which states:

> ... In FIPS mode ... SSH2 uses the aes128-ctr, aes192-ctr, aes256-ctr, aes128-gcm, aes256-gcm, aes128-cbc, and aes256-cbc encryption algorithms in descending order of priority for algorithm negotiation.

Which seems to indicate:

* Prefer CTR, then GCM, then finally CBC.
* Use as small a key as the SSH server will allow.

There might be good reason or documentation for that.

src/libguac/fips.c
```diff
@@ -0,0 +1,67 @@
...
+#include <openssl/evp.h>
+
+int guac_fips_enabled() {
+
+	return EVP_default_properties_is_fips_enabled(NULL);
+
+}
+
+/* For OpenSSL versions between 0.9.7e and 3.0, use FIPS_mode() */
+#else
```
mike-jumper, line 48: I'm finding this somewhat hard to read, and I think that's in part because the frequency of empty lines makes this effectively double-spaced.

I have some ideas that might clean things up a smidge:

1. Nix the extra spacing within the one-line func:

   ```c
   int foo() {
       return 0;
   }
   ```

   or ...
2. ... Leverage macros to avoid having to redeclare everything. For example:

   ```c
   #if thing1
   #include <foo.h>
   #define GUAC_FIPS_ENABLED EVP_default_properties_is_fips_enabled(NULL)
   #elif thing2
   #include <bar.h>
   #define GUAC_FIPS_ENABLED FIPS_mode()
   #else
   #define GUAC_FIPS_ENABLED 0
   #endif

   int guac_fips_enabled() {
       return GUAC_FIPS_ENABLED;
   }

   ```

src/libguac/fips.c
```diff
@@ -0,0 +1,67 @@
...
+}
+
+/* OpenSSL 3+ uses EVP_default_properties_is_fips_enabled() */
+#elif defined(OPENSSL_VERSION_MAJOR) && (OPENSSL_VERSION_MAJOR >= 3)
+
+#include <openssl/evp.h>
+
+int guac_fips_enabled() {
+
+	return EVP_default_properties_is_fips_enabled(NULL);
```
mike-jumper, line 43: Beware tabs have snuck in.

necouchman:
I should clarify that I wasn't necessarily saying that we should allow configuration of individual cipher/key algorithms, just whether FIPS mode is enabled or not. But, I'll go with whatever you guys think on that - it was an honest question, not a leading one :-).

src/libguac/fips.c
jmuehlner: Ah, good call! That is much better, thanks. My C preprocessor skills are apparently quite rusty so this is a good learning experience.

src/libguac/fips.c
jmuehlner: I'll get em.

src/libguac/guacamole/fips.h
```diff
@@ -0,0 +1,33 @@
...
+#ifndef GUAC_FIPS_H
+#define GUAC_FIPS_H
+
+/**
+ * Returns a positive value if FIPS mode is enabled for OpenSSL, or zero if
+ * FIPS mode is not enabled.
+ *
+ * @return
+ *      A positive value if FIPS mode is enabled for OpenSSL, or zero if FIPS
+ *      mode is not enabled.
```
mike-jumper, line 29: Rather than bake into the API that this value must be positive when true, I suggest instead documenting the value as non-zero.

jmuehlner:
> 



> > Luckily for FIPS, there's a pretty small set of options that are both FIPS-compliant, AND libssh2-supported. I just listed those from biggest key sizes to smallest. I guess you could imagine somebody wanting to prefer smaller key sizes for performance reasons, but I'd guess that nobody would ever use such an option ¯_(ツ)_/¯
> 
> Maybe there's a way we can check what OpenSSH considers its order of priority in a FIPS environment?
> 
> Googling around, I found https://techhub.hpe.com/eginfolib/networking/docs/switches/5130ei/5200-3946_security_cg/content/485048549.htm which states:
> 
> > ... In FIPS mode ... SSH2 uses the aes128-ctr, aes192-ctr, aes256-ctr, aes128-gcm, aes256-gcm, aes128-cbc, and aes256-cbc encryption algorithms in descending order of priority for algorithm negotiation.
> 
> Which seems to indicate:
> 
> * Prefer CTR, then GCM, then finally CBC.
> * Use as small a key as the SSH server will allow.
> 
> There might be good reason or documentation for that.



> > Luckily for FIPS, there's a pretty small set of options that are both FIPS-compliant, AND libssh2-supported. I just listed those from biggest key sizes to smallest. I guess you could imagine somebody wanting to prefer smaller key sizes for performance reasons, but I'd guess that nobody would ever use such an option ¯_(ツ)_/¯
> 
> Maybe there's a way we can check what OpenSSH considers its order of priority in a FIPS environment?
> 
> Googling around, I found https://techhub.hpe.com/eginfolib/networking/docs/switches/5130ei/5200-3946_security_cg/content/485048549.htm which states:
> 
> > ... In FIPS mode ... SSH2 uses the aes128-ctr, aes192-ctr, aes256-ctr, aes128-gcm, aes256-gcm, aes128-cbc, and aes256-cbc encryption algorithms in descending order of priority for algorithm negotiation.
> 
> Which seems to indicate:
> 
> * Prefer CTR, then GCM, then finally CBC.
> * Use as small a key as the SSH server will allow.
> 
> There might be good reason or documentation for that.

Ah yeah, that's a good call. I'll switch the order to lowest keysize first. That's what libssh2 was doing with the other algorithms first anyway, so it's a good call. FWIW, the GCM algorithms don't seem to be supported by libssh2, so I just left those out https://ssh-comparison.quendi.de/comparison/cipher.html

jmuehlner:
Ah, gotcha. Yeah I don't think I see a use case where you'd want to manually be turning this off and on.

> I should clarify that I wasn't necessarily saying that we should allow configuration of individual cipher/key algorithms, just whether FIPS mode is enabled or not. But, I'll go with whatever you guys think on that - it was an honest question, not a leading one :-).

mike-jumper (commented):
I think we should perhaps avoid tying the definition of `guac_fips_enabled()` to OpenSSL, and let it be independent of the fact that we currently only consider OpenSSL for this.

Otherwise, LGTM.

src/libguac/guacamole/fips.h
```diff
@@ -0,0 +1,33 @@
...
+#ifndef GUAC_FIPS_H
+#define GUAC_FIPS_H
+
+/**
+ * Returns a non-zero value if FIPS mode is enabled for OpenSSL, or zero if
+ * FIPS mode is not enabled.
+ *
+ * @return
+ *      A non-zero value if FIPS mode is enabled for OpenSSL, or zero if FIPS
+ *      mode is not enabled.
```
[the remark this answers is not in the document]
jmuehlner: Yeah, that's reasonable. I got rid of the OpenSSL stuff.

mike-jumper (approved):

PR #395, opened 2022-09-28
GUACAMOLE-1682: Normalize clipboard for ssh sessions.

aleitner:


necouchman:
@aleitner : This pull request needs to be associated with a Jira issue, and be tagged as such. Take a look at other pull requests for examples, and see the following page for full contribution guidelines: https://guacamole.apache.org/open-source/.

src/protocols/ssh/clipboard.c
```diff
@@ -47,14 +48,30 @@ int guac_ssh_clipboard_blob_handler(guac_user* user, guac_stream* stream,
...
 
-    /* Nothing to do - clipboard is implemented within client */
+    guac_client* client = user->client;
+    guac_ssh_client* ssh_client = (guac_ssh_client*) client->data;
+
+    char output_data[GUAC_COMMON_CLIPBOARD_MAX_LENGTH];
+
+    const char* input = ssh_client->clipboard->buffer;
+    char* output = output_data;
+    guac_iconv_write* writer = ssh_client->settings->clipboard_crlf ? GUAC_WRITE_CP1252_CRLF : GUAC_WRITE_CP1252;
```
mike-jumper, line 65: This will result in the clipboard contents being transformed into [Windows CP-1252](https://en.wikipedia.org/wiki/Windows-1252).

src/protocols/ssh/settings.h
```diff
@@ -178,6 +178,19 @@ typedef struct guac_ssh_settings {
...
+     * normalized to Unix-style newline characters.
+     */
+    int normalize_clipboard;
+
+    /**
+     * Whether Unix-style newline characters within the clipboard should be
+     * automatically translated to CRLF sequences before transmission to the
+     * SSH server.
+     */
+    int clipboard_crlf;
```
mike-jumper, line 192: Is there any case where the terminal clipboard contents shouldn't be normalized to UNIX-style newlines? When data is pasted within a terminal emulator session, it's passed through to the remote end as if entered by the user. In all cases, if a user presses the "Enter" or "Return" key, only a single character will be sent:

https://github.com/apache/guacamole-server/blob/b20afa275a78ca4a54f4e088413e99c93901a3bc/src/terminal/terminal.c#L1675

If we selectively normalize or provide an option to transform these newlines into CRLF pairs, we'd be adding an additional character and producing a sequence that isn't normally produced by user input.

src/protocols/ssh/client.c
```diff
@@ -44,6 +44,9 @@ int guac_client_init(guac_client* client) {
     guac_ssh_client* ssh_client = calloc(1, sizeof(guac_ssh_client));
     client->data = ssh_client;
 
+    /* Init clipboard */
+    ssh_client->clipboard = guac_common_clipboard_alloc();
```
mike-jumper, line 48: I don't think we should avoid using the clipboard provided by the terminal:

* The underlying issue here likely affects all terminal-related protocols, not just SSH.
* Switching to a dedicated terminal here would mean that options like `disable-copy` and `disable-paste` would need to be manually reimplemented.

Instead, this should be done at the terminal level. If there needs to be an option for this, then a terminal option can be provided, though I suspect we can safely normalize to UNIX-style newlines for all terminal protocols because of how the terminal clipboard functions.

mike-jumper (changes requested):

src/common/clipboard.c
```diff
@@ -154,13 +155,20 @@ void guac_common_clipboard_append(guac_common_clipboard* clipboard, const char*
     if (remaining < length)
         length = remaining;
 
+    char output_data[length];
+    memset(output_data, '\0', length);
```
mike-jumper, line 159: Why?

src/common/clipboard.c
```diff
@@ -154,13 +155,20 @@ void guac_common_clipboard_append(guac_common_clipboard* clipboard, const char*
...
+    char* output = output_data;
+
+    /* Convert clipboard contents */
+    guac_iconv(GUAC_READ_UTF8_NORMALIZED, &data, length,
+               GUAC_WRITE_UTF8, &output, length);
+
     /* Append to buffer */
-    memcpy(clipboard->buffer + clipboard->length, data, length);
+    memcpy(clipboard->buffer + clipboard->length, output_data, length);
```
mike-jumper, line 167: Changing this here within `guac_common_clipboard` will affect all current protocols, including VNC and RDP. Is this intended?

aleitner: Ahh that's true, moved to terminal

src/common/clipboard.c
aleitner: I put it in when I was debugging the variable. Removed

src/terminal/terminal.c
```diff
@@ -2165,11 +2166,19 @@ void guac_terminal_clipboard_reset(guac_terminal* terminal,
 
 void guac_terminal_clipboard_append(guac_terminal* terminal,
         const char* data, int length) {
-    guac_common_clipboard_append(terminal->clipboard, data, length);
+
+    char output_data[length];
+    char* output = output_data;
```
mike-jumper, line 2171: While `guac_common_clipboard_append()` will be inherently bounded by the remaining clipboard length, this will be limited only by the inherent per-instruction size limits on received `blob` instructions. Rather than trust the received `length`, especially for something that's likely allocated on the stack, I think we should enforce an upper bound here.

For example:

https://github.com/apache/guacamole-server/blob/d8073f9b17c3cb09b79591dbe3d0de82f0ae7004/src/common/clipboard.c#L152-L155

Alternatively, if we altered `guac_common_clipboard_append()` to accept an input encoding (`guac_iconv()` reader), then that could directly perform this conversion instead of its existing `memcpy()` for cases where the input is not `GUAC_READ_UTF8`, and we avoid the intermediate buffer entirely.

mike-jumper (changes requested):

src/terminal/terminal.c
```diff
@@ -2165,11 +2166,21 @@ void guac_terminal_clipboard_reset(guac_terminal* terminal,
 
 void guac_terminal_clipboard_append(guac_terminal* terminal,
         const char* data, int length) {
-    guac_common_clipboard_append(terminal->clipboard, data, length);
+
+    /* Allocate and clear space for the converted data */
+    char output_data[GUAC_COMMON_CLIPBOARD_MAX_LENGTH];
+    memset(output_data, '\0', GUAC_COMMON_CLIPBOARD_MAX_LENGTH);
```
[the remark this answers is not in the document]
mike-jumper: You shouldn't need to do a full `memset()`. This sounds like a null-termination issue.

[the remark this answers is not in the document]
jmuehlner: From the documentation of `guac_iconv()`
> The input and output string pointers will be updated based on the number of bytes read or written.

Looks like you could just check how many characters the string pointer has been incremented to get your answer.

Also probably want to make sure that you handle the case where `guac_iconv()` returns `0`. (You wouldn't want to call `strlen()` in that case).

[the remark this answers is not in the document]
aleitner: I completely overlooked that the pointer gets incremented. Thanks!

mike-jumper (changes requested):
LGTM, but please revert the changes to `src/common/clipboard.c`, which is unmodified except for the deletion of a blank line.

mike-jumper (approved):

PR #399, opened 2022-11-07
GUACAMOLE-1714: Update guacenc for const parameters/values introduced in FFmpeg 5.0.

mike-jumper:
This change is simply a direct copy of the changes proposed via #379, amended to include the JIRA issue in the commit message.

This closes #379.


mike-jumper:
This should fix the failures currently being seen for the `guacamole-server-master-ubuntu` and `guacamole-server-master-debian` builds in our CI.

PR #400, opened 2022-11-17
2 problems fixed

Saiyeux:
1. When wrapped line texts selected, no more "\n" written between rows.
2. Try to keep consistency with what vim shows when selecting and clipping.

### DIFF
--- a/src/protocols/rdp/keymaps/cs-cz-qwertz.keymap
+++ b/src/protocols/rdp/keymaps/cs-cz-qwertz.keymap
@@ -40,7 +40,7 @@ map +caps -altgr -shift      0x10..0x1B      ~ "QWERTZUIOPÚ)"
 map +caps -altgr -shift      0x1E..0x28 0x2B ~ "ASDFGHJKLŮ§¨"
 map +caps -altgr -shift      0x2C..0x35      ~ "YXCVBNM,.-"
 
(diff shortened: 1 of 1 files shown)


mike-jumper (changes requested):
Thanks, @Saiyeux. Please see our contribution guidelines:

https://github.com/apache/guacamole-server/blob/master/CONTRIBUTING

In particular:

* Each change needs a corresponding issue in JIRA. If there is no existing JIRA issue for this and you don't have an account on the ASF JIRA, let us know and we'll help set that up.
* Commits should be made in logical units and must reference the JIRA issue number (see established git history for examples). Please avoid commits which cover multiple, distinct goals that could (and should) be handled separately.

mike-jumper:
Also: please do not commit IDE- or OS-specific files like `.DS_Store`. It's OK to add these files to `.gitignore` as part of your changes, but please do not include the files themselves.

Saiyeux:
Ok, I'll do it one by one. By the way, it seems that no related issues appear on jira and I don't have access to jira for apache. Could you please help me with this so that I can pull my request in a formal way according to guidelines? Thanks a lot.

mike-jumper:
@Saiyeux - sure, if you shoot me an email at mjumper@apache.org, I can get a JIRA account created for you.

PR #402, opened 2022-11-21
GUACAMOLE-1586: Wrapped lines now clipped into 1.

Saiyeux:
When an extra long wrapped line is selected, no more \n added to clipboard.

### DIFF
--- a/src/terminal/select.c
+++ b/src/terminal/select.c
@@ -360,16 +360,19 @@ void guac_terminal_select_end(guac_terminal* terminal) {
     else {
 
         /* Store first row */
-        guac_terminal_clipboard_append_row(terminal, start_row, start_col, -1);
+        guac_terminal_buffer_row* buffer_row = guac_terminal_buffer_get_row(terminal->buffer, start_row, 0);
+        guac_terminal_clipboard_append_row(terminal, start_row, start_col, buffer_row->length - 2);
 
         /* Store all middle rows */
         for (int row = start_row + 1; row < end_row; row++) {
-            guac_common_clipboard_append(terminal->clipboard, "\n", 1);
-            guac_terminal_clipboard_append_row(terminal, row, 0, -1);
+            guac_terminal_buffer_row* buffer_row = guac_terminal_buffer_get_row(terminal->buffer, start_row, 0);
+            guac_terminal_clipboard_append_row(terminal, row, 0, buffer_row->length - 2);
+            if(!(buffer_row->characters[buffer_row->length - 1].value == 0 && buffer_row->characters[buffer_row->length - 2].value > 0))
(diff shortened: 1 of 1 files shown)


src/terminal/select.c
```diff
@@ -360,16 +360,19 @@ void guac_terminal_select_end(guac_terminal* terminal) {
...
+        guac_terminal_buffer_row* buffer_row = guac_terminal_buffer_get_row(terminal->buffer, start_row, 0);
+        guac_terminal_clipboard_append_row(terminal, start_row, start_col, buffer_row->length - 2);
 
         /* Store all middle rows */
         for (int row = start_row + 1; row < end_row; row++) {
-            guac_common_clipboard_append(terminal->clipboard, "\n", 1);
-            guac_terminal_clipboard_append_row(terminal, row, 0, -1);
+            guac_terminal_buffer_row* buffer_row = guac_terminal_buffer_get_row(terminal->buffer, start_row, 0);
+            guac_terminal_clipboard_append_row(terminal, row, 0, buffer_row->length - 2);
+            if(!(buffer_row->characters[buffer_row->length - 1].value == 0 && buffer_row->characters[buffer_row->length - 2].value > 0))
```
mike-jumper, line 370: To match the style of established code, please include a space between `if` and the opening paren.

src/terminal/select.c
```diff
@@ -360,16 +360,19 @@ void guac_terminal_select_end(guac_terminal* terminal) {
     else {
 
         /* Store first row */
-        guac_terminal_clipboard_append_row(terminal, start_row, start_col, -1);
+        guac_terminal_buffer_row* buffer_row = guac_terminal_buffer_get_row(terminal->buffer, start_row, 0);
+        guac_terminal_clipboard_append_row(terminal, start_row, start_col, buffer_row->length - 2);
```
mike-jumper, line 364: 1. `buffer_row->length` is not guaranteed to be at least 2 here unless the call to `guac_terminal_buffer_get_row()` requests this.
2. Why are we universally skipping the last two characters?

mike-jumper (changes requested):
I'm also unsure what you mean by:

> Wrapped lines now clipped into 1.

Can you clarify?

src/terminal/select.c
Saiyeux: My purpose it to omit the line break when an extra long line of text is separated because of insufficient length of the terminal simulator. I managed to use buffer_row->length - 1 with server version 1.3.0, but have to cut down another character with 1.4.0. Currently, the horizontal index of bound is length - 2, I'll figure it out with annotations in next commit.

mike-jumper:
Closing in favor of the alternative approach proposed and merged via #512.

PR #410, opened 2023-01-17
GUACAMOLE-1633: Adding alternate screen buffer for non-linux terminal type

Saiyeux:
The default terminal type of guacamole-server is linux. This change optimized screen content when using xterm-like terminal which supports alternate screen buffer.


necouchman:
This has been implemented by #521

PR #411, opened 2023-01-18
GUACAMOLE-377: Send a sync instruction to users when synchronizing surfaces.

aleitner:
After the introduction of RDPGFX the joining client when joining a shared connection appears to hang (does not receive a copy of the current display) until something changes graphically within the session.

This pull request:
* Updates `guac_common_display_dup` to call `guac_protocol_send_sync` to send a sync message to all connected users which will make the joined client's display appear immediately.

### DIFF
--- a/src/common/display.c
+++ b/src/common/display.c
@@ -166,6 +166,8 @@ void guac_common_display_free(guac_common_display* display) {
 void guac_common_display_dup(guac_common_display* display, guac_user* user,
         guac_socket* socket) {
 
+    guac_client* client = user->client;
+
     pthread_mutex_lock(&display->_lock);
 
     /* Sunchronize shared cursor */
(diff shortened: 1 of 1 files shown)


src/common/display.c
```diff
@@ -178,6 +180,9 @@ void guac_common_display_dup(guac_common_display* display, guac_user* user,
     guac_common_display_dup_layers(display->layers, user, socket);
     guac_common_display_dup_layers(display->buffers, user, socket);
 
+    /* Synchronize all connected users */
```
mike-jumper, line 183: This particular line does not synchronize all connected users, but rather marks the boundary of the first frame.

mike-jumper (changes requested):

mike-jumper (approved):

PR #414, opened 2023-02-16
GUACAMOLE-600: Add support for setting SSH and SFTP timeouts.

necouchman:
This pull request adds support for setting the connection timeout for SSH and SFTP connections via connection parameters.


necouchman:
Actually, converting this back to a draft. Since I used a previous change to collect the common socket code together in a single place, this needs to be reworked a bit...

necouchman:
Okay, I think this is ready, again, for review.

mike-jumper (approved):

PR #415, opened 2023-02-16
GUACAMOLE-600: Add support for setting the Telnet connection timeout.

necouchman:
Add support for specifying the connection timeout for Telnet connections.

### DIFF
--- a/src/protocols/telnet/telnet.c
+++ b/src/protocols/telnet/telnet.c
@@ -424,9 +424,29 @@ static telnet_t* __guac_telnet_create_session(guac_client* client) {
                 NI_NUMERICHOST | NI_NUMERICSERV)))
             guac_client_log(client, GUAC_LOG_DEBUG, "Unable to resolve host: %s", gai_strerror(retval));
 
-        /* Connect */
-        if (connect(fd, current_address->ai_addr,
-                        current_address->ai_addrlen) == 0) {
+        fd_set fdset;
+        FD_ZERO(&fdset);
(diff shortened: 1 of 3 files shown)


src/protocols/telnet/telnet.c
```diff
@@ -424,9 +424,25 @@ static telnet_t* __guac_telnet_create_session(guac_client* client) {
...
-                        current_address->ai_addrlen) == 0) {
+        fd_set fdset;
+        FD_ZERO(&fdset);
+        FD_SET(fd, &fdset);
+
+        struct timeval timeout_tv;
+        timeout_tv.tv_sec = settings->timeout;
+        timeout_tv.tv_usec = 0;
+
+        connect(fd, current_address->ai_addr, current_address->ai_addrlen);
```
mike-jumper, line 435: How does this work together with `select()`? Is it OK to ignore the return value of `connect()`?

necouchman: I'll have to re-locate the sources I was using when I was researching how to do this, but my recollection is that adding the `fd` to `fd_set` before calling `connect()` essentially makes that into an asynchronous call, and the `fd_set` is then monitored for activity via the `select()` call, including handling the timeout.

I suppose it may still be worthwhile to check the return value of `connect()` just to make that something unexpected doesn't occur, so I should probably put that back in there.

necouchman: Okay, added back some error checking on the `connect()` call.

mike-jumper (approved):

PR #418, opened 2023-03-17
GUACAMOLE-1717: Fix RDP cursor use of uninitialized memory

khoek:
As kindly described by Mike Powers in [GUACAMOLE-1717](https://issues.apache.org/jira/browse/GUACAMOLE-1717), using GNOME Remote Desktop (GRD) with gaucd causes a clearly junk/corrupt cursor to be rendered. Mike tracked this down to use of uninitialized memory (`data`) in [`src/protocols/rdp/pointer.c#guac_rdp_pointer_new()`](https://github.com/apache/guacamole-server/blob/ce27936ed539872d55039b6a1b9aa54d8388659e/src/protocols/rdp/pointer.c#L35) when the call to `freerdp_image_copy_from_pointer_data()` is skipped because connecting to GRD causes `pointer->andMaskData` to be `NULL`.

As described in the ticket, the `if` guard (on line 49) causing this to happen is not necessary because FreeRDP protects against this case internally (at least in 2.0 versions), and at any rate there is a bug because if FreeRDP can't handle this situation we should at least zero `data` instead of returning garbage.

Tested on Ubuntu 22.10 with the fix applied on top of master.

### DIFF
--- a/src/protocols/rdp/pointer.c
+++ b/src/protocols/rdp/pointer.c
@@ -42,18 +42,17 @@ BOOL guac_rdp_pointer_new(rdpContext* context, rdpPointer* pointer) {
             rdp_client->display, pointer->width, pointer->height);
 
     /* Allocate data for image */
-    unsigned char* data = _aligned_malloc(pointer->width * pointer->height * 4, 16);
+    unsigned char* data = _aligned_recalloc(NULL, 1, pointer->width * pointer->height * 4, 16);
 
     cairo_surface_t* surface;
 
-    /* Convert to alpha cursor if mask data present */
-    if (pointer->andMaskData && pointer->xorMaskData)
-        freerdp_image_copy_from_pointer_data(data,
-                guac_rdp_get_native_pixel_format(TRUE), 0, 0, 0,
-                pointer->width, pointer->height, pointer->xorMaskData,
-                pointer->lengthXorMask, pointer->andMaskData,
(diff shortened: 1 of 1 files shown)


necouchman (commented):
Based on your comment, does it make more sense to remove the `if` check altogether, or would it be better to zero out `data` somewhere else to avoid returning garbage? I think I'd be more inclined to leave the `if` check here and zero out data at some point, if that's feasible.

Also, you tested against Ubuntu - did you also test against other RDP servers to make sure this doesn't cause any regressions in those?

khoek:
@necouchman I think that the `if` should definitely be removed in any case, because it's doing something wrong---if we just zero the buffer then e.g. in my setup with GRD I won't see any cursor, but with the `if` removed I see the correctly transmitted cursor from the server.

But I agree we might as well zero it, so I pushed a change.

As for testing I've also tried the RDP server which ships with an up to date copy of Windows 11, but that's all. I don't think we can cause a regression because:
- The only situation where could introduce one is the case where the `if` condition is `false` and we skip the call, hence return the uninitialized weird-looking junk cursor, and
- so the only way we could cause a regression is by making this *worse*, i.e. the call to `freerdp_image_copy_from_pointer_data()` now crashes the program.
But the `freerdp_image_copy_from_pointer_data()` function correctly handles the cases where `xorMaskData` or `andMaskData` are `NULL` (in the former case it just errors and returns `false`, and in the latter case the empty and mask is correctly handled).

PR #419, opened 2023-03-23
GUACAMOLE-1746: docker enable usage of custom certificate

ronansalmon:
Docker: Added a start.sh script that parse env variables
Added documentation for standalone and docker version on how to create and use a custom certificate.

### DIFF
--- a/src/guacd-docker/sbin/start.sh
+++ b/src/guacd-docker/sbin/start.sh
@@ -25,8 +25,27 @@
 ##
 
 
-# guacd default value if not overridden
+##
+## Adds SSL command-line options to guacd
+##
+enable_ssl() {
+    if [ -z "$GUACD_CERTIFICATE_CRT" ] || [ -z "$GUACD_CERTIFICATE_KEY" ]; then
+        cat <<END
+FATAL: Missing required environment variables
+-------------------------------------------------------------------------------
+If using a SSL encryption, you must provide both GUACD_CERTIFICATE_CRT
+and GUACD_CERTIFICATE_KEY environment variables.
 
+END
+        exit 1;
+    fi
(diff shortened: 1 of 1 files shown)


src/guacd-docker/sbin/start.sh
```diff
@@ -0,0 +1,51 @@
...
+
+args=( -b "${GUACD_LISTEN_ADDRESS}" -L "${GUACD_LOG_LEVEL}" -l "${GUACD_PORT}" -f )
+
+# guacd certificate files
+if [ -n "$GUACD_CERTIFICATE_CRT" ]; then
+    args+=( -C "$GUACD_CERTIFICATE_CRT" )
+fi
+if [ -n "$GUACD_CERTIFICATE_KEY" ]; then
+    args+=( -K "$GUACD_CERTIFICATE_KEY" )
+fi
```
necouchman, line 48: Does it make sense to process these individually? I believe they are both required for SSL to function, so seems like this check should be for both of them, and fail if one is defined without the other.

necouchman (changes requested):
Overall I think this is fine - just one thing that I think should be changed.

@mike-jumper Any concerns with the direction of this change? It adds a start.sh script similar to the guacamole-client Docker image, but seems like Ronan has tried to maintain the same sort of script on this side, so looks okay to me.

src/guacd-docker/sbin/start.sh
ronansalmon: makes sense that they are both required for SSL to work.

necouchman:
@ronansalmon This now how some conflicts that need to be resolved.

@mike-jumper Any feedback on this one?

PR #428, opened 2023-04-23
Update rdp.c

FirstLicz:
remoteapp has no input method when opened since version 1.0


mike-jumper:
@FirstLicz Is there supposed to be something else here? These changes only add a comment.

PR #429, opened 2023-04-24
GUACAMOLE-1776: Batch up base64 encoding to reduce syscalls.

neandrake:
This introduces a buffer for binary data to be encoded as base64 as well as a buffer for the base64 encoded data, prior to any attempt to writing it to the socket. Doing this significantly reduces the number of system calls made by the socket-fd implementation. Instead of `clock_gettime`, `pthread_mutex_rdlock`, and `pthread_mutex_unlock` being called for every 3 bytes it's now called for every ~768 bytes.

### DIFF
--- a/src/libguac/socket.c
+++ b/src/libguac/socket.c
@@ -237,10 +237,32 @@ ssize_t guac_socket_write_string(guac_socket* socket, const char* str) {
 
 }
 
-ssize_t __guac_socket_write_base64_triplet(guac_socket* socket,
-        int a, int b, int c) {
-
-    char output[4];
+/**
+ * Encodes one to three bytes of data as four characters in base64 encoding.
+ *
+ * This function takes int arguments even though it's working with bytes to
+ * allow -1 to be used as distinct sentinel value for missing data.
+ *
+ * @param a An int holding the first byte of data to encode. Only the
+ *          least-significant byte will be used. This will always be inserted
+ *          into the output.
+ *
+ * @param b An int holding the second byte of data to encode. Only the
+ *          least-significant byte will be used. If this is less than zero,
+ *          the second and third bytes will be ignored and the last two
+ *          characters of the output will be '=' padding characters.
+ *
+ * @param c An int holding the third byte of data to encode. Only the
+ *          least-significant byte will be used. If this is less than zero,
+ *          the third byte will be ignored and the last character of the
+ *          output will be a '=' padding character.
+ *
+ * @param output Character buffer to hold the output. Exactly four characters
+ *               will be written to the buffer starting at this location.
+ *
+ * @return Returns zero.
+ */
+ssize_t __guac_socket_encode_base64(int a, int b, int c, char* output) {
 
     /* Byte 0:[AAAAAA] AABBBB BBBBCC CCCCCC */
     output[0] = __guac_socket_BASE64_CHARACTERS[(a & 0xFC) >> 2];
@@ -250,7 +272,7 @@ ssize_t __guac_socket_write_base64_triplet(guac_socket* socket,
         /* Byte 1: AAAAAA [AABBBB] BBBBCC CCCCCC */
         output[1] = __guac_socket_BASE64_CHARACTERS[((a & 0x03) << 4) | ((b & 0xF0) >> 4)];
 
-        /* 
+        /*
          * Bytes 2 and 3, zero characters of padding:
          *
          * AAAAAA  AABBBB [BBBBCC] CCCCCC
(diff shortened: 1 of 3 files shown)


src/libguac/socket.c
```diff
@@ -286,56 +308,70 @@ ssize_t __guac_socket_write_base64_triplet(guac_socket* socket,
...
+ssize_t guac_socket_flush_base64(guac_socket* socket) {
+    const unsigned char* src = socket->__ready_buf;
 
-}
+    int encodedCount = 0;
+    int remaining = socket->__ready;
 
-ssize_t __guac_socket_write_base64_byte(guac_socket* socket, int buf) {
+    // Encode bytes in groups of three
+    while(remaining > 2) {
```
mike-jumper, line 321: Space between `while` and `(` please (see surrounding code).

src/libguac/socket.c
```diff
@@ -286,56 +308,70 @@ ssize_t __guac_socket_write_base64_triplet(guac_socket* socket,
...
 
-    int* __ready_buf = socket->__ready_buf;
+        remaining -= 3;
+        src += 3;
+        encodedCount += 4;
+    }
 
-    int retval;
+    // Take care of partial remnants
+    if(remaining == 2) {
```
mike-jumper, line 330: Space between `if` and `(` please (see surrounding code).

src/libguac/socket.c
```diff
@@ -286,56 +308,70 @@ ssize_t __guac_socket_write_base64_triplet(guac_socket* socket,
...
+        __guac_socket_encode_base64(src[0], src[1], -1, socket->__encoded_buf + encodedCount);
+        encodedCount += 4;
+    } else if(remaining == 1) {
```
mike-jumper, line 333: 1. Space between `else if` and `(` please (see surrounding code).
2. Please do not cuddle the `else if` (see https://guacamole.apache.org/guac-style/#braces)

src/libguac/socket.c
```diff
@@ -286,56 +308,70 @@ ssize_t __guac_socket_write_base64_triplet(guac_socket* socket,
...
 
-        retval = __guac_socket_write_base64_byte(socket, *(char_buf++));
-        if (retval < 0)
-            return retval;
+        socket->__ready += len;
+        src += len;
+        remaining -= len;
 
+        // Flush ready buffer when full
+        if(socket->__ready == GUAC_SOCKET_BASE64_READY_BUFFER_SIZE) {
```
mike-jumper, line 370: Here, as well: space between `if` and `(`.

src/libguac/socket.c
```diff
@@ -286,56 +308,70 @@ ssize_t __guac_socket_write_base64_triplet(guac_socket* socket,
...
+    size_t remaining = count;
+    int len;
     int retval;
 
-    const unsigned char* char_buf = (const unsigned char*) buf;
-    const unsigned char* end = char_buf + count;
+    while (remaining > 0) {
+        // Fill ready buffer as much as possible
+        len = GUAC_SOCKET_BASE64_READY_BUFFER_SIZE - socket->__ready;
+        if(remaining < len) {
```
mike-jumper, line 359: `if(` -> `if (`

mike-jumper (changes requested):
LGTM - only some style nitpicks.

neandrake:
Thank you for reviewing! I made the updates and make sure to reference the style guidelines in the future.

src/libguac/socket.c
```diff
@@ -286,56 +308,70 @@ ssize_t __guac_socket_write_base64_triplet(guac_socket* socket,
...
-    return 3;
+ssize_t guac_socket_flush_base64(guac_socket* socket) {
+    const unsigned char* src = socket->__ready_buf;
 
-}
+    int encodedCount = 0;
+    int remaining = socket->__ready;
 
-ssize_t __guac_socket_write_base64_byte(guac_socket* socket, int buf) {
+    // Encode bytes in groups of three
```
mike-jumper, line 320: Sorry - didn't notice this before, but please use:

```c
/* C-style comments */
```

instead of:

```
// C++-style comments
```

Same elsewhere in these changes.

neandrake:
Updated. Thanks for clarifying the preferred landing style. I've been under the impression most git/github flows prefer the merging in most cases. For further clarification when looking through the git history, are merges reserved only for version branches, whereas for feature changes/branches rebase should be used. Is there a preference for squashing commits within a PR?

mike-jumper:
We use merge commits to represent merges of PRs to their target branch and merges of release branches (`staging/*`) back to `master`, so we avoid including merge commits within the set of commits in the PR itself.

As for the commits on a PR, we don't specifically ask that contributors squash their commits except in cases where we end up with a set of commits where only one is meaningful, like:

```
GUACAMOLE-1234: Add support for [some specific high-level feature].
Update README.md.
Update README.md.
```

In those cases, we point folks at https://www.codelord.net/2015/03/16/bad-commit-messages-hall-of-shame/ and ask that they rebase/squash the commits such that they are either meaningful or do not exist.

neandrake:
Ah okay, I actually hadn't considered the difference of merging PR vs. including merge commit in the PR. I went ahead and squashed the style correction commits.

mike-jumper (approved):

PR #435, opened 2023-05-08
GUACAMOLE-1784: Migrate usage of guac_rdp_audio_buffer_ack() to guac_client_for_user().

mike-jumper:
The function `guac_rdp_audio_buffer_ack()` is the only location in guacamole-server where a `guac_user` instance is accessed outside the context of a user-specific callback, a no-no in the libguac API. This change migrates that function such that it _is_ a callback, invoked by `guac_client_for_user()` (the function provided by libguac for exactly this purpose).

### DIFF
--- a/src/protocols/rdp/channels/audio-input/audio-buffer.c
+++ b/src/protocols/rdp/channels/audio-input/audio-buffer.c
@@ -275,6 +275,34 @@ guac_rdp_audio_buffer* guac_rdp_audio_buffer_alloc(guac_client* client) {
 }
 
 /**
+ * Parameters describing an "ack" instruction to be sent to the current user of
+ * an inbound audio stream (guac_rdp_audio_buffer).
+ */
+typedef struct guac_rdp_audio_buffer_ack_params {
+
+    /**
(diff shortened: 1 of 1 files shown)


src/protocols/rdp/channels/audio-input/audio-buffer.c
```diff
@@ -274,6 +274,34 @@ guac_rdp_audio_buffer* guac_rdp_audio_buffer_alloc(guac_client* client) {
...
+/**
+ * Parameters describing an "ack" instruction to be sent to the current user of
+ * an inbound audio stream (guac_rdp_audio_buffer).
+ */
+typedef struct guac_rdp_audio_buffer_ack_params {
+
+    /**
+     * The audio buffer associated with the guac_stream for which the "ack"
+     * instruction should be sent, if any. If there is no associated
+     * guac_stream, this function has no effect.
```
jmuehlner, line 286: Which function has no effect?

mike-jumper: Ah, I'll fix that.

jmuehlner (approved):

PR #443, opened 2023-05-18
GUACAMOLE-1754: Run unit tests within Docker build.

mike-jumper:


mike-jumper:
The PR build failure is due to libwebsockets.org not functioning as expected:

```
Cloning into 'libwebsockets'...
error: RPC failed; curl 18 HTTP/2 stream 3 was reset
fetch-pack: unexpected disconnect while reading sideband packet
fatal: early EOF
fatal: fetch-pack: invalid index-pack output
```

which should be addressed by #442.

mike-jumper:
Rebased off recently-merged #442 and all seems well again.

PR #446, opened 2023-06-07
GUACAMOLE-1804: Support mac style paste shortcut.

jmuehlner:


src/terminal/terminal.c
```diff
@@ -1576,6 +1577,8 @@ static int __guac_terminal_send_key(guac_terminal* term, int keysym, int pressed
     /* Track modifiers */
     if (keysym == 0xFFE3)
         term->mod_ctrl = pressed;
+    if (keysym == 0xFFE7)
```
jmuehlner, line 1580: This is left meta/cmd only - I'd defer adding support for right-side modifiers to https://issues.apache.org/jira/browse/GUACAMOLE-1113.

src/terminal/terminal.c
```diff
@@ -1576,6 +1577,8 @@ static int __guac_terminal_send_key(guac_terminal* term, int keysym, int pressed
     /* Track modifiers */
     if (keysym == 0xFFE3)
         term->mod_ctrl = pressed;
+    else if (keysym == 0xFFE7)
```
jmuehlner, line 1580: This is left meta/cmd only - I'd defer adding support for right-side modifiers to https://issues.apache.org/jira/browse/GUACAMOLE-1113.

src/terminal/terminal.c
```diff
@@ -1584,8 +1587,8 @@ static int __guac_terminal_send_key(guac_terminal* term, int keysym, int pressed
     /* If key pressed */
     else if (pressed) {
 
-        /* Ctrl+Shift+V shortcut for paste */
-        if (keysym == 'V' && term->mod_ctrl)
+        /* Ctrl+Shift+V or Cmd+v (mac style) shortcuts for paste */
+        if ((keysym == 'V' && term->mod_ctrl) || (keysym = 'v' && term->mod_meta))
```
mike-jumper, line 1591: This an assignment of `'v'` to `keysym`. It should instead be `keysym == 'v'`.

mike-jumper (changes requested):

src/terminal/terminal.c
jmuehlner: Oops, that would explain the weird behavior I saw when retesting this. Fixed!

mike-jumper (approved):

PR #448, opened 2023-06-08
GUACAMOLE-1113: Support right hand modifiers in terminal.

jmuehlner:


src/terminal/terminal.c
```diff
@@ -1575,13 +1575,13 @@ static int __guac_terminal_send_key(guac_terminal* term, int keysym, int pressed
...
 
     /* Track modifiers */
-    if (keysym == 0xFFE3)
+    if (keysym == 0xFFE3 || keysym == 0xFFE4)
         term->mod_ctrl = pressed;
-    else if (keysym == 0xFFE7)
+    else if (keysym == 0xFFE7 || keysym == 0xFFE8)
         term->mod_meta = pressed;
-    else if (keysym == 0xFFE9)
+    else if (keysym == 0xFFE9 || keysym == 0xFFEA)
```
jmuehlner, line 1582: There's some trickiness associated with right alt and AltGr that I don't fully understand, so I'm marking this as draft for now. This particular keysym `0xFFEA` for right alt is automatically translated to AltGr by the frontend in any case.

jmuehlner: Ok, I _think_ after looking at this a bit more, that this should be ok. We might want to un-automatically translate right alt to AltGr on the frontend, but in either case, this change should be safe here.

@mike-jumper If you could confirm that this makes sense that would be cool because I'm not quite as up on all the ins and outs of keyboard layout stuff.

mike-jumper: Yep, I agree that:

* It makes sense to handle right alt here, even though the frontend currently assumes the right alt key is always AltGr.
* It makes sense that the frontend should cease assuming the right alt key is AltGr, and instead do so only if the browser agrees that the local keyboard layout defines that key as AltGr (via the details of the key event).

mike-jumper (approved):

PR #450, opened 2023-06-28
GUACAMOLE-1290: Add support for ssh certificate Authentication

Tejaswikandula:


necouchman:
@Tejaswikandula : Your pull request and commit messages need to have the GUACAMOLE-1290: prefix.

src/common-ssh/common-ssh/user.h
```diff
@@ -44,6 +44,12 @@ typedef struct guac_common_ssh_user {
      */
     guac_common_ssh_key* private_key;
 
+    /**
+     * The public key which should be used to authenticate this user, if any,
+     * or NULL if a password or just a private key will be used instead.
+     */
+    char* public_key;
```
necouchman, line 51: Is "public key" the correct term for this? To me, a "public key" is the half of a key-pair that you share with someone in order to connect or facilitate encryption. Per the Jira issue, shouldn't this be `ssh_certificate` or something like that?

Tejaswikandula: @necouchman It is a public key but a CA signed public key. But, I kept in generic. Should I rename to ca_public_key instead?

Tejaswikandula:
> @Tejaswikandula : Your pull request and commit messages need to have the GUACAMOLE-1290: prefix.

Edited the commit message

src/common-ssh/common-ssh/user.h
necouchman: Okay, I see - I had to read the the Red Hat page linked on the Jira issue to actually understand the point of what was going on, here. I think the naming as you have it is fine.

src/common-ssh/user.c
```diff
@@ -80,3 +82,15 @@ int guac_common_ssh_user_import_key(guac_common_ssh_user* user,
 
 }
 
+int guac_common_ssh_user_import_public_key(guac_common_ssh_user* user,
+        char* public_key) {
+
+    /* Free existing public key, if present */
+    free(user->public_key);
```
necouchman, line 89: Should this perhaps use the same logic as the above function, with a check to see if the key is null or not and then a call to `guac_common_ssh_key_free()`?

Tejaswikandula: I skipped the null check because free in C is null pointer safe. I can still add a null check though, it shouldn't do any harm but thought that's not required. 
User->private_key is of type guac_common_ssh_key which includes both key and the optional passphrase so we call guac_common_ssh_key_free() to free the key whereas user->public_key is a simple char* type.

necouchman: Okay, then it's probably fine as-is.

src/common-ssh/common-ssh/user.h
```diff
@@ -104,5 +110,22 @@ void guac_common_ssh_user_set_password(guac_common_ssh_user* user,
...
 
+/**
+ * Imports the given public key, associating that key with the given user.
+ * If the public key is imported successfully, it will be used for
+ * future authentication attempts.
+ *
+ * @param user
+ *     The user to associate with the given private key.
+ *
+ * @param public_key
```
necouchman, line 121: This looks to be missing the documentation on the parameter.

src/protocols/ssh/settings.h
```diff
@@ -93,6 +93,12 @@ typedef struct guac_ssh_settings {
      */
     char* key_passphrase;
 
+    /**
+     * The public key, if any. If no public key is
+     * specified, this will be NULL.
+     */
```
necouchman, line 99: For consistency, can you please:
* Mention the base64-encoding (as above with `key_base64`
* Extended the length of the lines for the comment to the same length as the ones around it (roughly 80 characters).

necouchman (changes requested):
Overall looks pretty good - two minor things to clean up, and I think it will be ready to merge.

src/common-ssh/common-ssh/user.h
Tejaswikandula: Added missing documentation

necouchman (approved):

necouchman:
Thanks @Tejaswikandula - could you also add the documentation to apache/guacamole-manual and do a pull request for that?

adb014:
Was this ever linked to guacamole-client. I'd like to use OpenBao/Hashicorp signed ssh certificates with short TTL's automatically within guacamole and the cguacamole-client changes are needed for this

adb014:
> Was this ever linked to [guacamole-client](https://issues.apache.org/jira/browse/GUACAMOLE-client). I'd like to use OpenBao/Hashicorp signed ssh certificates with short TTL's automatically within guacamole and the cguacamole-client changes are needed for this

Forget it, I'm an idiot. You called it "Public Key" in the UI rather than signed certificate

PR #453, opened 2023-07-21
GUACAMOLE-1332: Add support for certificate fingerprints and auto-accept.

necouchman:
This pull requests adds options to support the following additional certificate handling in RDP connections:
* Trust on First Use (tofu): If a host is not "known" to the underlying FreeRDP library, the certificate will be accepted and the fingerprint stored. This is equivalent to the xfreerdp command line option `/cert:tofu`.
* Fingerprint + Hash: You can provide a comma-separated list of certificate hash and fingerprint combinations that FreeRDP will accept for the connection. It's worth noting that I actually don't know the proper way to specify these at the moment, so I've been unable to actually prove this works, but I'm getting identical behavior on the xfreerdp command line as in Guacamole. I've an e-mail out to the FreeRDP mailing list to try to get some guidance on it.

### DIFF
--- a/src/protocols/rdp/settings.c
+++ b/src/protocols/rdp/settings.c
@@ -76,6 +76,8 @@ const char* GUAC_RDP_CLIENT_ARGS[] = {
     "server-layout",
     "security",
     "ignore-cert",
+    "cert-tofu",
+    "cert-fingerprints",
     "disable-auth",
     "remote-app",
     "remote-app-dir",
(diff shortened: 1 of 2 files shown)


src/protocols/rdp/settings.c
```diff
@@ -708,6 +725,16 @@ guac_rdp_settings* guac_rdp_parse_args(guac_user* user,
...
 
+    /* Add new certificates to trust list */
+    settings->certificate_tofu =
+        guac_user_parse_args_boolean(user, GUAC_RDP_CLIENT_ARGS, argv,
+                IDX_CERTIFICATE_TOFU, 0);
+
+    /* Fingerprints of certificates that should be trusted */
+    settings->certificate_fingerprints =
+        guac_user_parse_args_string(user, GUAC_RDP_CLIENT_ARGS, argv,
+                IDX_CERTIFICATE_FINGERPRINTS, NULL);
```
mike-jumper, line 736: Does this need to be freed later in `guac_rdp_settings_free()`?

necouchman: Uh, yeah, that'd probably be a good idea...

necouchman: Fixed via rebase.

mike-jumper (approved):

PR #455, opened 2023-08-22
GUACAMOLE-1846: Synchronize new users to connections in batches to fix join race condition.

jmuehlner:
This change turned out to be pretty involved, but at their core, these changes are addressing the race condition described in [GUACAMOLE-1846](https://issues.apache.org/jira/browse/GUACAMOLE-1846). 

This change addresses this problem by batching up all newly connected users and synchronizing connection state to them using a newly defined `join_pending_handler`, at which time they'll be promoted to full users, and the broadcast socket will write to them.

This synchronization is done using a new broadcast socket that writes to all synchronized users at once, instead of once at a time.

Pending users are synchronized once every quarter second, which should result in little noticeable delay upon joining a connection.

Thanks to @mike-jumper for helping to work through the logic here and help figure out various bugs along the way.

### DIFF
--- a/src/common/common/cursor.h
+++ b/src/common/common/cursor.h
@@ -153,14 +153,14 @@ void guac_common_cursor_free(guac_common_cursor* cursor);
  * @param cursor
  *     The cursor to send.
  *
- * @param user
+ * @param client
  *     The user receiving the updated cursor.
  *
  * @param socket
  *     The socket over which the updated cursor should be sent.
  */
-void guac_common_cursor_dup(guac_common_cursor* cursor, guac_user* user,
-        guac_socket* socket);
+void guac_common_cursor_dup(
+        guac_common_cursor* cursor, guac_client* client, guac_socket* socket);
 
 /**
  * Updates the current position and button state of the mouse cursor, marking
--- a/src/common/list.c
+++ b/src/common/list.c
@@ -34,8 +34,26 @@ guac_common_list* guac_common_list_alloc() {
 
 }
 
-void guac_common_list_free(guac_common_list* list) {
+void guac_common_list_free(
+        guac_common_list* list,
+        guac_common_list_element_free_handler* free_element_handler) {
+
+    /* Free every element of the list */
+    guac_common_list_element* element = list->head;
+    while(element != NULL) {
+
+        guac_common_list_element* next = element->next;
+
+        if (free_element_handler != NULL)
+            free_element_handler(element->data);
+
+        free(element);
+        element = next;
+    }
+
+    /* Free the list itself */
     free(list);
+
 }
 
 guac_common_list_element* guac_common_list_add(guac_common_list* list,
--- a/src/guacd/daemon.c
+++ b/src/guacd/daemon.c
@@ -43,6 +43,7 @@
 #include <sys/socket.h>
 #include <sys/stat.h>
 #include <sys/types.h>
+#include <sys/wait.h>
 #include <unistd.h>
 
 #define GUACD_DEV_NULL "/dev/null"
@@ -244,6 +245,49 @@ static void guacd_openssl_free_locks(int count) {
 }
 #endif
 #endif
+
+/**
+ * A flag that, if non-zero, indicates that the daemon should immediately stop
+ * accepting new connections.
+ */
+int stop_everything = 0;
+
+/**
+ * A signal handler that will set a flag telling the daemon to immediately stop
+ * accepting new connections. Note that the signal itself will cause any pending
+ * accept() calls to be interrupted, causing the daemon to unlock and begin
+ * cleaning up.
+ *
+ * @param signal
+ *     The signal that was received. Unused in this function since only
+ *     signals that should result in stopping the daemon should invoke this.
+ */
+static void signal_stop_handler(int signal) {
+
+    /* Instruct the daemon to stop accepting new connections */
+    stop_everything = 1;
+
+}
+
+/**
+ * A callback for guacd_proc_map_foreach which will stop every process in the
+ * map.
+ *
+ * @param proc
+ *     The guacd process to stop.
+ *
+ * @param data
+ *     Unused.
+ */
+static void stop_process_callback(guacd_proc* proc, void* data) {
+
+    guacd_log(GUAC_LOG_DEBUG,
+            "Killing connection %s (%i)\n",
+            proc->client->connection_id, (int) proc->pid);
+
+    guacd_proc_stop(proc);
+
+}
 
 int main(int argc, char* argv[]) {
 
@@ -452,6 +496,12 @@ int main(int argc, char* argv[]) {
                 "Child processes may pile up in the process table.");
     }
 
+    /* Clean up and exit if SIGINT or SIGTERM signals are caught */
+    struct sigaction signal_stop_action = { 0 };
+    signal_stop_action.sa_handler = signal_stop_handler;
+    sigaction(SIGINT, &signal_stop_action, NULL);
+    sigaction(SIGTERM, &signal_stop_action, NULL);
+
     /* Log listening status */
     guacd_log(GUAC_LOG_INFO, "Listening on host %s, port %s", bound_address, bound_port);
 
@@ -465,7 +515,7 @@ int main(int argc, char* argv[]) {
     }
 
     /* Daemon loop */
-    for (;;) {
+    while (!stop_everything) {
 
         pthread_t child_thread;
 
@@ -475,7 +525,10 @@ int main(int argc, char* argv[]) {
                 (struct sockaddr*) &client_addr, &client_addr_len);
 
         if (connected_socket_fd < 0) {
-            guacd_log(GUAC_LOG_ERROR, "Could not accept client connection: %s", strerror(errno));
+            if (errno == EINTR)
+                guacd_log(GUAC_LOG_DEBUG, "Accepting of further client connection(s) interrupted by signal.");
+            else
+                guacd_log(GUAC_LOG_ERROR, "Could not accept client connection: %s", strerror(errno));
             continue;
         }
 
@@ -496,6 +549,26 @@ int main(int argc, char* argv[]) {
         /* Spawn thread to handle connection */
         pthread_create(&child_thread, NULL, guacd_connection_thread, params);
         pthread_detach(child_thread);
+
+    }
+
+    /* Stop all connections */
+    if (map != NULL) {
+
+        guacd_proc_map_foreach(map, stop_process_callback, NULL);
+
+        /*
+         * FIXME: Clean up the proc map. This is not as straightforward as it
+         * might seem, since the detached connection threads will attempt to
+         * remove the connection proccesses from the map when they complete,
+         * which will also happen upon shutdown. So there's a good chance that
+         * this map cleanup will happen at the same time as the thread cleanup.
+         * The map _does_ have locking mechanisms in place for ensuring thread
+         * safety, but cleaning up the map also requires destroying those locks,
+         * making them unusable for this case. One potential fix could be to
+         * join every one of the connection threads instead of detaching them,
+         * but that does complicate the cleanup of thread resources.
+         */
 
     }
 
--- a/src/guacd/proc.c
+++ b/src/guacd/proc.c
@@ -288,6 +288,26 @@ static int guacd_timed_client_free(guac_client* client, int timeout) {
 }
 
 /**
+ * A reference to the current guacd process.
+ */
+guacd_proc* guacd_proc_self = NULL;
+
+/**
+ * A signal handler that will be invoked when a signal is caught telling this
+ * guacd process to immediately exit.
+ *
+ * @param signal
+ *     The signal that was received. Unused in this function since only
+ *     signals that should result in stopping the proc should invoke this.
+ */
+static void signal_stop_handler(int signal) {
+
+    /* Stop the current guacd proc */
+    guacd_proc_stop(guacd_proc_self);
+
+}
+
+/**
  * Starts protocol-specific handling on the given process by loading the client
  * plugin for that protocol. This function does NOT return. It initializes the
  * process with protocol-specific handlers and then runs until the guacd_proc's
@@ -332,6 +352,14 @@ static void guacd_exec_proc(guacd_proc* proc, const char* protocol) {
 
     /* Enable keep alive on the broadcast socket */
     guac_socket_require_keep_alive(client->socket);
+
+    guacd_proc_self = proc;
+
+    /* Clean up and exit if SIGINT or SIGTERM signals are caught */
+    struct sigaction signal_stop_action = { 0 };
+    signal_stop_action.sa_handler = signal_stop_handler;
+    sigaction(SIGINT, &signal_stop_action, NULL);
+    sigaction(SIGTERM, &signal_stop_action, NULL);
 
     /* Add each received file descriptor as a new user */
     int received_fd;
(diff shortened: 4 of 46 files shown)


neandrake:
Oh cool, this issue/fix sounds like it might be related to what I was seeing in https://github.com/apache/guacamole-server/pull/316#issuecomment-1483939885

src/libguac/guacamole/client-fntypes.h
```diff
@@ -79,5 +91,22 @@ typedef void guac_client_log_handler(guac_client* client,
...
+ *     The guac_client associated with the users to broadcast to.
+ *
+ * @param callback
+ *     A callback that should be invoked with each broadcasted user.
+ *
+ * @param data
+ *     Arbitrary data that may be used to broadcast to the subset of users.
+ */
+typedef void guac_client_broadcast_handler(
+        guac_client* client, guac_user_callback* callback, void* data);
```
mike-jumper, line 109: This particular function type ...

* ... is specific to `guac_socket`, and should instead be named `guac_socket_broadcast_handler` within files specific to `guac_socket`
* ... is internal to the broadcast socket implementation and need not be exposed publicly. It should instead be declared strictly internally and omitted from the public API.

src/libguac/guacamole/client.h
```diff
@@ -162,14 +173,65 @@ struct guac_client {
...
+     * first user joins the connection.
+     */
+    struct sigevent __pending_users_timer_signal_config;
+
+    /**
+     * Specification for the timing of the pending users timer. Only for
+     * internal use within the client. This will be NULL until the
+     * first user joins the connection.
+     */
+    struct itimerspec __pending_users_time_spec;
```
mike-jumper, line 221: Do these need to be included within the `guac_client` structure? I see them used only during initialization.

src/libguac/guacamole/client.h
```diff
@@ -162,14 +173,65 @@ struct guac_client {
...
+
+    /**
+     * A flag that indicates whether the pending users timer event thread is
+     * currently running.
+     */
+    volatile atomic_flag __pending_timer_event_active;
```
mike-jumper, line 227: Which standard defines `atomic_flag`? The guacamole-server build currently targets C99: https://github.com/apache/guacamole-server/blob/80598ae857b4ccf696b5adff8c48a9ecbfe4a647/configure.ac#L37

src/libguac/guacamole/client.h
```diff
@@ -162,14 +173,65 @@ struct guac_client {
...
+    /**
+     * Specification for the timing of the pending users timer. Only for
+     * internal use within the client. This will be NULL until the
+     * first user joins the connection.
+     */
+    struct itimerspec __pending_users_time_spec;
+
+    /**
+     * A flag that indicates whether the pending users timer event thread is
+     * currently running.
```
mike-jumper, line 225: * We don't necessarily know that the timer is implemented with threads.
* This indicates whether the timer event callback is currently running, not whether the timer itself is running, yes? The documentation for `__pending_users_timer_running` uses the same terminology to refer to whether the timer has been registered.

src/libguac/guacamole/client.h
mike-jumper, line 227: Would it make sense to combine this with `__pending_users_timer_running` such that there is only one flag+mutex indicating the current state of the timer (unregistered, registered, triggered)?

src/libguac/guacamole/client-fntypes.h
```diff
@@ -48,6 +50,16 @@
...
 
+/**
+ * Handler that will run before immediately before pending users are promoted
+ * to full users. The pending user socket should be used to communicate to the
+ * pending users.
+ *
+ * @param client
+ *     The client whose handler was invoked.
+ */
+typedef void guac_join_pending_handler(guac_client* client);
```
mike-jumper, line 61: Should this function have error-handling semantics, as well, in line with things like the join handler?

mike-jumper, line 61: This should be `guac_client_join_pending_handler`.

mike-jumper (changes requested):

src/libguac/local-lock.c
```diff
@@ -0,0 +1,218 @@
...
+     */
+    if (key_value == GUAC_LOCAL_LOCK_READ_LOCK)
+        pthread_rwlock_unlock(&(local_lock->lock));
+
+    /* Acquire the write lock */
+    pthread_rwlock_wrlock(&(local_lock->lock));
+
+    /* Mark that the current thread has the lock, and increment the count */
+    pthread_setspecific(local_lock->key, get_value_from_flag_and_count(
+            GUAC_LOCAL_LOCK_WRITE_LOCK, count + 1));
```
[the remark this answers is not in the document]
jmuehlner: Well, it'll roll over back to 0, which might rely in the lock being acquired twice (resulting in undefined behavior), or the lock getting released early, resulting in potential simultaneous access to pieces of code that should only have one thread working with them at a time.

Are you worried enough about this that you want to add checks in the code for this condition?

src/libguac/local-lock.h
```diff
@@ -0,0 +1,118 @@
...
+ */
+
+/**
+ * A structure packaging together a pthread rwlock along with a key to a
+ * thread-local property to keep track of the current status of the lock,
+ * allowing the functions defined in this header to provide reentrant behavior.
+ * Note that both the lock and key must be initialized before being provided
+ * to any of these functions.
+ */
+typedef struct guac_local_lock {
```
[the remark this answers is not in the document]
jmuehlner: Yeah, the name definitely needs cleanup. Will do.

src/libguac/guacamole/client.h
jmuehlner: Probably not? When constructing the timer, you pass pointers to these various configuration-time things. The API docs don't say or even imply whether you need to keep them around afterwards, though it seems unlikely that any implementation would actually expect them to stick around for the whole timer lifetime.

I can just allocate them on the stack and not worry about it if it works correctly in my dev environment, I guess.

src/libguac/guacamole/client-fntypes.h
jmuehlner: Yeah, I guess I can just refuse to promote any users in a batch if the handler fails.

src/libguac/guacamole/client.h
jmuehlner: Oops, yeah this is a C11 thing. I'll get rid of it.

src/terminal/terminal/terminal.h
```diff
@@ -632,6 +635,22 @@ int guac_terminal_sendf(guac_terminal* term, const char* format, ...);
 void guac_terminal_dup(guac_terminal* term, guac_user* user,
         guac_socket* socket);
 
+/**
+ * Synchronizes the current display state for all users assocaited with a
+ * provided socket and client.
```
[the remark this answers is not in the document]
jmuehlner: Ah yes, that's a lot clearer.

src/libguac/local-lock.c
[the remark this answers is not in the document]
mike-jumper: Moderately, yes. It seems unlikely that the counter would ever reach this magnitude, but I think at least that we should have defined behavior for that case.

The `pthread_mutex_lock` function's behavior for this case is to return `EAGAIN`:

> <dl>
>     <dt><code>EAGAIN</code></dt>
>     <dd>The mutex could not be acquired because the maximum number of recursive locks for mutex has been exceeded.</dd>
> </dl>

We could do similar (provide some sort of error handling) and leave it up to the caller to deal with the error condition.

src/libguac/guacamole/client.h
jmuehlner: Yes it would.

src/common/common/cursor.h
```diff
@@ -150,6 +150,10 @@ void guac_common_cursor_free(guac_common_cursor* cursor);
  * the current cursor image. The resulting cursor on the remote display will
  * be visible.
  *
+ * @deprecated guac_common_cursor_dup_batch should be used instead of this
+ * user-specific method. Unlike this function, it supports sending the cursor
+ * to multiple users with a broadcast socket.
```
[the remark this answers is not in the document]
jmuehlner: Ok, those should all be removed now.

src/libguac/local-lock.c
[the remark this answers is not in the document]
jmuehlner: Ok, I added some error codes. Are you ok with not checking for the errors in the calling libguac code? We're not checking to see whether acquisition attempts of normal lock types work currently, and I happen to know that the maximum number of levels of reentrancy that we'll currently reach is `2,` which is significantly less than `2^60`.

src/libguac/guacamole/client-fntypes.h
jmuehlner: Ok, I've added it. Currently there aren't any handlers that use it, but it's there.

jmuehlner:
Marking as draft while I investigate an issue.

src/guacd/daemon.c
```diff
@@ -245,6 +246,49 @@ static void guacd_openssl_free_locks(int count) {
...
+ * A signal handler that will set a flag telling the daemon to immediately stop
+ * accepting new connections. Note that the signal itself will cause any pending
+ * accept() calls to be interrupted, causing the daemon to unlock and begin
+ * cleaning up.
+ *
+ * @param signal
+ *     The signal that was received. Unused in this function since only
+ *     signals that should result in stopping the daemon should invoke this.
+ */
+static void signal_stop_handler(int signal) {
```
jmuehlner, line 265: These various signal handling mechanisms are to ensure that the daemon and associated child processes will definitely stop, even if one of the children is locked up due to deadlock or similar.

jmuehlner:
Ok, I think this should be ready to go again. I'm still seeing the client occasionally get locked up when shutting down, but at least it's guaranteed to always shut down cleanly with the new shutdown handling introduced in https://github.com/apache/guacamole-server/pull/455/commits/eaca340a94c9aaeb235a6e30fa660ffded43cfef.

src/guacd/daemon.c
```diff
@@ -504,6 +555,26 @@ int main(int argc, char* argv[]) {
 
     }
 
+    /* Stop all connections */
+    if (map != NULL) {
+
+        guacd_proc_map_foreach(map, stop_process_callback, NULL);
+
+        /*
+         * FIXME: Clean up the proc map. This is not as straightforward as it
```
jmuehlner, line 564: The daemon was never really set up to perform cleanup, and it's less easy to get everything cleaned up than I had hoped. In the service of avoiding further complexity ballooning, I'm going to leave this for future improvement.

The map was never cleaned up before either, so at least this isn't making things worse than it used to be.

src/libguac/local-lock.c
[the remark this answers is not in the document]
mike-jumper: I think it's OK to leave as-is, unless there is any locking within a loop ... though 1,152,921,504,606,846,976 would be a _lot_ of locks.

src/common/list.c
```diff
@@ -23,6 +23,8 @@
 #include <stdlib.h>
 #include <pthread.h>
 
+#include <stdio.h>
```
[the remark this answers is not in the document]
jmuehlner: Indeed not. Removed.

jmuehlner:
This was an issue with lock acquisition/release order and has been fixed. 

> I'm still seeing the client occasionally get locked up when shutting down and fail to shut down cleanly

src/libguac/client.c
```diff
@@ -169,35 +290,56 @@ guac_client* guac_client_alloc() {
...
     client->socket = guac_socket_broadcast(client);
+    client->pending_socket = guac_socket_broadcast_pending(client);
 
     return client;
 
 }
 
 void guac_client_free(guac_client* client) {
 
+    /* Acquire write locks before referencing user pointers */
```
jmuehlner, line 317: Added this extra locking layer here because it's not really valid to reference the `__users` or `__pending_users` fields without holding these locks. 

I had thought about using `guac_client_remove_user` as a callback for `guac_client_foreach_user` and `guac_client_foreach_pending_user`, which is a more standard way of doing things, but that would also mean that'd be acquiring the read lock first, and upgrading to the write lock inside `guac_client_remove_user()`, which should technically work but should probably be avoided.

src/libguac/local-lock.c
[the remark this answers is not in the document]
jmuehlner: Yeah, this is probably the least of your problems if you find yourself acquiring that many locks.

src/terminal/terminal/scrollbar.h
```diff
@@ -260,6 +260,10 @@ void guac_terminal_scrollbar_flush(guac_terminal_scrollbar* scrollbar);
  * completely recreate and redraw the scrollbar rendering over the given
  * socket.
  *
+ * @deprecated guac_terminal_scrollbar_dup_batch() allows synchronizing
+ * multiple joining users at once using a broadcast socket, and should be
+ * used instead of this function
```
mike-jumper, line 265: Is there any need to maintain and deprecate this variant? Checking `Makefile.am`, this header looks to be part of the private API.

src/terminal/terminal/display.h
```diff
@@ -340,6 +340,25 @@ void guac_terminal_display_flush(guac_terminal_display* display);
...
+ *     socket.
+ *
+ * @param client
+ *     The client whose users are joining.
+ *
+ * @param socket
+ *     The socket over which any necessary instructions should be sent.
+ */
+void guac_terminal_display_dup_batch(
+        guac_terminal_display* display, guac_client* client, guac_socket* socket);
```
mike-jumper, line 360: Same here - I think there's no need to maintain the old version of this function, unless it has some utility internally within `guac_terminal`. It's not part of the public API.

src/libguac/client.c
```diff
@@ -277,36 +435,135 @@ void guac_client_abort(guac_client* client, guac_protocol_status status,
...
-
-        if (client->__users != NULL)
-            client->__users->__prev = user;
-
-        client->__users = user;
-        client->connected_users++;
+        /*
+        * Add the user to the list of pending users, to have their connection
+        * state synchronized asynchronously.
+        */
```
mike-jumper, line 558: The `*` characters within this comment appear misaligned.

mike-jumper (changes requested):

src/libguac/guacamole/client-fntypes.h
```diff
@@ -30,7 +30,9 @@
 #include "client-types.h"
 #include "object-types.h"
 #include "protocol-types.h"
+#include "socket.h"
 #include "stream-types.h"
+#include "user-fntypes.h"
```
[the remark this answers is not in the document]
jmuehlner: Oops, sorry these are vestigal. I'll remove.

src/libguac/guacamole/client.h
```diff
@@ -40,22 +41,32 @@
 
 #include <cairo/cairo.h>
 
-#include <pthread.h>
+#include <signal.h>
 #include <stdarg.h>
+#include <stdatomic.h>
```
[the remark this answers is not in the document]
jmuehlner: Nope - will remove.

src/common/list.c
```diff
@@ -34,8 +34,32 @@ guac_common_list* guac_common_list_alloc() {
...
+
+        if (free_element_handler != NULL)
+            free_element_handler(element->data);
+
+        free(element);
+        element = next;
+    }
+
+    /* Destroy the mutex controlling access to the list */
+    pthread_mutex_unlock(&(list->_lock));
```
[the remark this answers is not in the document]
jmuehlner: Ok, will remove.

src/terminal/terminal/display.h
jmuehlner: Oops, yeah I'll get rid of this. I had mistakenly thought it was part of the public API.

src/protocols/kubernetes/argv.h
```diff
@@ -70,5 +71,23 @@ guac_argv_callback guac_kubernetes_argv_callback;
...
+ *
+ * @param client
+ *     The client associated with the users who should receive the values of
+ *     all non-sensitive parameters which may be set while the connection is
+ *     running.
+ *
+ * @return
+ *     Always NULL.
+ */
+void* guac_kubernetes_send_current_argv_batch(
```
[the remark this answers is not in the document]
mike-jumper: Ah, OK.

src/terminal/terminal/scrollbar.h
jmuehlner: Nope. I got rid of it.

src/guacd/daemon.c
```diff
@@ -457,6 +501,10 @@ int main(int argc, char* argv[]) {
                 "Child processes may pile up in the process table.");
     }
 
+    /* Clean up and exist if SIGINT or SIGTERM signals are caught */
+    signal(SIGINT, signal_stop_handler);
+    signal(SIGTERM, signal_stop_handler);
```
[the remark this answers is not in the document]
jmuehlner: Yes, for sure. I'll switch to `sigaction`.

src/guacd/proc.c
```diff
@@ -333,6 +354,10 @@ static void guacd_exec_proc(guacd_proc* proc, const char* protocol) {
     /* Enable keep alive on the broadcast socket */
     guac_socket_require_keep_alive(client->socket);
 
+    guacd_proc_self = proc;
+    signal(SIGINT, signal_stop_handler);
+    signal(SIGTERM, signal_stop_handler);
```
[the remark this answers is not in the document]
jmuehlner: Done.

mike-jumper (approved):

mike-jumper:
@jmuehlner, as a bug fix should this be against `staging/1.5.4`?

jmuehlner:
> @jmuehlner, as a bug fix should this be against `staging/1.5.4`?

Uh, yeah, I suppose so. Let me retarget it.

mike-jumper (approved):

PR #456, opened 2023-08-29
GUACAMOLE-1850: Add locking around cursor to ensure cairo display is not used while being recreated.

jmuehlner:
I am seeing the issue described in [GUACAMOLE-1850](https://issues.apache.org/jira/browse/GUACAMOLE-1850) more often when testing [this change](https://github.com/apache/guacamole-server/pull/455).

Adding a mutex to control concurrent access to the surface underlying the socket appears to resolve this issue.

### DIFF
--- a/src/common/common/cursor.h
+++ b/src/common/common/cursor.h
@@ -123,6 +123,14 @@ typedef struct guac_common_cursor {
      */
     guac_timestamp timestamp;
 
+    /**
+     * Lock which restricts simultaneous access to the cursor, guaranteeing
+     * ordered modifications to the cursor and that incompatible operations
+     * do not occur simultaneously. This lock is for internal use within the
+     * cursor only.
+     */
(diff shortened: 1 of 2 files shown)


src/common/common/cursor.h
```diff
@@ -123,6 +123,14 @@ typedef struct guac_common_cursor {
      */
     guac_timestamp timestamp;
 
+    /**
+     * Lock which restricts simultaneous access to the cursor, guaranteeing
+     * ordered modifications to the cursor and that incompatible operations
+     * do not occur simultaneously. This lock is for internal use within the
+     * cursor only.
+     */
+    pthread_mutex_t lock;
```
neandrake, line 132: Is it customary for private/internal fields to be named with an underscore or double-underscore?
```suggestion
    pthread_mutex_t __lock;
```

mike-jumper (approved):

PR #457, opened 2023-08-29
GUACAMOLE-1846: Ensure successful builds on older GCC.

jmuehlner:
This fixes some build issues introduced in https://github.com/apache/guacamole-server/pull/455.


mike-jumper (changes requested):
Please rebase against `staging/1.5.4`

If the reentrant rwlock implementation will become part of the public API within libguac, I recommend:

1. Renaming the header to `rwlock.h`
2. Renaming the structure to `guac_rwlock` and related functions to `guac_rwlock_*`
3. Mirroring the naming and style of pthreads such that it's largely a drop-in replacement (and such that we have extra confirmation that our API aligns with well-established use).

jmuehlner:
> If the reentrant rwlock implementation will become part of the public API within libguac, I recommend

Good suggestions. All done now.

src/libguac/guacamole/client.h
```diff
@@ -30,7 +30,7 @@
 #include "client-types.h"
 #include "client-constants.h"
 #include "layer-types.h"
-#include "reentrant-rwlock.h"
+#include "guacamole/rwlock.h"
```
mike-jumper, line 33: This should just be:

```c
#include "rwlock.h"
```

within files that are already within the `guacamole/` include path.

src/libguac/guacamole/rwlock.h
```diff
@@ -42,15 +42,15 @@
...
 
 /**
  * The lock cannot be acquired because the lock has been already been
  * reentrantly acquired too many times, exhausting the capacity of this library
- * to track this lock. The lock must be released using guac_release_lock()
+ * to track this lock. The lock must be released using guac_rwlock_release_lock()
  * before it can be reacquired.
  */
-#define GUAC_REEANTRANT_LOCK_ERROR_TOO_MANY 2
+#define GUAC_RWLOCK_ERROR_TOO_MANY 2
```
mike-jumper, line 53: Rather than define new error codes and constants, I think it would be best to:

1. Use a non-zero return generically to represent that the operation failed.
2. Use `guac_error` (the libguac analogy to `errno`), `guac_error_message`, and the established `guac_status` constants (like `GUAC_STATUS_INVALID_ARGUMENT` and `GUAC_STATUS_TOO_MANY`) to make error details available.

src/libguac/guacamole/rwlock.h
```diff
@@ -82,15 +82,15 @@ typedef struct guac_reentrant_rwlock {
...
+void guac_init_rwlock(guac_rwlock* lock);
 
 /**
  * Clean up and destroy the provided guac reentrant rwlock.
  *
  * @param lock
  *     The guac reentrant rwlock to be destroyed.
  */
-void guac_destroy_reentrant_rwlock(guac_reentrant_rwlock* lock);
+void guac_destroy_rwlock(guac_rwlock* lock);
```
mike-jumper, line 93: To match convention, these should be `guac_rwlock_init()` and `guac_rwlock_destroy()` respectively.

mike-jumper (changes requested):

mike-jumper (approved):

PR #459, opened 2023-09-12
GUACAMOLE-1846: Skip user promotion entirely if there are no pending users.

jmuehlner:


jmuehlner:
This addresses an issue where unneeded display flushes would occur even when no users are being promoted.

src/libguac/client.c
```diff
@@ -185,6 +185,12 @@ static void guac_client_promote_pending_users(union sigval data) {
     /* Acquire the lock for reading and modifying the list of pending users */
     guac_rwlock_acquire_write_lock(&(client->__pending_users_lock));
 
+    /* Skip user promotion entirely if there's no pending users */
+    if (client->__pending_users == NULL) {
+        guac_rwlock_release_lock(&(client->__pending_users_lock));
+        goto promotion_complete;
```
mike-jumper, line 191: To be more DRY, should the same be done for:

https://github.com/apache/guacamole-server/blob/974816a5caae45f5ecde5b4f2b258b9a513f2c06/src/libguac/client.c#L200-L205

src/libguac/client.c
```diff
@@ -185,6 +185,12 @@ static void guac_client_promote_pending_users(union sigval data) {
     /* Acquire the lock for reading and modifying the list of pending users */
     guac_rwlock_acquire_write_lock(&(client->__pending_users_lock));
 
+    /* Skip user promotion entirely if there's no pending users */
+    if (client->__pending_users == NULL) {
+        guac_rwlock_release_lock(&(client->__pending_users_lock));
```
mike-jumper, line 190: Any reason to not move the `promotion_complete` label up a line and rely on the `guac_rwlock_release_lock()` call that already exists there?

mike-jumper (changes requested):
As with the other GUACAMOLE-1846 changes, this should probably also be against `staging/1.5.4`.

necouchman:
@jmuehlner It looks like #457 went into `staging/1.5.4`, so my guess is this should be based against that?

mike-jumper:
Wow ... reviewed at nearly _exactly_ the same time.

src/libguac/client.c
jmuehlner: Good call - it's done.

src/libguac/client.c
jmuehlner: Nope!

mike-jumper (approved):

PR #460, opened 2023-09-29
GUACAMOLE-1846: Check for NULLs before attempting to sync state in join pending handlers.

jmuehlner:
The user promotion thread is started before the client initialization is fully complete, so we should not expect that all client resources will be available when the pending handlers run. 

If the resources _are_ uninitialized, the pending handlers can still run successfully because there isn't any state to sync anyway.


necouchman (changes requested):
Looks good, except it should be re-based against `staging/1.5.4`, at least according to the Jira issue.

jmuehlner:
> Looks good, except it should be re-based against `staging/1.5.4`, at least according to the Jira issue.

oops

necouchman (approved):

PR #462, opened 2023-10-20
GUACAMOLE-1867: Add libguac convenience functions for memory allocation and management.

mike-jumper:
This change adds several convenience functions for management of memory, each analogous to standard functions but with the added benefits of checked arithmetic and automatic `NULL`-ing of freed pointers:

 Standard function | libguac function
------------------ | ----------------
`malloc()`         | `guac_mem_alloc()`
`calloc()`         | `guac_mem_zalloc()`
`realloc()`        | `guac_mem_realloc()`
`free()`           | `guac_mem_free()`

These new functions allow us to take problematic code like:

```c
void* foo = malloc(a * b * c);
```

and instead do:

```c
void* foo = guac_mem_alloc(a, b, c);
```

Or take code like:

```c
free(foo);
/* foo is dangling here */
```

and instead do:

```c
guac_mem_free(foo);
/* foo is NOT dangling here */
```

Additional functions are provided for performing arithmetic prior to allocating memory, as is often necessary in more complex cases. These functions are based off the C23 standard draft that includes new macros for checked arithmetic:

* `guac_mem_ckd_mul()` - Multiplies a series of `size_t` integers together, storing the result in a provided `size_t` variable and returning whether overflow occurred.
* `guac_mem_ckd_add()` - Adds a series of `size_t` integers together, storing the result in a provided `size_t` variable and returning whether overflow occurred.
* `guac_mem_ckd_sub()` - Subtracts a series of `size_t` integers from each other, storing the result in a provided `size_t` variable and returning whether overflow occurred (NOTE: this is necessary vs. simply adding negative values as `size_t` is an unsigned type).

As it's often inconvenient to repeatedly store intermediate results, `*_or_die()` variants of the above are also provided. Each of these is identical in behavior to the original, except that they will `abort()` the entire process in the event of overflow:

* `guac_mem_ckd_mul_or_die()`
* `guac_mem_ckd_add_or_die()`
* `guac_mem_ckd_sub_or_die()`

There will definitely be conflicts against `master` after merge, but I think these are pretty nifty and a step in the right direction in terms of cleanliness and inherent safeguards / hardening. While the set of changes itself is large, the changes themselves are mostly documentation, unit tests, and trivial replacements.

(Opening as a draft for now while doing at least a quick smoke test of each protocol...)


src/libguac/mem.c
```diff
@@ -0,0 +1,237 @@
...
+        return NULL;
+    }
+
+    /* Resize to 0 is equivalent to free() */
+    if (size == 0) {
+        guac_mem_free(mem);
+        return NULL;
+    }
+
+    void* resized_mem = realloc(mem, size);
```
aleitner, line 222: Should we free the mem if the realloc fails?

src/guacenc/ffmpeg-compat.c
```diff
@@ -106,10 +107,10 @@ int guacenc_avcodec_encode_video(guacenc_video* video, AVFrame* frame) {
     AVCodecContext* context = video->context;
 
     /* Calculate appropriate buffer size */
-    int length = FF_MIN_BUFFER_SIZE + 12 * context->width * context->height;
```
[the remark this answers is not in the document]
mike-jumper: Sure.

src/libguac/mem.c
mike-jumper: No, it's on the caller to decide whether to keep the original memory if it's not possible to reallocate. This is necessary to ensure the caller can still decide to take some other action in the case of allocation failure. We can't deviate from that without breaking compatibility with the core expectations of a call to `realloc()`.

I think a `guac_mem_realloc_or_die()` could make sense, though.

aleitner: that's a solid point and I like the idea of `guac_mem_realloc_or_die` to go with the rest of the `or_die` 💀 functions

src/guacenc/ffmpeg-compat.c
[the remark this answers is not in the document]
mike-jumper: Fixed via rebase.

src/libguac/mem.c
mike-jumper: Added via rebase, along with unit tests for both variants.

src/libguac/Makefile.am
```diff
@@ -57,6 +58,7 @@ libguacinc_HEADERS =                  \
     guacamole/plugin.h                \
     guacamole/pool.h                  \
     guacamole/pool-types.h            \
+    guacamole/private/mem.h           \
```
jmuehlner, line 61: Should this be exposed publicly?

src/libguac/guacamole/private/mem.h
```diff
@@ -0,0 +1,317 @@
...
+ *     limits of a size_t, non-zero otherwise.
+ */
+int PRIV_guac_mem_ckd_mul(size_t* result, size_t factor_count, const size_t* factors);
+
+/**
+ * Adds together each of the given values, storing the result in a size_t
+ * variable via the provided pointer. If the result of the addition overflows
+ * the limits of a size_t, non-zero is returned to signal failure.
+ *
+ * If the addition operation fails, the nature of any result stored the
```
jmuehlner, line 126: Looks like a missing word here: "...of any result stored the provided pointer..."

src/libguac/mem.c
```diff
@@ -0,0 +1,252 @@
...
+ *
+ * It is further OK for guac_mem_free() to be incompatible with free() and only
+ * usable on memory blocks allocated through guac_mem_alloc() and similar.
+ *
+ * ============================================================================
+ */
+
+int PRIV_guac_mem_ckd_mul(size_t* result, size_t factor_count, const size_t* factors) {
+
+    /* Consider calculation invalid if no factors are provided at all */
```
jmuehlner, line 49: Shouldn't this be returning a non-zero value if this is an invalid calculation?

src/libguac/mem.c
```diff
@@ -0,0 +1,252 @@
...
+    ssize_t size = *(factors++);
+    while (--factor_count && size) {
+
+        size_t factor = *(factors++);
+
+        /* Fail with GUAC_STATUS_NO_MEMORY if including this additional factor
+         * would exceed SIZE_MAX */
+        size_t max_factor = SIZE_MAX / size;
+        if (factor > max_factor)
+            return 1;
```
jmuehlner, line 63: Why not just use the `GUAC_STATUS_NO_MEMORY` constant here?

jmuehlner: Or am I misunderstanding what "Fail with GUAC_STATUS_NO_MEMORY" means?

src/libguac/tests/mem/ckd_add.c
```diff
@@ -0,0 +1,103 @@
...
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+#include <CUnit/CUnit.h>
+#include <guacamole/mem.h>
+#include <stdint.h>
+
+/**
+ * Test which verifies that guac_mem_ckd_add() calculates zero values for all
```
jmuehlner, line 25: This comment looks wrong. Add will definitely sometimes return non-zero values even if one argument is zero.

src/libguac/tests/mem/ckd_sub.c
```diff
@@ -0,0 +1,105 @@
...
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+#include <CUnit/CUnit.h>
+#include <guacamole/mem.h>
+#include <stdint.h>
+
+/**
+ * Test which verifies that guac_mem_ckd_sub() calculates zero values for all
```
jmuehlner, line 25: Again, this test is not expecting zero values for all inputs, nor should it.

src/libguac/tests/mem/ckd_sub.c
```diff
@@ -0,0 +1,105 @@
...
+ * overflow beyond zero.
+ */
+void test_mem__ckd_sub_small() {
+
+    size_t result = SIZE_MAX;
+
+    CU_ASSERT_FALSE_FATAL(guac_mem_ckd_sub(&result, 123));
+    CU_ASSERT_EQUAL(result, 123);
+
+    CU_ASSERT_FALSE_FATAL(guac_mem_ckd_sub(&result, 456 - 123));
```
jmuehlner, line 73: Shouldn't `456 - 123` be passed as separate arguments?

src/libguac/tests/mem/realloc_or_die.c
```diff
@@ -0,0 +1,98 @@
...
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+#include <CUnit/CUnit.h>
+#include <guacamole/mem.h>
+#include <stdint.h>
+
+/**
+ * Test which verifies that guac_mem_realloc_or_die() returns NULL for all
```
jmuehlner, line 25: I don't suppose theres's any way to test that the "or die" part works successfully?

src/libguac/tests/mem/ckd_sub.c
mike-jumper: Yes, good catch.

src/libguac/mem.c
mike-jumper: This comment is copypasta. `GUAC_STATUS_NO_MEMORY` within these changes is only relevant to some of the allocation functions, and then only as a value that may be assigned to `guac_error`.

src/libguac/tests/mem/realloc_or_die.c
mike-jumper: Hm ... I suppose we could try `fork()`ing and verifying that the process exits as expected. I'll give it a shot.

mike-jumper: Yep - that seems to work well.

Looks like I forgot to add `*_or_die()` tests for the other arithmetic operations. I'll do that, too.

src/libguac/mem.c
mike-jumper: Yes, that it should.

src/libguac/guacamole/private/mem.h
mike-jumper: Fixed via rebase.

src/libguac/mem.c
mike-jumper: Fixed via rebase.

src/libguac/mem.c
mike-jumper: Fixed via rebase.

src/libguac/tests/mem/ckd_sub.c
mike-jumper: Fixed via rebase.

src/libguac/tests/mem/ckd_sub.c
mike-jumper: Fixed via rebase.

src/libguac/tests/mem/ckd_add.c
mike-jumper: Fixed via rebase.

src/libguac/tests/mem/realloc_or_die.c
mike-jumper: Test for the `*_or_die()` variants of other arithmetic ops has now been added via rebase.

mike-jumper:
Did some basic tests confirming VNC/RDP/SSH, and all seems OK.

jmuehlner:
Everything LGTM. @aleitner any remaining concerns?

aleitner:
LGTM also!

PR #467, opened 2023-10-29
GUACAMOLE-1658: Remove reference to non-existing ssh_key header.

necouchman:
Removes the reference to the non-existent `ssh_key.h` file, which breaks when you try to build with SSH agent support (`--enable-ssh-agent`).

Submitting this is a bug fix to 1.5.4.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -1067,11 +1067,6 @@ AC_ARG_WITH([ssh],
             [],
             [with_ssh=check])
 
-AC_ARG_ENABLE(ssh_agent,
-              [AS_HELP_STRING([--enable-ssh-agent],
-                              [enable built-in ssh-agent])
-              ],enable_ssh_agent=yes)
-
 if test "x$with_ssh" != "xno"
 then
     have_libssh2=yes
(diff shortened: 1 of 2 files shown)


necouchman:
Converting to a draft for the moment - this needs a bit more work to fix the other issues that are present because of it.

jmuehlner:
Are there still issues with this PR? It looks pretty straightforward, and the issue appears to be still valid as well.

necouchman:
@jmuehlner Yeah, unfortunately this is going to require re-implementing the overall SSH agent/key support that got removed. There are notes in the Jira issue - https://issues.apache.org/jira/browse/GUACAMOLE-1658.

Yugnk:
I can't ssh with priv key. When I put the priv key in Guacamole. It doesn't connect. It gives the error "Log in failed. Please reconnect and try again."

necouchman:
@Yugnk This is not the place to ask that question - this is a (draft) pull request, not a mailing list or support forum.

DenisVerkhovsky:
Hi, 
Is there any update regarding the fix?

PR #470, opened 2023-12-28
GUACAMOLE-1686: Make WOL functionality capable of testing connectivity to the server.

necouchman:
This PR adds to the Wake-on-LAN functionality, implementing a new function that checks the connection to the destination server both before sending the WOL packet and, again, after sending the packet. If the host responds before sending the WOL packet, there is no need to actually send the WOL packet. If the packet does get sent, then the function will attempt to connect, then wait for the specified time period and then re-try the connection.

As a by-product of this implementation, I've also added a couple of other things:
* Centralized the TCP socket connection functionality into a common function, and updated the protocols that use it (SSH and Telnet, at the moment) to use that function. I'm not sure I got this in the right place, so if there are opinions about where to move it, I'm happy to adjust.
* Implemented a shared function to convert an integer to a string.

### DIFF
--- a/src/libguac/string.c
+++ b/src/libguac/string.c
@@ -22,6 +22,7 @@
 #include "guacamole/mem.h"
 
 #include <stddef.h>
+#include <stdio.h>
 #include <string.h>
 
 /**
@@ -43,6 +44,20 @@
  *     size of the buffer, zero will be returned.
  */
 #define REMAINING(n, length) (((n) < (length)) ? 0 : ((n) - (length)))
+
+int guac_itoa(char* restrict dest, unsigned int integer) {
+
+    /* Determine size of string. */
+    int str_size = snprintf(dest, 0, "%i", integer);
+
+    /* If an error occurs, just return that and skip the conversion. */
+    if (str_size < 0)
+        return str_size;
+
+    /* Do the conversion and return. */
+    return snprintf(dest, (str_size + 1), "%i", integer);
+
+}
 
 size_t guac_strlcpy(char* restrict dest, const char* restrict src, size_t n) {
 
--- a/src/common-ssh/ssh.c
+++ b/src/common-ssh/ssh.c
@@ -24,6 +24,7 @@
 #include <guacamole/client.h>
 #include <guacamole/fips.h>
 #include <guacamole/mem.h>
+#include <guacamole/socket-tcp.h>
 #include <guacamole/string.h>
 #include <libssh2.h>
 
@@ -34,15 +35,11 @@
 #include <openssl/err.h>
 #include <openssl/ssl.h>
 
-#include <errno.h>
-#include <netdb.h>
-#include <netinet/in.h>
 #include <pthread.h>
 #include <pwd.h>
 #include <stddef.h>
 #include <stdlib.h>
 #include <string.h>
-#include <sys/socket.h>
 #include <unistd.h>
 
 #ifdef LIBSSH2_USES_GCRYPT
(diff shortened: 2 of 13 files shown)


src/libguac/string.c
```diff
@@ -44,6 +45,22 @@
  */
 #define REMAINING(n, length) (((n) < (length)) ? 0 : ((n) - (length)))
 
+size_t guac_itoa(char* restrict dest, int integer) {
+
+    /* Make sure memory is cleared. */
+    guac_mem_free(dest);
```
jmuehlner, line 51: What if somebody wants to write an integer value into something that can't be freed, like a stack-allocated character buffer?

necouchman: Yeah, that's a fair point. I could remove the freeing, but then I'd also have to remove the automatic allocation, and make the assumption that the caller has already allocated the correct amount of memory, right? I was trying to roll as much of that into this "convenience" function as possible, but I'm open to other suggestions if that makes sense.

From what little I can gather, it isn't possible to detect if something is heap or stack allocated?

jmuehlner: I don't think it's unreasonable to expect the caller to have to allocate enough space - that's pretty normal. 

If you do want to have this allocate space itself, consider having the length of the string passed in as a pointer the way that `dest` currently is, and having it return the newly allocated string. That way there won't be any chance of somebody passing something un-freeable into the function.

necouchman: > On balance I think the first option is still better though.

By which you mean, just ditch the automatic allocation and assume that they've allocated the space for it to begin with?

jmuehlner: Yes. That would allow the greatest flexibility in how this function could be used.

necouchman: Sounds good - should be implemented as suggested.

src/libguac/string.c
```diff
@@ -44,6 +45,16 @@
  */
 #define REMAINING(n, length) (((n) < (length)) ? 0 : ((n) - (length)))
 
+size_t guac_itoa(char* restrict dest, int integer) {
+
+    /* Determine size of string. */
+    int str_size = snprintf(dest, 0, "%d", integer);
```
jmuehlner, line 51: This is a signed integer specifier (as opposed to `%i` for unsigned) but neither the type nor the documentation for this function indicate that fact.

Is that intentional?

necouchman: Nah - I should probably be consistent with the return type of the function and just use `size_t`.

necouchman: Ah, nevermind, misunderstood what you were saying. Will switch it over to an unsigned integer.

necouchman: Okay, changed it to "%i" and updated the function documentation and the type to make it `unsigned`.

jmuehlner: Ugh, sorry I typed that backwards when I commented. `%i` is signed, and `%d` is unsigned. So this is still wrong. My bad.

necouchman: Actually, according to the man page for the `*printf` functions:
```
       d, i   The  int  argument  is  converted  to signed decimal notation.  The precision, if any, gives the minimum number of digits that must appear; if the converted value requires
              fewer digits, it is padded on the left with zeros.  The default precision is 1.  When 0 is printed with an explicit precision 0, the output is empty.

       o, u, x, X
              The unsigned int argument is converted to unsigned octal (o), unsigned decimal (u), or unsigned hexadecimal (x and X) notation.  The letters abcdef are used for x  conver‐
              sions;  the  letters  ABCDEF are used for X conversions.  The precision, if any, gives the minimum number of digits that must appear; if the converted value requires fewer
              digits, it is padded on the left with zeros.  The default precision is 1.  When 0 is printed with an explicit precision 0, the output is empty.
```

So, seems like maybe I should try "u", instead??

src/libguac/string.c
jmuehlner: I'm still seeing the `free()` call - did a commit get lost?

necouchman: No, just didn't remove it...fixed via rebase.

jmuehlner: Looks like it's still allocating the memory? I think we'd decided to let the caller allocate the memory for it, yeah?

necouchman: Yeah, sorry, you're correct. Let me fix that up...

necouchman: Okay, maybe now I've done it correctly?! LoL :facepalm:

kohrar:
Thanks for the PR. I gave it a try after merging it with master but can't seem to get my VNC connection to work as designed.

It seems as though the guacd thread stops immediately after sending the WoL packet, possibly after calling  `guac_wol_wake_and_wait`. This is what I see in the debug logs:
```
guacd[216]: INFO:       User "@129a3167-dfc7-4826-80cd-89b938108cd8" joined connection "$9068079c-9de1-4f0c-b29e-f8651ba25cf1" (1 users now present)
guacd[216]: DEBUG:      Sending Wake-on-LAN packet, and pausing for 2 seconds.
guacd[216]: DEBUG:      Client is using protocol version "VERSION_1_5_0"
guacd[1]: INFO: Connection "$9068079c-9de1-4f0c-b29e-f8651ba25cf1" removed.
guacd[1]: DEBUG:        Unable to request termination of client process: No such process
guacd[1]: DEBUG:        All child processes for connection "$9068079c-9de1-4f0c-b29e-f8651ba25cf1" have been terminated.

```

If I set wol-wake-time to 0 which (which bypasses the call to `guac_wol_wake_and_wait`), the VNC connection works once the machine comes up.

kohrar:
Just a quick update -- the segfault comes from [`str_port` being uninitialized before getting passed to `guac_itoa`](https://github.com/necouchman/guacamole-server/blob/working/wol-improvements/src/protocols/vnc/vnc.c#L289).  After fixing that (I just hard coded a malloc of 6 bytes to it) and it seems to be working OK.

necouchman:
Thanks, @kohrar! I suspect that after I converted the `guac_itoa()` function to not do the allocation itself I forgot to go back and allocate. I'll fix this up...

necouchman:
> Just a quick update -- the segfault comes from [`str_port` being uninitialized before getting passed to `guac_itoa`](https://github.com/necouchman/guacamole-server/blob/working/wol-improvements/src/protocols/vnc/vnc.c#L289). After fixing that (I just hard coded a malloc of 6 bytes to it) and it seems to be working OK.

Apparently I had already fixed this in my git code, but not pushed it :facepalm:. Should be fixed, now.

NickVStepin (changes requested):

src/libguac/string.c
```diff
@@ -44,6 +45,16 @@
  */
 #define REMAINING(n, length) (((n) < (length)) ? 0 : ((n) - (length)))
 
+int guac_itoa(char* restrict dest, unsigned int integer) {
+
+    /* Determine size of string. */
+    int str_size = snprintf(dest, 0, "%i", integer);
```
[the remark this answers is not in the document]
necouchman: Fixed via rebase.

necouchman:
@jmuehlner Any further changes that need to be made on this one?

NickVStepin:
I don't have any other changes.

 

С наилучшими пожеланиями,

Николай В.Степин

 

Директор информационной службы

ООО "Декор"

 <http://www.evroplast.ru/> www.evroplast.ru

 ***@***.***> ***@***.***

+7(495) 789-6270 д.3108

 

Post hoc, non est propter hoc.

jmuehlner:
> @jmuehlner Any further changes that need to be made on this one?

Nope, looks reasonable to me.

PR #471, opened 2023-12-28
GUACAMOLE-1140: Address RDP pipe svc race condition.

necouchman:
This PR attempts to address the race condition associated with the RDP client join handler and the pipe svc attempt to access the `available_svc` member of the `rdp_client` data structure. I've taken both suggestions in one of the Jira tickets for 1) acquiring the overall lock and 2) doing a `NULL` check around `available_svc` before attempting to access it.

It's worth noting that simply acquiring the lock on its own within the `guac_rdp_join_pending_handler()` function does not actually resolve the problem - the `NULL` check is required. So, I'm not sure in the lock is actually worth anything, or if there's somewhere else (around the SFTP connection, maybe?) that the lock should also be acquired?


jmuehlner:
Looks like this branch wasn't rebased against `staging/1.5.5`...

necouchman:
Well, that somehow got completely messed up. Closing this PR to re-open against staging/1.5.5.

PR #474, opened 2024-01-01
GUACAMOLE-1231: Implement basic support for restoring minimized RAIL windows

necouchman:
This is a really basic update for the Guacamole RAIL channel that handles situations where RAIL windows are minimized by simply immediately sending the `SC_RESTORE` command for that window back to the RDP server, causing the window to immediately re-appear. This resolves, at a very basic level, the issue with RAIL windows that are minimized becoming inaccessible within the Guacamole interface.

Looking at both the MS-RDPERP specification and the xfreerdp client, I suspect there is a lot more potential for some really cool functionality within the RAIL channel that could be implemented within Guacamole - like the ability to track what RAIL windows are open for a particular client, perhaps even open them in new/separate windows, and perhaps even integrate with the current multi-connection support to provide an icon representation of "minimized" RAIL windows. But, I thought I'd go the simpler route, first, and just solve the problem of not being able to access things that get [mistakenly] minimized.

### DIFF
--- a/src/protocols/rdp/channels/rail.c
+++ b/src/protocols/rdp/channels/rail.c
@@ -26,26 +26,14 @@
 #include <freerdp/event.h>
 #include <freerdp/freerdp.h>
 #include <freerdp/rail.h>
+#include <freerdp/window.h>
 #include <guacamole/client.h>
+#include <guacamole/mem.h>
 #include <winpr/wtypes.h>
 #include <winpr/wtsapi.h>
 
 #include <stddef.h>
 #include <string.h>
-
-#ifdef FREERDP_RAIL_CALLBACKS_REQUIRE_CONST
-/**
- * FreeRDP 2.0.0-rc4 and newer requires the final argument for all RAIL
- * callbacks to be const.
- */
-#define RAIL_CONST const
-#else
-/**
- * FreeRDP 2.0.0-rc3 and older requires the final argument for all RAIL
- * callbacks to NOT be const.
- */
-#define RAIL_CONST
-#endif
 
 /**
  * Completes initialization of the RemoteApp session, responding to the server
@@ -86,6 +74,7 @@ static UINT guac_rdp_rail_complete_handshake(RailClientContext* rail) {
     };
 
     /* Send client handshake response */
+    guac_client_log(client, GUAC_LOG_TRACE, "Sending RAIL handshake.");
     pthread_mutex_lock(&(rdp_client->message_lock));
     status = rail->ClientHandshake(rail, &handshake);
     pthread_mutex_unlock(&(rdp_client->message_lock));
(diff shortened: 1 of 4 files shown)


mike-jumper:
> This is a really basic update for the Guacamole RAIL channel that handles situations where RAIL windows are minimized by simply immediately sending the `SC_RESTORE` command for that window back to the RDP server, causing the window to immediately re-appear. ...

I rather like this. It immediately mitigates the issue at hand without requiring huge changes, but also doesn't get in the way of some future representation for minimized windows.

src/protocols/rdp/channels/rail.c
```diff
@@ -81,11 +69,12 @@ static UINT guac_rdp_rail_complete_handshake(RailClientContext* rail) {
         /* Build number 7600 (0x1DB0) apparently represents Windows 7 and
          * compatibility with RDP 7.0. As of this writing, this is the same
          * build number sent for RAIL connections by xfreerdp. */
-        .buildNumber = 7600
+        .buildNumber = 9600
```
jmuehlner, line 72: This comment is now out of date.

src/protocols/rdp/channels/rail.c
```diff
@@ -213,9 +208,63 @@ static UINT guac_rdp_rail_handshake(RailClientContext* rail,
...
+ * @param orderInfo
+ *     A pointer to the data structure that contains information about what
+ *     window was updated what updates were performed.
+ *
+ * @param windowState
+ *     A pointer to the data structure that contains details of the updates
+ *     to the window, as indicated by flags in the orderInfo field.
+ *
+ * @return
+ *     TRUE if the client-side processing of the updates as successful; otherwise
```
jmuehlner, line 233: Looks to me like it always returns true,

necouchman: Clarified via rebase.

src/protocols/rdp/channels/rail.c
necouchman: Reverted the build number to `7600` - looks like even the current xfreerdp code still uses 7600 (and sometimes 7601), so we'll just stick with that. I do not believe it impacts functionality at all and isn't really required for this change.

src/protocols/rdp/channels/rail.c
```diff
@@ -86,6 +74,7 @@ static UINT guac_rdp_rail_complete_handshake(RailClientContext* rail) {
     };
 
     /* Send client handshake response */
+    guac_client_log(client, GUAC_LOG_DEBUG, "Sending RAIL handshake.");
```
jmuehlner, line 77: The DEBUG-level messages in this file seem more TRACEy than DEBUGy, no? 

Having these identical DEBUG messages appear in the log every single time the handshake is completed feels like overkill to me.

necouchman: Yeah, probably so...I'll switch them over to `TRACE`, instead.

necouchman: Fixed via rebase.

src/protocols/rdp/channels/rail.h
```diff
@@ -20,7 +20,34 @@
...
 #define GUAC_RDP_CHANNELS_RAIL_H
 
+#include "config.h"
+
 #include <freerdp/freerdp.h>
+#include <freerdp/window.h>
+
+#ifdef FREERDP_RAIL_CALLBACKS_REQUIRE_CONST
+/**
+ * FreeRDP 2.0.0-rc4 and newer requires the final argument for all RAIL
```
jmuehlner, line 30: Isn't it the last 2 arguments? That's what I see in the actual function signature.

necouchman: Well, it depends on how many arguments the functions have. To your point, the `WindowUpdate` callback has two arguments, both of which need to be `const`; however, the `ServerHandshake` and `ServerHandshakeEx` callbacks only have a single argument that is `const`.

I've tweaked the comment via rebase.

PR #475, opened 2024-01-06
add-optional-permissions: add ability to change permissions of screen recording file and path

alailsonko:
### Description

Add record-path-permissions and record-file-permissions optional fields when starting record of session.

### DIFF
--- a/src/libguac/guacamole/recording.h
+++ b/src/libguac/guacamole/recording.h
@@ -22,6 +22,8 @@
 
 #include <guacamole/client.h>
 
+#include <sys/types.h>
+
 /**
  * Provides functions and structures to be use for session recording.
  *
@@ -122,6 +124,14 @@ typedef struct guac_recording {
  *     written, or non-zero if the path should be created if it does not yet
  *     exist.
  *
+ * @param file_permissions
+ *     The permissions to apply for the recording file created within the specified
+ *     path.
+ *
+ * @param path_permissions
+ *     The permissions to apply for the recording path created within the specified
+ *     path.
+ *
  * @param include_output
  *     Non-zero if output which is broadcast to each connected client
  *     (graphics, streams, etc.) should be included in the session recording,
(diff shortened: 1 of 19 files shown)


necouchman:
@alailsonko: Thanks for the pull request and contribution to the project. You'll need to get a Jira account and create a Jira issue for this, and then tag both the pull request and the commit messages with the Jira issue (GUACAMOLE-XXXX:) before we'll accept and merge this.

src/libguac/guacamole/recording.h
```diff
@@ -154,8 +164,8 @@ typedef struct guac_recording {
  *     recording will be written, NULL otherwise.
  */
 guac_recording* guac_recording_create(guac_client* client,
-        const char* path, const char* name, int create_path,
-        int include_output, int include_mouse, int include_touch,
+        const char* path, const char* name, int create_path, mode_t file_permissions,
+        mode_t path_permissions, int include_output, int include_mouse, int include_touch,
```
mike-jumper, line 168: I'm not sure about directly coupling the API for screen recordings to POSIX-specific `mode_t` when libguac is generally intended to support Windows as a platform. Perhaps there's a way to implement this without making `guac_recording_create()` non-portable?

If it were possible to implement this in a way that doesn't require API/ABI-breaking changes, that would also be ideal.

alailsonko: i dont have the jira account for create issues at the moment.

thanks for letting me know and giving attention to this PR, i understand it is very delicate change.

we could change from mode_t to int, i think it would be agnostic. or creating own mode_t type.

i was reading documentation of c in microsoft and they use type int for mode.

- https://learn.microsoft.com/en-us/cpp/c-runtime-library/reference/open-wopen?view=msvc-170
```c
int _open(
   const char *filename,
   int oflag [,
   int pmode]
);
int _wopen(
   const wchar_t *filename,
   int oflag [,
   int pmode]
);
```
- https://learn.microsoft.com/en-us/cpp/c-runtime-library/reference/sopen-s-wsopen-s?view=msvc-170
```c
errno_t _sopen_s(
   int* pfh,
   const char *filename,
   int oflag,
   int shflag,
   int pmode
);
errno_t _wsopen_s(
   int* pfh,
   const wchar_t *filename,
   int oflag,
   int shflag,
   int pmode,
);
```

PR #476, opened 2024-01-10
GUACAMOLE-1892: Fix the build issues related to librt for Ubuntu 22.04.

myjimmy:
I've fixed the build issues related to librt for Ubuntu 22.04.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -68,6 +68,14 @@ AC_CHECK_LIB([pthread], [pthread_create], [PTHREAD_LIBS=-lpthread
               AC_DEFINE([HAVE_LIBPTHREAD],,
                         [Whether libpthread was found])])
 
+# librt
+AC_CHECK_FUNC([timer_create], [AC_MSG_RESULT([timer_create was found without librt.])],
+              [AC_CHECK_LIB([rt], [timer_create],
+                            [AC_MSG_RESULT([timer_create was found in librt.])
+                                           RT_LIBS=-lrt],
+                            [AC_MSG_ERROR([timer_create could not be found.])])
+              ])
+
 # Include libdl for dlopen() if necessary
 AC_CHECK_LIB([dl], [dlopen],
              [DL_LIBS=-ldl],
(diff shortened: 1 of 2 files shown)


configure.ac
```diff
@@ -68,6 +68,10 @@ AC_CHECK_LIB([pthread], [pthread_create], [PTHREAD_LIBS=-lpthread
               AC_DEFINE([HAVE_LIBPTHREAD],,
                         [Whether libpthread was found])])
 
+# librt
+AC_CHECK_LIB([rt], [timer_create], [RT_LIBS=-lrt],
+             AC_MSG_ERROR("librt is required for performing timer-related operations"))
```
mike-jumper, line 73: It's not _universally_ required. This check should only be fatal where there is no `timer_create()` without linking in librt. If the test determines that `timer_create()` does not require linking with librt, this check should pass and `RT_LIBS` should be empty.

src/libguac/Makefile.am
```diff
@@ -171,5 +171,6 @@ libguac_la_LDFLAGS =     \
     @UUID_LIBS@          \
     @VORBIS_LIBS@        \
     @WEBP_LIBS@          \
-    @WINSOCK_LIBS@
+    @WINSOCK_LIBS@       \
+    @RT_LIBS@
```
mike-jumper, line 175: Please maintain alphabetical ordering.

mike-jumper (changes requested):

src/libguac/Makefile.am
myjimmy: OK, I've changed code according to your opinion.

configure.ac
```diff
@@ -68,6 +68,11 @@ AC_CHECK_LIB([pthread], [pthread_create], [PTHREAD_LIBS=-lpthread
               AC_DEFINE([HAVE_LIBPTHREAD],,
                         [Whether libpthread was found])])
 
+# librt
+RT_LIBS=
+AC_CHECK_LIB([rt], [timer_create], [RT_LIBS=-lrt],
+             AC_MSG_ERROR("librt is required for performing timer-related operations"))
```
mike-jumper, line 74: Please see https://github.com/apache/guacamole-server/pull/476#discussion_r1447776140:

> It's not _universally_ required. This check should only be fatal where there is no `timer_create()` without linking in librt. If the test determines that `timer_create()` does not require linking with librt, this check should pass and `RT_LIBS` should be empty.

As written, this test will fail the build if librt is absent, even if the platform in question doesn't require librt for `timer_create()`.

mike-jumper (changes requested):

configure.ac
myjimmy: @mike-jumper I've fixed the above code specified by you.

myjimmy: @mike-jumper I've fixed the above code specified by you.

mike-jumper (approved):

PR #477, opened 2024-01-12
GUACAMOLE-1900: Fix a segfault issue caused by uninitialized memory.

myjimmy:
I've fixed a segfault issue related to the uninitialized memory when resizing the browser window.


necouchman (approved):
Nice find @myjimmy!

PR #478, opened 2024-01-15
GUACAMOLE-1907: Removed the double line breaks.

myjimmy:
I have removed the double line breaks incompatible to the coding style.


README
```diff
@@ -14,7 +14,6 @@ A full manual is available as well:
 
     http://guacamole.apache.org/doc/gug/
 
-
```
necouchman, line 17: We might want to consider leaving them in the `README` file.

necouchman (changes requested):
A couple of things:
* I think the double line breaks should probably be left in the `README`
* Most of the places where you've removed double line breaks are actually in the top of header files, between the comment containing the license, and the first `ifdef` statement. I'm not sure if these are the desired style or are exceptions to the desired style.  @mike-jumper or @jmuehlner, any opinions on this one?

myjimmy:
@necouchman I've reverted the changes in the `README` file by following your 1st opinion.
And, please let me know what I need to do for handling your 2nd opinion.

necouchman (approved):

PR #480, opened 2024-01-18
GUACAMOLE-1903: set rcMask to NULL

svogl:
set rcMask to NULL to prevent double-free errors occuring in libvncclient 0.9.13 and up.


necouchman:
@svogl Thanks for this - I'm good overall with the change, but have two requests:
* Both the PR and the commit message need to be adjusted slightly so that there is a ":" after the Jira issue - e.g. "GUACAMOLE-1903: Set rcMask to NULL".
* Could you also rebase this change, and adjust the pull request, against the staging/1.5.5 branch, instead of master, so that we can pull it in as a bugfix to the 1.5.5 release?

svogl:
@necouchman thanks for the suggestions, done!

necouchman:
> @necouchman thanks for the suggestions, done!

Unfortunately I think you'll need to rebase in git and then push - the PR now has 129 commits on it ;-)

svogl:
hmm, ok - never rebased a PR.... will create a new PR against the staging branch and dump this one...

PR #482, opened 2024-01-18
GUACAMOLE-1910: Add locking around TLS socket

neandrake:
When TLS is turned on the file descriptor for the socket does not use locking when writing to it. It results in corrupted guacamole instructions being written to the stream when there are multiple connected clients.

This adds a mutex in the same manner done in `socket-fd.c`.

### DIFF
--- a/src/libguac/guacamole/socket-ssl.h
+++ b/src/libguac/guacamole/socket-ssl.h
@@ -31,6 +31,7 @@
 #include "socket-types.h"
 
 #include <openssl/ssl.h>
+#include <pthread.h>
 
 /**
  * SSL socket-specific data.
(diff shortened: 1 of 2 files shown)


mike-jumper (commented):
LGTM! Mind rebasing against `staging/1.5.5` so this can be pulled into the upcoming 1.5.5 release?

neandrake:
> LGTM! Mind rebasing against `staging/1.5.5` so this can be pulled into the upcoming 1.5.5 release?

Okay I think I rebased it properly. Please take a look.

I pulled `upstream/staging/1.5.5`, rebased `GUAC-1910` on to it and force-pushed to my forked branch. I then updated this pull request to merge onto `staging/1.5.5` instead of `master`.

neandrake:
Do you prefer bringing up issues in the mailing list prior to filing an issue on jira and submitting a pull request?

mike-jumper (approved):

PR #484, opened 2024-01-22
GUACAMOLE-1912: Fix the spell errors in the identifiers of some constants.

myjimmy:
Fixed the spell errors in the identifiers of some constants.

### DIFF
--- a/src/libguac/error.c
+++ b/src/libguac/error.c
@@ -49,7 +49,7 @@ const char* __GUAC_STATUS_PERMISSION_DENIED_STR = "Permission denied";
 const char* __GUAC_STATUS_BUSY_STR              = "Resource busy";
 const char* __GUAC_STATUS_NOT_AVAILABLE_STR     = "Resource not available";
 const char* __GUAC_STATUS_NOT_SUPPORTED_STR     = "Not supported";
-const char* __GUAC_STATUS_NOT_INPLEMENTED_STR   = "Not implemented";
+const char* __GUAC_STATUS_NOT_IMPLEMENTED_STR   = "Not implemented";
 const char* __GUAC_STATUS_TRY_AGAIN_STR         = "Temporary failure";
 const char* __GUAC_STATUS_PROTOCOL_ERROR_STR    = "Protocol violation";
 const char* __GUAC_STATUS_NOT_FOUND_STR         = "Not found";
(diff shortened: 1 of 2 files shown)


necouchman:
Thanks, @myjimmy.

@mike-jumper Is there any chance this will cause ABI/API incompatibilities with library linking? I can't remember what conditions that occurs under...

mike-jumper:
> @mike-jumper Is there any chance this will cause ABI/API incompatibilities with library linking? I can't remember what conditions that occurs under...

Mildly, yes. The changes here are ABI-compatible (enums don't end up as symbols in the library - they are just resolved to integers by the C compiler at compile time based on the definition in the header), but they are not API-compatible.

To maintain API compatibility, we would need to:

1. Keep `GUAC_STATUS_NOT_INPLEMENTED` (sic), but mark it as deprecated.
2. Add the corrected `GUAC_STATUS_NOT_IMPLEMENTED` constant, explicitly mapped to the same value as old `GUAC_STATUS_NOT_INPLEMENTED` (this will require adding explicit values to all constants in the `enum`)

So we'd end up with something like:

```c
    ...

    /**
     * The requested operation is not supported.
     */
    GUAC_STATUS_NOT_SUPPORTED = 14,

    /**
     * Support for the requested operation is not yet implemented.
     *
     * @deprecated This constant contains a typo in its name and will be
     * removed in a future release. Use GUAC_STATUS_NOT_IMPLEMENTED instead.
     */
    GUAC_STATUS_NOT_INPLEMENTED = 15,

    /**
     * Support for the requested operation is not yet implemented.
     */
    GUAC_STATUS_NOT_IMPLEMENTED = 15,

    /**
     * The operation is temporarily unable to be performed, but may succeed if
     * reattempted.
     */
    GUAC_STATUS_TRY_AGAIN = 16,

    ...
```

That would ensure that code that continues to reference the old `GUAC_STATUS_NOT_INPLEMENTED` constant will still compile.

necouchman:
> Mildly, yes. The changes here are ABI-compatible (enums don't end up as symbols in the library - they are just resolved to integers by the C compiler at compile time based on the definition in the header), but they are not API-compatible.

So, we'll also need to rev the libguac version as part of GUACAMOLE-1906, assuming this change goes into the 1.5.5 release.

mike-jumper:
> > Mildly, yes. The changes here are ABI-compatible (enums don't end up as symbols in the library - they are just resolved to integers by the C compiler at compile time based on the definition in the header), but they are not API-compatible.
> 
> So, we'll also need to rev the libguac version as part of GUACAMOLE-1906, assuming this change goes into the 1.5.5 release.

Yep, good catch.

myjimmy:
@necouchman, for merging this PR into the `staging/1.5.5` branch, please let me know what I'll work on.

necouchman (approved):
Looks good. I'll open a change under GUACAMOLE-1906 to rev the libguac version.

PR #490, opened 2024-02-05
GUACAMOLE-1921: Process argv handler even when the connection is read-only.

necouchman:
This simple change processes the `argv` handler for the VNC protocol even when the connection is read-only, which is required to prompt for credentials for read-only connections. As these are the only parameters that are processed by the `argv` handler, there is no additional impact on the code.

It's worth noting that, if other parameters are added in the future, the `argv` handler may need to check the `read_only` parameter and make decisions about which parameters should be allowed to be processed and which should be ignored or blocked.

### DIFF
--- a/src/protocols/vnc/argv.h
+++ b/src/protocols/vnc/argv.h
@@ -28,6 +28,12 @@
 /**
  * Handles a received argument value from a Guacamole "argv" instruction,
  * updating the given connection parameter.
+ *
+ * As noted in the user.c file, care should be taken when updating this
+ * callback to make sure that arguments are handled correctly when
+ * a connection is marked as read-only, and to make sure that any
+ * usage of this callback for non-owner users of a connection does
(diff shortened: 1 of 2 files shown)


mike-jumper (changes requested):
To avoid any future issues, I think we should add some comments here noting:

* That `read-only` may need to be considered for handling of future argument values received via `argv` that are not related to authentication.
* That changing things such that `argv` is handled for users other than the owner may have security implications unless the handling of `argv` is modified to only accept `username`/`password` from the owner.

Otherwise, LGTM!

necouchman:
@mike-jumper I have added some comments - let me know if the style of those is okay, as they're a bit of a one-off in terms of the types of comments used throughout the code. In addition to the comment in `user.c`, I added a note to the `argv.h` header file where the user callback is documented.

mike-jumper (approved):
Very nice.

PR #491, opened 2024-02-09
GUACAMOLE-1924: Remove vestigial pthread_key_create calls during client initialization.

jmuehlner:
`guac_rwlock_init()` already initializes these keys - these extra calls will slowly eat up system resources until `pthread_key_create()` calls start failing. 

Removing them will fix that issue.


jmuehlner:
wrong ticket number - closed in favor of https://github.com/apache/guacamole-server/pull/492

PR #492, opened 2024-02-09
GUACAMOLE-1920: Remove vestigial pthread_key_create calls during client initialization.

jmuehlner:
`guac_rwlock_init()` already initializes these keys - these extra calls will slowly eat up system resources until `pthread_key_create()` calls start failing. 

Removing fixes that issue.


mike-jumper:
Wow, good catch!

mike-jumper (approved):

PR #494, opened 2024-02-25
GUACAMOLE-1906: Bump soversion for liguac after API changes.

necouchman:
After the changes in #484, the version of libguac needs to be bumped.

### DIFF
--- a/src/libguac/Makefile.am
+++ b/src/libguac/Makefile.am
@@ -160,7 +160,7 @@ libguac_la_CFLAGS = \
     -Werror -Wall -pedantic
 
 libguac_la_LDFLAGS =     \
-    -version-info 23:0:0 \
+    -version-info 24:0:0 \
     -no-undefined        \
     @CAIRO_LIBS@         \
     @DL_LIBS@            \


src/libguac/Makefile.am
```diff
@@ -160,7 +160,7 @@ libguac_la_CFLAGS = \
     -Werror -Wall -pedantic
 
 libguac_la_LDFLAGS =     \
-    -version-info 23:0:0 \
+    -version-info 24:0:0 \
```
mike-jumper, line 163: I'm late to the party here, but are the changes between 1.5.4 and 1.5.5 API-breaking?

If not, I think this should be `24:0:1` (interfaces added, not changed or removed). Doing a quick diff between `1.5.4` and `staging/1.5.5`, I'm not seeing anything stand out as breaking changes.

necouchman: Nah, this was probably my bad, here - I was trying to figure out the right one to do.

Also, do you mean `24:0:1`  or `23:0:1`?

I can open a new PR to fix it...

mike-jumper: Never mind - I think you were right to update to `24:0:0`. In the case of 1.5.5, the only change I see to any interface is the additional `socket_lock` member added to `guac_socket_ssl_data`, and that does qualify as a "changed interface".

PR #496, opened 2024-03-09
GUACAMOLE-1732: Prevent deadlock and enhance cleanup in RDP client.

aleitner:
Resolve authentication timeout and deadlock issues. There was a timeout occurring approximately 12 seconds into an RDP connection when using NLA and resources were not properly being cleaned up.

To address this, `guac_argv_stop` has been introduced. It ensures that when a client disconnects, any threads waiting for user input (such as username or password), can effectively abort their wait, preventing the connection from hanging indefinitely.
A deadlock has also been mitigated by modifying the lock state to allow read access to other threads when it is safe to do so.

### DIFF
--- a/src/protocols/rdp/rdp.c
+++ b/src/protocols/rdp/rdp.c
@@ -108,8 +108,14 @@ BOOL rdp_freerdp_pre_connect(freerdp* instance) {
 
     /* Load "AUDIO_INPUT" plugin for audio input*/
     if (settings->enable_audio_input) {
+        /* Upgrade the lock to write temporarily for setting the newly allocated audio buffer */
+        guac_rwlock_acquire_write_lock(&(rdp_client->lock));
         rdp_client->audio_input = guac_rdp_audio_buffer_alloc(client);
         guac_rdp_audio_load_plugin(instance->context);
+
+        /* Downgrade the lock to allow for concurrent read access */
+        guac_rwlock_release_lock(&(rdp_client->lock));
+        guac_rwlock_acquire_read_lock(&(rdp_client->lock));
     }
 
     /* Load "cliprdr" service if not disabled */
@@ -470,7 +476,7 @@ static int guac_rdp_handle_connection(guac_client* client) {
     /* Init random number generator */
     srandom(time(NULL));
 
-    pthread_rwlock_wrlock(&(rdp_client->lock));
+    guac_rwlock_acquire_write_lock(&(rdp_client->lock));
 
     /* Create display */
     rdp_client->display = guac_common_display_alloc(client,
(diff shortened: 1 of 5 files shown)


src/protocols/rdp/rdp.c
```diff
@@ -525,12 +525,24 @@ static int guac_rdp_handle_connection(guac_client* client) {
     /* Set default pointer */
     guac_common_cursor_set_pointer(rdp_client->display->cursor);
 
+    /* 
+     * Downgrade the lock to allow for concurrent read access
+     * Access to read locks need to be made available for other processes to use 
+     * while we await credentials from the user.
+     */
+    pthread_rwlock_unlock(&(rdp_client->lock));
```
jmuehlner, line 533: I like this approach, but I think there's at least one assignment that should probably still be guarded by the write lock - namely:

https://github.com/apache/guacamole-server/blob/main/src/protocols/rdp/rdp.c#L112

The lock itself is a bit unclear on exactly what it's guarding, and it _might_ be ok to have this particular assignment unguarded since it's only set once, so the only stale value that any other thread/process could read would be `NULL`.

jmuehlner: Maybe it would make sense to guard just that one assignment using the write lock? Possibly using a `guac_rwlock`?

Thoughts @mike-jumper?

aleitner: On one hand I like the idea of adding a separate lock for around setting the audio handler, but on the other hand I'm not keen on nesting more locks

jmuehlner: Sorry, I was unclear here. What I meant was using `guac_rwlock` to convert the entire `rdp_client->lock` into a reentrant rwlock, so you could keep the current read lock around the `freerdp_connect()` call, and then additionally grab the write lock, but only when the `rdp_client->audio_input` (or any assignment) is done.

aleitner: Oh I just looked through the guac_rwlock api and that seems much cleaner. I can try and implement that

jmuehlner: Cool, looks good. I realize now that this doesn't _really_ require reentrancy (acquiring the write lock while holding the read lock, as you're doing now is equivalent to just releasing the read lock and acquiring the write lock).

But it certainly won't hurt to use the reentrant locks either.

jmuehlner:
Hey @aleitner this looks like a good candidate for the upcoming 1.5.5 release. It's definitely a bug fix. Want to target it at `staging/1.5.5` instead?

aleitner:
> Hey @aleitner this looks like a good candidate for the upcoming 1.5.5 release. It's definitely a bug fix. Want to target it at `staging/1.5.5` instead?

No problem!

jmuehlner (approved):
LGTM

PR #497, opened 2024-03-13
GUACAMOLE-1732: Free dangling locks and correct and remove unnecessary locks.

aleitner:
This pull request corrects issues identified by Coverity Scan related to a misunderstanding of lock upgrading in `rwlock.h`. I refined the locking logic to ensure consistent lock acquisition and release, following the intended use of the `rwlock.h`

### DIFF
--- a/src/protocols/rdp/rdp.c
+++ b/src/protocols/rdp/rdp.c
@@ -115,7 +115,6 @@ BOOL rdp_freerdp_pre_connect(freerdp* instance) {
 
         /* Downgrade the lock to allow for concurrent read access */
         guac_rwlock_release_lock(&(rdp_client->lock));
-        guac_rwlock_acquire_read_lock(&(rdp_client->lock));
     }
 
     /* Load "cliprdr" service if not disabled */
@@ -537,6 +536,7 @@ static int guac_rdp_handle_connection(guac_client* client) {
     }
 
     /* Upgrade to write lock again for further exclusive operations */
+    guac_rwlock_release_lock(&(rdp_client->lock));
     guac_rwlock_acquire_write_lock(&(rdp_client->lock));
 
     /* Connection complete */
@@ -843,4 +843,3 @@ void* guac_rdp_client_thread(void* data) {
     return NULL;
 
 }
-


src/protocols/rdp/rdp.c
```diff
@@ -537,6 +536,7 @@ static int guac_rdp_handle_connection(guac_client* client) {
     }
 
     /* Upgrade to write lock again for further exclusive operations */
+    guac_rwlock_release_lock(&(rdp_client->lock));
```
jmuehlner, line 539: This shouldn't be necessary - the guac rwlock is set up to handle upgrading from a read lock to a write lock just by calling `guac_rwlock_acquire_write_lock`. I don't think coverity fully understands how all this works (either that, or _I'm_ confused).

src/protocols/rdp/rdp.c
```diff
@@ -115,7 +115,6 @@ BOOL rdp_freerdp_pre_connect(freerdp* instance) {
 
         /* Downgrade the lock to allow for concurrent read access */
         guac_rwlock_release_lock(&(rdp_client->lock));
-        guac_rwlock_acquire_read_lock(&(rdp_client->lock));
```
jmuehlner, line 118: Isn't this going to just result in no lock being held at all? It sure looks like the read lock is still needed here.

jmuehlner:
I think at least some of what coverity's reporting here are actually false positives. It doesn't seem to understand the `guac_rwlock` functionality very well, and how it will lock/unlock based on the value of the thread-local.

Additionally, I think it might be confused about how the various FreeRDP callbacks work (i.e. `rdp_freerdp_pre_connect`). It's understandably upset that the function acquires a lock and never releases it, because it doesn't know that it's invoked from within a section of code that already holds that lock...

jmuehlner:
Of all of these findings, `CID 1593181` does look valid - `guac_client_free()` _does_ attempt to acquire the cursor lock and the users lock in the opposite order that `guac_common_cursor_update()` does - but this is also not a new behavior. Perhaps coverity just now updated and knows to look for this.

src/protocols/rdp/rdp.c
aleitner: If I understand the way the locks work the read lock still exists and releasing the write lock switches back to using the read lock.

src/protocols/rdp/rdp.c
aleitner: From what I read it seems like upgrading to a write lock adds a layer on top of the read locks and the still requires a free for each layer. I figured it was more straight forward an not necessary in this case to layer the locks

src/protocols/rdp/rdp.c
jmuehlner: Ah, right you are. Ignore me.

src/protocols/rdp/rdp.c
jmuehlner: Nevermind, I am confused.

PR #498, opened 2024-03-18
GUACAMOLE-1931: Allow writing to existing recording files.

jmuehlner:
See also https://github.com/apache/guacamole-client/pull/961


aleitner (approved):
This as well lgtm

PR #502, opened 2024-04-05
GUACAMOLE-1084: Implement basic handling for unsuccessful RAIL execution results.

necouchman:
This implements some basic logging and client handling (abort) for situations where a RAIL command is not executed successfully. Without this change, the RAIL session just hangs open until the server times out and closes the RDP connection, which can take several minutes.

### DIFF
--- a/src/protocols/rdp/channels/rail.c
+++ b/src/protocols/rdp/channels/rail.c
@@ -169,6 +169,37 @@ static UINT guac_rdp_rail_complete_handshake(RailClientContext* rail) {
 }
 
 /**
+ * A callback function that is invoked when the RDP server sends the result
+ * of the Remote App (RAIL) execution command back to the client, so that the
+ * client can handle any required actions associated with the result.
+ * 
+ * @param context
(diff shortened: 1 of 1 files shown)


src/protocols/rdp/channels/rail.c
```diff
@@ -168,6 +168,36 @@ static UINT guac_rdp_rail_complete_handshake(RailClientContext* rail) {
...
+ * 
+ * @param context
+ *     A pointer to the RAIL data structure associated with the current
+ *     RDP connection.
+ *
+ * @param execResult
+ *     A data structure containing the result of the RAIL command.
+ *
+ * @return
+ *     CHANNEL_RC_OK (zero) if the result was handled successfully, otherwise
```
jmuehlner, line 184: Looks to me like it always returns `CHANNEL_RC_OK`

necouchman: Clarified the return clause via rebase.

jmuehlner (approved):

PR #406, opened 2022-12-21
GUACAMOLE-192: Double and triple click function

Saiyeux:
Two files added to /src/protocols/ssh and small changes to input.c. Now we can double click a character and select the whole word. Also triple to the whole line.

PR #504, opened 2024-04-15
GUACAMOLE-192: Select word in terminal on double click

corentin-soriano:
Allows selecting a word on double-click and a complete line on triple-click.

Based on the unmaintained PR by @Saiyeux (https://github.com/apache/guacamole-server/pull/406) :

- All "rewritten" functions have been removed in favor of existing functions.
	-> Code simplification.
	-> Fixes several unexpected behaviors when scrolling upwards.
- The following characters are now considered part of a "word": [0-9\\/A-Za-z_\\~\\&\\-\\.\\?\\\\%=:\$]
	-> Allows selecting an IP address, a path or URL (ex: https://www.test.tld/file.php?param1=value1&param2=value%202)
- Improved condition in the guac_terminal_select_update (select.c) function to allow selecting a "word" of a single character on double-click (no impact on single-click).

### DIFF
--- a/src/terminal/terminal.c
+++ b/src/terminal/terminal.c
@@ -590,6 +590,10 @@ guac_terminal* guac_terminal_create(guac_client* client,
     /* Configure backspace */
     term->backspace = options->backspace;
 
+    /* Initialize mouse latest click time and counter */
+    term->click_timer = 0;
+    term->click_counter = 0;
+
     return term;
 
 }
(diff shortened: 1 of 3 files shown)


src/terminal/terminal.c
```diff
@@ -2224,3 +2254,80 @@ void guac_terminal_remove_user(guac_terminal* terminal, guac_user* user) {
...
+        (ascii_char == '/') ||
+        (ascii_char == ':') ||
+        (ascii_char == '=') ||
+        (ascii_char == '?') ||
+        (ascii_char == '\\') ||
+        (ascii_char == '_') ||
+        (ascii_char == '~'))
+        return true;
+    else
+        return false;
```
necouchman, line 2276: I believe this could be something like:
```
return ((ascii_char >= '0' && ascii_char <= '9') || 
        (ascii_char >= 'A' && ascii_char <= 'Z') || 
        (ascii_char >= 'a' && ascii_char <= 'z') ||
        (ascii_char == '$') ||
        (ascii_char == '%') ||
        (ascii_char == '&') ||
        (ascii_char == '-') ||
        (ascii_char == '.') ||
        (ascii_char == '/') ||
        (ascii_char == ':') ||
        (ascii_char == '=') ||
        (ascii_char == '?') ||
        (ascii_char == '\\') ||
        (ascii_char == '_') ||
        (ascii_char == '~'));
```

?

corentin-soriano: Yes it's more readable, I also change in the function just below !

src/terminal/terminal/terminal-priv.h
```diff
@@ -656,5 +666,27 @@ void guac_terminal_copy_rows(guac_terminal* terminal,
...
+/**
+ * Get the char (int ASCII code) at the row/col requested.
+ */
+int guac_terminal_get_char(guac_terminal* terminal, int row, int col);
+
+/**
+ * Selection on double click event
+ */
+void guac_terminal_double_click(guac_terminal* terminal, int row, int col);
+
```
necouchman, line 690: I don't think any of these functions are used outside of the `terminal.c` file, correct? If so, there's no reason to put them in the header file - you can just define them in the source file and also mark them as `static`.

Also, for the documentation for the functions - whether here or in the `termianl.c` file - you'll need to do more complete documentation - should include the overall description, the parameters, and the return value (if any).

corentin-soriano: - Declarations removed from header file.
- Functions marked as `static`.
- Function documentation moved to `terminal.c`.
- Improved documentation.

src/terminal/terminal/terminal-priv.h
```diff
@@ -465,6 +465,16 @@ struct guac_terminal {
      */
     bool disable_copy;
 
+    /**
+     * The time betwen two left clicks.
+     */
+    guac_timestamp click_timer;
+
+    /**
+     * Counter for left clicks :
```
jmuehlner, line 474: Incorrect punctuation here.

jmuehlner:
This looks good to me, aside from a minor punctuation mistake. My only other request would be that some of these commits be reworded and / or squashed.

Each commit should:
* Follow the format `GUACAMOLE-1234: Clarify commit message expectations.`, with the first word capitalized, and a period after the message.
* Cleary explain what the commit is intended to do.

 Something like `GUACAMOLE-192: revert removed line` is unclear about what line is being reverted, and whether it's a substantive change or not. If it's just a minor whitespace fix, I'd suggest just squashing it into whatever commit its related to.

src/terminal/terminal/terminal-priv.h
corentin-soriano: Fixed!

corentin-soriano:
> This looks good to me, aside from a minor punctuation mistake. My only other request would be that some of these commits be reworded and / or squashed.
> 
> Each commit should:
> 
> * Follow the format `GUACAMOLE-1234: Clarify commit message expectations.`, with the first word capitalized, and a period after the message.
> * Cleary explain what the commit is intended to do.
> 
> Something like `GUACAMOLE-192: revert removed line` is unclear about what line is being reverted, and whether it's a substantive change or not. If it's just a minor whitespace fix, I'd suggest just squashing it into whatever commit its related to.

Thank you for your review!

The punctuation mistake is corrected.
It was in fact not really useful to keep all of these commits which were only corrections of the first one. I squashed them.

jmuehlner (approved):

PR #508, opened 2024-04-16
GUACAMOLE-1940: Correct flag comparison in "guac_rwlock_acquire_write_lock" function

pourfar:
Corrected flag comparison in "guac_rwlock_acquire_write_lock" function


necouchman:
@pourfar Nice catch, thank you. One request - could you please modify both the PR title and the commit message to reflect what you're actually doing (correcting the flag comparison) rather than the problem?

pourfar:
@necouchman Done

necouchman (approved):

PR #510, opened 2024-04-17
GUACAMOLE-1945: Changed OpenSSL to build using source code

wojiushixiaobai:
- https://issues.apache.org/jira/browse/GUACAMOLE-1945

Build OpenSSL from source code to adapt to newer operating systems.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -22,9 +22,7 @@
 #
 
 # The Alpine Linux image that should be used as the basis for the guacd image
-# NOTE: Using 3.18 because the required openssl1.1-compat-dev package was
-# removed in more recent versions.
-ARG ALPINE_BASE_IMAGE=3.18
+ARG ALPINE_BASE_IMAGE=latest
 FROM alpine:${ALPINE_BASE_IMAGE} AS builder
 
 # Install build dependencies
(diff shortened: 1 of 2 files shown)


mike-jumper:
I'm not sure I understand why it's necessary for the Docker images themselves to build their own copies of OpenSSL. We need to do this for things like FreeRDP and libssh2, where distributions tend to lag behind in terms of updates, but OpenSSL is typically widely supported and rapidly patched for security.

Wouldn't it be better to switch to whichever OpenSSL package is provided and supported by the latest Alpine?

necouchman:
> I'm not sure I understand why it's necessary for the Docker images themselves to build their own copies of OpenSSL. We need to do this for things like FreeRDP and libssh2, where distributions tend to lag behind in terms of updates, but OpenSSL is typically widely supported and rapidly patched for security.
> 
> Wouldn't it be better to switch to whichever OpenSSL package is provided and supported by the latest Alpine?

I went back and looked at the history on this in the related Jira issues, and it looks like the situation is that Alpine 3.19 drops the OpenSSL 1.x series in favor of OpenSSL 3.x, but there's a but in the libvncclient library that, when linked against OpenSSL 3.x, results in VeNCrypt logins failing. Thus, we reverted to 1.x, and then had to pin the Alpine image to 3.18, which is the last version that includes support for the OpenSSL 1.x packages.

wojiushixiaobai:
OK. I've got it.

PR #511, opened 2024-04-20
Create Depend-on-debian.sh

CooperDActor:
i made a auto dependincies installer so they run  the bash and unattended install

thanks 

cooper 12yo


necouchman:
@CooperDActor A few concerns about this:
* While we provide guidance for installing in our manual, we generally do not provide distribution-specific code or scripts. I'm hesitant to start doing this - it feels like a very slippery slope into demands for other distributions to be included and maintained, and then having to make sure that it works across different versions of different distributions.
* Assuming we decided we'd like to include something like this, I don't think the root directory of the source tree is the right place. I'd envision setting up some sort of `contrib/` directory or something similar to put things like this in, that was clearly marked for items that aren't only tangentially related to the source code.
* You always need a Jira ticket to go with any changes, so, even if we decided to move forward with including this, so you'd need to get a Jira account, create the ticket, and mark the pull request and commit messages with that Jira ticket. See: https://guacamole.apache.org/open-source/.

mike-jumper:
Same here. I'm not necessarily against providing more automation around the build process, but I don't think we should start including scripts that just run a series of distribution-specific install commands to install dependencies without prompting.

If we were to adopt a practice like this, I think the script would need to be more robust and intelligent: identifying what needs to be installed, not assuming that using `sudo` is appropriate, and prompting the user to confirm that they are OK with installing those dependencies.

CooperDActor-bytes:
right will do thanks 

never done a pull request before 


will update and make a jiva request 

I am open to an idea of a folder and want names for it


thanks 

Cooper D'ANDILLY 
12yo

CooperDActor:
making a new one

PR #512, opened 2024-04-20
GUACAMOLE-1586: Don't add \n in clipboard if the line isn't finished.

corentin-soriano:
In the case of selecting multiple lines, do not add \n if the line is not really finished.

**Changes:**
- Correct buffer_row length on terminal init.
- Using the last column of the buffer_row (out of screen) to store a 1 if the word wrap was forced (0 is written when initializing the row).
- In case of selection on several lines, we add a \n only if the value of the last column is 0.

### DIFF
--- a/src/terminal/buffer.c
+++ b/src/terminal/buffer.c
@@ -48,6 +48,7 @@ guac_terminal_buffer* guac_terminal_buffer_alloc(int rows, guac_terminal_char* d
         /* Allocate row  */
         row->available = 256;
         row->length = 0;
+        row->wrapped_row = false;
         row->characters = guac_mem_alloc(sizeof(guac_terminal_char), row->available);
 
         /* Next row */
@@ -166,6 +167,10 @@ void guac_terminal_buffer_copy_rows(guac_terminal_buffer* buffer,
         /* Copy data */
         memcpy(dst_row->characters, src_row->characters, sizeof(guac_terminal_char) * src_row->length);
         dst_row->length = src_row->length;
+        dst_row->wrapped_row = src_row->wrapped_row;
+
+        /* Reset src wrapped_row */
+        src_row->wrapped_row = false;
 
         /* Next current_row */
         current_row += step;
--- a/src/terminal/select.c
+++ b/src/terminal/select.c
@@ -371,17 +371,31 @@ void guac_terminal_select_end(guac_terminal* terminal) {
     /* Otherwise, copy multiple rows */
     else {
 
-        /* Store first row */
+        /* Get the first selected row */
+        guac_terminal_buffer_row* buffer_row = guac_terminal_buffer_get_row(terminal->buffer, start_row, 0);
+
+        /* Store first row from start_col to last available col */
         guac_terminal_clipboard_append_row(terminal, start_row, start_col, -1);
 
         /* Store all middle rows */
         for (int row = start_row + 1; row < end_row; row++) {
-            guac_common_clipboard_append(terminal->clipboard, "\n", 1);
+
+            /* Add a new line only if the line was not wrapped */
+            if (buffer_row->wrapped_row == false)
+                guac_common_clipboard_append(terminal->clipboard, "\n", 1);
+
+            /* Store middle row */
             guac_terminal_clipboard_append_row(terminal, row, 0, -1);
+
+            /* Get next buffer row */
+            buffer_row = guac_terminal_buffer_get_row(terminal->buffer, row, 0);
         }
 
-        /* Store last row */
-        guac_common_clipboard_append(terminal->clipboard, "\n", 1);
+        /* Add a new line only if the line was not wrapped */
+        if (buffer_row->wrapped_row == false)
+            guac_common_clipboard_append(terminal->clipboard, "\n", 1);
+
+        /* Store last row from col 0 to end_col */
         guac_terminal_clipboard_append_row(terminal, end_row, 0, end_col);
 
     }
--- a/src/terminal/terminal-handlers.c
+++ b/src/terminal/terminal-handlers.c
@@ -57,8 +57,16 @@
  *
  * @param term
  *     The guac_terminal whose cursor should be advanced to the next row.
+ *
+ * @param force_wrap
+ *     True if the line wrap was forced, false otherwise
  */
-static void guac_terminal_linefeed(guac_terminal* term) {
+static void guac_terminal_linefeed(guac_terminal* term, bool force_wrap) {
+
+    /* Assign in wrapped_row: 1 to avoid \n in clipboard or 0 to add \n */
+    guac_terminal_buffer_row* buffer_row = 
+            guac_terminal_buffer_get_row(term->buffer, term->cursor_row, 0);
+    buffer_row->wrapped_row = force_wrap;
 
     /* Scroll up if necessary */
     if (term->cursor_row == term->scroll_end)
@@ -221,7 +229,7 @@ int guac_terminal_echo(guac_terminal* term, unsigned char c) {
         case 0x0C: /* FF */
 
             /* Advance to next row */
-            guac_terminal_linefeed(term);
+            guac_terminal_linefeed(term, false);
 
             /* If automatic carriage return, fall through to CR handler */
             if (!term->automatic_carriage_return)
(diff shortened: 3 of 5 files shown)


necouchman (commented):
I'm wondering if this entire section, including the `if` statement immediately above here, could be simplified to something like this:
```
guac_terminal_select_normalized_range(terminal, &start_row, &start_col, &end_row, &end_col);
int row = start_row;

do {

  // Copy the current row.
  guac_terminal_clipboard_append_row(terminal, row, start_col, end_col);
  guac_terminal_buffer_row* buffer_row = guac_terminal_buffer_get_row(terminal->buffer, row, 0);
  int last_char = buffer_row->characters[terminal->term_width - 1].value;

  if (last_char == 0)
      guac_common_clipboard_append(terminal->clipboard, "\n", 1);

  // Increment row
  row++;
}
while (row <= end_row);
```
?

I haven't actually tested that block of code, but it works in my head :-). I'm not sure if there's some reason that the code needs the extra `if...else` blocks, plus the loop? Maybe there are some corner cases that don't work - like if the last line does or does not end with a line break??

corentin-soriano:
We must keep this existing logic :
- first row = start_col to $
- middle rows = ^ to $
- last row = ^ to end_col

With `guac_terminal_clipboard_append_row(terminal, row, start_col, end_col);` for each row, we risk having a rectangular selection of this type :
image
We need to keep this selection format here :
image


However, I noticed that if we select a row that has exactly the same number of columns as the display, it will be missing a \n :
Example :
image
Content of clipboard :
```
[root@vm01 ~]# echo -e "eeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeee\ntest\ntest2"
eeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeetest
test2
```
This case is very rare but can exist.

corentin-soriano:
I've reworked this PR, it's ready for review!

necouchman (changes requested):
@corentin-soriano The changes look good to me, although @mike-jumper and @jmuehlner are more familiar with the terminal implementation, so I'm curious if either of them has any feedback on it.

My only other request is to rebase this against the `patch` branch, as I think this is more of a bug fix than a feature implementation - having a stray newline added to the clipboard when it is not part of the actual text your copying seems like a (minor) bug.

corentin-soriano:
> @corentin-soriano The changes look good to me, although @mike-jumper and @jmuehlner are more familiar with the terminal implementation, so I'm curious if either of them has any feedback on it.
> 
> My only other request is to rebase this against the `patch` branch, as I think this is more of a bug fix than a feature implementation - having a stray newline added to the clipboard when it is not part of the actual text your copying seems like a (minor) bug.

Rebased from patch branch.

hugaleno:
Just passing here to thank you @corentin-soriano  for this contribution and also for the work on Guacamole-192.
And of course want to thank you all guacamole team to the great work being done.
Keep up the good work!

src/terminal/buffer.c
```diff
@@ -164,7 +164,7 @@ void guac_terminal_buffer_copy_rows(guac_terminal_buffer* buffer,
         guac_terminal_buffer_row* dst_row = guac_terminal_buffer_get_row(buffer, current_row + offset, src_row->length);
 
         /* Copy data */
-        memcpy(dst_row->characters, src_row->characters, sizeof(guac_terminal_char) * src_row->length);
+        memcpy(dst_row->characters, src_row->characters, sizeof(guac_terminal_char) * src_row->length + 1);
```
jmuehlner, line 167: It doesn't look like the storage space in `dst_row->characters` actually got any bigger - you'll need to allocate additional space for this newly added character.

src/terminal/terminal-handlers.c
```diff
@@ -58,7 +58,12 @@
  * @param term
  *     The guac_terminal whose cursor should be advanced to the next row.
  */
-static void guac_terminal_linefeed(guac_terminal* term) {
+static void guac_terminal_linefeed(guac_terminal* term, bool force_wrap) {
+
+    /* Assign in column out of screen: 1 to avoid \n in clipboard or 0 to add \n */
+    guac_terminal_buffer_row* buffer_row =
```
jmuehlner, line 64: The fact that this structure is being modified is not obvious from the function documentation - it should be documented there.

src/terminal/buffer.c
jmuehlner: Or rather, even better, this should probably just be a completely independent flag on the `guac_terminal_buffer_row`, rather than a secret flag shoved onto the end of the character data, since it's not actually character data at all.

corentin-soriano: Yes it's a good idea, I changed it.

src/terminal/terminal-handlers.c
corentin-soriano: Documentation updated.

src/terminal/terminal.c
```diff
@@ -243,7 +243,7 @@ void guac_terminal_reset(guac_terminal* term) {
 
     /* Clear terminal */
     for (row=0; row<term->term_height; row++)
-        guac_terminal_set_columns(term, row, 0, term->term_width, &(term->default_char));
+        guac_terminal_set_columns(term, row, 0, term->term_width-1, &(term->default_char));
```
jmuehlner, line 246: This is wrong now, yeah? Since the flag was moved out of the buffer and into a separate field.

corentin-soriano: Without the -1 here, row->length is larger on the initial screen than on new lines once you get to the bottom of the screen.
Everywhere else in the code initializations are done on term_width-1 (probably because the scrollbar is on top).

jmuehlner: In that case, would you mind adding a comment to the code explaining this?

corentin-soriano: Comment added to the code.

src/terminal/select.c
```diff
@@ -371,17 +371,31 @@ void guac_terminal_select_end(guac_terminal* terminal) {
     /* Otherwise, copy multiple rows */
     else {
 
+        /* Get last char of the row */
+        guac_terminal_buffer_row* buffer_row = guac_terminal_buffer_get_row(terminal->buffer, start_row, 0);
```
necouchman, line 375: I'm a bit confused by the comment, here - wouldn't this particular call get the _first_ character of the given row (index 0 of the row), not the last char of the row?

src/terminal/select.c
```diff
@@ -371,17 +371,31 @@ void guac_terminal_select_end(guac_terminal* terminal) {
...
+
+            /* Add a new line only if the line was not wrapped */
+            if (buffer_row->wrapped_row == false)
+                guac_common_clipboard_append(terminal->clipboard, "\n", 1);
+
+            /* Store middle row */
             guac_terminal_clipboard_append_row(terminal, row, 0, -1);
+
+            /* Update to last char of current row */
+            buffer_row = guac_terminal_buffer_get_row(terminal->buffer, row, 0);
```
necouchman, line 391: As above, how is this getting the last char of the current row? It seems to be to be getting the first char?

src/terminal/select.c
corentin-soriano: Yes, there was a small inconsistency since we retrieved a buffer row and not a char.
I pushed the correction.

src/terminal/select.c
corentin-soriano: Same fix.

hugaleno:
@necouchman Is this merge going to be included on 1.6.0 release? I think that would be great

jmuehlner:
> @necouchman, @jmuehlner, have you any other comments on this PR?

It's looking pretty good to me - though I do wonder if it should be targeted at 1.6.0 - it is a bug fix, afterall.

necouchman:
> > @necouchman, @jmuehlner, have you any other comments on this PR?
> 
> It's looking pretty good to me - though I do wonder if it should be targeted at 1.6.0 - it is a bug fix, afterall.

Agreed - @corentin-soriano, if you could rebase against `staging/1.6.0`, I think this belongs in the next release.

corentin-soriano:
Thank you for your feedback!
I changed the base branch to staging/1.6.0.

necouchman (approved):

mike-jumper (approved):

PR #513, opened 2024-04-26
GUACAMOLE-1719: Fix errors in autoconf file that cause strange build behavior

necouchman:
This takes care of a some issues in `configure.ac` that cause some build issues.

I'm still a bit stumped by the original issue, though, that is mentioned in the ticket - when you run `./configure --with-libavcodec=no`, several strange things happen:
* PANGO isn't found
* RDP, SSH, and Telnet don't build

I'm having a very hard time tracking down what's going on - if anyone can pinpoint the issue, I'd be grateful. It's almost as if there's some mis-matched quotes or brackets or something that are triggered specifically in the case of libavcodec being disabled. Interestingly, if I removed the libavcodec section from `configure.ac`, the behavior follows to the next block (libavformat).

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -37,6 +37,9 @@ AC_PROG_CC
 AC_PROG_CC_C99
 AC_PROG_LIBTOOL
 
+# Initialize pkg-config support
+PKG_PROG_PKG_CONFIG()
+
 # Headers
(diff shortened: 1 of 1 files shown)


necouchman:
Okay, thanks to some help from Libor on the Jira issue, I think I've got the autoconf issues taken care of and ready to merge.

PR #515, opened 2024-04-27
Create A Pull request

CooperDActor:
I made an easy installer you just have to do this

`cd apache-guacamole`

`cd depend-on`

`bash Bash.sh`

I couldn't create a jiva ticket too hard for a 12yo

thanks Cooper D'Andilly 12yo


Depend-on/Bash.sh
```diff
@@ -0,0 +1,221 @@
...
+                            ;;
+                        *)
+                            echo "Installation cancelled."
+                            ;;
+                    esac
+                else
+                    echo "Neither apt-get nor dpkg found. Unsupported package manager."
+                fi
+                ;;
+            *)
```
necouchman, line 147: I don't think making the assumption that anything that isn't Ubuntu or EL is automatically SuSE is the right thing to do. While EL-based, Debian-based, and SuSE-based covers most available distributions, there are others out there (like Alpine, which is what we use for the Docker image) and other ways of managing packages.

If you/we don't want to build out and maintain package commands for each of these, that's fine, but then the `*)` should throw an error, not assume that everything not EL or Ubuntu-based is SuSE.

Depend-on/Bash.sh
```diff
@@ -0,0 +1,221 @@
...
+    else
+        echo "Unsupported operating system."
+    fi
+}
+
+identify_and_run
+
+
+#done
+#Rick Astley
```
necouchman, line 221: We try not Rick-roll people as part of the source code ;-).

Depend-on/Bash.sh
```diff
@@ -0,0 +1,221 @@
+#!/bin/bash
+
+identify_and_run() {
+    os=$(uname)
+    if [ "$os" = "Linux" ]; then
+        distro=$(lsb_release -si)
+        case "$distro" in
+            CentOS|RedHatEnterpriseServer)
```
necouchman, line 8: What about:
* Rocky
* Oracle
* Fedora
* Scientific Linux

I'm sure I'm missing others.

Depend-on/Bash.sh
```diff
@@ -0,0 +1,221 @@
...
+                            ;;
+                        *)
+                            echo "Installation cancelled."
+                            ;;
+                    esac
+                else
+                    echo "Neither dnf nor yum found. Unsupported package manager."
+                fi
+                ;;
+            Ubuntu|Debian)
```
necouchman, line 83: Xubuntu, Lubuntu. I'm not as familiar with Debian-based distros, but there may be others.

necouchman (changes requested):
@CooperDActor :
Thanks, again, for jumping in and contributing.

You will definitely need a Jira ticket for this. Creating an account and ticket is pretty easy:
* Go to our Jira page (https://issues.apache.org/jira/projects/GUACAMOLE/issues).
* At the top of the page, click the "Self serve sign-up page" link and request an account.
* Once your account is created, log in and create the Jira issue.
* Once you've created the issue, update the pull request and commit messages with the Jira issue (GUACAMOLE-XXXX:).

Beyond that, I have some comments below for the actual script. And, in general, I still have a couple of concerns about this:
* I'm not crazy about the name `Depend-on/Bash.sh` - first, IMHO, everything should be lower-case, and, second, I'd much prefer a directory called "contrib" or something like that if we're going to start collecting items like this. I can imagine that there may be other things, like RPM spec files, that would fit into this category of items that people might want to work on and contribute to the project, and I'd rather create a broader place to catch that information.
* You'll also need to add the Apache 2.0 license header to the top of the shell script - see https://github.com/apache/guacamole-server/blob/main/src/guacd-docker/bin/build-all.sh as an example.
* The title of the pull request should be descriptive of what the changes are you are trying to make - for example, something like "Create a shell script to help manage build dependencies"
* The commit messages should also be descriptive and helpful in determining what has changed. See https://www.codelord.net/2015/03/16/bad-commit-messages-hall-of-shame/.
* Finally, 7 commits for a single shell script seems a bit excessive - you might consider squashing down to one or two commits.

Depend-on/Bash.sh
```diff
@@ -0,0 +1,221 @@
+#!/bin/bash
+
+identify_and_run() {
+    os=$(uname)
+    if [ "$os" = "Linux" ]; then
+        distro=$(lsb_release -si)
```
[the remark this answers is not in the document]
CooperDActor: well they're screwed for the moment in future I will test with all operating systems but maybe now we could keep it like that?

Depend-on/Bash.sh
```diff
@@ -0,0 +1,221 @@
...
+                    echo "libssh2-devel"
+                    echo "libtelnet-devel"
+                    echo "libvncserver-devel"
+                    echo "libwebsockets-devel"
+                    echo "pulseaudio-libs-devel"
+                    echo "openssl-devel"
+                    echo "libvorbis-devel"
+                    echo "libwebp-devel"
+
+                    read -p "Do you understand? (yes/no): " choice
```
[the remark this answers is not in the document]
CooperDActor: it makes apache look more professional and cool from a kids perspective

Depend-on/Bash.sh
```diff
@@ -0,0 +1,221 @@
...
+                            sudo dnf install -y freerdp2-devel
+                            sudo dnf install -y pango-devel
+                            sudo dnf install -y libssh2-devel
+                            sudo dnf install -y libtelnet-devel
+                            sudo dnf install -y libvncserver-devel
+                            sudo dnf install -y libwebsockets-devel
+                            sudo dnf install -y pulseaudio-libs-devel
+                            sudo dnf install -y openssl-devel
+                            sudo dnf install -y libvorbis-devel
+                            sudo dnf install -y libwebp-devel
```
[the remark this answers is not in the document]
CooperDActor: then I can say to my mates at school I made a 100-200 line code script lol 
nah because I am a pain in my OWN arse 

though what I am more scared about is not having the packages on dnf ect ect and

Depend-on/Bash.sh
CooperDActor: thats just for the moment when I become more familiar with them I'll add them

(im only 12)

Depend-on/Bash.sh
CooperDActor: thats the Debian tag that should pick it up but if not I might add a redundnt ask so for example if its not found you type in the package manager and thats what im thinking

Depend-on/Bash.sh
CooperDActor: I will put his video there nah when you say no I will follow that to the T

Depend-on/Bash.sh
CooperDActor: any thing simpler for my 12yo brain to understand?


what I believe your talking about is the pkg managers and not working if it errors all of the time 

see the redundancy talk of my idea

CooperDActor:
> @CooperDActor : Thanks, again, for jumping in and contributing.
> 
> You will definitely need a Jira ticket for this. Creating an account and ticket is pretty easy:
> 
> * Go to our Jira page (https://issues.apache.org/jira/projects/GUACAMOLE/issues).---- under issues get me a t-issue okay if you dont like that joke u have a cold heart but really under issues or where?? thats why I wasn't doing it
> * At the top of the page, click the "Self serve sign-up page" link and request an account.
> * Once your account is created, log in and create the Jira issue.
> * Once you've created the issue, update the pull request and commit messages with the Jira issue (GUACAMOLE-XXXX:).
> 
> Beyond that, I have some comments below for the actual script. And, in general, I still have a couple of concerns about this:
> 
> * I'm not crazy about the name `Depend-on/Bash.sh` - first, IMHO, everything should be lower-case, and, second, I'd much prefer a directory called "contrib" or something like that if we're going to start collecting items like this. I can imagine that there may be other things, like RPM spec files, that would fit into this category of items that people might want to work on and contribute to the project, and I'd rather create a broader place to catch that information. --- this is not an insult but think of better names here's how I think of it contrib wait contribute maybe someone put malicious code In it NUP NUP NUP just like my mother   I could put dependencies but autocorrect saved me there
> * You'll also need to add the Apache 2.0 license header to the top of the shell script - see https://github.com/apache/guacamole-server/blob/main/src/guacd-docker/bin/build-all.sh as an example.---- okay
> * The title of the pull request should be descriptive of what the changes are you are trying to make - for example, something like "Create a shell script to help manage build dependencies" -- next time or if I can I will edit it
> * The commit messages should also be descriptive and helpful in determining what has changed. See https://www.codelord.net/2015/03/16/bad-commit-messages-hall-of-shame/.-- will do from now on
> * Finally, 7 commits for a single shell script seems a bit excessive - you might consider squashing down to one or two commits. ----- I made it simple first time they wanted more if you can as I dont know how. you could squash it down (adding you to the contrib on my guac folder so you can edit) if you need/want

Depend-on/Bash.sh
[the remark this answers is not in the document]
necouchman: Even if the plan is to stick with `lsb_release` for this detection, this just highlights the other comment I made - that the default case for the subsequent `case` statement should not be to assume that SuSE is in use, but to provide an error that tells the user something useful: "Your distribution was not detected, please make sure your distribution supports the lsb_release command" - or something like that.

Depend-on/Bash.sh
necouchman: My preference would be to create a "SuSE" section of the `case` statement above this `*` (default) handler, and put the SuSE commands in that section, and then use the `*` area to provide feedback that the distribution is either undetected or unsupported. Roughly, something like this:
```
case "distro" in
    CentOS|RedHatEnterpriseServer|Rocky)
        // Run EL commands here
    Debian|Ubuntu|Xubuntu)
        // Run Debian commands here
    SuSE)
        // Run SuSE commands here
    *)
        // Throw error output and exit - distro not detected, not supported, etc.
esac
```

PR #519, opened 2024-05-13
IP addresses can be input as variables

sunjiankun126:
The IP address can be used as a variable input to make the image more flexible


necouchman:
@sunjiankun126 - Thank you for the contribution. In order to consider/merge this request, you'll also need to request a Jira account (https://issues.apache.org/jira/projects/GUACAMOLE), put in a Jira issue for this change, and then update the pull request and commit messages with the Jira issue. You can look at the other pull requests and commits for examples of what that looks like.

See the following page for the full run-down on contributing: https://guacamole.apache.org/open-source/.

sunjiankun126:
Okay, I'm going to apply for a Jira account

PR #521, opened 2024-06-05
GUACAMOLE-1633: Add support of alternate screen buffer

corentin-soriano:
Add support of alternate screen buffer for tools like less/vi/man/...

Before:
image

After:
image
image
image

### DIFF
--- a/src/terminal/terminal-handlers.c
+++ b/src/terminal/terminal-handlers.c
@@ -51,6 +51,11 @@
 #define GUAC_TERMINAL_OK          "\x1B[0n"
 
 /**
+ * Alternative buffer CSI sequence.
+ */
+#define GUAC_TERMINAL_ALT_BUFFER   1049
+
+/**
  * Advances the cursor to the next row, scrolling if the cursor would otherwise
  * leave the scrolling region. If the cursor is already outside the scrolling
  * region, the cursor is prevented from leaving the terminal bounds.
@@ -885,6 +890,10 @@ int guac_terminal_csi(guac_terminal* term, unsigned char c) {
                 if (flag != NULL)
                     *flag = true;
 
+                /* Open alternate screen buffer */
+                if (argv[0] == GUAC_TERMINAL_ALT_BUFFER)
+                    guac_terminal_switch_buffers(term, true);
+
                 break;
 
             /* l: Reset Mode */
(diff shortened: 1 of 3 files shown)


src/terminal/terminal-handlers.c
```diff
@@ -885,6 +885,10 @@ int guac_terminal_csi(guac_terminal* term, unsigned char c) {
                 if (flag != NULL)
                     *flag = true;
 
+                /* Open alternate screen buffer */
+                if (argv[0] == 1049)
```
necouchman, line 889: Might be worth `#define` a constant for this value...

hugaleno:
This is some how related to [GUACAMOLE-1256](https://issues.apache.org/jira/browse/GUACAMOLE-1256)?

corentin-soriano:
> This is some how related to [GUACAMOLE-1256](https://issues.apache.org/jira/browse/GUACAMOLE-1256)?

This pull request is not related to this ticket.

This PR only avoid this display on editor close:
image

hugaleno:
I was wondering if this alternate screen buffer would also solve that other problem, but thank you for clarifying

src/terminal/terminal-handlers.c
corentin-soriano: It's done!

src/terminal/terminal/terminal-priv.h
```diff
@@ -269,6 +269,11 @@ struct guac_terminal {
      */
     int visible_cursor_col;
 
+    /**
+     * Backup of cursor position when using alternate buffer.
+     */
+    int cursor_row_alt, cursor_col_alt, visible_cursor_row_alt, visible_cursor_col_alt;
+
```
necouchman, line 276: As the matching primary members are declared individually, I think this should probably follow the same convention and declare each of these individually.

necouchman (changes requested):
One change, other than that looks okay to me.

Only other question is - does it make sense to allocate the alternate buffer only if/when it's required? I'm good either way, just curious...

corentin-soriano:
> One change, other than that looks okay to me.
> 
> Only other question is - does it make sense to allocate the alternate buffer only if/when it's required? I'm good either way, just curious...

Indeed, it helps to limit memory usage. I made the modification, no impact on performance.
I took the opportunity to declare it without scrollback since it's unnecessary on the alternate buffer.

necouchman (approved):
Looks fine to me.

PR #525, opened 2024-06-11
GUACAMOLE-377: Add libguac "guac_display" API for easy and efficient rendering.

mike-jumper:
This change builds off the internal `guac_common_display` and `guac_common_surface` structures to create a new public `guac_display` structure and surrounding API. The approach within `guac_display` improves on that of `guac_common_surface` in several ways:

* It makes use of a pool of worker threads to parallelize the encoding process. The appropriate number of worker threads is detected automatically when the `guac_display` is created.
* **Changes due to scrolling, etc. are automatically detected in real time** using a combination of a hash table, a hash function designed to avoid inefficient memory traversal, and a 2D variant of [Rabin-Karp](https://en.m.wikipedia.org/wiki/Rabin%E2%80%93Karp_algorithm). Such drawing operations are automatically transformed into copy operations.
* Changes that can be represented as a simple `rect` followed by `cfill` are automatically recognized and sent out as such.
* Dirty rectangles covering the changes made to the display are automatically broken up and tightened to more efficiently represent the actual changes made, regardless of how large the dirty rectangle submitted to `guac_display` may have been.
* Adjustment for client-side processing lag is handled automatically.
* Encoding of images for a previous frame happens asynchronously while the next frame is being assembled, reducing latency.
* Any number of additional frames may be assembled while waiting for the previous frame to encode. If a previous frame is still being encoded, further frames are combined together into a pending frame that will be flushed when encoding is complete.

With these changes, you can now throw virtually anything at the display and it will magically get decomposed into an efficient combination of copies, draws, and rectangles, even if the information regarding the nature of those updates is unavailable (ie: RDPGFX and SPICE).

Metrics covering rendering and optimization performance of `guac_display` are logged at the "trace" level.

I'm opening this as a draft for now, as these changes are currently incomplete:

* Only VNC has been migrated to the new API, and then only partially.
* I suspect that the internals of the terminal emulator can be simplified, given that scrolling is now automatically detected.
* I have not yet added a mechanism for callers to "hint" that a change is due to a copy (this will be necessary for handling RDP's bitmap cache).

Once the above is finished, I'll delete the old `guac_common_display`, `guac_common_surface`, etc. and this will be ready.


mike-jumper:
I've been using the following patch to observe the impact these changes:

https://gist.github.com/mike-jumper/0afd41c9fbcbc764f719890fd0dd4e3b

With the above patch in place, you can set `Guacamole.Display.DEBUG` to `true` to enable a debug mode in which draw operations to the Guacamole display are highlighted in colored rectangles, where the color varies by the type of operation:

* **RED:** Draw of compressed image data (the most expensive operation)
* **BLUE:** Copy of existing image data (ie: an optimized scroll, restoring data from a cache - much cheaper than encoding, sending, and decoding image data)
* **GREEN:** Solid-color rectangular draw (a combination of `rect` and `cfill` - very cheap).

### Behavior of `guac_common_surface`

Here, everything is red. Dirty rects are generally nicely split and tightened around the regions actually changed, but that's about as deep as the processing goes. Without information from the remote desktop server explicitly stating that a particular update is a copy, we can't detect that a copy would be better.

https://github.com/apache/guacamole-server/assets/4632905/b0792ae3-f1a7-4bc6-8ce1-932832c32a21

### Behavior of `guac_display`

Now, things get much more colorful. Things that can be represented more efficiently as copies or rectangles are automatically optimized. This includes cases when the nature of those updates would prevent even the remote desktop server from knowing that a copy has occurred.

https://github.com/apache/guacamole-server/assets/4632905/6fb37e05-37c3-4eef-ba42-88f4c89442a7

.gitignore
```diff
@@ -48,3 +48,7 @@ doc/*/doxygen-output
 
 # IDE metadata
 nbproject/
+
+# Compilation database, as may be generated by tools like Bear
```
jmuehlner, line 52: Huh, had never heard of Bear before. TIL

mike-jumper: Another case where GitHub's selection of only 8 emojis for reactions feels woefully inadequate. :bear:

jmuehlner:
```
   ,__ ., __, ,,,,
   '--/,,\--'\*\%\*\
     //  \\\'\'%.\'%\
      '..'//'%\.\%/\\'.^
         \\'/'/%''/\'
           ||     ||
           "      "
```

src/libguac/display-cursor.c
```diff
@@ -0,0 +1,102 @@
...
+            break;
+
+    }
+
+    /* Resize cursor to fit requested icon */
+    guac_display_layer* cursor_layer = guac_display_cursor(display);
+    guac_display_layer_resize(cursor_layer, cursor->width, cursor->height);
+
+    /* Copy over graphical content of cursor icon ... */
+
```
jmuehlner, line 73: Did you mean to have a newline here?

src/libguac/display-cursor.c
```diff
@@ -0,0 +1,102 @@
...
+        memcpy(dst_cursor_row, src_cursor_row, row_length);
+        src_cursor_row += cursor->stride;
+        dst_cursor_row += context->stride;
+    }
+
+    /* ... and cursor hotspot */
+    guac_display_set_cursor_hotspot(display, cursor->hotspot_x, cursor->hotspot_y);
+
+    /* Update to cursor icon is now complete - notify display */
+
```
jmuehlner, line 90: Or here?

src/libguac/display-cursor.c
jmuehlner: Based on what I'm seeing elsewhere, I'm guessing yes.

src/libguac/tests/rect/align.c
```diff
@@ -0,0 +1,57 @@
...
+
+    /* Adjustment with moving of rect */
+    guac_rect_init(&rect, 75, 75, 25, 25);
+    guac_rect_align(&rect, cell_size);
+    CU_ASSERT_EQUAL(64, rect.left);
+    CU_ASSERT_EQUAL(64, rect.top);
+    CU_ASSERT_EQUAL(112, rect.right);
+    CU_ASSERT_EQUAL(112, rect.bottom);
+
+    guac_rect_init(&rect, -5, -5, 25, 25);
```
jmuehlner, line 49: Probably worth mentioning what this test is doing as well.

src/libguac/tests/rect/intersects.c
```diff
@@ -0,0 +1,96 @@
...
+    res = guac_rect_intersects(&rect, &min);
+    CU_ASSERT_TRUE(res);
+
+    /* Rectangle intersection - complete
+     * rect intersects along UL but inside */
+    guac_rect_init(&rect, 10, 10, 5, 5);
+    res = guac_rect_intersects(&rect, &min);
+    CU_ASSERT_TRUE(res);
+
+    /* Rectangle intersection - partial
```
jmuehlner, line 71: This terminoloy here about a partial rectangle intersection is sort of confusing - it looks like these rectangles are _touching_ at an edge, but not actually overlapping, right?

Maybe consider rewording to make it clearer what these comments mean.

src/protocols/rdp/bitmap.c
```diff
@@ -79,54 +82,53 @@ BOOL guac_rdp_bitmap_paint(rdpContext* context, rdpBitmap* bitmap) {
...
-        cairo_surface_destroy(image);
+    /* If cached, retrieve from cache */
+    if (buffer != NULL) {
+        guac_display_layer_raw_context* src_context = guac_display_layer_open_raw(buffer);
+        guac_display_layer_raw_context_put(dst_context, &dst_rect, src_context->buffer, src_context->stride);
+    }
 
+    /* Otherwise, draw with stored image data */
+    else if (bitmap->data != NULL) {
+        guac_display_layer_raw_context_put(dst_context, &dst_rect, bitmap->data, 4 * bitmap->width);
```
jmuehlner, line 111: Why 4?

src/protocols/rdp/rdp.c
```diff
@@ -143,6 +142,7 @@ BOOL rdp_freerdp_pre_connect(freerdp* instance) {
     if (!gdi_init(instance, guac_rdp_get_native_pixel_format(FALSE)))
         return FALSE;
 
+#if 0
```
jmuehlner, line 145: Looks like I should wait for this stuff to be finalized before reviewing the protocols

src/protocols/rdp/bitmap.c
mike-jumper: Each pixel within the image data at `bitmap->data` is 32-bit.

jmuehlner: Probably worth commenting on

src/protocols/rdp/rdp.c
mike-jumper: RDP and VNC should now be complete and no longer contain any of these `#if 0` blocks.

I'm now just finishing up the last terminal emulator bits and retesting.

mike-jumper:
**_FINALLY_** ready for review. :relieved:

necouchman:
> **_FINALLY_** ready for review. 😌

Whew, just almost 10K lines of code to review. Shouldn't take more than a few minutes ;-).

src/common-ssh/key.c
```diff
@@ -132,18 +132,25 @@ guac_common_ssh_key* guac_common_ssh_key_alloc(char* data, int length,
...
+     * been fixed, but as of this writing the fix has not yet been released.
+     * Below, we add our own null terminator to ensure that such calls to
+     * strlen() will work without issue. We can remove this workaround once
+     * copies of libssh2 that use strlen() on key data are not in common use. */
+
     /* Copy private key to structure */
     key->private_key_length = length;
-    key->private_key = guac_mem_alloc(length);
+    key->private_key = guac_mem_alloc(guac_mem_ckd_add_or_die(length, 1)); /* Extra byte added here for null terminator (see above) */
     memcpy(key->private_key, data, length);
```
jmuehlner, line 152: Why not use `guac_strlcpy()` here?

mike-jumper: Here we're copying binary data, not a string. The null terminator is added purely because libssh2 may incorrectly call `strlen()` on that data even though it's not a string.

src/libguac/guacamole/rect.h
```diff
@@ -0,0 +1,256 @@
...
+ * @return
+ *     The memory address of given rectangle within the given buffer.
+ */
+#define GUAC_RECT_MUTABLE_BUFFER(rect, buffer, stride, bpp) ((void*) (        \
+            ((unsigned char*) (buffer))                                       \
+                + guac_mem_ckd_mul_or_die((rect).top, stride)                 \
+                + guac_mem_ckd_mul_or_die((rect).left, bpp)))
+
+/**
+ * Returns the memory address of given rectangle within the given immutable
```
jmuehlner, line 61: Should this be "a given rectangle" or "the given rectangle"? Reads a bit odd as is.

src/terminal/buffer.c
```diff
@@ -19,13 +19,85 @@
...
 
+#include <guacamole/assert.h>
 #include <guacamole/mem.h>
 
+#include <stdbool.h>
 #include <stdlib.h>
 #include <string.h>
 
-guac_terminal_buffer* guac_terminal_buffer_alloc(int rows, guac_terminal_char* default_character) {
+#define GUAC_TERMINAL_BUFFER_ROW_MIN_SIZE 256
```
jmuehlner, line 31: This constant should be documented.

src/libguac/guacamole/rect.h
mike-jumper: Yep, I'll fix that.

src/terminal/buffer.c
mike-jumper: Oopsies.

src/libguac/guacamole/rect.h
mike-jumper: Ooof, I say "of given rectangle" nearly everywhere... I'll fix all the occurrences.

mike-jumper: Fixed via rebase.

src/terminal/buffer.c
mike-jumper: Fixed via rebase.

src/terminal/terminal/terminal-priv.h
```diff
@@ -77,32 +91,20 @@ struct guac_terminal {
      * a download of a given remote file.
      */
     guac_terminal_file_download_handler* file_download_handler;
-
+
```
jmuehlner, line 94: It doesn't look like this new whitespace belongs here.

src/terminal/terminal/terminal-priv.h
```diff
@@ -325,23 +290,31 @@ struct guac_terminal {
...
+     * operations is determined by the current value of current_buffer.
      */
-    guac_terminal_buffer* buffer;
+    guac_terminal_buffer* normal_buffer;
 
     /**
-     * Alternate buffer.
+     * The non-default, "alternate" buffer containing all characters that should
+     * be displayed within the terminal emulator while not using the normal
+     * buffer. Unless switched to the alternate buffer, all terminal operations
```
jmuehlner, line 304: Copypasta? "Unless switched to the alternate buffer, all terminal operations will involve this buffer"/

jmuehlner:
Ok, read through everything, and it all looks reasonable except for a few minor formatting / wording issues. I'm looking forward to using this!

aleitner:
Looks good on my initial pass, and code ran without issues as well

mike-jumper:
I'm seeing some performance regressions in the terminal emulator, which makes some sense given that `guac_display` is very pixel-oriented.

I'll dig a bit deeper, but if nothing pops up as being a clear bottleneck specific to the terminal, I think it's probably best to roll the terminal emulator back to using its own change tracking until such future time as `guac_display` can support text-based displays without having to consider each pixel of the characters in the text.

mike-jumper:
OK - I think I have a handle on the performance regression. The issue seems to have been mainly a memory bottleneck, with the solution being:

1. Use `memcpy()` where possible
2. Provide to mechanism to point a `guac_display_layer` at an external buffer (to avoid having to copy from the framebuffer of a client library)

I'm putting the above in place, plus a partial revert of the terminal emulator changes (migrating the display back to the character-based display).

jmuehlner:
In testing this, I'm seeing some visual artifacts where windows dragged around in VNC have bits of the window content spread around, e.g.
image

src/libguac/guacamole/rect.h
```diff
@@ -0,0 +1,256 @@
...
+ *
+ * @param rect
+ *     The rectangle to collapse.
+ *
+ * @param max
+ *     The maximum area in which the given rectangle can exist.
+ */
+void guac_rect_constrain(guac_rect* rect, const guac_rect* max);
+
+void guac_rect_shrink(guac_rect* rect, int max_width, int max_height);
```
jmuehlner, line 206: Missing doc here

mike-jumper:
> In testing this, I'm seeing some visual artifacts where windows dragged around in VNC have bits of the window content spread around, e.g. image

This should now be resolved - it was caused by the `GotCopyRect()` function having been overridden. Previously, this was no issue because we used our own buffer. Now that we're reusing libvncclient's buffer, the original `GotCopyRect()` needs to be called or that buffer won't be updated with respect to received `CopyRect` updates.

jmuehlner:
> > In testing this, I'm seeing some visual artifacts where windows dragged around in VNC have bits of the window content spread around, e.g. image
> 
> This should now be resolved - it was caused by the `GotCopyRect()` function having been overridden. Previously, this was no issue because we used our own buffer. Now that we're reusing libvncclient's buffer, the original `GotCopyRect()` needs to be called or that buffer won't be updated with respect to received `CopyRect` updates.

Nice - I can confirm that this issue looks to be fixed.

src/libguac/client.c
```diff
@@ -132,7 +132,8 @@ guac_stream* guac_client_alloc_stream(guac_client* client) {
         return NULL;
 
     /* Allocate stream */
-    stream_index = guac_pool_next_int(client->__stream_pool);
+    stream_index = guac_pool_next_int_below_or_die(client->__stream_pool,
```
jmuehlner, line 135: Right above this we do a non-threadsafe check against `GUAC_CLIENT_MAX_STREAMS`, which will cause the function to return `NULL` - do we still need that? Does the new check not correctly handle this?

And if the documented behavior of this function is to return `NULL` when out of streams, why are we dying here instead?

It seems like we'd want to maintain that `NULL` return for backwards compatibility, but also I'd point out that none of the callers of these `alloc_stream()` functions look to actually check if the return value is non-NULL before using it...

corentin-soriano (commented):
I haven't had any problems with RDP/VNC.
In terminal, cat a large file is very slow and can in some cases disconnect me from guacamole and the display is not restored when exiting the alternate buffer.

src/terminal/terminal-handlers.c
```diff
@@ -918,31 +1012,87 @@ int guac_terminal_csi(guac_terminal* term, unsigned char c) {
...
+                                -guac_terminal_get_available_scroll(term), 0);
+
+                    }
+
+                    /* Clear normal buffer only if we were previously using
+                     * the alternate buffer */
+                    if (argv[0] == GUAC_TERMINAL_DECRST_USE_NORMAL_BUFFER_AND_CLEAR) {
+                        guac_terminal_clear_range(term,
+                                0, 0, term->term_height - 1, term->term_width - 1);
+                    }
```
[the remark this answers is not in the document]
mike-jumper: > I think we should check if we receive the code `GUAC_TERMINAL_DECRST_USE_NORMAL_BUFFER_AND_RESTORE_CURSOR`.

We typically only receive that if `TERM` is set to `xterm-256color` within the SSH session.

> However guac_terminal_clear_range completely clears the screen instead of restoring the contents of the normal buffer so we have a black screen with only the cursor row shown

I believe that's required behavior for this particular DECRST code (1047), which is defined as clearing the normal buffer after switching to it. From https://www.xfree86.org/current/ctlseqs.html:

> P s = 1 0 4 7 → Use Normal Screen Buffer, clearing screen first if in the Alternate Screen (unless disabled by the titeInhibit resource)

jmuehlner:
I've fully reviewed this and don't see any issues with the code as is - does anyone object if I merge this? @mike-jumper did you have any more that you wanted to add?

necouchman:
> I've fully reviewed this and don't see any issues with the code as is - does anyone object if I merge this? @mike-jumper did you have any more that you wanted to add?

There are a couple of unresolved comments above - one about a missing document block for a function, and one about a non-thread-safe check. Have these been resolved, even though Github is not marking them as outdated?

mike-jumper:
@necouchman I'll take a look and make sure I've not missed anything.

@jmuehlner I just wrote a threaded display render loop that looks like it may improve things yet further for RDP. If that checks out, I'll push it in a moment.

mike-jumper:
> I haven't had any problems with RDP/VNC. In terminal, cat a large file is very slow and can in some cases disconnect me from guacamole and the display is not restored when exiting the alternate buffer.

@corentin-soriano Thanks for checking this. The remaining performance trouble with the terminal seems rooted in GUACAMOLE-1256, so if there isn't an obvious fix, I think we'll just end up rolling back those changes and revisiting next release. The changes here should no longer affect that.

src/libguac/client.c
mike-jumper: > Right above this we do a non-threadsafe check against GUAC_CLIENT_MAX_STREAMS, which will cause the function to return NULL - do we still need that? Does the new check not correctly handle this?

Yes - very good catch. I'll remove the superfluous check.

> And if the documented behavior of this function is to return NULL when out of streams, why are we dying here instead?

Also fair - will fix.

src/libguac/guacamole/rect.h
mike-jumper: Documented via rebase.

src/libguac/client.c
mike-jumper: Both should now be fixed.

mike-jumper:
@jmuehlner The "easy mode" render loop is ready, working, and seems substantially better than the previous render loops. I've updated both VNC and RDP accordingly. If anyone needs something more complex than the provided loop, they can just invoke `guac_display_end_multiple_frames()` on their own as they see fit.

@necouchman As of right this moment, I think I've resolved all comments.

src/libguac/guacamole/rect.h
```diff
@@ -0,0 +1,271 @@
...
+ *     The rectangle to collapse.
+ *
+ * @param max
+ *     The maximum area in which the given rectangle can exist.
+ */
+void guac_rect_constrain(guac_rect* rect, const guac_rect* max);
+
+/**
+ * Reduces the size of the given rectangle such that it does not exceed the
+ * given width and height. The aspect ration of the given rectangle is
```
jmuehlner, line 208: "aspect ration" ... sounds ... tasty?

jmuehlner:
Aside from a couple of minor typos in the comments, this looks good to me.

src/libguac/display-render-thread.c
```diff
@@ -0,0 +1,173 @@
...
+#include "display-priv.h"
+#include "guacamole/client.h"
+#include "guacamole/display.h"
+#include "guacamole/flag.h"
+#include "guacamole/mem.h"
+#include "guacamole/timestamp.h"
+
+/**
+ * The maximum duration of a frame in milliseconds. This ensures we at least
+ * meet a reasonable minimum framerate in the case that the RDP server provides
```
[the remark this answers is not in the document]
mike-jumper: Oops, yep.

src/libguac/guacamole/rect.h
mike-jumper: We have to conserve the aspect!

mike-jumper: Fixed via rebase.

src/libguac/display-render-thread.c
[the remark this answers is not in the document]
mike-jumper: Fixed via rebase.

corentin-soriano:
> > I haven't had any problems with RDP/VNC. In terminal, cat a large file is very slow and can in some cases disconnect me from guacamole and the display is not restored when exiting the alternate buffer.
> 
> 
> 
> @corentin-soriano Thanks for checking this. The remaining performance trouble with the terminal seems rooted in GUACAMOLE-1256, so if there isn't an obvious fix, I think we'll just end up rolling back those changes and revisiting next release. The changes here should no longer affect that.

I misunderstood your comment in the ticket, I thought you had completely replaced the display_flush and that the problem was solved.
In this case, what seems best is to remove it in the scroll_up function while keeping it in scroll_down. The latencies would no longer be present, and the scrolling bug in vi would also be gone.
I will create the PR on Monday.
Thanks for the clarification.

jmuehlner (approved):
Everything looks good to me! Let's light this candle.

PR #527, opened 2024-06-17
Kerberos

cezary-marczak:


src/guacd-docker/krb5.conf
```diff
@@ -0,0 +1,16 @@
...
+    PROCYON.COM = {
+        kdc = cez-private.procyon.com
+        admin_server = cez-private.procyon.com
+        master_kdc = cez-private.procyon.com
+        default_domain = procyon.com
+    }
+
+[domain_realm]
+    .procyon.com = PROCYON.COM
+    procyon.com = PROCYON.COM
```
necouchman, line 16: I think this file should be removed from the commit - this looks like something internal to your configuration. I'm not sure that we should provide a template for this - maybe just document instructions for it.

Dockerfile
```diff
@@ -186,6 +203,7 @@ ARG UID=1000
 ARG GID=10001
 RUN groupadd --gid $GID guacd
 RUN useradd --system --create-home --shell /sbin/nologin --uid $UID --gid $GID guacd
+COPY ./src/guacd-docker/krb5.conf /etc/krb5.conf
```
necouchman, line 206: I'm not sure we should provide this - maybe just document the instructions for it.

necouchman (changes requested):
Aside from changes mentioned below, you need to do the following things before this pull request can be accepted:
* Create a Jira account and open a new Jira tickets for this issue.
* Tag your pull request and commit messages with the ticket name.

See the following links:
https://guacamole.apache.org/open-source/
https://issues.apache.org/jira/browse/GUACAMOLE/

PR #528, opened 2024-06-18
GUACAMOLE-1256: Correct terminal scroll with text editors.

corentin-soriano:
- Add support of 'S' and 'T' CSI sequences (scroll up/down by amount).
- Force display flush when scrolling to avoid conflict when cursor_commit use the set_collumns function which change the operation_type.

### DIFF
--- a/src/terminal/terminal-handlers.c
+++ b/src/terminal/terminal-handlers.c
@@ -835,6 +835,32 @@ int guac_terminal_csi(guac_terminal* term, unsigned char c) {
 
                 break;
 
+            /* S: Scroll Up by amount */
+            case 'S':
+
+                /* Get move amount */
+                amount = argv[0];
+                if (amount == 0) amount = 1;
+
+                /* Scroll up */
+                guac_terminal_scroll_up(term, term->scroll_start,
+                        term->scroll_end, amount);
+
+                break;
+
+            /* T: Scroll Down by amount */
+            case 'T':
+
+                /* Get move amount */
+                amount = argv[0];
+                if (amount == 0) amount = 1;
+
+                /* Scroll Down */
+                guac_terminal_scroll_down(term, term->scroll_start,
+                        term->scroll_end, amount);
+
+                break;
+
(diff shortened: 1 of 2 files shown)


src/terminal/terminal-handlers.c
```diff
@@ -835,6 +835,28 @@ int guac_terminal_csi(guac_terminal* term, unsigned char c) {
...
                 break;
 
+            /* Scroll Up by amount */
+            case 'S':
+
+                /* Multiple linefeed */
+                for (int i = 0; i < argv[0]; i++)
+                    guac_terminal_linefeed(term, false);
+
+                term->char_handler = guac_terminal_echo;
```
necouchman, line 845: Why is this required? I don't see this assignment required in any of the other `case` statements - is there some reason why it needs to be assigned for these two new instances?

corentin-soriano: Yes you are right, it is not useful since it is already assigned at line 1117 (in this branch).
I removed them.

src/terminal/terminal-handlers.c
```diff
@@ -835,6 +835,24 @@ int guac_terminal_csi(guac_terminal* term, unsigned char c) {
 
                 break;
 
+            /* Scroll Up by amount */
+            case 'S':
+
+                /* Multiple linefeed */
+                for (int i = 0; i < argv[0]; i++)
+                    guac_terminal_linefeed(term, false);
```
necouchman, line 843: One other question - above, in the sections for inserting/deleting lines, there are functions `guac_terminal_scroll_up` and `guac_terminal_scroll_down` - is there a reason not to use these, versus using `for` loops you have here?

corentin-soriano: Yes, we can use these functions directly instead of doing a for loop.
I changed and tested again: the result is the same.

hugaleno:
Thank you once again @corentin-soriano. Another amazing contribution

necouchman (approved):

PR #530, opened 2024-07-19
GUACAMOLE-1969: Implement recording-include-clipboard connection parameter

lukasraska:
This PR introduces new `recording-include-clipboard` into all supported protocols that dumps the `clipboard` instruction, along with the relevant stream blobs, into recording.

Few key notes:
- because clipboard can contain sensitive data, it's opt-in in the same way as `recording-include-keys` and administrators need to explicitly set the parameter to `true`
- as the recording components are responsible for handling the recording dumps, `guac_recording_create` includes the new parameter (i've opted for making it the last argument, but let me know if I should switch it to be next to the `include_keys`)
- the keyboard dumps are hooked into user clipboard handler, so should only handle changes from user side and not from the server side
- as clipboard instruction merely creates stream, even the simplest paste is essentially at least 3 instructions (clipboard, 1+ blobs, end), which can be seen like this in some testing recording
```
$ sed 's/;/\n/g' recording-clipboard | grep clipboard -A 2
9.clipboard,1.0,10.text/plain
4.blob,1.0,20.UHVsbCBjb21wbGV0ZQ==
3.end,1.0
```

Once this PR is approved from technical perspective, I will create respective PRs for guacamole-client and guacamole-manual to accomodate for both documentation change as well the client side option

### DIFF
--- a/src/protocols/kubernetes/settings.c
+++ b/src/protocols/kubernetes/settings.c
@@ -51,6 +51,7 @@ const char* GUAC_KUBERNETES_CLIENT_ARGS[] = {
     "recording-exclude-output",
     "recording-exclude-mouse",
     "recording-include-keys",
+    "recording-include-clipboard",
     "create-recording-path",
     "recording-write-existing",
     "read-only",
@@ -212,6 +213,16 @@ enum KUBERNETES_ARGS_IDX {
      * as passwords, credit card numbers, etc.
      */
     IDX_RECORDING_INCLUDE_KEYS,
+
+    /**
+     * Whether clipboard paste data should be included in the session recording.
+     * Clipboard data is NOT included by default within the recording,
+     * as doing so has privacy and security implications. Including clipboard data
+     * may be necessary in certain auditing contexts, but should only be done
+     * with caution. Clipboard data can easily contain sensitive information, such
+     * as passwords, credit card numbers, etc.
+     */
+    IDX_RECORDING_INCLUDE_CLIPBOARD,
 
     /**
      * Whether the specified screen recording path should automatically be
@@ -408,6 +419,11 @@ guac_kubernetes_settings* guac_kubernetes_parse_args(guac_user* user,
     settings->recording_include_keys =
         guac_user_parse_args_boolean(user, GUAC_KUBERNETES_CLIENT_ARGS, argv,
                 IDX_RECORDING_INCLUDE_KEYS, false);
+
+    /* Parse clipboard inclusion flag */
+    settings->recording_include_clipboard =
+        guac_user_parse_args_boolean(user, GUAC_KUBERNETES_CLIENT_ARGS, argv,
+                IDX_RECORDING_INCLUDE_CLIPBOARD, false);
 
     /* Parse path creation flag */
     settings->create_recording_path =
--- a/src/protocols/kubernetes/settings.h
+++ b/src/protocols/kubernetes/settings.h
@@ -246,6 +246,16 @@ typedef struct guac_kubernetes_settings {
     bool recording_include_keys;
 
     /**
+     * Whether clipboard paste data should be included in the session recording.
+     * Clipboard data is NOT included by default within the recording,
+     * as doing so has privacy and security implications. Including clipboard data
+     * may be necessary in certain auditing contexts, but should only be done
+     * with caution. Clipboard data can easily contain sensitive information, such
+     * as passwords, credit card numbers, etc.
+     */
+    bool recording_include_clipboard;
+
+    /**
      * Whether existing files should be appended to when creating a new recording.
      * Disabled by default.
      */
--- a/src/protocols/rdp/settings.c
+++ b/src/protocols/rdp/settings.c
@@ -125,6 +125,7 @@ const char* GUAC_RDP_CLIENT_ARGS[] = {
     "recording-exclude-mouse",
     "recording-exclude-touch",
     "recording-include-keys",
+    "recording-include-clipboard",
     "create-recording-path",
     "recording-write-existing",
     "resize-method",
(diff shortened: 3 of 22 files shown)


src/protocols/kubernetes/settings.c
```diff
@@ -212,6 +213,16 @@ enum KUBERNETES_ARGS_IDX {
      */
     IDX_RECORDING_INCLUDE_KEYS,
 
+    /**
+     * Whether clipboard paste data should be included in the session recording.
```
necouchman, line 218: Is it only "paste" data, or would it also include copy events?

src/protocols/kubernetes/settings.c
```diff
@@ -212,6 +213,16 @@ enum KUBERNETES_ARGS_IDX {
...
+     * Key events are NOT included by default within the recording,
```
necouchman: Maybe some copy-pasta, here - `Key events` -> `Clipboard events`?

src/protocols/kubernetes/settings.h
```diff
@@ -240,6 +240,16 @@ typedef struct guac_kubernetes_settings {
      */
     bool recording_include_keys;
 
+    /**
+     * Whether clipboard paste data should be included in the session recording.
+     * Key events are NOT included by default within the recording,
```
necouchman: `clipboard paste data` or just `clipboard data`?

`key events" -> `Clipboard events`?

src/protocols/rdp/settings.c
```diff
@@ -560,6 +561,16 @@ enum RDP_ARGS_IDX {
      */
     IDX_RECORDING_INCLUDE_KEYS,
 
+    /**
+     * Whether clipboard paste data should be included in the session recording.
+     * Key events are NOT included by default within the recording,
```
necouchman: And, again, here - with `paste data` and `Key events`...

src/protocols/rdp/settings.h
```diff
@@ -559,6 +559,16 @@ typedef struct guac_rdp_settings {
      */
     int recording_include_keys;
 
+    /**
+     * Whether clipboard paste data should be included in the session recording.
+     * Key events are NOT included by default within the recording,
```
necouchman: And here.

src/protocols/ssh/settings.c
```diff
@@ -243,6 +244,16 @@ enum SSH_ARGS_IDX {
      */
     IDX_RECORDING_INCLUDE_KEYS,
 
+    /**
+     * Whether clipboard paste data should be included in the session recording.
+     * Key events are NOT included by default within the recording,
```
necouchman: And here.

src/protocols/ssh/settings.h
```diff
@@ -257,6 +257,16 @@ typedef struct guac_ssh_settings {
      */
     bool recording_include_keys;
 
+    /**
+     * Whether clipboard paste data should be included in the session recording.
+     * Key events are NOT included by default within the recording,
```
necouchman: Another instance.

src/protocols/telnet/settings.c
```diff
@@ -195,6 +196,16 @@ enum TELNET_ARGS_IDX {
      */
     IDX_RECORDING_INCLUDE_KEYS,
 
+    /**
+     * Whether clipboard paste data should be included in the session recording.
+     * Key events are NOT included by default within the recording,
```
necouchman: And another one.

src/protocols/telnet/settings.h
```diff
@@ -245,6 +245,16 @@ typedef struct guac_telnet_settings {
      */
     bool recording_include_keys;
 
+    /**
+     * Whether clipboard paste data should be included in the session recording.
+     * Key events are NOT included by default within the recording,
```
necouchman: And here.

src/protocols/vnc/settings.c
```diff
@@ -329,6 +330,16 @@ enum VNC_ARGS_IDX {
      */
     IDX_RECORDING_INCLUDE_KEYS,
 
+    /**
+     * Whether clipboard paste data should be included in the session recording.
+     * Key events are NOT included by default within the recording,
```
necouchman: Here.

src/protocols/vnc/settings.h
```diff
@@ -285,6 +285,16 @@ typedef struct guac_vnc_settings {
      */
     bool recording_include_keys;
 
+    /**
+     * Whether clipboard paste data should be included in the session recording.
+     * Key events are NOT included by default within the recording,
```
necouchman: And here.

necouchman (changes requested):
A handful of places where I think some copy-pasta slipped in and got replicated across all of the files you modified. Other than that, I think this looks good.

src/protocols/kubernetes/settings.c
lukasraska: Yeah, tried to be overly consistent with the wording for "recording-include-keys" and well, got too similar to it. Fixed in all occurences.

src/protocols/kubernetes/settings.c
lukasraska: As of now, it only reports paste data -> those within the user-originating clipboard blobs. I aimed this to give better insight into what is the actual data transfered in, since currently if somebody pastes some command via clipboard, it's not really visible in the recordings (merely the UI changes are rendered, whatever happens based on that).

If also server-originating clipboard changes should be logged, from brief check it should probably be easy to hook it up to `__send_user_clipboard` (or probably the respective method that calls this for each connected user), but that recording could get quite big in such cases (so if this would be desired, I would probably aim for two separate args, like with the disable-copy/paste options, so users can choose which one to log.

necouchman: In that case, I wonder if it would be worth changing the overall parameter name to better reflect what we're capturing, in case we do want to add something different in the future. Your point about the size of the recording capturing certain clipboard data is well-taken, and I'm not sure it makes sense to add that (now), I just think we want to be very clear about what data we're capturing.

lukasraska: Sure, what would you suggest as an appropriate parameter name? `recording-include-clipboard-paste` / `recording-include-user-clipboard` or something else? I'm not sure how long these args can be in practice and I would like to avoid overly long parameter names, but then it should be fairly descriptive as you mention. Thanks

necouchman: I'm fine with either of those options - they tend to line up with the existing ones (e.g. `recording-include-keys`).

necouchman:
Close by #630.

PR #531, opened 2024-07-19
GUACAMOLE-1971: Safely multiply values to be used for memory allocation.

aleitner:
CodeQL has flagged a potential integer overflow so I used `guac_mem_ckd_mul_or_die()` to safely multiple the values.

### DIFF
--- a/src/protocols/rdp/pointer.c
+++ b/src/protocols/rdp/pointer.c
@@ -30,6 +30,7 @@
 #include <freerdp/freerdp.h>
 #include <freerdp/gdi/gdi.h>
 #include <guacamole/client.h>
+#include <guacamole/mem.h>
 #include <winpr/crt.h>
 
 BOOL guac_rdp_pointer_new(rdpContext* context, rdpPointer* pointer) {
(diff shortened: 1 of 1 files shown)


jmuehlner (approved):

necouchman:
@aleitner and @jmuehlner Should this get fixed in the `patch` branch, and then applied to the changes in `main`, as well?

jmuehlner:
@necouchman this fix is related to the FreeRDP 3 code, which isn't in `patch`.

necouchman:
Okay. I asked because the call to `_aligned_malloc` is in there in the previous version:

https://github.com/apache/guacamole-server/blob/c715f16b0d93ccd189fe7ddf5391049388e33d5d/src/protocols/rdp/pointer.c#L42-L45

and I think the changes made for FreeRDP3 commits just wrap that call based on whether it should be `winpr_aligned_malloc` or `_aligned_malloc`:

https://github.com/apache/guacamole-server/blob/2aabe2847100f67a8c6da512dd5974ef4118dad6/src/protocols/rdp/rdp.h#L55-L61

I wasn't sure if it only affects one case and not the previous one...

jmuehlner:
Ah, good call. Yeah this should be fixed in `patch` as well, for the old call.

mike-jumper:
FWIW, I've looked through the FreeRDP code surrounding this, and the values in question are strongly limited to 32x32 or 96x96 (`_update_read_pointer_color()`) or 384x384 (`_update_read_pointer_large()`).

It's probably still worth fixing on `patch`.

PR #533, opened 2024-07-30
GUACAMOLE-1973: Add support for XTerm bracketed-paste mode

scottp-dpaw:


necouchman:
@scottp-dpaw Since this is a new feature, and not a bug fix, this can be based against the `main` branch of the Apache repo rather than the `patch` branch. :-)

src/terminal/terminal.c
```diff
@@ -1500,6 +1501,87 @@ int guac_terminal_send_string(guac_terminal* term, const char* data) {
 
 }
 
+int guac_terminal_send_clipboard(guac_terminal *term) {
+    char *filtered = guac_mem_alloc(term->clipboard->length + 12);
```
necouchman, line 1505: It may be completely obvious to you why you need to allocate 12 bytes in addition to the clipboard length, here, but it isn't to me. This is one example of a place where 1) more comments throughout the code might be helpful, and 2) this may be a good place for a constant to replace the literal `12`.

necouchman (changes requested):
I have two overall concerns with the code, here, as illustrated by the one comment I've left inline, below:
1) You've done a reasonable job of documenting things within the code, but there are a few places that could use slightly more documentation.
2) There are several places where constants may be useful.

Other than that, I'll let @mike-jumper and @jmuehlner take a closer look, since they're much more familiar with the terminal code than I am.

scottp-dpaw:
No worries. In lieu of making constants for every bitmask, I've added clearer comments describing the sequence of bits that we're trying to find for each codepoint. Hopefully that should provide enough context to the reader about what we're trying to do, if not I'm open to suggestions.
Gotta say writing out more comments was a worthwhile exercise; I found two errors with my original logic. Turns out Unicode is hard!

PR #538, opened 2024-08-26
GUACAMOLE-1290: Add SFTP support for public keys and correct SSH protocol issue.

necouchman:
This pull request fixes up a couple of issues with public key imports for SSH/SFTP connections:
* Adds the required parameters so that public keys can be used for SFTP connections in addition to SSH connections.
* Moves the code that checks for the presence of the public key inside the block that checks for private key authentication. I'm still a little fuzzy on the purpose of importing a public key, but, if I understand the libssh2 functions correctly, there's no time that you'd want to try to import or use a public key if you weren't also authenticating with a private key. The code, as merged before, imported the public key regardless of whether a private key was present, and the `if/else` logic use also messed up the detection between public key and password-based authentication.

### DIFF
--- a/src/protocols/rdp/rdp.c
+++ b/src/protocols/rdp/rdp.c
@@ -800,6 +800,33 @@ void* guac_rdp_client_thread(void* data) {
                 return NULL;
             }
 
+            /* Import the public key, if that is specified. */
+            if (settings->sftp_public_key != NULL) {
+
+                guac_client_log(client, GUAC_LOG_DEBUG,
+                        "Attempting public key import");
+
+                /* Attempt to read public key */
+                if (guac_common_ssh_user_import_public_key(rdp_client->sftp_user,
+                            settings->sftp_public_key)) {
+
+                    /* Public key import fails. */
+                    guac_client_abort(client,
+                           GUAC_PROTOCOL_STATUS_CLIENT_UNAUTHORIZED,
+                           "Failed to import public key: %s",
+                            guac_common_ssh_key_error());
+
+                    guac_common_ssh_destroy_user(rdp_client->sftp_user);
+                    return NULL;
+
+                }
+
+                /* Success */
+                guac_client_log(client, GUAC_LOG_INFO,
+                        "Public key successfully imported.");
+
+            }
+
         }
 
         /* Otherwise, use specified password */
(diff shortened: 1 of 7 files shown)


src/protocols/rdp/rdp.c
```diff
@@ -800,6 +800,33 @@ void* guac_rdp_client_thread(void* data) {
...
+
+                guac_client_log(client, GUAC_LOG_DEBUG,
+                        "Attempting public key import");
+
+                /* Attempt to read public key */
+                if (guac_common_ssh_user_import_public_key(rdp_client->sftp_user,
+                            settings->sftp_public_key)) {
+
+                     /* Public key import fails. */
+                         guac_client_abort(client,
```
mike-jumper, line 814: Indentation went slightly wonky here.

src/protocols/ssh/ssh.c
```diff
@@ -134,33 +134,34 @@ static guac_common_ssh_user* guac_ssh_get_user(guac_client* client) {
...
-                "Attempting public key import");
+            /* Attempt to read public key */
+            if (guac_common_ssh_user_import_public_key(user,
+                        settings->public_key_base64)) {
 
-        /* Attempt to read public key */
-        if (guac_common_ssh_user_import_public_key(user,
-                    settings->public_key_base64)) {
+                 /* If failing*/
+                     guac_client_abort(client,
```
mike-jumper, line 148: Indentation got skewed here, too.

src/protocols/vnc/vnc.c
```diff
@@ -397,6 +397,33 @@ void* guac_vnc_client_thread(void* data) {
...
+
+                guac_client_log(client, GUAC_LOG_DEBUG,
+                        "Attempting public key import");
+
+                /* Attempt to read public key */
+                if (guac_common_ssh_user_import_public_key(vnc_client->sftp_user,
+                            settings->sftp_public_key)) {
+
+                     /* Public key import fails. */
+                         guac_client_abort(client,
```
mike-jumper, line 411: :upside_down_face:

mike-jumper (changes requested):
LGTM - just some borked indentation.

src/protocols/rdp/rdp.c
necouchman: Oops. Fixed via rebase.

src/protocols/ssh/ssh.c
necouchman: Fixed via rebase.

src/protocols/vnc/vnc.c
necouchman: Can we say copy pasta?!

Fixed via rebase.

mike-jumper (approved):

PR #541, opened 2024-08-28
GUACAMOLE-1980: Bump server version to 1.6.0.

jmuehlner:


mike-jumper (changes requested):
Looks like this needs to be rebased against `staging/1.6.0`, but otherwise LGTM.

jmuehlner:
> Looks like this needs to be rebased against `staging/1.6.0`, but otherwise LGTM.

oops - fixed

mike-jumper (approved):

PR #542, opened 2024-08-29
GUACAMOLE-600: Correct handling of non-blocking socket for timeout.

necouchman:
Okay, I think I've got the issue figured out after some Google Research on how to handle non-blocking sockets in Linux. Let me know if this looks sane or needs any work. I re-tested and am able to establish SSH connections...

### DIFF
--- a/src/libguac/tcp.c
+++ b/src/libguac/tcp.c
@@ -54,8 +54,7 @@ int guac_tcp_connect(const char* hostname, const char* port, const int timeout) 
     }
 
     /* Attempt connection to each address until success */
-    current_address = addresses;
-    while (current_address != NULL) {
+    for (current_address = addresses; current_address != NULL; current_address = current_address->ai_next) {
 
         /* Resolve hostname */
         if ((retval = getnameinfo(current_address->ai_addr,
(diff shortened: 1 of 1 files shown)


necouchman:
Tested with a couple of different timeout values (5s and 30s) against an IP address that isn't up/doesn't respond, and it appears to work as expected.

mike-jumper:
Should we restore the default (blocking) behavior prior to returning the socket, so that callers are not surprised to find that the socket is non-blocking?

neandrake:
I was looking into something similar, flipping a socket to non-blocking and back, for when read-only connections are created and sync display so it doesn’t cause the main connection to lag then. I found it challenging to think through the behavior of the socket. Would the socket need polled until all expected data comes in before switching back to blocking?

necouchman:
> Should we restore the default (blocking) behavior prior to returning the socket, so that callers are not surprised to find that the socket is non-blocking?

Yeah, probably a good idea. I've added a block in to do that, too.

mike-jumper:
> Would the socket need polled until all expected data comes in before switching back to blocking?

I think the blocking vs. non-blocking behavior comes into play only when actually servicing a system call, like a `read()`. The socket should buffer any received data regardless of whether `O_NONBLOCK` is set.

mike-jumper (approved):

PR #544, opened 2024-09-05
GUACAMOLE-1196: Fix VNC resizing behavior.

jmuehlner:
VNC resizing was not working for me - I'm getting this message every time I attempt to resize the window, as well as on initial load:

`ERROR:	Screen has not been initialized, cannot send resize.`

I dug into it, and I found two problems:

1. The RFB client screen ID was never set, which caused any message from the server specifying a screen size to be ignored by the client.
2. The guacamole display was never resized after sending the resize request to the VNC server.

I've fixed these two issues, though I'm not sure if this is the cleanest way of doing it. 
@necouchman mind having a look? I know you're much more familliar with this code than me.

### DIFF
--- a/src/protocols/vnc/display.c
+++ b/src/protocols/vnc/display.c
@@ -182,22 +182,28 @@ static rfbBool guac_vnc_send_desktop_size(rfbClient* client, int width, int heig
     /* Get the Guacamole client data */
     guac_client* gc = rfbClientGetClientData(client, GUAC_VNC_CLIENT_KEY);
 
+    guac_client_log(gc, GUAC_LOG_TRACE,
+            "Current screen size is %ix%i; setting new size %ix%i\n",
+            rfbClientSwap16IfLE(client->screen.width),
+            rfbClientSwap16IfLE(client->screen.height),
+            width, height);
+
 #ifdef LIBVNC_CLIENT_HAS_SCREEN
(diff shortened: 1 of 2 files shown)


src/protocols/vnc/display.c
```diff
@@ -182,22 +182,22 @@ static rfbBool guac_vnc_send_desktop_size(rfbClient* client, int width, int heig
...
     guac_client* gc = rfbClientGetClientData(client, GUAC_VNC_CLIENT_KEY);
 
+    guac_client_log(gc, GUAC_LOG_TRACE,
+            "Current screen size is %ix%i; setting new size %ix%i\n",
+            rfbClientSwap16IfLE(client->screen.width),
+            rfbClientSwap16IfLE(client->screen.height),
+            width, height);
+
 #ifdef LIBVNC_CLIENT_HAS_SCREEN
-    /* Don't send an update if the sreen appears to be uninitialized. */
```
jmuehlner, line 186: I removed this because the screen size seems to be never initialized by the time that the owner attempts to set the initial screen size, so any initial screen resize will fail.

The screen size will be updated when a message comes back from the server specifying the updated size, but this in practice does not happen before the initial resize attempt.

In any case, removing this check fixes the initial resize issue, and doesn't seem to cause any problems that I can see.

necouchman (approved):
Thanks @jmuehlner !

PR #546, opened 2024-09-06
GUACAMOLE-1026: Use LoadChannels callback method to load plugins with FreeRDP3.

aleitner:
Plugins were not working properly when building Guacamole with FreeRDP3 due to a change in how they are loaded. This PR implements the LoadChannels callback function for loading plugins.

https://github.com/FreeRDP/FreeRDP/wiki/FreeRDP3-migration-notes
> Channel loading has been moved from PreConnect to LoadChannels. This has a default implementation in client/common and only step required is removing calls to freerdp_client_load_addins

### DIFF
--- a/src/protocols/rdp/rdp.c
+++ b/src/protocols/rdp/rdp.c
@@ -79,20 +79,28 @@
 #include <stdlib.h>
 #include <time.h>
 
-BOOL rdp_freerdp_pre_connect(freerdp* instance) {
-
+/**
+ * Initializes and loads the necessary FreeRDP plugins based on the current
+ * RDP session settings. This function is designed to work in environments
+ * where the FreeRDP instance expects a LoadChannels callback to be set
+ * otherwise it can becalled directly from our pre_connect callback. It
+ * configures various features such as display resizing, multi-touch support,
+ * audio input, clipboard synchronization, device redirection, and graphics
+ * pipeline, by loading their corresponding plugins if they are enabled in the
+ * session settings.
+ *
+ * @param instance
+ *     The FreeRDP instance to be prepared, containing all context and
+ *     settings for the session.
+ *
+ * @return
+ *     Always TRUE.
+ */
+static BOOL rdp_freerdp_load_channels(freerdp* instance) {
     rdpContext* context = GUAC_RDP_CONTEXT(instance);
-    rdpGraphics* graphics = context->graphics;
-
     guac_client* client = ((rdp_freerdp_context*) context)->client;
     guac_rdp_client* rdp_client = (guac_rdp_client*) client->data;
     guac_rdp_settings* settings = rdp_client->settings;
-
-    /* Push desired settings to FreeRDP */
-    guac_rdp_push_settings(client, settings, instance);
-
-    /* Init FreeRDP add-in provider */
-    freerdp_register_addin_provider(freerdp_channels_load_static_addin_entry, 0);
 
     /* Load "disp" plugin for display update */
     if (settings->resize_method == GUAC_RESIZE_DISPLAY_UPDATE)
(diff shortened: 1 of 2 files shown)


src/protocols/rdp/rdp.c
```diff
@@ -194,21 +247,11 @@ BOOL rdp_freerdp_pre_connect(freerdp* instance) {
...
-    /* Load plugin providing Dynamic Virtual Channel support, if required */
-    if (freerdp_settings_get_bool(GUAC_RDP_CONTEXT(instance)->settings, FreeRDP_SupportDynamicChannels) &&
-            guac_freerdp_channels_load_plugin(context, "drdynvc",
-                GUAC_RDP_CONTEXT(instance)->settings)) {
-        guac_client_log(client, GUAC_LOG_WARNING,
-                "Failed to load drdynvc plugin. Display update and audio "
-                "input support will be disabled.");
-    }
+    #ifndef RDP_INST_HAS_LOAD_CHANNELS
+        rdp_freerdp_load_channels(instance);
```
jmuehlner, line 251: Does this only run for FreeRDP2? The function docs say that, but it looks to me like it always runs?

aleitner: It does always run, but where it is called is changed depending on freerdp2 or 3. I'll update the docs to show that

aleitner: Err actually I think that this documentation is enough that I already had in the function comment?

>  * This function is designed to be used as the LoadChannels
>  * callback within FreeRDP3, but is also called directly from our pre_connect
>  * function in FreeRDP2 environments.

jmuehlner: This function is called directly from the pre_connect function _only_ in FreeRDP2 environments?

aleitner: Yeah

aleitner: Maybe I can reword to 

> 

        This function is designed to be used as the LoadChannels
        callback within FreeRDP3, or is called directly from our pre_connect
        function in FreeRDP2 environments.

jmuehlner: Oops, I misread this. Please ignore me.

jmuehlner (approved):

src/protocols/rdp/rdp.c
jmuehlner: Eh, I think it's ok as is, but if you'd prefer to reword it, go for it.

aleitner: If it was misread then I probably wrote it not too clearly so I updated it

jmuehlner: The point of confusion was the relationship of the `RDP_INST_HAS_LOAD_CHANNELS` symbol to the FreeRDP version. It might be worth mentioning that it's only going to be defined for FreeRDP3, or just refer to what the configure check is actually checking, not the FreeRDP version itself, which isn't obvious at a first glance.

aleitner: Ahhh gotcha, added some comments next to each of the ifdefs

jmuehlner: That's better, but I think it might be worth just saying something about using this callback if available, not tying this behavior to a specific version that's not even checked in the configure check.

There's no guarantee that FreeRDP could release a 2.x version that _does_ have this callback if they wanted to, in which case this comment would be wrong.

jmuehlner: That looks better - mind updating the documentation for `rdp_freerdp_pre_connect()`, which still ties this to the FreeRDP version?

aleitner: Removed mentions of freerdp versions

src/protocols/rdp/rdp.c
```diff
@@ -79,21 +79,29 @@
...
-BOOL rdp_freerdp_pre_connect(freerdp* instance) {
-
+/**
+ * Initializes and loads the necessary FreeRDP plugins based on the current
+ * RDP session settings. This function is designed to work in environments
+ * where the FreeRDP instance expects a LoadChannels callback to be set
+ * otherwise it can becalled directly from our pre_connect callback. It
+ * configures various features such as display resizing, multi-touch support,
+ * audio input, clipboard synchronization, device redirection, and graphics
+ * pipeline, by loading their correspondingplugins if they are enabled in the
```
jmuehlner, line 89: Missing space: "correspondingplugins"

jmuehlner (approved):
LGTM

PR #547, opened 2024-09-06
GUACAMOLE-1196: Resize guacamole display only when VNC screen size is changed by the server.

jmuehlner:
Before, I was immediately resizing the guac display size right after sending the resize request to the server. This works fine as long as the VNC server actually _does_ the resize as requested, but we can't assume that it will. 

This change updates it to only update the display size when the screen size has changed.

### DIFF
--- a/src/protocols/vnc/display.c
+++ b/src/protocols/vnc/display.c
@@ -51,6 +51,19 @@ void guac_vnc_update(rfbClient* client, int x, int y, int w, int h) {
     guac_client* gc = rfbClientGetClientData(client, GUAC_VNC_CLIENT_KEY);
     guac_vnc_client* vnc_client = (guac_vnc_client*) gc->data;
 
+    /* Resize the surface if VNC screen size has changed */
+    int old_height = vnc_client->display->default_surface->height;
+    int old_width  = vnc_client->display->default_surface->width;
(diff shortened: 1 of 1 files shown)


src/protocols/vnc/display.c
```diff
@@ -51,6 +51,18 @@ void guac_vnc_update(rfbClient* client, int x, int y, int w, int h) {
...
+    int old_height = vnc_client->display->default_surface->height;
+    int old_width  = vnc_client->display->default_surface->width;
+    int new_height = rfbClientSwap16IfLE(client->screen.height);
+    int new_width  = rfbClientSwap16IfLE(client->screen.width);
+    if (
+            new_height > 0 && new_width > 0
+            && (new_height != old_height || new_width != old_width)
+    )
+        guac_common_surface_resize(vnc_client->display->default_surface,
+                new_width, new_height);
```
mike-jumper, line 64: Since the condition in this `if` block is lengthy, I think we should include braces around the body for readability.

mike-jumper (changes requested):

src/protocols/vnc/display.c
jmuehlner: Sure

mike-jumper (approved):

PR #550, opened 2024-09-11
GUACAMOLE-1196: Read VNC dimensions directly from the client, not the screen.

jmuehlner:
Per https://libvnc.github.io/doc/html/structrfb_client.html#a46b40d74f04da6f6c302370156cc46c0, the screen field of the VNC client should only be used for setting intended dimensions, not reading updated / current dimensions.

### DIFF
--- a/src/protocols/vnc/display.c
+++ b/src/protocols/vnc/display.c
@@ -51,16 +51,10 @@ void guac_vnc_update(rfbClient* client, int x, int y, int w, int h) {
     guac_client* gc = rfbClientGetClientData(client, GUAC_VNC_CLIENT_KEY);
     guac_vnc_client* vnc_client = (guac_vnc_client*) gc->data;
 
-#ifdef LIBVNC_CLIENT_HAS_SCREEN
-    int new_height = rfbClientSwap16IfLE(client->screen.height);
-    int new_width  = rfbClientSwap16IfLE(client->screen.width);
-#else
-    int new_height = rfbClientSwap16IfLE(client->height);
-    int new_width  = rfbClientSwap16IfLE(client->width);
-#endif
-
     /* Resize the surface if VNC screen size has changed */
+    int new_height = client->height;
     int old_height = vnc_client->display->default_surface->height;
+    int new_width  = client->width;
     int old_width  = vnc_client->display->default_surface->width;
     if (
             new_height > 0 && new_width > 0


src/protocols/vnc/display.c
```diff
@@ -51,16 +51,10 @@ void guac_vnc_update(rfbClient* client, int x, int y, int w, int h) {
     guac_client* gc = rfbClientGetClientData(client, GUAC_VNC_CLIENT_KEY);
     guac_vnc_client* vnc_client = (guac_vnc_client*) gc->data;
 
-#ifdef LIBVNC_CLIENT_HAS_SCREEN
-    int new_height = rfbClientSwap16IfLE(client->screen.height);
-    int new_width  = rfbClientSwap16IfLE(client->screen.width);
-#else
-    int new_height = rfbClientSwap16IfLE(client->height);
```
jmuehlner, line 58: This was wrong anyway - the bit swap should not happen on the client height and width.

corentin-soriano:
This partially fixes the issue I mentioned yesterday in the ticket.
With disable_display_resize checked, no more display problems, but without check disable_display_resize, I was disconnected instantly upon connection.

Is there no way to automatically detect if the remote vnc server supports dimension changes?

necouchman:
> 
> Is there no way to automatically detect if the remote vnc server supports dimension changes?

In my research, there is no consistent way to detect support for it - there is no standard handshake or capabilities message between the client and the server that passes this information, it's just up to the client to try and the server to either respond by changing the display size or ignore the request.

jmuehlner:
> > Is there no way to automatically detect if the remote vnc server supports dimension changes?
> 
> In my research, there is no consistent way to detect support for it - there is no standard handshake or capabilities message between the client and the server that passes this information, it's just up to the client to try and the server to either respond by changing the display size or ignore the request.

Sounds like there's not much left to do here, yeah? People will just have to disable VNC resizing for any server that can't handle the resize request.

necouchman:
> > > Is there no way to automatically detect if the remote vnc server supports dimension changes?
> > 
> > 
> > In my research, there is no consistent way to detect support for it - there is no standard handshake or capabilities message between the client and the server that passes this information, it's just up to the client to try and the server to either respond by changing the display size or ignore the request.
> 
> Sounds like there's not much left to do here, yeah? People will just have to disable VNC resizing for any server that can't handle the resize request.

Agreed. I'll merge this and we'll probably have to call it good, unless someone else comes up with a more reliable way to test for server-side support of resizing.

necouchman (approved):

PR #555, opened 2024-09-30
GUACAMOLE-377: Fix issue where display content is not updated when switching back to normal buffer.

corentin-soriano:
The contents of the alternative buffer are still visible when returning to the normal buffer.

Example with vi:
image
  
After close it (selection remains too):
image

I reused `guac_terminal_redraw_default_layer` function on `apply_color_scheme` and `apply_font` functions because the code is the same.

### DIFF
--- a/src/terminal/terminal-handlers.c
+++ b/src/terminal/terminal-handlers.c
@@ -1070,6 +1070,13 @@ int guac_terminal_csi(guac_terminal* term, unsigned char c) {
                         guac_terminal_scrollbar_set_bounds(term->scrollbar,
                                 -guac_terminal_get_available_scroll(term), 0);
 
+                        /* Redraw normal buffer content */
+                        guac_terminal_redraw_default_layer(term);
+                        
+                        /* Clear selection */
+                        term->text_selected = false;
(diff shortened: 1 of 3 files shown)


src/terminal/terminal-handlers.c
```diff
@@ -1070,6 +1070,14 @@ int guac_terminal_csi(guac_terminal* term, unsigned char c) {
                         guac_terminal_scrollbar_set_bounds(term->scrollbar,
                                 -guac_terminal_get_available_scroll(term), 0);
 
+                        /* Redraw normal buffer content */
+                        guac_terminal_redraw_default_layer(term);
+                        
+                        /* Clear selection */
+                        term->text_selected = false;
+                        term->selection_committed = false;
+                        guac_terminal_notify(term);
```
mike-jumper, line 1079: Is this call needed? There is an overall call to `guac_terminal_notify()` that occurs after any data is written:

https://github.com/apache/guacamole-server/blob/1c1bcc87c8c47d5005d9400c651afa4980e8cc18/src/terminal/terminal.c#L817

mike-jumper (approved):

PR #556, opened 2024-10-15
GUACAMOLE-1961: improve selection

corentin-soriano:


src/terminal/display.c
```diff
@@ -976,38 +976,67 @@ void guac_terminal_display_select(guac_terminal_display* display,
             start_row = end_row;
             end_row = temp;
 
+            /* Don't swap columns if it's a rectangular selection and start_col
+             * is less than end_col */
+            if (!rectangle || start_col > end_col) {
+                temp = start_col;
+                start_col = end_col;
+                end_col = temp;
+            }
```
necouchman: I'm not sure the comment here matches what's actually being done - it seems like it's the opposite. The comment mentions that if it's rectangular and the start column is less than the end column, but the `if` statement is if it's *not* rectangular or the start column is greater than the end column.

Maybe comment just needs to be updated to say, `If the selection is rectangular and the start column is greater than the end column, swap the columns`.

Or does the `if` statement need to be updated?

src/terminal/display.c
```diff
@@ -976,38 +976,67 @@ void guac_terminal_display_select(guac_terminal_display* display,
...
+
+            /* Last row */
+            guac_protocol_send_rect(socket, select_layer,
+
+                    0,
+                    end_row * display->char_height,
+
+                    (end_col + 1) * display->char_width,
+                    display->char_height);
+        }
```
necouchman, line 999: There are a lot of extra lines in this section - for example, when making a multi-line function call, it would be:
```
guac_protocol_send)rect(socket, select_layer,
        0,
        end_row * display->char_height,
        (end_col + 1) * display->char_width,
        display->char_height);
```

src/terminal/terminal.c
```diff
@@ -1804,18 +1804,45 @@ static bool guac_terminal_is_part_of_word(int ascii_char) {
     return ((ascii_char >= '0' && ascii_char <= '9') || 
             (ascii_char >= 'A' && ascii_char <= 'Z') || 
             (ascii_char >= 'a' && ascii_char <= 'z') ||
+            (ascii_char >= GUAC_TERMINAL_LATIN1_CAPITAL_AGRAVE &&
+             ascii_char <= GUAC_TERMINAL_LATIN1_Y_UMLAUT) ||
             (ascii_char == '$') ||
-            (ascii_char == '%') ||
-            (ascii_char == '&') ||
```
necouchman, line 1809: Why are these being removed from the `is_part_of_word()` function? I can understand adding them to the `is_part_of_uri` function, but shouldn't they still be considered part of normal words?

src/terminal/terminal.c
```diff
@@ -1832,6 +1859,215 @@ static bool guac_terminal_is_blank(int ascii_char) {
...
+        *col_head = terminal->selection_initial_column;
+    }
+}
+
+/**
+ * Get word/URI/blank boundaries on a terminal buffer depending on given position:
+ * existing selection to update or mouse position. Gets the boundaries in the
+ * left/up direction and then in the right/down direction, continuing if the
+ * row is wrapped.
+ * A sequence of the same character whatever it is will be treated as a word.
```
necouchman: This part of the comment should be on the line above - no reason for a new line, here.

src/terminal/display.c
corentin-soriano: I agree, the comment is not very clear.
Even more interesting: the entire `if (start_row > end_row) {` block has become unnecessary since #607 because the swap is performed in `guac_terminal_select_update`:
https://github.com/apache/guacamole-server/blob/53ec1f49bdfb7412d0d94c1fab48d18166a00cad/src/terminal/select.c#L165-L170

src/terminal/display.c
corentin-soriano: I changed them.

src/terminal/terminal.c
corentin-soriano: Done

src/terminal/terminal.c
corentin-soriano: I initially placed them here to be able to match a URL. Outside of the context of a URL, I haven't found any other use for them.

necouchman (approved):

PR #558, opened 2024-10-15
Guacamole 1990 Add Canadian Multilingual Standard (cms) layout to RDP

Tejaswikandula:


necouchman:
@Tejaswikandula: Please correct the Pull Request title - it should be in the same format as the commit messages (GUACAMOLE-1990: Add Canadian...).

Also, you have two commits with the same commit message - the commit messages should be helpful and descriptive, and possibly squashed together.

Tejaswikandula:
> @Tejaswikandula: Please correct the Pull Request title - it should be in the same format as the commit messages (GUACAMOLE-1990: Add Canadian...).
> 
> Also, you have two commits with the same commit message - the commit messages should be helpful and descriptive, and possibly squashed together.

I have fixed the title

Tejaswikandula:
On the Canadian Multilingual Standard Keyboard, the "Right-CTRL" key serves a distinct function from the "Left-CTRL" key, as it's used in combination with other keys to generate special characters. This is different from other layouts currently supported by guacamole. I noticed that the +ctrl modifier doesn’t seem to be supported. Adding this support would require modifying [generate.pl](https://github.com/apache/guacamole-server/blob/main/src/protocols/rdp/keymaps/generate.pl) to handle the +ctrl modifier, along with other changes and updates to existing keymap files. Could this be the recommended approach, or are there alternative methods to achieve this? I’d appreciate any insights. Thank you!

Tejaswikandula:
> On the Canadian Multilingual Standard Keyboard, the "Right-CTRL" key serves a distinct function from the "Left-CTRL" key, as it's used in combination with other keys to generate special characters. This is different from other layouts currently supported by guacamole. I noticed that the +ctrl modifier doesn’t seem to be supported. Adding this support would require modifying [generate.pl](https://github.com/apache/guacamole-server/blob/main/src/protocols/rdp/keymaps/generate.pl) to handle the +ctrl modifier, along with other changes and updates to existing keymap files. Could this be the recommended approach, or are there alternative methods to achieve this? I’d appreciate any insights. Thank you!

@necouchman Thoughts?

PR #559, opened 2024-10-17
GUACAMOLE-1167: Add server settings for RD gateway type.

necouchman:


src/protocols/rdp/settings.c
```diff
@@ -1263,6 +1271,20 @@ guac_rdp_settings* guac_rdp_parse_args(guac_user* user,
     settings->gateway_port =
         guac_user_parse_args_int(user, GUAC_RDP_CLIENT_ARGS, argv,
                 IDX_GATEWAY_PORT, 443);
+    
+    char* gateway_type = guac_user_parse_args_string(user, GUAC_RDP_CLIENT_ARGS,
```
jmuehlner, line 1275: What happens if this isn't set to any of these values?

jmuehlner: To be more clear, it'd be nice if there was some documentation somewhere in here that briefly explains what happens in the default case where the `gateway-type` parameter isn't set (or is set to some invalid value).

PR #560, opened 2024-10-20
GUACAMOLE-288: Add support for multi-monitor connections.

corentin-soriano:


src/protocols/rdp/channels/disp.c
```diff
@@ -191,6 +193,15 @@ void guac_rdp_disp_update_size(guac_rdp_disp* disp,
 
     int width = disp->requested_width;
     int height = disp->requested_height;
+    int monitors_count = disp->requested_monitors;
+
+    /* Prevent opening too many monitors than allowed */
+    if (settings->max_secondary_monitors + 1 < monitors_count)
```
neandrake, line 199: This condition is odd and might be simpler to read as
```c
int requested_monitors = disp->requested_monitors;

if (requested_monitors > settings->max_secondary_monitors) {
//...
}
```

src/protocols/rdp/channels/disp.c
```diff
@@ -191,6 +193,15 @@ void guac_rdp_disp_update_size(guac_rdp_disp* disp,
...
+        monitors_count = settings->max_secondary_monitors + 1;
```
neandrake, line 200: This makes it seem like `max_secondary_monitors` is 0-based instead of 1-based. Is that the case? It should be 1-based since it's a count and not an index.

src/protocols/rdp/channels/disp.c
```diff
@@ -204,7 +215,7 @@ void guac_rdp_disp_update_size(guac_rdp_disp* disp,
 
     /* Do NOT send requests unless the size will change */
     if (rdp_inst != NULL
-            && width == guac_rdp_get_width(rdp_inst)
+            && width * monitors_count == guac_rdp_get_width(rdp_inst)
             && height == guac_rdp_get_height(rdp_inst))
         return;
```
neandrake, line 220: This makes the assumption that monitors are horizontally aligned and have the same resolution. Is that a constraint for this change? It's probably a common setup it would be nice to be flexible in its support.

src/protocols/rdp/channels/disp.c
corentin-soriano: Yes it's more readable, I modified it in this way.

src/protocols/rdp/channels/disp.c
corentin-soriano: The +1 corresponds to the primary monitor.
For better readability I added a variable:
```c
/* Add one to account for the primary monitor */
int max_monitors = settings->max_secondary_monitors + 1;
```

src/protocols/rdp/channels/disp.c
corentin-soriano: Yes, all monitors have the same dimensions and are aligned left to right (the left one is the primary).
I haven't looked to do this better yet but there is no limitation of FreeRDP.

Pierre-Gronau-ndaal:
it will be great if we bring that to main

necouchman:
> it will be great if we bring that to main

Yes, that is the plan, once the review is finalized.

src/protocols/rdp/channels/disp.c
corentin-soriano: We can now have different sizes and shift a monitor on the Y axis on the guacd side!

corentin-soriano:
> Frequently resizing a monitor window, e.g. by using Maximize and Restore down of the window in quick succession, can cause a RDP disconnect with the following message in the guacd logs:
> 
> ```
> GUAC_ASSERT in guac_rdp_gdi_desktop_resize() failed at gdi.c:172.
> ```
> 
> [Assertion code](https://github.com/apache/guacamole-server/blob/5294c1b803cc454261c889aad1eabe3ec30a4c84/src/protocols/rdp/gdi.c#L172)
> 
> Is it possible to wait for the flush or just discard the affected window instead of crashing the connection?

Which version of freerdp are you using?
If you are on v3, can you update it to at least version 3.8.0?

Vertganti:
I'm using the Dockerfile form the repository which [defaults to FreeRDP 2](https://github.com/corentin-soriano/guacamole-server/blob/GUACAMOLE-288_multi_monitors/Dockerfile#L41). The exact version is `2.11.7`.  I can try using FreeRDP 3 to see if that improves anything.

corentin-soriano:
> I'm using the Dockerfile form the repository which [defaults to FreeRDP 2](https://github.com/corentin-soriano/guacamole-server/blob/GUACAMOLE-288_multi_monitors/Dockerfile#L41). The exact version is `2.11.7`. I can try using FreeRDP 3 to see if that improves anything.

This is potentially a bug in FreeRDP. This assertion is mandatory in version 2.
Are you sure you can't reproduce this with a single monitor?

Vertganti:
I don't think I tried it with one monitor. Will do that as well once I have the time.

Vertganti:
> > I'm using the Dockerfile form the repository which [defaults to FreeRDP 2](https://github.com/corentin-soriano/guacamole-server/blob/GUACAMOLE-288_multi_monitors/Dockerfile#L41). The exact version is `2.11.7`. I can try using FreeRDP 3 to see if that improves anything.
> 
> This is potentially a bug in FreeRDP. This assertion is mandatory in version 2. Are you sure you can't reproduce this with a single monitor?

I was able to reproduce it with one monitor with the FreeRDP 2 build.

src/protocols/rdp/channels/disp.c
```diff
@@ -221,32 +387,40 @@ void guac_rdp_disp_update_size(guac_rdp_disp* disp,
...
+            monitors[i].PhysicalWidth = 0;
+            monitors[i].PhysicalHeight = 0;
         }
 
+        /* Send display update notification */
+        guac_client* client = disp->client;
+        guac_rdp_client* rdp_client = (guac_rdp_client*) client->data;
+
+        pthread_mutex_lock(&(rdp_client->message_lock));
+        disp->disp->SendMonitorLayout(disp->disp, monitors_count, monitors);
```
weberr13, line 419: I may not understand the API, but naively I feel like when a pointer (monitors) is sent via a signal that memory must either be given to the receiver (eg it is now responsible for freeing monitors) or it would have to signal it is "done" with the memory somehow (which I cannot see happening).  I believe from line 422 that there is a possible segmentation violation with the receiver of SendMonitorLayout reading free'ed memory.

src/protocols/rdp/client.c
```diff
@@ -130,6 +131,16 @@ static int guac_rdp_join_pending_handler(guac_client* client) {
...
 
+    /* Get max secondary monitors */
+    char* max_monitors = guac_mem_alloc(12);
+    guac_itoa(max_monitors,
+            (unsigned int) rdp_client->settings->max_secondary_monitors);
+
+    /* Send current max allowed secondary monitors */
+    guac_client_stream_argv(client, broadcast_socket, "text/plain",
+            "secondary-monitors", max_monitors);
+    guac_mem_free(max_monitors);
```
weberr13, line 142: same concern as above, who "owns" the char* max_monitors?  does guac_client_stream_argv copy from max_monitors?

src/protocols/rdp/gdi.c
```diff
@@ -190,6 +193,45 @@ BOOL guac_rdp_gdi_desktop_resize(rdpContext* context) {
 
     guac_display_layer_close_raw(default_layer, current_context);
 
+    /* Make json string containing monitor information */
+    char json[JSON_BUFFER_SIZE];
+    int pos = 0;
+    pos += snprintf(json + pos, JSON_BUFFER_SIZE - pos, "{");
```
weberr13, line 199: nit: While there may be concerns about grabbing an external dependency, when it comes to "never re-invent the wheel" any manual string manipulation to do json generation/parsing is king.  Libraries like

https://github.com/nlohmann/json

will allow "C++ like" json parsing and generation that covers all the corner cases that JSON hides from the average user.

src/protocols/rdp/gdi.c
```diff
@@ -190,6 +193,45 @@ BOOL guac_rdp_gdi_desktop_resize(rdpContext* context) {
...
+    for (int i = 0; i < rdp_client->disp->monitors_count; i++) {
+
+        /* Skip monitors that have not been initialized yet */
+        if (rdp_client->disp->monitors[i].requested_width == 0 ||
+            rdp_client->disp->monitors[i].requested_height == 0) {
+            continue;
+        }
+
+        /* Append monitor information to JSON string */
+        pos += snprintf(json + pos, JSON_BUFFER_SIZE - pos,
```
weberr13, line 210: A small piece of my soul is severed and sent to the depths of Hades every time I see sprintf generating json.

lirannisani:
@necouchman @corentin-soriano 
Thanks for adding this feature!
Do you plan to merge it into the main branch? I’d like to use the **multimon:i:1** capability in Guacamole, and it looks like this PR would enable that functionality.
Or maybe meanwhile there’s another way to achieve multimon:i:1 support in Guacamole?

src/protocols/rdp/channels/disp.c
corentin-soriano: The FreeRDP function (SendMonitorLayout) is synchronous and copies the data.
The original pointer is not retained.

[FreeRDP source file](https://github.com/FreeRDP/FreeRDP/blob/master/channels/disp/client/disp_main.c#L103)

src/protocols/rdp/client.c
corentin-soriano: The same applies here; the function is synchronous and will make a copy of the data before sending it over the socket.

PR #565, opened 2024-12-01
GUACAMOLE-1256: Fix a display issue when scrolling up in vi as well as latency in less.

corentin-soriano:
Allows better synchronization between the buffer and the display, which avoids "mixed" displayed data when highlighting braces, parentheses and others in vi.
Seems to fix a latency issue when scrolling up in less.

### DIFF
--- a/src/terminal/display.c
+++ b/src/terminal/display.c
@@ -480,6 +480,11 @@ void guac_terminal_display_set_columns(guac_terminal_display* display, int row,
     /* For each column in range */
     for (int col = start_column; col <= end_column; col += character->width) {
 
+        /* Flush pending copy operation before adding new SET operation. This
+         * avoid operation conflicts that cause inconsistent display. */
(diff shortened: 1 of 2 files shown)


necouchman:
Looks okay to me. @mike-jumper or @jmuehlner, any thoughts/comments?

mike-jumper:
This looks at least equivalent to the workaround that was already put in place, and with the understanding that the workaround does appear to address the behavior, I'm happy with the change.

More broadly, I'm not sure about the logic of the workaround. If the issue is that the terminal's `GUAC_CHAR_COPY` can end up pointing at unexpected data, I'd think the correction would be to flush the copy whenever the _source_ of that copy is changed, not when character containing `GUAC_CHAR_COPY` changes (which would just mean that the copy is being replaced and is no longer necessary).

mike-jumper (approved):

PR #567, opened 2024-12-28
Added Russian keyboard layout

VAGurko:


VAGurko:
Added Russian keyboard layout

PR #570, opened 2025-01-04
GUACAMOLE-925: Add Russian keymap support "ru-ru-qwerty"

VAGurko:
The keymap "ru_ru_qwerty.keymap" has been added to the "src/protocols/rdp/Makefile.am" file to support the Russian keyboard layout.
The file "src/protocols/rdp/keymaps/ru_ru_qwerty.keymap" has been added to support the Russian keyboard layout.

### DIFF
--- a/src/protocols/rdp/keymaps/ru_ru_qwerty.keymap
+++ b/src/protocols/rdp/keymaps/ru_ru_qwerty.keymap
@@ -1,0 +1,63 @@
+#
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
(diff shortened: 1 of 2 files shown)


src/protocols/rdp/keymaps/ru_ru_qwerty.keymap
```diff
@@ -0,0 +1,63 @@
...
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+#
+
+parent  "base"
+name    "ru-ru-qwerty"
+# We use the "KBD_US" layout because we switch the language, not the layout
+freerdp "KBD_US"
```
mike-jumper, line 23: Can you clarify why/how this is the case?

I would think one of the following mutually-exclusive possibilities has to be true:

* This is, in fact, a keyboard layout and needs to point to the RDP layout so that it can be sent to the RDP server. I think the correct value here is likely `KBD_RUSSIAN`.
* This isn't a keyboard layout and we shouldn't be defining it at all.

Assuming this is indeed a keyboard layout that needs a definition, I'm looking around to find examples of that keyboard, and so far they don't exactly align: https://en.wikipedia.org/wiki/JCUKEN (just Cyrillic keys with no Latin-only mode)

VAGurko: You're right, you need to use "KBD_RUSSIAN", I'm going to fix it.

PR #572, opened 2025-01-07
GUACAMOLE-2012: Fix SSH connection to FIPS servers which only offer AES GCM

eugen-keeper:
* Added AES GCM ciphers to the FIPS compliant cipher list
* Added some code to check if the hardcoded FIPS compliant ciphers are really supported by libssh2

### DIFF
--- a/src/common-ssh/ssh.c
+++ b/src/common-ssh/ssh.c
@@ -17,6 +17,8 @@
  * under the License.
  */
 
+#include "common/string.h"
+
 #include "common-ssh/key.h"
 #include "common-ssh/ssh.h"
 #include "common-ssh/user.h"
@@ -39,6 +41,7 @@
 #include <netinet/in.h>
 #include <pthread.h>
 #include <pwd.h>
+#include <stdbool.h>
 #include <stddef.h>
 #include <stdlib.h>
 #include <string.h>
@@ -61,7 +64,7 @@ GCRY_THREAD_OPTION_PTHREAD_IMPL;
 /**
  * A list of ciphers that are both FIPS-compliant, and OpenSSL-supported.
  */
-#define FIPS_COMPLIANT_CIPHERS "aes128-ctr,aes192-ctr,aes256-ctr,aes128-cbc,aes192-cbc,aes256-cbc"
+#define FIPS_COMPLIANT_CIPHERS "aes256-gcm@openssh.com,aes128-gcm@openssh.com,aes128-ctr,aes192-ctr,aes256-ctr,aes128-cbc,aes192-cbc,aes256-cbc"
 
 #ifdef OPENSSL_REQUIRES_THREADING_CALLBACKS
 /**
@@ -403,6 +406,59 @@ static int guac_common_ssh_authenticate(guac_common_ssh_session* common_session)
 
 }
 
+/**
+ * Verifies if given algorithms are supported by libssh2.
+ * Writes log messages if an algorithm is not supported or
+ * could not get the list of supported algorithms from libssh2.
+ *
+ * @param client
+ *     The Guacamole client that is using SSH.
+ * 
+ * @param session
+ *     The session associated with the user to be authenticated.
+ *
+ * @param method_type
+ *      One of the libssh2 Method Type constants for libssh2_session_method_pref().
+ * 
+ * @param algs
+ *      A string with preferred list of algorithms, for example FIPS_COMPLIANT_CIPHERS.
+ *
+ */
+static void check_if_algs_are_supported(guac_client* client, LIBSSH2_SESSION* session,
+        int method_type, const char* algs) {
+
+    /* Request the list of supported algorithms/cyphers from libssh2. */
+    const char** supported_algs;
+    int supported_algs_count =
+        libssh2_session_supported_algs(session, method_type, &supported_algs);
+
+    if (supported_algs_count > 0) {
+        char** preferred_algs = guac_split(algs, ',');
+        for (int i = 0; preferred_algs[i]; i++) {
+            bool found = false;
+            /* Check if the algorithm is found in the libssh2 supported list. */
+            for (int j = 0; j < supported_algs_count; j++) {
+                if (strcmp(preferred_algs[i], supported_algs[j]) == 0) {
+                    found = true;
+                    break;
+                }
+            }
+            if (!found) {
+                guac_client_log(client, GUAC_LOG_WARNING,
+                    "Preferred algorithm/cipher '%s' is not supported by libssh2", preferred_algs[i]);
+            }
+        }
+        guac_mem_free(preferred_algs);
+        /* should free if supported_algs_count is a positive number. */
+        libssh2_free(session, supported_algs);
+    }
+    else {
+        guac_client_log(client, GUAC_LOG_WARNING,
+            "libssh2 reports that no ciphers/algorithms are supported. This may be a bug in libssh2. "
+            "If the SSH connection fails, it may not be possible to log the cause here.");
+    }
+}
+
 guac_common_ssh_session* guac_common_ssh_create_session(guac_client* client,
         const char* hostname, const char* port, guac_common_ssh_user* user,
         int keepalive, const char* host_key,
@@ -507,8 +563,16 @@ guac_common_ssh_session* guac_common_ssh_create_session(guac_client* client,
      * https://csrc.nist.gov/CSRC/media/projects/cryptographic-module-validation-program/documents/security-policies/140sp2906.pdf
      */
     if (guac_fips_enabled()) {
+        /*
+         * The following algorithm check is only to simplify debugging.
+         * libssh2_session_method_pref() ignores unsupported methods.
+         * So they are not sent to the remote host during protocol negotiation anyways.
+         */
+        check_if_algs_are_supported(client, session, LIBSSH2_METHOD_KEX, FIPS_COMPLIANT_KEX_ALGORITHMS);
         libssh2_session_method_pref(session, LIBSSH2_METHOD_KEX, FIPS_COMPLIANT_KEX_ALGORITHMS);
+        check_if_algs_are_supported(client, session, LIBSSH2_METHOD_CRYPT_CS, FIPS_COMPLIANT_CIPHERS);
         libssh2_session_method_pref(session, LIBSSH2_METHOD_CRYPT_CS, FIPS_COMPLIANT_CIPHERS);
+        check_if_algs_are_supported(client, session, LIBSSH2_METHOD_CRYPT_SC, FIPS_COMPLIANT_CIPHERS);
         libssh2_session_method_pref(session, LIBSSH2_METHOD_CRYPT_SC, FIPS_COMPLIANT_CIPHERS);
     }
 


necouchman (changes requested):
Looks pretty good to me - the only thing that I see is an opportunity for some comments in the new function you implemented - the existing code around that has bits of documentation throughout, so maybe just brush that up a bit in the new code?

src/common-ssh/ssh.c
```diff
@@ -412,6 +428,84 @@ static int guac_common_ssh_authenticate(guac_common_ssh_session* common_session)
...
+                algs_str_size += MAX(req_size, 1024);
+                algs_str = guac_mem_realloc(algs_str, algs_str_size);
+            }
+            /* Add comma only if necessary. */
+            if (req_comma) {
+                strcat(algs_str, ",");
+            }
+            /* Add the algorithm name. */
+            strcat(algs_str, algs[i]);  
+        } else {
```
mike-jumper, line 499: Please don't cuddle the `else`. See: https://cwiki.apache.org/confluence/display/GUAC/Contribution+and+Style+Guidelines#ContributionandStyleGuidelines-Braces

src/common-ssh/ssh.c
```diff
@@ -412,6 +428,84 @@ static int guac_common_ssh_authenticate(guac_common_ssh_session* common_session)
...
+        libssh2_session_supported_algs(session, method_type, &supported_algs);
+
+    /* This is not a show stopper. */
+    if (supported_algs_count <= 0) {
+        guac_client_log(client, GUAC_LOG_WARNING,
+            "Could not verify if preferred algorithms/ciphers are supported by libssh2");
+    }
+
+    int i = 0;
+    while (algs[i]) {
```
mike-jumper, line 474: This loop:

```c
int i = 0;
while (algs[i]) {
   ...
   i++;
}
```

is equivalent to:

```c
for (int i = 0; algs[i]; i++) {
    ...
}
```

src/common-ssh/ssh.c
```diff
@@ -412,6 +428,84 @@ static int guac_common_ssh_authenticate(guac_common_ssh_session* common_session)
...
+                /* Either 1024 to not reallocate often or required size if bigger. */
+                algs_str_size += MAX(req_size, 1024);
+                algs_str = guac_mem_realloc(algs_str, algs_str_size);
+            }
+            /* Add comma only if necessary. */
+            if (req_comma) {
+                strcat(algs_str, ",");
+            }
+            /* Add the algorithm name. */
+            strcat(algs_str, algs[i]);
```
mike-jumper, line 498: Each call to `strcat()` is going to search through `algs_str` to find the end of the string. Doing this in a loop, we're going to be unnecessarily searching through `algs_str` repeatedly.

src/common-ssh/ssh.c
```diff
@@ -412,6 +428,84 @@ static int guac_common_ssh_authenticate(guac_common_ssh_session* common_session)
...
+ *     The session associated with the user to be authenticated.
+ *
+ * @param method_type
+ *      One of the libssh2 Method Type constants for libssh2_session_method_pref().
+ * 
+ * @param algs
+ *      A preferred list of algorithms, for example FIPS_COMPLIANT_CIPHERS.
+ * 
+ * @return
+ *     The |prefs| string to use with libssh2_session_method_pref().
```
mike-jumper, line 451: The fact that this pointer must eventually be freed with a call to `guac_mem_free()` should be documented.

src/common-ssh/ssh.c
```diff
@@ -412,6 +428,84 @@ static int guac_common_ssh_authenticate(guac_common_ssh_session* common_session)
...
+             if (strcmp(algs[i], supported_algs[j]) == 0) {
+                found = true;
+                break;
+            }
+        }
+        /* Add the algorithm to the string if found or libssh2_session_supported_algs() failed. */
+        if (found || supported_algs_count <= 0) {
+            bool req_comma = algs_str[0] != 0;
+            /* Size of the algorithm name plus 1 for comma (if required) plus 1 for '\0'. */
+            int req_size = strlen(algs[i]) + (req_comma ? 1 : 0) + 1;
```
mike-jumper, line 487: This could wrap around due to a large value from `strlen()` and produce a negative value. It would be better to use `guac_mem_ckd_add_or_die()`:

https://guacamole.apache.org/doc/libguac/mem_8h.html#a57af96611f73a3f2300ba9fd64b0167a

src/common-ssh/ssh.c
```diff
@@ -412,6 +428,84 @@ static int guac_common_ssh_authenticate(guac_common_ssh_session* common_session)
...
+            if (strlen(algs_str) + req_size > algs_str_size) {
+                /* Either 1024 to not reallocate often or required size if bigger. */
+                algs_str_size += MAX(req_size, 1024);
```
mike-jumper, line 490: `MAX()` is not a macro defined by C or POSIX - it's a GNU and BSD extension. We can't rely on this macro being available.

src/common-ssh/ssh.c
```diff
@@ -412,6 +428,84 @@ static int guac_common_ssh_authenticate(guac_common_ssh_session* common_session)
...
+                algs_str = guac_mem_realloc(algs_str, algs_str_size);
```
mike-jumper, line 491: This can fail, which will overwrite `algs_str` with `NULL` and leave a dangling block of memory. `guac_mem_realloc_or_die()` is a better choice if always discarding the original pointer.

src/common-ssh/ssh.c
```diff
@@ -412,6 +428,84 @@ static int guac_common_ssh_authenticate(guac_common_ssh_session* common_session)
...
+
+    /* Request the list of supported algorithms/cyphers from libssh2. */
+    const char **supported_algs;
+    int supported_algs_count =
+        libssh2_session_supported_algs(session, method_type, &supported_algs);
+
+    /* This is not a show stopper. */
+    if (supported_algs_count <= 0) {
+        guac_client_log(client, GUAC_LOG_WARNING,
+            "Could not verify if preferred algorithms/ciphers are supported by libssh2");
```
mike-jumper, line 470: I like this warning, but we should include wording that makes it clear to the admin reading these logs what this means in practical terms. Will this potentially result in the connection failing? Or in lack of information regarding future failures? Or ... ?

src/common-ssh/ssh.c
```diff
@@ -412,6 +428,84 @@ static int guac_common_ssh_authenticate(guac_common_ssh_session* common_session)
...
+    int i = 0;
+    while (algs[i]) {
+        bool found = false;
+        /* Check if the algorithm is found in the libssh2 supported list. */
+        for (int j = 0; j < supported_algs_count; j++) {
+             if (strcmp(algs[i], supported_algs[j]) == 0) {
+                found = true;
+                break;
+            }
+        }
```
mike-jumper, line 482: I'm wary of doing a brute-force search in a loop like this, but if we're certain the list cannot grow to any substantial length, then this is fine.

If the list might grow to become large, we should consider sorting the list and doing a `bsearch()`.

src/common-ssh/ssh.c
```diff
@@ -445,9 +539,20 @@ guac_common_ssh_session* guac_common_ssh_create_session(guac_client* client,
...
+
+        char* crypt_cs_str = get_method_pref_string(client,
+            session, LIBSSH2_METHOD_CRYPT_CS, FIPS_COMPLIANT_CIPHERS);
+        libssh2_session_method_pref(session, LIBSSH2_METHOD_CRYPT_CS, crypt_cs_str);
+        guac_mem_free(crypt_cs_str);
+      
+        char* crypt_sc_str = get_method_pref_string(client,
+            session, LIBSSH2_METHOD_CRYPT_SC, FIPS_COMPLIANT_CIPHERS);
+        libssh2_session_method_pref(session, LIBSSH2_METHOD_CRYPT_SC, crypt_sc_str);
+        guac_mem_free(crypt_sc_str);
```
mike-jumper, line 555: According to https://libssh2.org/libssh2_session_method_pref.html, the `libssh2_session_method_pref()` function will ignore any preferred cipers/algorithms/etc. if not supported by libssh2:

> ... If a method is listed which is not supported by libssh2 it will be ignored and not sent to the remote host during protocol negotiation.

So, in case this helps simplify things, it's not necessary to dynamically assemble a string based on what libssh2 has available (though it's still helpful to log whether something isn't supported).

mike-jumper (changes requested):

src/common-ssh/ssh.c
```diff
@@ -412,6 +415,56 @@ static int guac_common_ssh_authenticate(guac_common_ssh_session* common_session)
...
+                    break;
+                }
+            }
+            if (!found) {
+                guac_client_log(client, GUAC_LOG_WARNING,
+                    "Preferred algorithm/cipher '%s' is not supported by libssh2", preferred_algs[i]);               
+            }
+        }
+        guac_mem_free(preferred_algs);
+    } else {
```
mike-jumper, line 461: Please don't cuddle the `else`. See: https://cwiki.apache.org/confluence/display/GUAC/Contribution+and+Style+Guidelines#ContributionandStyleGuidelines-Braces

mike-jumper (changes requested):

mike-jumper:
@eugen-keeper Please also:

1. Rebase to remove the merge commit.
2. Either add context to the commit messages ("Fix a size calculation issue" is a bit vague, for example) or squash them into a single commit.

src/common-ssh/ssh.c
```diff
@@ -403,6 +406,57 @@ static int guac_common_ssh_authenticate(guac_common_ssh_session* common_session)
...
+                guac_client_log(client, GUAC_LOG_WARNING,
+                    "Preferred algorithm/cipher '%s' is not supported by libssh2", preferred_algs[i]);
+            }
+        }
+        guac_mem_free(preferred_algs);
+    }
+    else {
+        guac_client_log(client, GUAC_LOG_WARNING,
+            "libssh2 reports that no ciphers/algorithms are supported. This may be a bug in libssh2."
+            "If the SSH connection fails, it may not be possible to log the cause here.");
```
mike-jumper, line 456: A space is needed at the end of the first string literal here. Without that, the relevant portion of the resulting string will be `libssh2.If` and not the intended `libssh2. If`.

src/common-ssh/ssh.c
```diff
@@ -403,6 +406,57 @@ static int guac_common_ssh_authenticate(guac_common_ssh_session* common_session)
...
+ *      A string with preferred list of algorithms, for example FIPS_COMPLIANT_CIPHERS.
+ *
+ */
+static void check_if_algs_are_supported(guac_client* client, LIBSSH2_SESSION* session,
+        int method_type, const char* algs) {
+
+    /* Request the list of supported algorithms/cyphers from libssh2. */
+    const char** supported_algs;
+    int supported_algs_count =
+        libssh2_session_supported_algs(session, method_type, &supported_algs);
```
mike-jumper, line 433: Based on the example at https://libssh2.org/libssh2_session_supported_algs.html, the memory pointed to by `supported_algs` after a successful call to `libssh2_session_supported_algs()` must eventually be freed by a call to `libssh2_free()`.

mike-jumper (changes requested):

src/common-ssh/ssh.c
eugen-keeper: Yes, I had in the original implementation but forgot to add after reimplementing the function

eugen-keeper: Fixed

src/common-ssh/ssh.c
eugen-keeper: Fixed

mike-jumper (approved):

mike-jumper:
LGTM.

@necouchman Any further thoughts?

necouchman (approved):
All good, here.

PR #575, opened 2025-02-14
GUACAMOLE-1374: Migrate to LOG_LEVEL and modular Docker build.

mike-jumper:
This change migrates to a Docker build leveraging multiple modular stages that cover each of guacamole-server's dependencies, avoiding unnecessary recompilation of dependencies that are not changing (something that is slowing down rebuilds for testing quite a bit, particularly if any download from GitHub times out).

For consistency's sake, I've also added a `LOG_LEVEL` variable to replace `GUACD_LOG_LEVEL`, in line with the guacamole-client image.


aaron-iles (approved):
LGTM. Testing the build now.

Dockerfile
```diff
@@ -127,19 +106,39 @@ ARG FREERDP_OPTS="\
     -DWITH_XV=OFF \
     -DWITH_ZLIB=ON"
 
+ARG FREERDP_X86_OPTS="-DWITH_SSE2=ON"
```
mattieb, line 109: You may be able to use WITH_SIMD across all architectures in modern versions of FreeRDP—it [replaces WITH_SSE2](https://github.com/FreeRDP/FreeRDP/blob/6a4bbfc04712885e4322a48988b4833cd3c46ea0/cmake/DetectIntrinsicSupport.cmake#L5) anyway.

src/guacd-docker/bin/autobuild.sh
```diff
@@ -0,0 +1,144 @@
...
+fi
+
+# Determine architecture that we're building on
+if [ -z "$BUILD_ARCHITECTURE" ]; then
+    case "$(uname -m)" in
+        armv6l|armv7l|aarch64)
+            BUILD_ARCHITECTURE="ARM"
+            ;;
+        x86_64)
+            BUILD_ARCHITECTURE="X86"
```
necouchman, line 91: I know 32-bit x86 has mostly gone away, but are there any known issues with Guacamole running in a 32-bit x86 environment? Should this be broadened to include both i686 and x86_64?

necouchman:
Just one comment from me, plus the one that @mattieb mentioned above, and this should be ready to merge.

Dockerfile
mike-jumper: Welp, this would still have to be `WITH_SSE2` for builds against FreeRDP 2.x, but thankfully it looks like both `WITH_SSE2` (2.x) and `WITH_SIMD` (3.x) are the default. I'll just remove the option.

src/guacd-docker/bin/autobuild.sh
mike-jumper: Sure, easy enough.

mike-jumper:
Went ahead and added `CPPFLAGS=-Wno-error=deprecated-declarations` to `GUACAMOLE_SERVER_OPTS`, as well, to work around the FreeRDP header warnings until addressed in FreeRDP or GCC.

necouchman (approved):

PR #577, opened 2025-02-24
GUACAMOLE-2035: Add start-up period to guacd health check in Dockerfile

jeloneal:
Added start-period to docker healthcheck. Should fix containers being stuck in `starting` state for a long time causing issues with certain docker compose setups.

Without the start-period parameter the healthcheck seems to run only after the set interval of 5 minutes. This is way too long if you use the guacamole container within a stack and the depends_on condition "service_healthy". The deployment of the stack will abort.

Tested this fix on my own setup and it fixes the problem for me. Might be useful for others as well.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -328,7 +328,7 @@ ENV LC_ALL=C.UTF-8
 ENV LD_LIBRARY_PATH=${PREFIX_DIR}/lib
 
 # Checks the operating status every 5 minutes with a timeout of 5 seconds
-HEALTHCHECK --interval=5m --timeout=5s CMD nc -z 127.0.0.1 4822 || exit 1
+HEALTHCHECK --interval=5m --timeout=5s --start-period=15s CMD nc -z 127.0.0.1 4822 || exit 1
 
 # Create a new user guacd
 ARG UID=1000


necouchman:
@jeloneal : Thanks for the contribution. In order to process this change you will need to get a Jira account and create a ticket, and then tag both the commit messages and the pull request with the Jira issue. Also, please update the commit messages and Jira issue with more descriptive - something along the lines of "Add start-up period to guacd health check in Dockerfile."

For a full list of contribution guidelines, please see the following site: https://guacamole.apache.org/open-source/.

necouchman:
Thanks @jeloneal - one more request, could you please re-base this against the `patch` branch instead of `main` - I think this qualifies as a patch/fix, and we'll get it merged into that branch.

necouchman:
@jeloneal I changed the PR request, but, as you can see, it now has 21 commits instead of just the one that you pushed. You'll need to do a `git rebase` within your local branch, and then a force `git push` to forcibly update the remote branch with those changes. This should re-align things such that the code will match the PR.

necouchman (changes requested):
@jeloneal You have a change from `GUACAMOLE-663` in this pull request, completely unrelated to the health check changes.

necouchman:
@jeloneal If you can get rid of the extra commit and rebase against `staging/1.6.1`, we can get this into the next release.

necouchman (approved):

PR #579, opened 2025-03-24
GUACAMOLE-1026: Correct inverted performance flag logic.

mike-jumper:
This change corrects the regression in handling of FreeRDP 3.x performance flags. Apparently, we simply inverted the logic through the magic of copypasta.

### DIFF
--- a/src/protocols/rdp/settings.c
+++ b/src/protocols/rdp/settings.c
@@ -1553,10 +1553,10 @@ void guac_rdp_push_settings(guac_client* client,
 
     /* Set individual flags - some FreeRDP versions overwrite flags set by guac_rdp_get_performance_flags() above */
     freerdp_settings_set_bool(rdp_settings, FreeRDP_AllowFontSmoothing, guac_settings->font_smoothing_enabled);
-    freerdp_settings_set_bool(rdp_settings, FreeRDP_DisableWallpaper, guac_settings->wallpaper_enabled);
-    freerdp_settings_set_bool(rdp_settings, FreeRDP_DisableFullWindowDrag, guac_settings->full_window_drag_enabled);
-    freerdp_settings_set_bool(rdp_settings, FreeRDP_DisableMenuAnims, guac_settings->menu_animations_enabled);
-    freerdp_settings_set_bool(rdp_settings, FreeRDP_DisableThemes, guac_settings->theming_enabled);
+    freerdp_settings_set_bool(rdp_settings, FreeRDP_DisableWallpaper, !guac_settings->wallpaper_enabled);
+    freerdp_settings_set_bool(rdp_settings, FreeRDP_DisableFullWindowDrag, !guac_settings->full_window_drag_enabled);
+    freerdp_settings_set_bool(rdp_settings, FreeRDP_DisableMenuAnims, !guac_settings->menu_animations_enabled);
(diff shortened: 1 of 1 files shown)


necouchman:
@mike-jumper The build seems to be failing with some messages regarding codecs:
```
/opt/guacamole/lib/pkgconfig/../../include/freerdp3/freerdp/codecs.h:93:9: error: 'codecs_free' is deprecated: [since 3.6.0] Use freerdp_client_codecs_free [-Werror=deprecated-declarations]
#9 221.2    93 |         WINPR_DEPRECATED_VAR("[since 3.6.0] Use freerdp_client_codecs_new",
#9 221.2       |         ^~~~~~~~~~~~~~~~~~~~
```

mike-jumper:
@necouchman Interestingly, we're not actually calling the functions warned about there. It looks like GCC just universally issues warnings for anything including the `codecs.h` header (which is effectively everything via `freerdp.h`).

I've opened FreeRDP/FreeRDP#11412 for the warnings, though am not sure this is intentional behavior on GCC's part.

necouchman:
> @necouchman Interestingly, we're not actually calling the functions warned about there. It looks like GCC just universally issues warnings for anything including the `codecs.h` header (which is effectively everything via `freerdp.h`).
> 
> I've opened [FreeRDP/FreeRDP#11412](https://github.com/FreeRDP/FreeRDP/issues/11412) for the warnings, though am not sure this is intentional behavior on GCC's part.

Okay - so are you comfortable moving forward with this merge even with the build failing? Or do you want to wait until that's sorted?

mike-jumper:
I think it's OK to move forward and possibly address the build failure separately (or rely instead on adding build flags until it's no longer an issue).

necouchman (approved):

drupol:
Hello,

I'm trying to package `guacamole-server` on Nix and having this issue. As this been fixed already ?

Follow the progress at https://github.com/NixOS/nixpkgs/pull/407726

necouchman:
@drupol: This code has been merged, but only for the coming 1.6.0 release. You can get it from the staging/1.6.0 branch of the repo.

The issue you mentioned with NixOS appears to be for Guacamole 1.5.5, so, no, it will not be fixed, there.

drupol:
Hey !

I tried using commit acb69735359d4d4a08f65d6eb0bde2a0da08f751 (from branch `staging/1.6.0`) but got the same issue. If I understand correctly, the issue stem from FreeRDP, and it looks like it hasn't been fixed yet.

necouchman:
@drupol No, this particular issue does not stem from any changes in FreeRDP, this was completely a mistake in the original code committed when we implemented support for FreeRDP 3.x, where we reversed the logic of the performance flag settings when sending those settings to FreeRDP.

drupol:
Right... I'll make further tests during the weekend and keep you posted.

necouchman:
@drupol Maybe you're referring, not to the inversion of the performance flags, but the build failure due to the deprecation warning in FreeRDP? This is actually not an issue with the Guacamole code at all - this is with the FreeRDP build. And this PR does not address that in any way - the only place we had to address that is in our build of the FreeRDP source code for our Docker images.

drupol:
I'm relating to the following build issue : https://logs.ofborg.org/?key=nixos/nixpkgs.407726&attempt_id=b4bb74d7-6b94-43db-83ad-6fd5c66eee1d

PR #580, opened 2025-03-27
GUACAMOLE-2045: Add ability to force prompting for missing credentials

pv2b:


pv2b:
So I wrote this code with the justification provided in GUACAMOLE-2045. However, when testing it, I discovered that this is probably entirely unneccessary, because there seems to be some change in behaviour where FreeRDP 3 already does this by itself. Because of that, I had to test this code against FreeRDP 2.

I'm putting the code out there because I already wrote it, but I think I need to figure out what's changed in FreeRDP to change this behaviour first, to see if this is even an idea that makes any sense.

There may be a different undesired behavioural change here, where it's not longer possible to get the graphical credential screen on a non-NLA server. I'm not entirely sure what's going on here yet, something doesn't make sense.

PR #582, opened 2025-05-01
GUACAMOLE-1026: Set ConnectionType to LAN to prevent auto-detection in FreeRDP from overriding performance flags.

aleitner:
This PR fixes an issue where RDP performance flags set in Guacamole were being ignored by newer versions of FreeRDP.

The problem occurs because FreeRDP now defaults to automatic detection of connection type, which causes the server to ignore explicitly set performance flags. 

I am currently testing this but it seems by explicitly setting the connection type to LAN, we can ensure that our performance flags are properly respected.

### DIFF
--- a/src/protocols/rdp/settings.c
+++ b/src/protocols/rdp/settings.c
@@ -1528,6 +1528,9 @@ void guac_rdp_push_settings(guac_client* client,
     /* Explicitly set flag value */
     freerdp_settings_set_uint32(rdp_settings, FreeRDP_PerformanceFlags, guac_rdp_get_performance_flags(guac_settings));
 
+    /* Set explicit connection type to LAN to prevent auto-detection */
+    freerdp_settings_set_uint32(rdp_settings, FreeRDP_ConnectionType, CONNECTION_TYPE_LAN);
+
     /* Always request frame markers */
     freerdp_settings_set_bool(rdp_settings, FreeRDP_FrameMarkerCommandEnabled, TRUE);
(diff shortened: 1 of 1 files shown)


corentin-soriano:
This seems to fix the issue!
Could you please rebase your branch from staging/1.6.0 ?

aleitner:
> This seems to fix the issue! Could you please rebase your branch from staging/1.6.0 ?

Whoops had a mishap, but rebased to staging. Glad to hear it's working in your preliminary testing!

src/protocols/rdp/settings.c
```diff
@@ -1770,6 +1773,9 @@ void guac_rdp_push_settings(guac_client* client,
     /* Explicitly set flag value */
     rdp_settings->PerformanceFlags = guac_rdp_get_performance_flags(guac_settings);
 
+    /* Set explicit connection type to LAN to prevent auto-detection */
+    freerdp_settings_set_uint32(rdp_settings, FreeRDP_ConnectionType, CONNECTION_TYPE_LAN);
```
mike-jumper, line 1777: Is this valid when `HAVE_SETTERS_GETTERS` is not defined?

mike-jumper: Confirmed - `freerdp_settings_set_uint32()` is present since FreeRDP 2.0.0, as are the `FreeRDP_ConnectionType` and `CONNECTION_TYPE_LAN` constants.

mike-jumper (approved):

PR #583, opened 2025-05-01
GUACAMOLE-377: Correct race condition in guac_display flush planning.

mike-jumper:
The actual addition of operations to the `ops` FIFO has to happen atomically, or else an earlier frame's operations may end up in the FIFO _after_ a later frame's operations.

This is the cause of the odd smearing artifacts during scrolling when the system is under heavy load.


corentin-soriano (approved):
This fix works fine, thank you!

PR #588, opened 2025-05-06
GUACAMOLE-2061: Report mouse movement as part of frame via render thread.

mike-jumper:
Doing this decouples reporting of mouse movement from the display locks and display processing, ensuring that any delays from processing the pending frame do not cause delays in processing further instructions (like "sync").

If delays in processing "sync" are tied to server-side processing, the assumptions behind the client processing lag calculations cause a feedback loop of further delays. Measurable delays in "sync" need to be purely due to client-side processing.


src/libguac/guacamole/display.h
```diff
@@ -254,6 +254,13 @@ void guac_display_notify_user_left(guac_display* display, guac_user* user);
  * mouse button. This function automatically invokes
  * guac_display_end_mouse_frame().
  *
+ * NOTE: If using guac_display_render_thread, the
+ * guac_display_render_thread_notify_user_moved_mouse() function instead. If
```
necouchman, line 258: I think this part of the comment is missing a couple of words - maybe `use the...function, instead`, or `the... function should be used, instead.` ??

mike-jumper: Oops, yes.

necouchman (approved):

PR #595, opened 2025-05-31
GUACAMOLE-2075: Ensure guac_display threads are stopped before attempting to free underlying external buffers.

mike-jumper:


src/libguac/display.c
```diff
@@ -162,15 +162,58 @@ guac_display* guac_display_alloc(guac_client* client) {
 
 }
 
-void guac_display_free(guac_display* display) {
+void guac_display_stop(guac_display* display) {
 
-    /* Stop further use of the operation FIFO */
-    guac_fifo_invalidate(&display->ops);
+    /* Ensure only one any number of concurrent calls to guac_display_stop()
```
necouchman, line 167: `only one any`?

necouchman (changes requested):
Just the one minor correction to the comment and I think this will be good to go.

src/libguac/display.c
mike-jumper: Oops, missing an "of"

necouchman (approved):

PR #597, opened 2025-06-03
GUACAMOLE-1256: Workaround for the terminal scroll down issue.

eugen-keeper:
**Vim** scroll down (i.e. scrolling to the beginning of text) is completely broken. It is annoying. We should call `guac_terminal_display_flush` before `guac_terminal_copy_rows`. Otherwise we have a wrong order of GUAC_CHAR_COPY/GUAC_CHAR_SET operations. I am aware that there was already an attempt to flush at the end of `guac_terminal_scroll_down` but it actually did not work. Then, it was removed as a part of the fix for performance degradation when dumping big text files. I believe `guac_terminal_scroll_down` is not called in that case. It mainly affects going to the beginning of text files using the **UP ARROW** key.

### DIFF
--- a/src/terminal/terminal.c
+++ b/src/terminal/terminal.c
@@ -868,15 +868,6 @@ void guac_terminal_scroll_up(guac_terminal* term,
 void guac_terminal_scroll_down(guac_terminal* term,
         int start_row, int end_row, int amount) {
 
-    /*
-     * We must flush all pending operations here because 
-     * guac_terminal_copy_rows will set new GUAC_CHAR_COPY for
-     * cells of shifted rows meanwhile preceding GUAC_CHAR_SET for these cells
-     * are not actually done yet. The GUAC_CHAR_COPY operation has the highest
-     * priority, so if we do not 'flush' here, wrong content will be copied.
-     */
-    guac_terminal_display_flush_operations(term->display);
-
     guac_terminal_copy_rows(term, start_row, end_row - amount, amount);
 
     /* Clear new area */
--- a/src/terminal/terminal/display.h
+++ b/src/terminal/terminal/display.h
@@ -225,6 +225,12 @@ typedef struct guac_terminal_display {
      */
     int selection_end_column;
 
+    /**
+     * Whether there are GUAC_CHAR_SET operations that need to be flushed
+     * to the display.
+     */
+    bool unflushed_set;
+
 } guac_terminal_display;
 
 /**
--- a/src/terminal/display.c
+++ b/src/terminal/display.c
@@ -258,6 +258,7 @@ guac_terminal_display* guac_terminal_display_alloc(guac_client* client,
     display->width = 0;
     display->height = 0;
     display->operations = NULL;
+    display->unflushed_set = false;
 
     /* Initially nothing selected */
     display->text_selected = false;
@@ -430,6 +431,12 @@ void guac_terminal_display_copy_rows(guac_terminal_display* display,
     guac_terminal_operation* src = &(display->operations[src_offset]);
     guac_terminal_operation* dst = &(display->operations[dst_offset]);
 
+    /* Flush operations if there are any unflushed SET operations. This is
+     * necessary to ensure that the copy operation does not conflict with any
+     * SET operations that may have been performed since the last flush. */
+    if (display->unflushed_set)
+        guac_terminal_display_flush_operations(display);
+
     /* Copy data */
     memmove(dst, src, guac_mem_ckd_mul_or_die(sizeof(guac_terminal_operation),
                 display->width, (end_row - start_row + 1)));
@@ -483,7 +490,7 @@ void guac_terminal_display_set_columns(guac_terminal_display* display, int row,
         /* Flush pending copy operation before adding new SET operation. This
          * avoid operation conflicts that cause inconsistent display. */
         if (current->type == GUAC_CHAR_COPY)
-            guac_terminal_display_flush(display);
+            guac_terminal_display_flush_operations(display);
 
         /* Set operation */
         current->type      = GUAC_CHAR_SET;
@@ -493,6 +500,13 @@ void guac_terminal_display_set_columns(guac_terminal_display* display, int row,
         current += character->width;
 
     }
+
+    /* Marks whether there are unflushed GUAC_CHAR_SET operations when the
+     * operation is not on the first or last row because flushing new lines
+     * added has a high performance cost. This flag is used to determine
+     * whether a flush is necessary before performing copy operations. */
+    if (row > 0 && row < display->height - 1)
+        display->unflushed_set = true;
 
 }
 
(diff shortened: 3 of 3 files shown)


src/terminal/terminal.c
```diff
@@ -868,6 +868,14 @@ void guac_terminal_scroll_up(guac_terminal* term,
 void guac_terminal_scroll_down(guac_terminal* term,
         int start_row, int end_row, int amount) {
 
+    /*
+     * This is necessary because guac_terminal_copy_rows will set new GUAC_CHAR_COPY
+     * for cells of shifted rows meanwhile preceding GUAC_CHAR_SET for these cells
+     * are not actually done yet. The GUAC_CHAR_COPY operation has the highest priority,
+     * so if we do not 'flush' here, wrong content will be copied.
+     */
+    guac_terminal_display_flush(term->display);
```
mike-jumper: A couple things:

1. While it might be necessary to get these operations handled/flushed within the terminal emulator, the internals of `guac_terminal_display_flush()` mean that this will still result in a full flush of a frame:

   https://github.com/apache/guacamole-server/blob/d80112c993a352c4e366317c0ae12919bfa61011/src/terminal/display.c#L863-L873

   If the need for this call is specifically that we need to ensure `GUAC_CHAR_COPY`, etc. are handled before performing yet more copies, then we should perform a partial flush that does just that (without flushing images via `guac_common_surface_flush()` and incurring performance penalties).
   
   If there is a way to _only_ flush the operations that need to be flushed (rather than the entire display), that would be better still.

2. Shouldn't this be at a lower level, like `guac_terminal_copy_rows()` or `guac_terminal_display_copy_rows()`? `guac_terminal_scroll_down()` is a specific instance of the issue, but it sounds like the problem is more fundamental.

eugen-keeper: I performed the research sometime ago. Even after I did it, I was confused a little about the behavior because we have mix of numbers and images there. It was difficult to simply sample frames or dump rows to understand every details. Anyway, I remember the following things.
1. `guac_common_surface_flush` is actually not necessary, the following calls are necessary:
- `__guac_terminal_display_flush_copy`
- `__guac_terminal_display_flush_clear`
- `__guac_terminal_display_flush_set`
2. Based on my research the solution could be `set` before `guac_terminal_copy_rows` and `copy/clear` at the end of `guac_terminal_scroll_down`. Although performing all the operations `guac_terminal_copy_rows` also helps 100%. Tested!
3.  I would prefer to avoid making changes inside `guac_terminal_copy_rows` because the issue happens only during `scrolling down` because the first line is set with new characters and then we have to copy them to one of the next lines. 
4. Overall, no matter how I tried to fix it a different way, as far as remember I found that the best solution is to flush between preceding `set` operations and guac_terminal_display_copy_rows.

mike-jumper: So, should we move:

https://github.com/apache/guacamole-server/blob/d80112c993a352c4e366317c0ae12919bfa61011/src/terminal/display.c#L865-L868

into its own function, perhaps something like `guac_terminal_display_flush_operations()`, and then call _that_ instead? That would at least avoid the unnecessary `guac_common_surface_flush()`.

eugen-keeper:
@necouchman, It does not affect dumping big files. Actually I was a little mistaken. Here we can see that the fix for big files did not change the function: https://github.com/apache/guacamole-server/pull/554/files
The flush in the function was removed here https://github.com/apache/guacamole-server/pull/565/files to improve latency in less. Obviously, my workaround will make scrolling to the beginning of texts slower but in practice you cannot press the **UP ARROW** too fast. Will it affect anything else?

mike-jumper:
Here's a quick script that prints the "lorem ipsum" test text 100 times, reversed and scrolling backwards via the reverse linefeed sequence (which leverages `guac_terminal_scroll_down()` in our terminal emulator implementation):

```bash
#!/bin/bash

print_reversed() {
    while read LINE; do
        echo "$LINE"
        echo -en '\x1BM\x1BM'
        sleep 0.01s
    done
}

clear

for N in `seq 1 100`; do
    print_reversed <<'EOF'
Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor
incididunt ut labore et dolore magna aliqua. Ut enim ad minim veniam, quis
nostrud exercitation ullamco laboris nisi ut aliquip ex ea commodo consequat.
Duis aute irure dolor in reprehenderit in voluptate velit esse cillum dolore eu
fugiat nulla pariatur. Excepteur sint occaecat cupidatat non proident, sunt in
culpa qui officia deserunt mollit anim id est laborum.
EOF
done
```

corentin-soriano (commented):
I ran some tests again:
- On 1.6.0 branch, I reproduce it on some VMs but not all.
- If I revert [565](https://github.com/apache/guacamole-server/pull/565/files), the issue systematically occurs when passing over braces, brackets or parentheses (highlighting the corresponding parenthesis), so there is definitely an improvement but incomplete.
- It seems that the bug only appears when holding the up arrow key.
- Doesn't seem to appear on plain text but only if there is bracket/brace/parentheses highlighting (in a json file for example).
- The change proposed here causes a lot of slowdowns.

Is the problem only related to SET operations that didn't have time to flush before the next ones, and the VMs where I can't reproduce would be slightly slower?

eugen-keeper:
> Is the problem only related to SET operations that didn't have time to flush before the next ones, and the VMs where I can't reproduce would be slightly slower?

> Yes, that is correct. I think you are right. It significantly depends on overall system/hardware performance. We have a flush in a different thread (if I remember correctly), so on some systems the issue is not reproduced. In my case I reproduce the issue on AWS virtual machines.

> And yes, there is a performance degradation in case of my changes but comparing with the first attempt of adding a flush into the function, my solution fixes the issue entirely (the original one didn't). Also, practically it will mostly affect the UP ARROW key. Where else you would use the reverse scrolling automatically. It will not affect dumping big files at all for example.

src/terminal/terminal/display.h
```diff
@@ -319,6 +319,12 @@ void guac_terminal_display_resize(guac_terminal_display* display, int width, int
 /**
  * Flushes all pending operations within the given guac_terminal_display.
  */
+void guac_terminal_display_flush_operations(guac_terminal_display* display);
```
mike-jumper, line 331: Please document the `display` parameter.

There is still old documentation in our codebase that omits documentation for parameters (like `guac_terminal_display_flush()` below), but we are including this documentation going forward at least for new/modified code.

mike-jumper (changes requested):

src/terminal/terminal/display.h
eugen-keeper: Added

mike-jumper (approved):

mike-jumper:
These changes look good to me as of the latest update. @corentin-soriano, can you confirm whether you're still seeing a performance impact with this branch and, if so, describe the steps to reproduce?

.gitignore
```diff
@@ -52,3 +52,6 @@ nbproject/
 # Compilation database, as may be generated by tools like Bear
 .cache/
 compile_commands.json
+
+# IDE
+.vscode/
```
eugen-keeper: Should this really be a part of this PR?

necouchman: No, it should not :-).

corentin-soriano: Done!

mike-jumper (approved):

PR #602, opened 2025-07-17
GUACAMOLE-2091: optimize cursor rendering and handle empty column ranges.

theorangeguo:
- Add check for empty column ranges in display_set_columns
- Improve cursor rendering by considering character width
- Adjust cursor redraw area to fit within visible columns
- Log debug message for empty column ranges

### DIFF
--- a/src/terminal/display.c
+++ b/src/terminal/display.c
@@ -486,6 +486,11 @@ void guac_terminal_display_set_columns(guac_terminal_display* display, int row,
     start_column = guac_terminal_fit_to_range(start_column, 0, display->width - 1);
     end_column   = guac_terminal_fit_to_range(end_column,   0, display->width - 1);
 
+    if (start_column > end_column) {
+        guac_client_log(display->client, GUAC_LOG_TRACE, "display_set_columns: Adjusted range is empty [%d,%d], returning.", start_column, end_column);
+        return;
+    }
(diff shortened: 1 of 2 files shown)


src/terminal/display.c
```diff
@@ -486,6 +486,10 @@ void guac_terminal_display_set_columns(guac_terminal_display* display, int row,
     start_column = guac_terminal_fit_to_range(start_column, 0, display->width - 1);
     end_column   = guac_terminal_fit_to_range(end_column,   0, display->width - 1);
 
+    if (start_column > end_column) {
+        guac_client_log(display->client, GUAC_LOG_DEBUG, "display_set_columns: Adjusted range is empty [%d,%d], returning.", start_column, end_column);
```
necouchman: This seems like it should be in `GUAC_LOG_TRACE` rather than `DEBUG`?

necouchman (changes requested):
Just one change from me, other than that looks okay. Also @orange-guo, if you could rebase and change the pull request to be against the `patch` branch instead of `main`? That way we can get the fix into the 1.6.1 release.

@mike-jumper Any comments from you?

orange-guo:
@necouchman  all comments have been resolved.

necouchman (approved):

PR #606, opened 2025-07-31
GUACAMOLE-2113: Improve warning when WoL is enabled but no MAC address is provided

svkurupath-keeper:
Proposed fix for [GUACAMOLE-2113](https://issues.apache.org/jira/browse/GUACAMOLE-2113)

- Removed extra comma(,) in guac_user_log function calls so that the full message is logged.
- Added a comma for clarity with in the log message for RDP.

Results after local testing:
```
For VNC
guacd[273689]: WARNING: Wake on LAN was requested, but no MAC address was specified.  WoL will not be sent.

For SSH
guacd[299356]: WARNING: WoL was enabled, but no MAC address was provided.  WoL will not be sent.

For Telnet
guacd[298001]: WARNING: Wake on LAN was requested, but no MAC address was specified.  WoL will not be sent.

For RDP
guacd[299506]: WARNING: WoL requested, but no MAC address specified.  WoL will not be sent.
```

### DIFF
--- a/src/protocols/rdp/settings.c
+++ b/src/protocols/rdp/settings.c
@@ -1331,8 +1331,8 @@ guac_rdp_settings* guac_rdp_parse_args(guac_user* user,
         
         /* If WoL has been requested but no MAC address given, log a warning. */
         if(strcmp(argv[IDX_WOL_MAC_ADDR], "") == 0) {
-            guac_user_log(user, GUAC_LOG_WARNING, "WoL requested, but no MAC "
-                    "address specified.  WoL will not be sent.");
+            guac_user_log(user, GUAC_LOG_WARNING, "WoL was enabled, but no "
+                    "MAC address was provided. WoL will not be sent.");
             settings->wol_send_packet = 0;
         }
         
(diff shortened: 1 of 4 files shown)


mike-jumper (changes requested):
With respect to the changes themselves, LGTM.

A couple things need to be addressed before this is mergeable:

* As a minor fix/improvement that does not impact documentation, this should be against `patch` instead of `main`.

* The commit message needs to be formatted to match established formatting (JIRA issue followed by a colon, etc.). See established git history for examples of this.

Since each of the messages is being modified as a part of this, these changes could be a good opportunity to make these messages consistent (ie: `WoL requested` vs. `WoL was enabled` vs. `Wake on LAN was requested` vs. ...). I don't think that's deal-breaking, though.

svkurupath-keeper:
> With respect to the changes themselves, LGTM.
> 
> A couple things need to be addressed before this is mergeable:
> 
> * As a minor fix/improvement that does not impact documentation, this should be against `patch` instead of `main`. (See [the documentation covering our branching scheme](https://cwiki.apache.org/confluence/pages/viewpage.action?pageId=293047342#VersionNumberingandBranchingScheme-BranchNames).)
> * The commit message needs to be formatted to match established formatting (JIRA issue followed by a colon, etc.).
> 
> Since each of the messages is being modified as a part of this, these changes could be a good opportunity to make these messages consistent (ie: `WoL requested` vs. `WoL was enabled` vs. `Wake on LAN was requested` vs. ...). I don't think that's deal-breaking, though.

Changed the target to `apache:patch`.
Changed the commit messages to reflect the established style for the project.
With the new changes, the same message will be used for logging each time when a WoL MAC Address is not specified.

PR #608, opened 2025-08-14
GUACAMOLE-2120: Support controlling console codes sent for function keys.

eugen-keeper:


src/terminal/terminal/terminal.h
```diff
@@ -139,6 +139,24 @@ typedef enum guac_terminal_cursor_type {
...
+ */
+typedef enum guac_terminal_func_keys_and_keypad {
+
+    /**
+     * Not set.
+     */
+    GUAC_TERMINAL_FUNC_KEYS_AND_KEYPAD_DEFAULT,
+
+    /**
+     * VT!00+
```
mike-jumper: Accidentally put an "uppercase 1" here. ;)

src/terminal/terminal/terminal.h
```diff
@@ -139,6 +139,24 @@ typedef enum guac_terminal_cursor_type {
...
+/**
+ * All possible families of codes (e.g. vt100) which can be used when you push
+ * the function and keypad keys.
+ */
+typedef enum guac_terminal_func_keys_and_keypad {
+
+    /**
+     * Not set.
+     */
+    GUAC_TERMINAL_FUNC_KEYS_AND_KEYPAD_DEFAULT,
```
mike-jumper: Rather than define this as simply the "default" mode, is there an actual named family of codes that is that default? Or, put another way: if VT100+ _were_ the default, what would this be called?

We should also explicitly define and handle a parameter value for whatever this default happens to be called, so that it can be explicitly selected by an administrator that doesn't want to rely on the default staying the same.

src/terminal/terminal.c
```diff
@@ -1529,7 +1537,56 @@ static int __guac_terminal_send_key(guac_terminal* term, int keysym, int pressed
...
+                    case '-':
+                        return guac_terminal_send_string(term, "\x1BOm");
+                    case '.':
+                        return guac_terminal_send_string(term, "\x1BOn");
+                    /*
+                     * There is no comma on the regular keypad, so we use '*' to send
+                     * the appropriate code.
+                     */
+                    case '*':
+                        return guac_terminal_send_string(term, "\x1BOl");
```
mike-jumper: Does VT100+ lack a `*` on the keypad, so this is safe?

src/terminal/terminal.c
```diff
@@ -1529,7 +1537,56 @@ static int __guac_terminal_send_key(guac_terminal* term, int keysym, int pressed
...
+                    /*
+                     * At least in some environments JS does not differentiate regular
+                     * and keypad ENTER and its location property is always 0. So we
+                     * use '+' to send the appropriate ENTER escape code.
+                     */
+                    case '+':
+                        return guac_terminal_send_string(term, "\x1BOM");
```
mike-jumper: Does VT100+ lack a `+` on the keypad, so this is safe?

src/terminal/terminal.c
```diff
@@ -1529,7 +1537,56 @@ static int __guac_terminal_send_key(guac_terminal* term, int keysym, int pressed
...
+                    case '5':
+                        return guac_terminal_send_string(term, "\x1BOu");
+                    case '6':
+                        return guac_terminal_send_string(term, "\x1BOv");
+                    case '7':
+                        return guac_terminal_send_string(term, "\x1BOw");
+                    case '8':
+                        return guac_terminal_send_string(term, "\x1BOx");
+                    case '9':
+                        return guac_terminal_send_string(term, "\x1BOy");
```
mike-jumper: Perhaps this could be calculated instead of explicitly listing each mapping from 0-9 onto p-y?

src/terminal/terminal.c
```diff
@@ -504,6 +505,13 @@ guac_terminal* guac_terminal_create(guac_client* client,
     /* Configure backspace */
     term->backspace = options->backspace;
 
+    /* Configure the family of codes for function keys and the keypad */
+    if (options->func_keys_and_keypad == NULL || options->func_keys_and_keypad[0] == 0) {
+        term->func_keys_and_keypad = GUAC_TERMINAL_FUNC_KEYS_AND_KEYPAD_DEFAULT;
+    } else if (strcmp(options->func_keys_and_keypad, "vt100plus") == 0) {
```
mike-jumper: Please don't cuddle the `else if`. See: https://cwiki.apache.org/confluence/display/GUAC/Contribution+and+Style+Guidelines#ContributionandStyleGuidelines-Braces

mike-jumper: Any particular reason why `vt100plus` instead of `vt100+` or `vt100`? The value `vt100` would be in line with the familiar value accepted by `TERM`.

src/terminal/terminal/terminal.h
eugen-keeper: Fixed

src/terminal/terminal/terminal.h
eugen-keeper: There is no specific name for the default escape codes. That's why I just called it default. I beleive ANSI would be too wide term because even VT100 is ANSI. So, I now called it GUAC_TERMINAL_FUNC_KEYS_AND_KEYPAD_ESC with the "ESC[n~" title same as Putty has.

src/terminal/terminal.c
eugen-keeper: Yes, that is safe. The VT100 keypad has 4 function keys, 10 numbers, dash, comma, period and ENTER.

src/terminal/terminal.c
eugen-keeper: Yes, there is no + on VT100 keypad.

src/terminal/terminal.c
eugen-keeper: Fixed

src/terminal/terminal.c
eugen-keeper: Fixed

eugen-keeper: Initially I called the new mode VT100+ because Putty uses the same name. It is possible they had added + because the original VT100 did not have color support, then the support was added. Or maybe because the original VT100 keyboard only had 4 function keys but modern keyboard have more function keys. I think anything we call it can be considered ambiguous. So for now we have the following names:
GUAC_TERMINAL_FUNC_KEYS_AND_KEYPAD_ESC -> "ESC[n~" (default; same as Putty has)
GUAC_TERMINAL_FUNC_KEYS_AND_KEYPAD_VT100 -> "vt100" (similar to what we have for the terminal; although it is important to remember that this is not related to a terminal type directly)

PR #609, opened 2025-08-15
GUACAMOLE-2107: Fix the error handling within the guac_tcp_connect() function.

necouchman:
This pull request makes a couple of changes within the `guac_tcp_connect()` function that correct issues that cause the Wake-on-LAN functionality to break, and probably also break other protocols that rely on this function:
* The function returned either the socket that was connected, or various error codes from other calls to functions like `connect()`, `select()`, etc. The callers of `guac_tcp_connect()` rely on the return value being either a socket or a negative integer upon failure, so I've updated many of the instances of the `fd` variable that might get returned to be -1 in the instances where it would be returned.
* The simple call to the `select()` function for waiting for a timeout when attempting to connect to a non-blocking socket requires a bit more checking to see if it has succeeded or failed, which involves querying the socket with `getsockopt()` and looking at the error message.

### DIFF
--- a/src/libguac/tcp.c
+++ b/src/libguac/tcp.c
@@ -25,6 +25,7 @@
 #include <fcntl.h>
 #include <netdb.h>
 #include <netinet/in.h>
+#include <string.h>
 #include <sys/select.h>
 #include <sys/socket.h>
 #include <unistd.h>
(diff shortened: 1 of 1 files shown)


src/libguac/tcp.c
```diff
@@ -108,12 +111,46 @@ int guac_tcp_connect(const char* hostname, const char* port, const int timeout)
...
                 
                 retval = select(fd + 1, NULL, &fdset, NULL, &tv);
+
+                if (retval > 0) {
+                    int so_error = 0;
+                    socklen_t so_error_len = sizeof(so_error);
+                    retval = getsockopt(fd, SOL_SOCKET, SO_ERROR, &so_error, &so_error_len);
+
+                    // Error occurs retrieving the socket status, so we set it
+                    // and move on.
```
mike-jumper: We should prefer `/* C-style block comments */` to `// C++-style comments` in C code (same elsewhere in these changes).

necouchman: Should be fixed!

mike-jumper (approved):

PR #610, opened 2025-08-21
GUACAMOLE-522: USB Device Redirection.

aleitner:
This PR implements the server-side protocol support for Web USB device redirection, complementing the [client-side PR](https://github.com/apache/guacamole-client/pull/1103). This provides the infrastructure for Guacamole protocol plugins (RDP, VNC, etc.) to handle USB device redirection from browser clients.

## Data Flow

1. Server receives `usbconnect` with device descriptors and interface/endpoint data
2. Protocol plugin's `usbconnect_handler` is invoked to establish redirection
3. `usbdata` messages flow between client device and server handler
4. Either side can initiate disconnection via `usbdisconnect`

### New Protocol Instructions
- **Server → Client**: Added `usbdata` and `usbdisconnect` instructions for sending data to USB devices and requesting disconnection
- **Client → Server**: Added handlers for receiving `usbconnect`, `usbdata`, and `usbdisconnect` instructions

### Handlers
- `guac_user_usbconnect_handler`: Processes USB device connections with full device metadata (vendor/product IDs, interfaces, endpoints)
- `guac_user_usbdata_handler`: Handles USB data transfer from the client with USB endpoint and transfer type information. 
- `guac_user_usbdisconnect_handler`: Manages USB device disconnection events.

### DIFF
--- a/src/libguac/guacamole/protocol.h
+++ b/src/libguac/guacamole/protocol.h
@@ -1155,5 +1155,44 @@ guac_protocol_version guac_protocol_string_to_version(const char* version_string
  */
 const char* guac_protocol_version_to_string(guac_protocol_version version);
 
-#endif
+/**
+ * Sends a usbdisconnect instruction over the given guac_socket connection,
+ * requesting that the client disconnect the specified USB device.
+ *
+ * @param socket
+ *     The guac_socket connection to use.
+ *
+ * @param device_id
+ *     The unique identifier of the USB device that should be disconnected.
+ *
+ * @return
+ *     Zero on success, non-zero on error.
+ */
+int guac_protocol_send_usbdisconnect(guac_socket* socket, 
+        const char* device_id);
 
+/**
+ * Sends a usbdata instruction over the given guac_socket connection,
+ * sending USB data to a specific endpoint on a client-side USB device.
+ *
+ * @param socket
+ *     The guac_socket connection to use.
+ *
+ * @param device_id
+ *     The unique identifier of the USB device.
+ *
+ * @param endpoint_number
+ *     The USB endpoint number identifying the logical communication channel
+ *     for this data transfer. The valid endpoint numbers for a device are
+ *     provided in the interface_data parameter during usbconnect.
+ *
+ * @param data
+ *     The base64-encoded data to send to the USB device.
+ *
+ * @return
+ *     Zero on success, non-zero on error.
+ */
+int guac_protocol_send_usbdata(guac_socket* socket, const char* device_id,
+        int endpoint_number, const char* data);
+
+#endif
--- a/src/libguac/guacamole/user-fntypes.h
+++ b/src/libguac/guacamole/user-fntypes.h
@@ -499,5 +499,95 @@ typedef int guac_user_get_handler(guac_user* user, guac_object* object,
 typedef int guac_user_put_handler(guac_user* user, guac_object* object,
         guac_stream* stream, char* mimetype, char* name);
 
-#endif
+/**
+ * Handler for Guacamole USB connect events, invoked when a "usbconnect" 
+ * instruction has been received from a user. This indicates that the user
+ * has connected a USB device via WebUSB and it is available for redirection.
+ *
+ * @param device_id
+ *     The unique identifier for the USB device. Required.
+ *
+ * @param vendor_id
+ *     The vendor ID of the USB device. Required.
+ *
+ * @param product_id
+ *     The product ID of the USB device. Required.
+ *
+ * @param device_name
+ *     The human-readable name of the device. Required (may be empty string).
+ *
+ * @param serial_number
+ *     The serial number of the device. Optional (may be NULL or empty string
+ *     if not available).
+ *
+ * @param device_class
+ *     The USB device class. Required.
+ *
+ * @param device_subclass
+ *     The USB device subclass. Required.
+ *
+ * @param device_protocol
+ *     The USB device protocol. Required.
+ *
+ * @param interface_data
+ *     Encoded string containing interface and endpoint information. Required.
+ *     Format: "ifaceNum:class:subclass:protocol:ep1Num:ep1Dir:ep1Type:ep1Size;ep2...,iface2..."
+ *     where multiple endpoints within an interface are separated by semicolons
+ *     and multiple interfaces are separated by commas.
+ *
+ * @return
+ *     Zero if the USB connect event was handled successfully, or non-zero if 
+ *     an error occurred.
+ */
+typedef int guac_user_usbconnect_handler(guac_user* user, const char* device_id,
+    int vendor_id, int product_id, const char* device_name, 
+    const char* serial_number, int device_class, int device_subclass,
+    int device_protocol, const char* interface_data);
 
+/**
+* Handler for Guacamole USB data events, invoked when a "usbdata" instruction
+* has been received from a user. This carries data from a client-side USB 
+* device to be processed on the server side.
+*
+* @param user
+*     The user that sent the USB data.
+*
+* @param device_id
+*     The unique identifier for the USB device.
+*
+* @param endpoint_number
+*     The USB endpoint number that the data originated from. Endpoint
+*     numbers correspond to the endpoints defined in the device's
+*     interface_data from the usbconnect instruction.
+*
+* @param data
+*     The base64-encoded USB data.
+*
+* @param transfer_type
+*     The type of USB transfer (bulk, interrupt, isochronous, control).
+*
+* @return
+*     Zero if the USB data was handled successfully, or non-zero if an error
+*     occurred.
+*/
+typedef int guac_user_usbdata_handler(guac_user* user, const char* device_id,
+    int endpoint_number, const char* data, const char* transfer_type);
+
+/**
+* Handler for Guacamole USB disconnect events, invoked when a "usbdisconnect"
+* instruction has been received from a user. This indicates that the user
+* has disconnected a USB device that was previously available for redirection.
+*
+* @param user
+*     The user that disconnected the USB device.
+*
+* @param device_id
+*     The unique identifier for the USB device that was disconnected.
+*
+* @return
+*     Zero if the USB disconnect event was handled successfully, or non-zero 
+*     if an error occurred.
+*/
+typedef int guac_user_usbdisconnect_handler(guac_user* user, const char* device_id);
+
+#endif
--- a/src/libguac/guacamole/user.h
+++ b/src/libguac/guacamole/user.h
@@ -538,6 +538,66 @@ struct guac_user {
      */
     guac_user_touch_handler* touch_handler;
 
+/**
+     * Handler for USB connect events sent by the Guacamole web-client.
+     *
+     * The handler takes the user that connected the device, the device ID
+     * (required), vendor ID (required), product ID (required), device name
+     * (required), serial number (optional - may be NULL or empty), device class
+     * (required), device subclass (required), device protocol (required), and
+     * interface data (required).
+     *
+     * Example:
+     * @code
+     *     int usbconnect_handler(guac_user* user, const char* device_id,
+     *             int vendor_id, int product_id, const char* device_name,
+     *             const char* serial_number, int device_class, 
+     *             int device_subclass, int device_protocol, 
+     *             const char* interface_data);
+     *
+     *     int guac_user_init(guac_user* user, int argc, char** argv) {
+     *         user->usbconnect_handler = usbconnect_handler;
+     *     }
+     * @endcode
+     */
+    guac_user_usbconnect_handler* usbconnect_handler;
+
+    /**
+     * Handler for USB data events sent by the Guacamole web-client.
+     *
+     * The handler takes the user that sent the data, the device ID,
+     * endpoint number, the base64-encoded data, and the transfer type.
+     *
+     * Example:
+     * @code
+     *     int usbdata_handler(guac_user* user, const char* device_id,
+     *             int endpoint_number, const char* data,
+     *             const char* transfer_type);
+     *
+     *     int guac_user_init(guac_user* user, int argc, char** argv) {
+     *         user->usbdata_handler = usbdata_handler;
+     *     }
+     * @endcode
+     */
+    guac_user_usbdata_handler* usbdata_handler;
+
+    /**
+     * Handler for USB disconnect events sent by the Guacamole web-client.
+     *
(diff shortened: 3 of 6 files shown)


necouchman:
@coolkingh Yes, this is the framework for it that provides methods for handling USB device redirection. This particular pull request does not implement it concretely for any of the protocols. Once this pull request is merged the changes can be made at the protocol level (RDP, SSH, etc.) that would actually enable this to function for a particular protocol.

coolkingh:
@necouchman  What's the point of merging if implementation is incomplete for any of the protocol.

src/libguac/guacamole/protocol.h
```diff
@@ -1155,5 +1155,42 @@ guac_protocol_version guac_protocol_string_to_version(const char* version_string
...
+ *     The endpoint number to send data to.
+ *
+ * @param data
+ *     The base64-encoded data to send to the USB device.
+ *
+ * @return
+ *     Zero on success, non-zero on error.
+ */
+int guac_protocol_send_usbdata(guac_socket* socket, const char* device_id,
+        int endpoint_number, const char* data);
```
necouchman, line 1196: Should there be a matching `connect` function implemented, here? I understand that the `disconnect` function is provided so that a user could disconnect a USB device while a connection is still active - I think it would also be useful to be able to connect a USB device after a connection has already started, which would likely require a `connect` function?

Or is there some other reason to implement a protocol-level `disconnect` but not a `connect`?

src/libguac/guacamole/user-fntypes.h
```diff
@@ -499,5 +499,93 @@ typedef int guac_user_get_handler(guac_user* user, guac_object* object,
...
+ *     format: "iface_num:class:subclass:protocol:ep_num:dir:type:size,..."
+ *
+ * @return
+ *     Zero if the USB connect event was handled successfully, or non-zero if 
+ *     an error occurred.
+ */
+typedef int guac_user_usbconnect_handler(guac_user* user, const char* device_id,
+    int vendor_id, int product_id, const char* device_name, 
+    const char* serial_number, int device_class, int device_subclass,
+    int device_protocol, const char* interface_data);
```
necouchman, line 545: It might be good to indicate more explicitly which of these arguments is required and which is optional?One of them - `serial_number` - says "if available", which indicates it may be optional - is that the only one that isn't required and is optional, and the rest are required?

src/libguac/guacamole/protocol.h
```diff
@@ -1155,5 +1155,42 @@ guac_protocol_version guac_protocol_string_to_version(const char* version_string
...
+ * sending USB data to a specific endpoint on a client-side USB device.
+ *
+ * @param socket
+ *     The guac_socket connection to use.
+ *
+ * @param device_id
+ *     The unique identifier of the USB device.
+ *
+ * @param endpoint_number
+ *     The endpoint number to send data to.
```
necouchman: I poked around in the code for anything else that refers to `endpoint_number`, and I don't see this mentioned anywhere else. This may need a little more documentation - the rest of the arguments, here, are pretty self-explanatory, but `endpoint_number` doesn't really mean much to me?

src/libguac/guacamole/user-fntypes.h
```diff
@@ -499,5 +499,93 @@ typedef int guac_user_get_handler(guac_user* user, guac_object* object,
...
+* device to be processed on the server side.
+*
+* @param user
+*     The user that sent the USB data.
+*
+* @param device_id
+*     The unique identifier for the USB device.
+*
+* @param endpoint_number
+*     The endpoint number the data originated from.
```
necouchman: As with the mention about in `protocol.h` - endpoint_number probably needs some more thorough documentation as it appears to be a newly-introduced concept, here.

src/libguac/guacamole/user.h
```diff
@@ -538,6 +538,64 @@ struct guac_user {
      */
     guac_user_touch_handler* touch_handler;
 
+/**
+     * Handler for USB connect events sent by the Guacamole web-client.
+     *
+     * The handler takes the user that connected the device, the device ID,
+     * vendor ID, product ID, device name, serial number, device class,
+     * device subclass, device protocol, and interface data.
```
necouchman: As mentioned, above, it might be good to document which of these is required and which may be optional.

src/libguac/user-handlers.h
```diff
@@ -231,6 +231,27 @@ __guac_instruction_handler __guac_handshake_name_handler;
  */
 __guac_instruction_handler __guac_handshake_timezone_handler;
 
+/**
+ * Internal initial handler for the usbconnect instruction. When a usbconnect
+ * instruction is received, this handler will be called. The client's
```
necouchman: `The client's` should probably be `The user's`, as this is a user handler, not a client handler?

src/libguac/user-handlers.h
```diff
@@ -231,6 +231,27 @@ __guac_instruction_handler __guac_handshake_name_handler;
...
+/**
+ * Internal initial handler for the usbconnect instruction. When a usbconnect
+ * instruction is received, this handler will be called. The client's 
+ * usbconnect handler will be invoked if defined.
+ */
+__guac_instruction_handler __guac_handle_usbconnect;
+
+/**
+ * Internal initial handler for the usbdata instruction. When a usbdata
+ * instruction is received, this handler will be called. The client's
```
necouchman: `client's` -> `user's`

src/libguac/user-handlers.h
```diff
@@ -231,6 +231,27 @@ __guac_instruction_handler __guac_handshake_name_handler;
...
+ * Internal initial handler for the usbdata instruction. When a usbdata
+ * instruction is received, this handler will be called. The client's 
+ * usbdata handler will be invoked if defined.
+ */
+__guac_instruction_handler __guac_handle_usbdata;
+
+/**
+ * Internal initial handler for the usbdisconnect instruction. When a 
+ * usbdisconnect instruction is received, this handler will be called. 
+ * The client's usbdisconnect handler will be invoked if defined.
```
necouchman: `client's` -> `user's`

necouchman (changes requested):
@aleitner Overall looks pretty good to me - my comments are mostly about documentation and not the actual implementation, aside from the one question about a protocol-level `usbconnect` function.

src/libguac/guacamole/protocol.h
aleitner: I should clarify the architecture here! The asymmetry you noticed is actually intended by design. USB devices are physically connected to the client and redirected TO the server, not the other way around. The flow works like this:

1. Client detects USB device and sends `usbconnect` instruction to server
2. Server's `usbconnect_handler` decides whether to accept the device
3. Once accepted, server can send `usbdata` instructions to communicate with the device
4. Either side can terminate the redirection with `usbdisconnect`

Since the server never initiates USB device connections on the client side (it can only respond to client-offered devices), there's no need for a `guac_protocol_send_usbconnect` function. THe disconnect function exists so the server can request the client to stop redirecting a device that was previously connected.

Let me know if I am misunderstanding what you are proposing!

src/libguac/guacamole/protocol.h
aleitner: I'll update documentation, but just to clarify USB endpoints are communication channels within a USB device. Each endpoint has a number and direction (IN/OUT). Multiple endpoints allow simultaneous communication channels over the same USB connection. The valid endpoint numbers for a device should be provided in the `interface_data` during usbconnect

src/libguac/guacamole/user-fntypes.h
aleitner: Currently serial_number is the only optional parameter (can be NULL or empty). All others are required at the moment. Will update the documentation.

src/libguac/user-handlers.h
aleitner: Updated comment.

src/libguac/user-handlers.h
aleitner: Updated comment.

src/libguac/user-handlers.h
aleitner: Updated comment.

src/libguac/guacamole/user.h
aleitner: Updated comment.

src/libguac/guacamole/user-fntypes.h
aleitner: Updated comment.

src/libguac/guacamole/protocol.h
necouchman: Thanks @aleitner, this is very helpful, and I believe addresses the question I was asking (Why there wouldn't be a server-side, protocol-level `usbconnect` function).

src/libguac/guacamole/user-fntypes.h
```diff
@@ -499,5 +499,95 @@ typedef int guac_user_get_handler(guac_user* user, guac_object* object,
...
+ *     The USB device subclass. Required.
+ *
+ * @param device_protocol
+ *     The USB device protocol. Required.
+ *
+ * @param interface_data
+ *     Encoded string containing interface and endpoint information. Required.
+ *     Format: "ifaceNum:class:subclass:protocol:ep1Num:ep1Dir:ep1Type:ep1Size;ep2...,iface2..."
+ *     where multiple endpoints within an interface are separated by semicolons
+ *     and multiple interfaces are separated by commas.
```
necouchman, line 536: A couple of more question from me, here, regarding this...

First, are the format and fields of this `interface_data` parameter something standard an well-known/documented when dealing with USB devices? Is this data that will be provided automatically by the web browser, or something that it would be expected that someone would know how and what to assemble for it? I can kind of guess what each of the fields means, but a few are a bit obscure.

Somewhat related, it looks like some of the data in this `interface_data` field is the same as the other parameters to the function (class, subclass, protocol). How does that work out between the parameters provided in the function and the interface data? Should they always be identical? Or are there situations where the individual interfaces will differ from the overall "device" being connected?

I ask for two reasons:
1. Curiosity, so I understand better how it works.
2. To make sure we're not sending duplicate data, both for efficiency but also because of the risk that it'll be mis-matched.

necouchman (changes requested):
@aleitner One more comment, but, also, it looks like you added the RDP functionality, here - not sure if that was intended or not? Also, with that change, the check of the PR is failing, as the build has an issue.

src/libguac/guacamole/user-fntypes.h
aleitner: The format `ifaceNum:class:subclass:protocol:ep1Num:ep1Dir:ep1Type:ep1Size;ep2...,iface2...` is not a USB standard. It's a custom encoding I set up for this Guacamole protocol.

The data itself comes from standard USB descriptors that the browser's WebUSB API provides automatically. We can see in [ManagedUSB.js](https://github.com/aleitner/guacamole-client/blob/82a69a62b4d126d0905763ba6b00d806e7ef3307/guacamole/src/main/frontend/src/app/client/types/ManagedUSB.js#L116) in our client PR that we're just reading properties that WebUSB gives us:

```iface.interfaceNumber           // Standard USB interface number
alt.interfaceClass              // Standard USB class code
alt.interfaceSubclass           // Standard USB subclass code  
alt.interfaceProtocol           // Standard USB protocol code
ep.endpointNumber               // Standard USB endpoint number
ep.direction                    // 'in' or 'out'
ep.type                         // 'bulk', 'interrupt', 'isochronous'
ep.packetSize                   // Max packet size in bytes
```

These are all standard USB descriptor fields defined in the USB specification. We're just encoding them into a string format for sending over the Guacamole protocol.

As you pointed out thoguh we can see that USB has class/subclass/protocol at two levels.

These describe the overall device and are often set to 0 for composite devices:
- `device_class`, `device_subclass`, `device_protocol`

Composite devices like webcams with microphones commonly use:
- Device class 0xEF (Miscellaneous Device)
- Device subclass 0x02 (Common Class)
- Device protocol 0x01 (Interface Association Descriptor)

-----

These describe each individual interface and is where the actual functionality is usually defined:
- `interfaceClass`, `interfaceSubclass`, `interfaceProtocol`  

HID devices like keyboards/mice commonly have:
- Device class 0x00 (defined at interface level)
- Interface class 0x03 (HID)
- Interface subclass 0x01 (Boot Interface)
- Interface protocol 0x01 (Keyboard) or 0x02 (Mouse)
- Interface 0: class=0x03 (HID), subclass=0x01 (Boot), protocol=0x01 (Keyboard)

necouchman: Okay - I'd suggest this all of this - both the encoding of the fields for Guacamole, and the way composite devices are handled - get documented _somewhere_ - I'm not really sure if the code is the best place, or the manual, but I think it'd be useful for folks looking to implement USB functionality with Guacamole to know how we're handling this data.

src/libguac/user-handlers.c
```diff
@@ -55,6 +55,9 @@ __guac_instruction_handler_mapping __guac_instruction_handler_map[] = {
    {"audio",      __guac_handle_audio},
    {"argv",       __guac_handle_argv},
    {"nop",        __guac_handle_nop},
+   {"usbconnect",    __guac_handle_usbconnect},
+   {"usbdata",       __guac_handle_usbdata},
+   {"usbdisconnect", __guac_handle_usbdisconnect},
```
necouchman, line 60: One minor style nit-pick, here, but the rest of the lines are lined up so nicely, so it probably makes sense to bump them all out to the new required level to line them up.

necouchman (changes requested):
One more style nitpick and I think this is ready to merge.

src/libguac/user-handlers.c
aleitner: realigned

DerekMcCauley2002:
Hello.
Regarding https://issues.apache.org/jira/browse/GUACAMOLE-522, does the comment involving Smartcard read get address to some degree within this implementation? This is a question stemming from a desire to have Smartcard certs (which seem to be natively shared between web client and web app) have USB passthrough of data directly to, most primarily, the RDP client and it's URBDRC functionality?

necouchman:
@DerekMcCauley2002:
As mentioned [above](https://github.com/apache/guacamole-server/pull/610#issuecomment-3723990594), this particular PR does not implement anything outside of the framework for protocol-level support for USB direction. The concrete details of things like Smartcard Readers and webcams will be implemented in future pull requests.

necouchman (approved):

PR #611, opened 2025-08-22
GUACAMOLE-2072: RAIL plugin should be loaded in the LoadPlugins callback.

necouchman:
This is likely the first of a couple of changes to correct issues with RAIL support in FreeRDP3 and in Guacamole's support of GFX. I finally stumbled upon the reason that the FreeRDP3 changes had broken RAIL support - the RAIL plugin was being loaded _only_ the `PreConnect` callback (`rdp_freerdp_pre_connect()`), and not in the `LoadChannels` callback (`rdp_freerdp_load_channels()`). The change in FreeRDP (https://github.com/FreeRDP/FreeRDP/commit/ef8c9e48e8acf61e16cb9fd547799067c87b20da) that broke this unified the reloading of channels into a single function that disconnected and closed all of the channels and then used the `LoadChannels` callback to reload all of them. Since RAIL was not being processed through that callback, it would get dropped but never reconnected. This moves the processing of the RAIL plugin, along with any user-defined static plugins, to the `LoadChannels` callback, which keeps it connected in recent FreeRDP 3.x versions.

I think these changes are solid and ready to be merged - it's pretty simple, and seems to fix at least the RAIL plugin connection issue. That said, I'm still seeing some issues with RAIL when GFX support is enabled (hangs/disconnects/freezes) that don't occur when you disable GFX support, so I think there may be additional items to address related to the intersection of GFX + RAIL.

### DIFF
--- a/src/protocols/rdp/channels/rail.c
+++ b/src/protocols/rdp/channels/rail.c
@@ -64,24 +64,6 @@ static UINT guac_rdp_rail_complete_handshake(RailClientContext* rail) {
     guac_client* client = (guac_client*) rail->custom;
     guac_rdp_client* rdp_client = (guac_rdp_client*) client->data;
 
-    RAIL_HANDSHAKE_ORDER handshake = {
-
-        /* Build number 7600 (0x1DB0) apparently represents Windows 7 and
-         * compatibility with RDP 7.0. As of this writing, this is the same
-         * build number sent for RAIL connections by xfreerdp. */
-        .buildNumber = 7600
-
-    };
-
-    /* Send client handshake response */
-    guac_client_log(client, GUAC_LOG_TRACE, "Sending RAIL handshake.");
-    pthread_mutex_lock(&(rdp_client->message_lock));
-    status = rail->ClientHandshake(rail, &handshake);
-    pthread_mutex_unlock(&(rdp_client->message_lock));
-
-    if (status != CHANNEL_RC_OK)
-        return status;
-
     RAIL_CLIENT_STATUS_ORDER client_status = {
         .flags =
                 TS_RAIL_CLIENTSTATUS_ALLOWLOCALMOVESIZE
--- a/src/protocols/rdp/settings.c
+++ b/src/protocols/rdp/settings.c
@@ -1684,6 +1684,7 @@ void guac_rdp_push_settings(guac_client* client,
         freerdp_settings_set_bool(rdp_settings, FreeRDP_Workarea, TRUE);
         freerdp_settings_set_bool(rdp_settings, FreeRDP_RemoteApplicationMode, TRUE);
         freerdp_settings_set_bool(rdp_settings, FreeRDP_RemoteAppLanguageBarSupported, TRUE);
+        freerdp_settings_set_bool(rdp_settings, FreeRDP_HiDefRemoteApp, guac_settings->enable_gfx);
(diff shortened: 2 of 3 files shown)


mike-jumper:
NICE!!! Looking through this now...

src/protocols/rdp/settings.c
```diff
@@ -1683,7 +1683,8 @@ void guac_rdp_push_settings(guac_client* client,
     if (guac_settings->remote_app != NULL) {
         freerdp_settings_set_bool(rdp_settings, FreeRDP_Workarea, TRUE);
         freerdp_settings_set_bool(rdp_settings, FreeRDP_RemoteApplicationMode, TRUE);
-        freerdp_settings_set_bool(rdp_settings, FreeRDP_RemoteAppLanguageBarSupported, TRUE);
+        freerdp_settings_set_bool(rdp_settings, FreeRDP_RemoteAppLanguageBarSupported, FALSE);
```
mike-jumper: Why is this changing to `FALSE`?

src/protocols/rdp/channels/rail.c
```diff
@@ -64,24 +64,6 @@ static UINT guac_rdp_rail_complete_handshake(RailClientContext* rail) {
...
-         * compatibility with RDP 7.0. As of this writing, this is the same
-         * build number sent for RAIL connections by xfreerdp. */
-        .buildNumber = 7600
-
-    };
-
-    /* Send client handshake response */
-    guac_client_log(client, GUAC_LOG_TRACE, "Sending RAIL handshake.");
-    pthread_mutex_lock(&(rdp_client->message_lock));
-    status = rail->ClientHandshake(rail, &handshake);
```
mike-jumper, line 79: Is this handled automatically by FreeRDP?

necouchman: Yes, see: https://github.com/FreeRDP/FreeRDP/issues/11785#issuecomment-3197896683

src/protocols/rdp/settings.c
necouchman: I think I changed it as part of the debugging I was doing, I can definitely flip that back...although, I'm wondering, do we actually support the language bar? I've never seen it pop up in my use of RAIL support, so not sure if we should be saying we support that or not?

necouchman: @mike-jumper Any further thoughts on reverting this or leaving disabled?

mike-jumper: I think we should revert this unless things do not work without it disabled.

IIRC, the language bar is the UI component of Windows that allows switching between input methods, and would be needed for languages like Chinese and Japanese.

necouchman: Okay, this should be reverted, now, via rebase.

mike-jumper (approved):

necouchman:
> > That said, I'm still seeing some issues with RAIL when GFX support is enabled (hangs/disconnects/freezes) that don't occur when you disable GFX support, so I think there may be additional items to address related to the intersection of GFX + RAIL.
> 
> I wonder if this might have something to do with:
> 
> ```c
> freerdp_settings_set_bool(rdp_settings, FreeRDP_HiDefRemoteApp, guac_settings->enable_gfx);
> ```
> 
> (https://github.com/necouchman/guacamole-server/blob/10f2ccf3050f4c720b94d3a07a4753083556b80d/src/protocols/rdp/settings.c#L1687)
> 
> I'm not sure why that would cause issues, but it is at least one direct connection between RemoteApp and whether GFX is enabled.

I did try playing with and toggling that setting, to no apparent avail.

PR #612, opened 2025-08-22
GUACAMOLE-2068: RAIL should only re-show minimized windows.

necouchman:
This PR removes the re-showing of HIDDEN windows, instead only re-showing MINIMIZED windows, which fixes issues with dialog boxes that have drop-downs that won't go away.


mike-jumper (approved):

src/protocols/rdp/channels/rail.c
```diff
@@ -280,8 +280,7 @@ static BOOL guac_rdp_rail_window_update(rdpContext* context,
         guac_client_log(client, GUAC_LOG_TRACE, "RAIL window visibility change: %d", windowState->showState);
 
         /* State is either hidden or minimized - send restore command. */
```
poliagos, line 282: You probably should update comment as well

PR #615, opened 2025-09-02
GUACAMOLE-2129: Migrate recording/typescript creation to shared function.

mike-jumper:
This change defines a new function, `guac_openat()`, which encapsulates the functionality relied upon by both recording and typescript creation:

* A numeric suffix can be appended automatically, if requested.
* The file can be locked appropriately upon return.
* The path containing the file is provided separately and can be automatically created.

~Opening as a draft while I implement unit tests and double-check the old vs. new approaches.~

### DIFF
--- a/src/libguac/tests/mem/free.c
+++ b/src/libguac/tests/mem/free.c
@@ -50,6 +50,7 @@ void test_mem__free_const() {
 void test_mem__free_null() {
     void* ptr = NULL;
     guac_mem_free(ptr);
+    CU_PASS("Nothing freed (segfault did not occur)");
 }
 
 /**
@@ -59,5 +60,6 @@ void test_mem__free_null() {
 void test_mem__free_null_const() {
     const void* ptr = NULL;
     guac_mem_free_const(ptr);
+    CU_PASS("Nothing freed (segfault did not occur)");
 }
 
--- a/util/generate-test-runner.pl
+++ b/util/generate-test-runner.pl
@@ -91,6 +91,11 @@ print <<'END';
 int tap_test_number = 1;
 
 /**
+ * The total number of assertions made prior to the current test.
+ */
+int last_asserts = 0;
+
+/**
  * Logs the status of a CUnit test which just completed. This implementation
  * logs test completion in TAP format.
  *
@@ -106,10 +111,19 @@ int tap_test_number = 1;
 static void tap_log_test_completed(const CU_pTest test,
         const CU_pSuite suite, const CU_pFailureRecord failure) {
 
+    const char* directive = "";
+
+    /* Determine whether the test was skipped (no assertions made at all) */
+    int current_asserts = CU_get_number_of_asserts();
+    if (current_asserts == last_asserts)
+        directive = " # SKIP";
+    else
+        last_asserts = current_asserts;
+
     /* Log success/failure in TAP format */
     if (failure == NULL)
-        printf("ok %i - [%s] %s: OK\n",
-            tap_test_number, suite->pName, test->pName);
+        printf("ok %i - [%s] %s: OK%s\n",
+            tap_test_number, suite->pName, test->pName, directive);
     else
         printf("not ok %i - [%s] %s: Assertion failed on %s:%i: %s\n",
             tap_test_number, suite->pName, test->pName,


src/libguac/file-private.h
```diff
@@ -0,0 +1,42 @@
...
+/**
+ * Tests whether the given string is a filename with no other path components
+ * present.
+ *
+ * @param filename
+ *     The filename to check.
+ *
+ * @return
+ *     Zero if the provided string is a filename without any other path
+ *     components, non-zero if at least one path component is present.
```
necouchman: I think this is actually backward from how the function is implemented - looking at the below code in `file.c`, it looks like it returns `0` if a path separate *is* found, and `1` if a path separator is *not* found. Which makes more sense to me given the function name `guac_is_filename()`.

mike-jumper: Ack! You're right. I'll fix that.

mike-jumper:
OK - I've added unit tests and the documentation for `guac_is_filename()` should be fixed now.

src/libguac/file.c
```diff
@@ -0,0 +1,245 @@
...
+    return mkdir(path, S_IRWXU | S_IRGRP | S_IXGRP);
+#else
+    return _mkdir(path);
+#endif
+}
+
+/**
+ * Attempts to acquire a lock on the file associated with the given file
+ * descriptor. The type of lock acquired is dictated by the read_lock flag. If
+ * the lock cannot be acquired. errno is set appropriately.
```
necouchman: `acquired. errno` -> `acquired, errno`

src/libguac/file.c
```diff
@@ -0,0 +1,245 @@
...
+ *
+ * This function currently has no effet under Windows and simply returns
```
necouchman: `effet` -> `effect`

src/libguac/file.c
```diff
@@ -0,0 +1,245 @@
...
+ * directory cannot be created, errno is set appropriately.
+ *
+ * @param path
+ *     The full path of the directory to create.
+ *
+ * @return
+ *     Zero if the directory was created successfully, non-zero otherwise.
+ */
+static int guac_mkdir(const char* path) {
+#ifndef __MINGW32__
```
necouchman: Just one question/comment on this - I see you've consistently used the `#ifndef __MINGW32__` for specifying UNIX variants. Is there some reason you're doing it this way rather than `#ifdef __MINGW32__` for specifying stuff that applies to Windows, and using the `#else` for everything else (= UNIX)? It seems to me that, since Windows is the exception, you'd want to to capture it as the specific case, falling back to POSIX-compliant interfaces for everything that, you know, follows standards :laughing:?

src/libguac/file.c
```diff
@@ -0,0 +1,245 @@
...
+
+}
+
+int guac_openat(const char* path, const char* filename,
+        const guac_open_how* how) {
+
+    int dir_fd = -1;
+    int fd = -1;
+
+    /* Verify only the path contains any path separators, etc. */
```
necouchman: This comment seems a little off to me - maybe `verify that the filename does not contain path separators`?

src/libguac/tests/file/openat.c
```diff
@@ -0,0 +1,407 @@
...
+    int dir_fd = open(path, O_RDONLY);
+    if (dir_fd == -1)
+        return 0;
+
+    if (unlinkat(dir_fd, filename, 0))
+        return 0;
+
+    close(dir_fd);
+    return 1;
+
```
necouchman: Does `dir_fd` get automatically closed in the case where the `unlinkat()` call fails? Or will this leave `dir_fd` hanging open? In the above implementation of `exists()` you recorded the result of the `fstatat()` call and then did the `close_if_necessary()` call, returning the result. Should the same logic be applied, here?

src/libguac/tests/file/openat.c
```diff
@@ -0,0 +1,407 @@
...
+ */
+static int has_mode(const char* path, const char* filename, mode_t mode) {
+
+
+    int dir_fd = open(path, O_RDONLY);
+    if (dir_fd == -1)
+        return 0;
+
+    struct stat file_info = { 0 };
+    int result = fstatat(dir_fd, filename, &file_info, 0);
```
necouchman, line 127: Are there any circumstances under which `filename` could be `NULL`? If so, will `fstatat()` handle that cleanly, or should there be a check for that (as in `exists()` above)?

necouchman (changes requested):
A few minor things, mostly in the tests...

src/libguac/file.c
mike-jumper: Nope, no particular reason except habit. I agree it makes more sense to have the exception be the exception...

src/libguac/tests/file/openat.c
mike-jumper: Yep, you're right - this would otherwise leak the file descriptor.

src/libguac/tests/file/openat.c
mike-jumper: Erg ... not sure where that `NULL` check came from, actually. Only `remove_file()` actually needs that handling. The other functions specific to these unit tests can all safely assume the filename will be provided.

PR #617, opened 2025-09-26
GUACAMOLE-2143: Improve process management to prevent zombie process accumulation

sschiffli:
## Problem
Guacamole daemon was not properly cleaning up orphaned client sub-processes, leading to zombie processes that accumulate over time and consume system resources.

## Root Cause
- Daemon did not adopt orphaned processes when parent client processes terminated unexpectedly
- Process termination only killed individual processes instead of entire process groups
- This left child processes running as orphans that couldn't be properly reaped

## Solution
Implemented comprehensive process lifecycle management:

1. **Added subreaper capability**: Configure daemon with `prctl(PR_SET_CHILD_SUBREAPER, 1)` to automatically adopt any orphaned client processes
2. **Process group termination**: Changed termination signal from `kill(proc->pid, SIGTERM)` to `kill(-proc->pid, SIGTERM)` to terminate entire process groups
3. **Proper cleanup chain**: Ensures all child processes are properly reaped when connections terminate

## Impact
- Prevents zombie process accumulation
- Improves system resource management
- Ensures clean process termination in all scenarios
- Maintains compatibility with existing process group creation (`setpgid(0, 0)`)

### DIFF
--- a/src/guacd/daemon.c
+++ b/src/guacd/daemon.c
@@ -43,6 +43,9 @@
 #include <stdlib.h>
 #include <string.h>
 #include <syslog.h>
+#ifdef HAVE_PRCTL
+#include <sys/prctl.h>
+#endif
 #include <sys/socket.h>
 #include <sys/stat.h>
 #include <sys/types.h>
@@ -133,7 +136,7 @@ static int daemonize() {
 
     /* Change to root directory */
     if (chdir(GUACD_ROOT) < 0) {
-        guacd_log(GUAC_LOG_ERROR, 
+        guacd_log(GUAC_LOG_ERROR,
                 "Unable to change working directory to "
                 GUACD_ROOT);
         return 1;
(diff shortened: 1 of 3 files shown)


src/guacd/daemon.c
```diff
@@ -508,6 +509,9 @@ int main(int argc, char* argv[]) {
     sigaction(SIGINT, &signal_stop_action, NULL);
     sigaction(SIGTERM, &signal_stop_action, NULL);
 
+    /* Adopt any orphan processes from clients so we can properly terminate and reap */
+    prctl(PR_SET_CHILD_SUBREAPER, 1);
```
mike-jumper, line 516: Is this specific to Linux (ie: not part of a standard like POSIX)?

sschiffli: Correct, not a POSIX standard. I guess I need to add AC_CHECK_FUNCS([prctl]) check and #ifdef around this?

mike-jumper (approved):

PR #630, opened 2025-12-15
GUACAMOLE-1969: Report Clipboard History.

aleitner:
Built off https://github.com/apache/guacamole-server/pull/530

Added additional changes to record when the clipboard is modified on the remote desktop itself.

Paired with https://github.com/apache/guacamole-client/pull/1137

### DIFF
--- a/src/libguac/guacamole/recording.h
+++ b/src/libguac/guacamole/recording.h
@@ -49,6 +49,11 @@
 #define GUAC_COMMON_RECORDING_MAX_NAME_LENGTH 2048
 
 /**
+ * The block size to use when sending clipboard data in a recording.
+ */
+#define GUAC_RECORDING_CLIPBOARD_BLOCK_SIZE 4096
+
+/**
  * An in-progress session recording, attached to a guac_client instance such
  * that output Guacamole instructions may be dynamically intercepted and
  * written to a file.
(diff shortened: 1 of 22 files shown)


src/libguac/guacamole/recording.h
```diff
@@ -95,6 +106,15 @@ typedef struct guac_recording {
      */
     int include_keys;
 
+    /**
+     * Non-zero if clipboard paste data should be included in the session
```
necouchman: It's not just `paste` data, right? When I first read this it gives the impression that this would only record data coming from the user into the Guacamole session, but I think this would also control data copied out of the session back to the user, correct?

necouchman (changes requested):
You have a merge conflict that needs to be fixed, here. Also, I'd generally change the comments that say `clipboard paste data` to just `clipboard data` - I think the `paste` word is unnecessary and risks causing confusion on what is being recorded.

src/libguac/guacamole/recording.h
aleitner: Good catch. I renamed the recording related "clipboard paste data" comments to just "clipboard data". Also rebased onto current main to resolve the conflict.

necouchman (approved):

PR #632, opened 2026-01-06
GUACAMOLE-1170: Use min_size with user stream pool to delay stream index reuse.

sschiffli:
Adding minimum size to user stream pool, so new stream integers are preferred over reused integers until the minimum size is reached.

This fixes a race condition with stream index reuse outlined in GUACAMOLE-1170.

### DIFF
--- a/src/libguac/user.c
+++ b/src/libguac/user.c
@@ -57,8 +57,11 @@ guac_user* guac_user_alloc() {
     user->processing_lag = 0;
     user->active = 1;
 
-    /* Allocate stream pool */
-    user->__stream_pool = guac_pool_alloc(0);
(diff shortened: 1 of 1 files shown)


mike-jumper (changes requested):
I think the same changes you've made to the user-specific stream pool also need to be made to the connection-wide stream pool:

https://github.com/apache/guacamole-server/blob/cf09d3a615ceb69933754f8fccc1524e87dc9089/src/libguac/client.c#L284-L285

Other than that, LGTM. Really glad this didn't require changes to `guac_pool`.

mike-jumper:
Good point. I think you're right that the same ambiguity doesn't apply to connection-wide streams.

PR #633, opened 2026-02-13
GUACAMOLE-2210: Add support for AAD authentication to RDP protocol.

aleitner:
- Adds Azure AD (Entra ID) authentication for RDP connections using FreeRDP 3's `AadSecurity` mode
- Uses the OAuth2 device authorization grant (RFC 8628), so no credentials are stored in or entered through Guacamole. The user completes sign-in on a separate device (typically a phone) by scanning a QR code
- HTTP requests use libcurl; token/device-code responses are parsed with WinPR's JSON API

## Authentication flow

When `security=aad` is set on an RDP connection:

1. FreeRDP begins the AAD handshake and invokes the `GetAccessToken` callback with the requested scope and a Proof-of-Possession key confirmation (`req_cnf`).
2. guacd requests a device code from Azure AD using the configured application (client) ID and scope, receiving a user code and a verification URI.
3. guacd opens a pipe stream to the connection owner carrying the user code and verification URI as JSON. The web client renders this as a QR code (paired change: apache/guacamole-client#1168).
4. guacd polls the token endpoint (`grant_type=urn:ietf:params:oauth:grant-type:device_code`) while the user authenticates on their own device, forwarding `req_cnf` so Azure issues a Proof-of-Possession token. Polling honors the server-provided interval and `slow_down`, and stops if the connection closes or the code expires.
5. When the flow completes, the pipe stream is closed (dismissing the prompt) and the access token is handed to FreeRDP, which finishes the AAD-authenticated RDP session.

## Configuration

New RDP parameters:

- `aad-client-id` (required for AAD) is the Azure AD application (client) ID to authenticate as. The app must have public client flows enabled so it can use the device code grant.
- `aad-scope` (optional) is the OAuth2 scope to request. If omitted, the scope FreeRDP derives for the target host is used.
- `aad-tenant-id` (optional) is the Azure AD tenant, defaulting to `organizations`.

## Notes

- The entire feature is gated on `HAVE_FREERDP_AAD_SUPPORT` (FreeRDP built with AAD) and the presence of libcurl. If either is unavailable it compiles out cleanly, and non-AAD RDP is unaffected.
- Adds a unit test for the percent-decoding helper.

Paired with [apache/guacamole-client#1168](https://github.com/apache/guacamole-client/pull/1168)

### DIFF
--- a/src/protocols/rdp/Makefile.am
+++ b/src/protocols/rdp/Makefile.am
@@ -38,6 +38,7 @@ nodist_libguac_client_rdp_la_SOURCES =  \
     _generated_keymaps.c
 
 libguac_client_rdp_la_SOURCES =                  \
+    aad.c                                        \
     argv.c                                       \
     beep.c                                       \
     channels/audio-input/audio-buffer.c          \
@@ -84,6 +85,7 @@ libguac_client_rdp_la_SOURCES =                  \
     user.c
 
 noinst_HEADERS =                                 \
+    aad.h                                        \
     argv.h                                       \
     beep.h                                       \
     channels/audio-input/audio-buffer.h          \
@@ -141,7 +143,8 @@ libguac_client_rdp_la_LDFLAGS = \
     -version-info 0:0:0         \
     @CAIRO_LIBS@                \
     @PTHREAD_LIBS@              \
-    @RDP_LIBS@
+    @RDP_LIBS@                  \
+    @CURL_LIBS@
 
 libguac_client_rdp_la_LIBADD = \
     @COMMON_LTLIB@             \
--- a/src/protocols/rdp/Makefile.am
+++ b/src/protocols/rdp/Makefile.am
@@ -40,6 +40,7 @@ nodist_libguac_client_rdp_la_SOURCES =  \
     _generated_keymaps.c
 
 libguac_client_rdp_la_SOURCES =                  \
+    aad.c                                        \
     argv.c                                       \
     beep.c                                       \
     channels/audio-input/audio-buffer.c          \
@@ -86,6 +87,7 @@ libguac_client_rdp_la_SOURCES =                  \
     user.c
 
 noinst_HEADERS =                                 \
+    aad.h                                        \
     argv.h                                       \
     beep.h                                       \
     channels/audio-input/audio-buffer.h          \
@@ -136,12 +138,14 @@ libguac_client_rdp_la_CFLAGS = \
     -Werror -Wall -Iinclude    \
     @COMMON_INCLUDE@           \
     @COMMON_SSH_INCLUDE@       \
+    @CURL_CFLAGS@              \
     @LIBGUAC_INCLUDE@          \
     @RDP_CFLAGS@
 
 libguac_client_rdp_la_LDFLAGS = \
     -version-info 0:0:0         \
     @CAIRO_LIBS@                \
+    @CURL_LIBS@                 \
     @PTHREAD_LIBS@              \
     @RDP_LIBS@
 
--- a/src/protocols/rdp/aad.c
+++ b/src/protocols/rdp/aad.c
@@ -1,0 +1,1072 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+#include "config.h"
+
+#ifdef HAVE_FREERDP_AAD_SUPPORT
+
+#include "aad.h"
+
+#include <guacamole/client.h>
+#include <guacamole/mem.h>
+#include <guacamole/string.h>
+
+#include <curl/curl.h>
+#include <stdlib.h>
+#include <string.h>
+
+/**
+ * Azure AD OAuth2 token endpoint URL format.
+ * The %s placeholder is replaced with the tenant ID.
+ */
+#define GUAC_AAD_TOKEN_ENDPOINT \
+    "https://login.microsoftonline.com/%s/oauth2/v2.0/token"
+
+/**
+ * Azure AD OAuth2 authorization endpoint URL format.
+ * The %s placeholder is replaced with the tenant ID.
+ */
+#define GUAC_AAD_AUTHORIZE_ENDPOINT \
+    "https://login.microsoftonline.com/%s/oauth2/v2.0/authorize"
+
+/**
+ * The native client redirect URI used for the authorization code flow.
+ * This is a special Microsoft-provided redirect URI for non-web applications.
+ */
+#define GUAC_AAD_NATIVE_REDIRECT_URI \
+    "https://login.microsoftonline.com/common/oauth2/nativeclient"
+
+/**
+ * Maximum size for the login page HTML response.
+ */
+#define GUAC_AAD_LOGIN_PAGE_MAX_SIZE (64 * 1024)
+
+/**
+ * HTTP request timeout in seconds.
+ */
+#define GUAC_AAD_HTTP_TIMEOUT_SECONDS 30
+
+/**
+ * User-Agent string sent with all HTTP requests to Microsoft login endpoints.
+ * A browser-like UA is required to avoid "unsupported browser" responses.
+ */
+#define GUAC_AAD_USER_AGENT \
+    "Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 " \
+    "(KHTML, like Gecko) Chrome/120.0.0.0 Safari/537.36"
+
+/**
+ * HTTP response structure for AAD requests.
+ */
+typedef struct guac_rdp_aad_response {
+
+    /**
+     * The response body data.
+     */
+    char* data;
+
+    /**
+     * The current size of the response data.
+     */
+    size_t size;
+
+} guac_rdp_aad_response;
+
+/**
+ * Callback function for libcurl to write received HTTP data into a
+ * guac_rdp_aad_response buffer.
+ *
+ * @param contents
+ *     Pointer to the received data.
+ *
+ * @param size
+ *     Size of each data element.
+ *
+ * @param nmemb
+ *     Number of data elements.
+ *
+ * @param userp
+ *     User-provided pointer (guac_rdp_aad_response structure).
+ *
+ * @return
+ *     The number of bytes processed.
+ */
+static size_t guac_rdp_aad_write_callback(void* contents, size_t size,
+        size_t nmemb, void* userp) {
+
+    size_t total_size = size * nmemb;
+    guac_rdp_aad_response* response = (guac_rdp_aad_response*) userp;
+
+    /* Reject responses that exceed the maximum login page size */
+    if (response->size + total_size > GUAC_AAD_LOGIN_PAGE_MAX_SIZE)
+        return 0;
+
+    /* Copy data into response buffer and null-terminate */
+    memcpy(response->data + response->size, contents, total_size);
+    response->size += total_size;
+    response->data[response->size] = '\0';
+
+    return total_size;
+}
+
+/**
+ * URL-encodes a string for use in HTTP POST data or query parameters.
+ *
+ * @param curl
+ *     The CURL handle to use for encoding.
+ *
+ * @param str
+ *     The string to encode.
+ *
+ * @return
+ *     A newly allocated URL-encoded string, or NULL on error. The caller
+ *     must free this string using curl_free().
+ */
+static char* guac_rdp_aad_urlencode(CURL* curl, const char* str) {
+    if (str == NULL)
+        return NULL;
+    return curl_easy_escape(curl, str, strlen(str));
+}
+
+/**
+ * Allocates and initializes a new guac_rdp_aad_response structure with a
+ * fixed buffer large enough to hold the maximum allowed response.
+ *
+ * @return
+ *     A newly allocated response structure, or NULL on allocation failure.
+ *     The caller must free this with guac_rdp_aad_response_free().
+ */
+static guac_rdp_aad_response* guac_rdp_aad_response_alloc(void) {
+
+    guac_rdp_aad_response* response =
+            guac_mem_zalloc(sizeof(guac_rdp_aad_response));
+
+    if (response == NULL)
+        return NULL;
+
+    /* Allocate the maximum allowed size upfront so the write callback
+     * never needs to reallocate */
+    response->data = guac_mem_alloc(GUAC_AAD_LOGIN_PAGE_MAX_SIZE + 1);
+
+    if (response->data == NULL) {
+        guac_mem_free(response);
+        return NULL;
+    }
+
+    response->data[0] = '\0';
+
+    return response;
+}
+
+/**
+ * Extracts a string value from the $Config JavaScript object embedded in
+ * the Microsoft login page HTML. Searches for the pattern "key":" and
+ * returns the value up to the next unescaped double-quote.
+ *
+ * @param html
+ *     The HTML string to search.
+ *
+ * @param key
+ *     The JSON key name to find (without quotes).
+ *
+ * @return
+ *     A newly allocated string containing the extracted value, or NULL if
+ *     the key was not found. The caller must free with guac_mem_free().
+ */
+static char* guac_rdp_aad_extract_config_value(const char* html,
+        const char* key) {
+
+    if (html == NULL || key == NULL)
+        return NULL;
+
+    char pattern[256];
+    snprintf(pattern, sizeof(pattern), "\"%s\":\"", key);
+
+    const char* value_start = strstr(html, pattern);
+    if (value_start == NULL)
+        return NULL;
+
+    value_start += strlen(pattern);
+
+    /* Find closing quote, skipping escaped characters */
+    const char* value_end = value_start;
+    while (*value_end != '\0') {
+        if (*value_end == '\\' && *(value_end + 1) != '\0') {
+            /* Skip escaped character */
+            value_end += 2;
+            continue;
+        }
+        if (*value_end == '"')
+            break;
+        value_end++;
+    }
+
+    if (*value_end != '"')
+        return NULL;
+
+    size_t value_len = value_end - value_start;
+    return guac_strndup(value_start, value_len);
+}
+
+char* guac_rdp_percent_decode(const char* str) {
+
+    if (str == NULL)
+        return NULL;
+
+    size_t len = strlen(str);
+    char* decoded = guac_mem_alloc(len + 1);
+    size_t out_pos = 0;
+
+    for (size_t i = 0; i < len; i++) {
+        if (str[i] == '%' && i + 2 < len) {
+            char hex[3] = { str[i + 1], str[i + 2], '\0' };
+            char* hex_end;
+            long byte_val = strtol(hex, &hex_end, 16);
+            if (hex_end == hex + 2) {
+                decoded[out_pos++] = (char) byte_val;
+                i += 2;
+                continue;
+            }
+        }
+        decoded[out_pos++] = str[i];
+    }
+
+    decoded[out_pos] = '\0';
+    return decoded;
+}
+
+/**
+ * Parses the JSON response from a token exchange request and extracts the
+ * access token. If the response contains an error_description field instead,
+ * that error is logged.
+ *
+ * @param client
+ *     The guac_client associated with the current RDP connection.
+ *
+ * @param json_response
+ *     The raw JSON response body from the token endpoint.
+ *
+ * @return
+ *     A newly allocated string containing the access token, or NULL if
+ *     parsing failed or the response contained an error. The caller must
+ *     free the returned string with guac_mem_free().
+ */
+static char* guac_rdp_aad_parse_token_response(guac_client* client,
+        const char* json_response) {
+
+    if (json_response == NULL)
+        return NULL;
+
+    /* Look for access_token in the JSON response */
+    const char* token_key = "\"access_token\"";
+    const char* token_pos = strstr(json_response, token_key);
+
+    if (token_pos == NULL) {
+
+        /* Log error description if present instead */
+        const char* error_desc_key = "\"error_description\"";
+        const char* error_desc_pos = strstr(json_response, error_desc_key);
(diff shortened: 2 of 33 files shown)


src/protocols/rdp/rdp.c
```diff
@@ -33,7 +33,6 @@
 #include "channels/rdpsnd/rdpsnd.h"
 #include "client.h"
 #include "color.h"
-#include "config.h"
```
necouchman: Why is this being removed, here, when it's been added to all the other source files?

necouchman (changes requested):
Looks like the build is failing with an error about the `switch()` statement and one of the enums.

src/protocols/rdp/rdp.c
aleitner: There is another include of the config.h at the top of this file

aleitner:
> Looks like the build is failing with an error about the `switch()` statement and one of the enums.

Ahh whoops! I was testing with only freerdp3 as that's when AAD support was implemented. Just added a case to the switch for freerdp2 so that it errors if AAD is selected

src/protocols/rdp/rdp.c
aleitner: When resizing the window, the desktop display stayed fixed at whatever size it connected at, and with certain timing we'd hit a segfault. After a bunch of pointer tracing I found that `rdp_client->disp` was non-NULL in `rdp.c` but `NULL` when read from `disp.c` using the exact same `rdp_client` pointer, which turned out to be a struct layout mismatch: `disp.c` and several other files in channels/ weren't including `config.h`, so some definitions were undefined and pointer fields were missing from their view of `guac_rdp_client`, shifting the offset of disp by some bytes to land on another `NULL` property. All I had to do was add `#include "config.h"` to all the affected files so every compilation unit sees the same struct

necouchman: Nice find!

src/protocols/rdp/settings.c
```diff
@@ -1645,8 +1660,30 @@ void guac_rdp_push_settings(guac_client* client,
...
+        #ifdef HAVE_FREERDP_AAD_SUPPORT
+            freerdp_settings_set_bool(rdp_settings, FreeRDP_AadSecurity, TRUE);
+            guac_client_log(client, GUAC_LOG_INFO,
+                "Azure AD authentication enabled");
+        #else
+            guac_client_log(client, GUAC_LOG_ERROR,
+                "Azure AD authentication requested but not supported by this version of FreeRDP. "
+                "FreeRDP 3.0 or later is required.");
+        #endif
+    break;
```
necouchman: I think this `break;` is not indented correctly.

src/protocols/rdp/settings.c
```diff
@@ -1885,6 +1926,13 @@ void guac_rdp_push_settings(guac_client* client,
             rdp_settings->VmConnectMode = TRUE;
             break;
 
+        /* Azure AD authentication (not supported on FreeRDP 2) */
+        case GUAC_SECURITY_AAD:
+            guac_client_log(client, GUAC_LOG_ERROR,
+                "Azure AD authentication is not supported by this "
+                "version of FreeRDP. FreeRDP 3.0 or later is required.");
+            break;
+
```
necouchman: It looks like you're assuming that, if `HAVE_SETTERS_GETTERS` is not defined, we must be on FreeRDP 2.x, and therefore AAD is not supported? I'm not sure that's a good assumption - it may be generally true, it may even be the case in most of the versions of FreeRDP, but I'm not sure it's the correct logic? I would think the proper way to do this would be either to just set the appropriate `rdp_settings` member, here, or use the correct `#ifdef` to check for `HAVE_FREERDP_AAD_SUPPORT`.

src/protocols/rdp/aad.c
```diff
@@ -0,0 +1,1072 @@
...
+ */
+#define GUAC_AAD_HTTP_TIMEOUT_SECONDS 30
+
+/**
+ * User-Agent string sent with all HTTP requests to Microsoft login endpoints.
+ * A browser-like UA is required to avoid "unsupported browser" responses.
+ */
+#define GUAC_AAD_USER_AGENT \
+    "Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 " \
+    "(KHTML, like Gecko) Chrome/120.0.0.0 Safari/537.36"
```
mike-jumper: Are we going to need to continually update this value to avoid being re-flagged as unsupported due to the claimed version being old?

mike-jumper: Not sure if this is an insane idea ... but we _could_ calculate the expected current Chrome release from the current date based on their published major release cadence ...

aleitner: Microsoft's endpoints don't appear to validate browser version freshness currently. On the other hand we could fetch the latest stable major version from Google's VersionHistory API:
https://versionhistory.googleapis.com/v1/chrome/platforms/linux/channels/stable/versions

aleitner: Made the changes to get the latest chrome version!

mike-jumper: Unfortunately, phoning Google from within guacd is going to be a privacy issue. I don't think we can go this route.

src/protocols/rdp/rdp.c
```diff
@@ -327,6 +328,149 @@ static BOOL rdp_freerdp_authenticate(freerdp* instance, char** username,
 
 }
 
+#ifdef HAVE_FREERDP_AAD_SUPPORT
+#include "aad.h"
```
necouchman: It seems like the generally more used/accepted way to do this is to have a separate `#ifdef` above in the includes up with the rest of them, rather than putting this `#include` this far down in the body?

src/protocols/rdp/rdp.c
```diff
@@ -327,6 +328,149 @@ static BOOL rdp_freerdp_authenticate(freerdp* instance, char** username,
...
+ * @param token
+ *     Pointer to a string which will receive the access token. This function
+ *     must allocate and populate this string.
+ *
+ * @param count
+ *     Number of additional variadic arguments.
+ *
+ * @param ...
+ *     Additional arguments (for AAD: scope and req_cnf)
+ *
```
necouchman, line 362: Are there other situations outside of AAD that we expect to use this for? Or is how FreeRDP defines the callback function? I'm just not sure about using a variadic function when we only have a single use-case??

src/protocols/rdp/rdp.c
```diff
@@ -327,6 +328,149 @@ static BOOL rdp_freerdp_authenticate(freerdp* instance, char** username,
...
+    guac_client* client = ((rdp_freerdp_context*) context)->client;
+    guac_rdp_client* rdp_client = (guac_rdp_client*) client->data;
+    guac_rdp_settings* settings = rdp_client->settings;
+
+    *token = NULL;
+
+    /* Only handle ACCESS_TOKEN_TYPE_AAD for now */
+    if (tokenType != ACCESS_TOKEN_TYPE_AAD) {
+        guac_client_log(client, GUAC_LOG_ERROR,
+            "AAD: Unsupported token type: %d", tokenType);
```
necouchman: I'm not sure this log message should be prefixed with `AAD` when, what it's saying is that `AAD` was _not_ requested?

src/protocols/rdp/rdp.c
```diff
@@ -327,6 +328,149 @@ static BOOL rdp_freerdp_authenticate(freerdp* instance, char** username,
...
+        return FALSE;
+    }
+
+    /* Extract scope and req_cnf from variadic arguments */
+    if (count < 2) {
+        guac_client_log(client, GUAC_LOG_ERROR,
+            "AAD: Expected 2 arguments (scope and req_cnf)");
```
necouchman, line 387: ...and, if this is intended to be re-used for things other than AAD, should these be prefixed with `AAD` at all?

src/protocols/rdp/rdp.c
```diff
@@ -327,6 +328,149 @@ static BOOL rdp_freerdp_authenticate(freerdp* instance, char** username,
...
+        guac_argv_await((const char**) required_params);
+
+        /* Check that credentials were provided */
+        if (settings->username == NULL || settings->password == NULL) {
+            guac_client_log(client, GUAC_LOG_ERROR,
+                    "AAD: Username and password are required for AAD "
+                    "authentication");
+            return FALSE;
+        }
+    }
```
necouchman: I'm not all that familiar with AAD logins outside of a more traditional domain context - is the `domain` parameter never required, or should it be checked and prompted for, here?

src/protocols/rdp/aad.c
```diff
@@ -0,0 +1,1072 @@
...
+ * @param str
+ *     The string to encode.
+ *
+ * @return
+ *     A newly allocated URL-encoded string, or NULL on error. The caller
+ *     must free this string using curl_free().
+ */
+static char* guac_rdp_aad_urlencode(CURL* curl, const char* str) {
+    if (str == NULL)
+        return NULL;
```
necouchman, line 270: If you're going to check for a `NULL` string, here, maybe it's also worth a quick check to see if `curl` is `NULL`?

src/protocols/rdp/aad.c
```diff
@@ -0,0 +1,1072 @@
...
+static void guac_rdp_aad_get_credential_type(guac_client* client,
+        CURL* curl, guac_rdp_aad_params* params, const char* auth_url,
+        char** flow_token, const char* ctx, const char* api_canary) {
+
+    guac_client_log(client, GUAC_LOG_DEBUG,
+            "AAD: Calling GetCredentialType API");
+
+    char gct_url[512];
+    snprintf(gct_url, sizeof(gct_url),
+            "https://login.microsoftonline.com/%s/GetCredentialType?mkt=en",
```
necouchman: Should this URL get a `#define` toward the top of this file like you've done for the others?

src/protocols/rdp/aad.c
```diff
@@ -0,0 +1,1072 @@
...
+            "&canary=%s"
+            "&ctx=%s"
+            "&flowtoken=%s"
+            "&type=11",
+            encoded_login,
+            encoded_login,
+            encoded_passwd,
+            encoded_canary,
+            encoded_ctx,
+            encoded_flowtoken);
```
necouchman: I see this sort of logic repeated a couple of times throughout the code. Maybe the creation of POST body structures is worth breaking into its own function? May not be worth it - just a thought.

necouchman (changes requested):
A few comments/changes. Also, generally:
* It seems like there might be a couple of opportunities for reusing code by creating a couple of new functions?
* Some of the documentation within the code is a bit sparse and could use a little more building out.

src/protocols/rdp/settings.c
aleitner: fixed

src/protocols/rdp/settings.c
aleitner: updated to now use `#ifdef HAVE_FREERDP_AAD_SUPPORT`

src/protocols/rdp/rdp.c
aleitner: Moved up alongside the existing `ENABLE_COMMON_SSH` conditional include block.

src/protocols/rdp/rdp.c
aleitner: Kept the variadic signature (FreeRDP imposes it) but renamed to `rdp_freerdp_get_aad_access_token` and updated the docstring to make clear the implementation is AAD-only

src/protocols/rdp/rdp.c
aleitner: Rephrased to `"Unsupported access token type: %d (only AAD is supported)"`

src/protocols/rdp/rdp.c
aleitner: Specialized the callback to AAD. the remaining `AAD: ` log prefixes inside the function are now appropriate since the function only handles AAD

src/protocols/rdp/rdp.c
aleitner: AAD identifies users by UPN (e.g. user@tenant.onmicrosoft.com), so the legacy RDP domain field has no role here. Microsoft's login form itself has no domain input, identity is derived from the UPN suffix. I also added a clarifying comment near the credential prompt

src/protocols/rdp/aad.c
aleitner: done

src/protocols/rdp/aad.c
aleitner: Added a define for `GUAC_AAD_GET_CREDENTIAL_TYPE_ENDPOINT`

src/protocols/rdp/aad.c
aleitner: added a `guac_rdp_aad_build_form_body` helper backed by a `guac_rdp_aad_form_field` key/value struct, and refactored the credential POST, token exchange, and the new AVD device-binding POST to use it.

aleitner:
Also fixed two runtime issues from changes Microsoft made to the AAD login flow. The credential POST field name needs to be flowToken (camelCase) per `$Config.sFTName`, and for AVD sessions Microsoft now serves a `/common/rdp/set` device-binding page that the browser auto-submits before the OAuth redirect, which we replicate with a second POST

aleitner:
This has been reworked from the original login-page automation to the OAuth2 device authorization grant (RFC 8628), so the user signs in on a separate device via a QR code and no credentials pass through Guacamole.

src/protocols/rdp/Makefile.am
bbennett-ks, line 143: +@CURL_CFLAGS@

Probably a good practice to include both autoconf `PKG_CHECK_MODULES`-generated variables (`CFLAGS` and `LIBS`) to avoid potential build breaks if package dependencies or installation paths change, rather than relying on implicitly available include or library paths.

src/protocols/rdp/Makefile.am
```diff
@@ -143,7 +145,8 @@ libguac_client_rdp_la_LDFLAGS = \
     -version-info 0:0:0         \
     @CAIRO_LIBS@                \
     @PTHREAD_LIBS@              \
-    @RDP_LIBS@
+    @RDP_LIBS@                  \
+    @CURL_LIBS@
```
bbennett-ks: Alphabetize.

src/protocols/rdp/aad.c
```diff
@@ -0,0 +1,934 @@
...
+ *     The form-encoded POST body.
+ *
+ * @return
+ *     A newly-allocated response on success, or NULL on failure. The caller
+ *     must free the result with guac_rdp_aad_response_free().
+ */
+static guac_rdp_aad_response* guac_rdp_aad_http_post(guac_client* client,
+        const char* url, const char* post_data) {
+
+    CURL* curl = curl_easy_init();
```
bbennett-ks: This is probably OK, but it maybe a good idea to call `curl_global_init()` in some early initialization code. Perhaps assume other components may use curl in the future?

From `https://linux.die.net/man/3/curl_easy_init`:
```
If you did not already call [curl_global_init](https://linux.die.net/man/3/curl_global_init)(3), curl_easy_init(3) does it automatically. This may be lethal in multi-threaded cases, since [curl_global_init](https://linux.die.net/man/3/curl_global_init)(3) is not thread-safe, and it may result in resource problems because there is no corresponding cleanup.

You are strongly advised to not allow this automatic behaviour, by calling [curl_global_init](https://linux.die.net/man/3/curl_global_init)(3) yourself properly. See the description in [libcurl](https://linux.die.net/man/3/libcurl)(3) of global environment requirements for details of how to use this function.
'''

File: src/protocols/rdp/aad.c [hunk 15]
```diff
@@ -0,0 +1,934 @@
...
+        return NULL;
+    }
+
+    guac_rdp_aad_response* response = guac_rdp_aad_response_alloc();
+    if (response == NULL) {
+        curl_easy_cleanup(curl);
+        return NULL;
+    }
+
+    struct curl_slist* headers = curl_slist_append(NULL,
```
bbennett-ks, line 501: `NULL` check here? I know much of the codebase intentionally ignores OOM-only failures, but `curl_slist_append()` can return NULL for any failure. The `libcurl` man page  recommends assigning the result to a temporary variable before replacing an existing list pointer, so the original list is not lost on failure.

src/protocols/rdp/aad.c
```diff
@@ -0,0 +1,934 @@
...
+    if (tenant_id == NULL || client_id == NULL || scope == NULL) {
+        guac_client_log(client, GUAC_LOG_ERROR,
+                "AAD: tenant, client ID, and scope are all required for "
+                "authentication");
+        return NULL;
+    }
+
+    /* Begin the device authorization grant */
+    guac_rdp_aad_device_code* dc = guac_rdp_aad_device_code_request(client,
+            tenant_id, client_id, scope);
```
bbennett-ks, line 969: Since `tenant_id` is inserted into a URL, we should validate it to only allow expected characters to prevent URL injection attacks.

bbennett-ks (changes requested):

src/protocols/rdp/aad.c
```diff
@@ -0,0 +1,934 @@
...
+
+    guac_stream* stream = guac_user_alloc_stream(user);
+    if (stream == NULL) {
+        guac_client_log(user->client, GUAC_LOG_WARNING,
+                "AAD: No stream available to display the sign-in prompt");
+        free(body);
+        return NULL;
+    }
+
+    int failed = guac_protocol_send_pipe(user->socket, stream,
```
bbennett-ks: If `guac_protocol_send_pipe()` fails, should we log a message?

src/protocols/rdp/aad.c
```diff
@@ -0,0 +1,934 @@
...
+                    GUAC_RDP_AAD_MIMETYPE, GUAC_RDP_AAD_PIPE_NAME)
+            || guac_protocol_send_blob(user->socket, stream, body, strlen(body))
+            || guac_socket_flush(user->socket);
+
+    free(body);
+
+    if (failed) {
+        guac_client_log(user->client, GUAC_LOG_DEBUG,
```
bbennett-ks: If `guac_protocol_send_pipe()` succeeds, should we call `guac_protocol_send_end()` if `guac_protocol_send_end()` fails()?

bbennett-ks (changes requested):

src/protocols/rdp/Makefile.am
aleitner: Added the curl cflags so both variables are included

src/protocols/rdp/Makefile.am
aleitner: Done

src/protocols/rdp/aad.c
aleitner: Now init curl once up front so it isn't done implicitly and racing across connections

src/protocols/rdp/aad.c
aleitner: Added a null check so we handle it failing instead of passing it on

src/protocols/rdp/aad.c
aleitner: Added validation on the tenant id so it can't inject anything into the url

src/protocols/rdp/aad.c
aleitner: The sends and flush are checked together now and logged if they fail

src/protocols/rdp/aad.c
aleitner: The close path checks the send and flush and logs if they fail

src/protocols/rdp/aad.c
bbennett-ks: Ah, thinking it through there's no reason for a temp variable as there is not list to lose.

bbennett-ks (approved):
LGTM!

src/protocols/rdp/aad.c
```diff
@@ -0,0 +1,1027 @@
...
+
+void guac_rdp_aad_send_prompt(guac_client* client, guac_socket* socket,
+        const char* body) {
+
+    /* Send the prompt as a complete pipe stream. The browser displays it and
+     * dismisses it on its own once the connection completes, so the stream is
+     * not held open. */
+    guac_stream* stream = guac_client_alloc_stream(client);
+    if (stream == NULL) {
+        guac_client_log(client, GUAC_LOG_DEBUG,
```
bbennett-ks: Should this be a warning?

src/protocols/rdp/aad.c
```diff
@@ -0,0 +1,1027 @@
...
+
+    if (guac_protocol_send_pipe(socket, stream, GUAC_RDP_AAD_MIMETYPE,
+                GUAC_RDP_AAD_PIPE_NAME)
+            || guac_protocol_send_blob(socket, stream, body, strlen(body))
+            || guac_protocol_send_end(socket, stream)
+            || guac_socket_flush(socket))
+        guac_client_log(client, GUAC_LOG_DEBUG,
+                "AAD: Failed to send the sign-in prompt");
+
+    guac_client_free_stream(client, stream);
```
bbennett-ks, line 905: If `guac_protocol_send_pipe()` succeeds, but `guac_protocol_send_blob()` fails, send & flush?

src/protocols/rdp/aad.c
```diff
@@ -0,0 +1,1027 @@
...
+    /* Begin the device authorization grant */
+    guac_rdp_aad_device_code* dc = guac_rdp_aad_device_code_request(client,
+            tenant_id, client_id, scope);
+    if (dc == NULL)
+        return NULL;
+
+    /* Store the prompt so users who join or reconnect while sign-in is pending
+     * are brought up to date, then show it to the current owner */
+    char* body = guac_rdp_aad_build_prompt_json(dc);
+    if (body != NULL) {
```
bbennett-ks, line 976: Log a warning if `body == NULL`?

src/protocols/rdp/aad.c
```diff
@@ -0,0 +1,1027 @@
...
+    time_t deadline = time(NULL) + dc->expires_in;
+    char* token = NULL;
+
+    while (client->state == GUAC_CLIENT_RUNNING && time(NULL) < deadline
+            && guac_client_for_owner(client, guac_rdp_aad_owner_present, NULL)) {
+
+        /* Sleep in short slices so a disconnect is noticed within ~1s rather
+         * than after the full polling interval */
+        for (int i = 0; i < interval && client->state == GUAC_CLIENT_RUNNING; i++)
+            sleep(1);
```
bbennett-ks, line 1008: In `for (int i` loop, maybe check for `time(NULL) < deadline` and `guac_client_for_owner(client, guac_rdp_aad_owner_present, NULL)`, like in the `while` loop?

bbennett-ks (changes requested):

src/protocols/rdp/aad.c
aleitner: Made it a warning.

src/protocols/rdp/aad.c
aleitner: Reworked so end and flush always run once the pipe is open, even if the blob send fails.

src/protocols/rdp/aad.c
aleitner: Added a warning on the null path.

src/protocols/rdp/aad.c
aleitner: Added the deadline and owner checks to the sleep loop too.

bbennett-ks (approved):
New changes LGTM!

PR #637, opened 2026-02-20
GUACAMOLE-2220: Support Shift/Alt/Ctrl/Meta sequences for terminal navigation and function keys. 

bbennett-ks:
### Related PRs
- [GUACAMOLE-2220 (client 1207)](https://github.com/apache/guacamole-client/pull/1207)
- [GUACAMOLE-2220 (manual 291)](https://github.com/apache/guacamole-manual/pull/291)

### Overview: Shift/Alt/Ctrl/Meta Support

This change improves terminal keyboard interoperability by making special keys modifier-aware across all terminal types.

`Shift, Alt, Ctrl, and Meta` combinations are now supported for arrow keys, function keys, cursor editing keys, and keypad keys using xterm-style CSI modifier encoding, _even for terminal types that do not natively support these extensions_.

The terminal type for SSH and Telnet sessions is now used to control keyboard behavior, ensuring that key sequences align with the expectations of the configured $TERM.

For non-`linux` terminal types, keyboard handling follows standard xterm and VT-style behavior, with modified keys emitting consistent CSI sequences.

For the `linux` terminal type, legacy Linux console keycodes are preserved for compatibility with $TERM=linux, including `F1` to `F5` and `Home` and `End`, while still benefiting from modifier-aware extensions. `F11` and `F12` sequences have also been corrected to match xterm and terminfo standards.

As a result, all terminal types, including older ones like `ANSI` and `VT100`, gain support for modern keyboard behavior such as modifier-aware arrow keys, function keys `F1` to `F12`, and cursor editing keys, while remaining backward compatible with existing behavior.

In addition:
- Modifier key tracking has been expanded to include `Super`, `Hyper`, and `AltGr` (MODE_SWITCH, ISO_LEVEL3_SHIFT, ISO_LEVEL5_SHIFT), which are now mapped to `Meta` and `Alt` respectively.
- Key handling has also been refactored to replace raw keysym literals with named constants and centralized helper logic, improving readability and maintainability.

### Overview: Kubernetes terminal-type Support
Unlike SSH and Telnet, Kubernetes sessions do not automatically pass a terminal type. Kubernetes exposes stdin, stdout, and stderr, but has no built-in way to propagate the TERM environment variable.

Kubernetes supports two connection modes:
- Attach: connect stdin/stdout/stderr to the main process already running in the container.
- Exec: start a new process in the container and connect to that process.

Because Kubernetes does not pass the terminal type itself, this change adds a server-side substitution mechanism for Exec sessions: Any **$TERM** in the Exec command line is replaced with the configured terminal type when the session starts. This lets users decide whether and how to pass the terminal type, for example:
```
env TERM=$TERM /bin/bash
```
Attach mode has no command line, so the terminal type is not propagated automatically.

### Testing
Use showkey -a inside SSH or Telnet session to verify output.
```showkey -a

  Press keys (Ctrl+D to exit)```

## Notes
  - Test both physical keyboard and Guacamole on-screen keyboard (OSK), since they can differ by platform/browser and help verify modifiers that are hard to produce physically.
  - Some combinations may be intercepted by host OS/window manager/browser before reaching Guacamole (common for Command+..., Ctrl+F4, global shortcuts, workspace switching, etc.).

### DIFF
--- a/src/protocols/kubernetes/url.c
+++ b/src/protocols/kubernetes/url.c
@@ -19,6 +19,7 @@
 
 #include "url.h"
 
+#include <ctype.h>
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
@@ -89,7 +90,87 @@ int guac_kubernetes_escape_url_component(char* output, int length,
 
 }
 
-int guac_kubernetes_append_endpoint_param(char* buffer, int length, 
+/**
+ * Splits exec_command on whitespace and appends each argument as a separate
+ * Kubernetes "command=..." query parameter.
+ *
+ * Quoting is not supported. "$TERM" is replaced with terminal_type unless it
+ * is part of a longer shell variable name, such as "$TERMINAL".
+ *
+ * @param buffer
+ *     The buffer that should receive the appended query parameters.
+ *
+ * @param length
+ *     The number of bytes available in the given buffer.
+ *
+ * @param exec_command
+ *     The command string to split into arguments.
+ *
+ * @param terminal_type
+ *     The replacement value for "$TERM", or NULL if "$TERM" should be
+ *     replaced with an empty string.
+ *
+ * @return
+ *     Zero on success, non-zero if the buffer is too small.
+ */
+static int guac_kubernetes_append_exec_command(char* buffer, int length,
+        const char* exec_command, const char* terminal_type) {
+
+    const char* replacement = terminal_type != NULL ? terminal_type : "";
+    size_t replacement_length = strlen(replacement);
+
+    char arg[GUAC_KUBERNETES_MAX_ENDPOINT_LENGTH];
+    size_t arg_length = 0;
+
+    /* Cast before ctype.h calls to avoid undefined behavior. */
+    while (*exec_command != '\0') {
+
+        while (isspace((unsigned char)*exec_command))
+            exec_command++;
+
+        arg_length = 0;
+
+        while (*exec_command != '\0' && !isspace((unsigned char)*exec_command)) {
+
+            if (strncmp(exec_command, "$TERM", 5) == 0) {
+
+                /* Match "$TERM" only when it is not part of a longer shell
+                 * variable name, like "$TERMINAL". */
+                char next = exec_command[5];
+
+                if (!(isalnum((unsigned char)next) || next == '_')) {
+
+                    /* Leave room for the final null terminator. */
+                    if (arg_length + replacement_length >= sizeof(arg))
+                        return 1;
+
+                    memcpy(arg + arg_length, replacement, replacement_length);
+                    arg_length += replacement_length;
+                    exec_command += 5;
+                    continue;
+                }
+            }
+
+            if (arg_length + 1 >= sizeof(arg))
+                return 1;
+
+            arg[arg_length++] = *exec_command++;
+        }
+
+        /* Avoid adding "command=" for trailing whitespace. */
+        if (arg_length > 0) {
+            arg[arg_length] = '\0';
+
+            if (guac_kubernetes_append_endpoint_param(buffer, length,
+                        "command", arg))
+                return 1;
+        }
+    }
+
+    return 0;
+}
+
+int guac_kubernetes_append_endpoint_param(char* buffer, int length,
         const char* param_name, const char* param_value) {
 
     char escaped_param_value[GUAC_KUBERNETES_MAX_ENDPOINT_LENGTH];
@@ -98,7 +179,7 @@ int guac_kubernetes_append_endpoint_param(char* buffer, int length,
     if (guac_kubernetes_escape_url_component(escaped_param_value,
                     sizeof(escaped_param_value), param_value))
             return 1;
-    
+
     char* str = buffer;
 
     int str_len = 0;
@@ -139,7 +220,8 @@ int guac_kubernetes_append_endpoint_param(char* buffer, int length,
 
 int guac_kubernetes_endpoint_uri(char* buffer, int length,
         const char* kubernetes_namespace, const char* kubernetes_pod,
-        const char* kubernetes_container, const char* exec_command) {
+        const char* kubernetes_container, const char* exec_command,
+        const char* terminal_type) {
 
     char escaped_namespace[GUAC_KUBERNETES_MAX_ENDPOINT_LENGTH];
     char escaped_pod[GUAC_KUBERNETES_MAX_ENDPOINT_LENGTH];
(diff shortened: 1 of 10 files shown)

### CODE REVIEW

File: src/terminal/terminal.c [hunk 1]
```diff
@@ -1423,6 +1544,188 @@ int guac_terminal_send_string(guac_terminal* term, const char* data) {
 
 }
 
+/**
+ * Returns true if the given keysym is an arrow key.
+ */
```
necouchman, line 1577: These functions should all be documented using the same JSDoc-style comments that are generally used throughout the code. Really this just means adding the `@param` and `@return` tags.

necouchman (changes requested):
Overall looks really good, and very helpful in adding the constants to the terminal code. The functions should be documented in a manner consistent with the rest of the code - other than that, should the setting also be added to the Kubernetes protocol? I'm not sure if the Linux keycodes apply there the same way they do with Telnet and SSH?

bbennett-ks:
> ... should the setting also be added to the Kubernetes protocol? I'm not sure if the Linux keycodes apply there the same way they do with Telnet and SSH?

Yep, sounds very reasonable to add terminal type to Kubernetes. I'll create a Kubernetes setup to confirm & validate.

bbennett-ks:
> ... should the setting also be added to the Kubernetes protocol? I'm not sure if the Linux keycodes apply there the same way they do with Telnet and SSH?
@necouchman @mike-jumper 
I've added `terminal-type` to `Kubernetes`. I've tested Byobu in a Kubernetes pod and it works fine. However, there is an issue ...

Guacamole supports two mechanisms to attaching to a (running) container:
1. `attach`: Uses the Kubernetes `Attach` API to attach to the containers already-running main process/tty stream. This is used when the `Command (exec)` field is blank. With this mechanism, there is no way to "automatically" pass `TERM`. Though the user can do this manually.
2. `exec`: Uses the `Exec` API to start a new process withing the container. For example, `/bin/bash`. With this mechanism, we can pass `TERM` by wrapping the command with `env TERM=`. However, this may cause issues with non-POSIX images.

**The core issue is #1: Unlike `ssh` and `Telnet`, we cannot pass TERM.**

I still think we should do it, to support Byobu, tmux... for power users.

I can see several ways to handle this, none really ideal.

1. For consistency (amongst Kubernetes), don't set `TERM` for `Exec`: Add note to Kubernetes setting `TERM` this must be done manually.
2. If `Command (exec)` is blank, use `/bin/bash`. Or even have a `Default shell` field. (Uhg)
3. Set `TERM` for `Exec`, and have a note that `TERM` needs to be set manually if `Command (exec)` is blank.
4. Not support `terminal-type` for Kubernetes.

~~I'm leaning towards #3...~~ I'm leaning towards #1.

bbennett-ks:
**Command specified: TERM passed to session.**
image

**Command not specified: TERM not passed  passed to session.**
image

I have a UI change that makes option #3 (Set TERM for Exec, and have a note that TERM needs to be set manually if Command (exec) is blank) more palatable. I've added support for the ability for the protocol Json files to dynamically change field labels base on the the value of other fields. e.g. a `terminal-type` field label `Terminal type (TERM must be set manually` is displayed if `Command (exec):` is empty. 

This done by adding `conditionalLabel` to the field definition, an optional condition that allows a field's label to vary based on the value of another field (in the same form).
```
                {
                    "name"      : "terminal-type",
                    "type"      : "ENUM",
                    "options"   : [ "", "xterm", "xterm-256color", "vt220", "vt100", "ansi", "linux" ],
                    "conditionalLabel" : {
                        "field"    : "exec-command",
                        "equals"   : "",
                        "labelKey" : "FIELD_HEADER_TERMINAL_TYPE_MANUAL_TERM"
                    }
                }
```

bbennett-ks:
I've pushed an additional commit, adding Kubernetes support for `terminal-type`. I've for option #1 (For consistency (amongst Kubernetes), don't set TERM), as passing `TERM` by `env` breaks non-POSIX containers and other containers that have minimal environments that don't support `env`.

Instead, `$TERM` can be used in the  Exec command to pass the `terminal-type` to the container. e.g. `env TERM=$TERM /bin/bash`. This will be included in the documentation. I corresponding client and manual PRs will be created.

src/terminal/terminal.c
```diff
@@ -55,6 +55,126 @@
...
+ * - X.Org keysyms: https://cgit.freedesktop.org/xorg/proto/x11proto/tree/keysymdef.h
+ * - xterm control sequences: https://invisible-island.net/xterm/ctlseqs/ctlseqs.html
+ * - ASCII control codes (C0/DEL): https://www.rfc-editor.org/rfc/rfc20
+ */
+
+/* See X11 keysymdef.h */
+#define GUAC_KEYSYM_UNICODE_MASK    0xFFFF0000 /* Mask to test for 0x0100xxxx Unicode-keysym prefix. */
+#define GUAC_KEYSYM_UNICODE_PREFIX  0x01000000 /* X11 Unicode keysym prefix (0x01000000 | codepoint). */
+
+#define GUAC_KEYSYM_KP_ASCII_OFFSET 0xFF80 /* Keypad keysyms start at 0xFF80 (XK_KP_Space). */
```
necouchman: I'd suggest that all three of these be properly documented rather than having the end-of-line documentation.

src/terminal/terminal.c
```diff
@@ -55,6 +55,126 @@
...
+
+typedef enum guac_terminal_keysym {
```
necouchman, line 118: Please document this `typedef`.

src/protocols/kubernetes/url.c
```diff
@@ -89,6 +90,72 @@ int guac_kubernetes_escape_url_component(char* output, int length,
...
 
+/**
+ * Splits exec_command on whitespace and appends each argument as a separate
+ * Kubernetes "command=..." query parameter.
+ *
+ * Quoting is not supported. "$TERM" is replaced with terminal_type unless it
+ * is part of a longer shell variable name, such as "$TERMINAL".
+ *
+ * @return Zero on success, non-zero if the buffer is too small.
+ */
```
necouchman, line 115: This documentation block needs:
* `@param` tags for the parameters.
* Proper formatting in the `@return` tag (newline).

necouchman (changes requested):
A few more style/comment issues. Almost there, I think :-).

src/terminal/terminal.c
bbennett-ks: Done!

src/terminal/terminal.c
bbennett-ks: Because of the density, I opted for the member comment style of typedef `enum guac_composite_mode`.  Let me know if you prefer the more prevalent style.

src/protocols/kubernetes/url.c
bbennett-ks: Done!

necouchman (approved):

PR #640, opened 2026-02-26
GUACAMOLE-2234: Fix race condition in non-opaque layer clearing across display worker threads.

sschiffli:
Adding path_lock to guac_display_layer, acquired around the rect+cfill pair in guac_display_layer_clear_non_opaque(). The lock is per-layer, so workers on different layers are unaffected.

### DIFF
--- a/src/libguac/display-layer-list.c
+++ b/src/libguac/display-layer-list.c
@@ -259,6 +259,8 @@ guac_display_layer* guac_display_add_layer(guac_display* display, guac_layer* la
     display_layer->layer = layer;
     display_layer->opaque = opaque;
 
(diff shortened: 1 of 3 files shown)


necouchman:
> Is there a macro in the FreeRDP version information header that includes the git hash? We could build in an exception for this specific version.

I do not see one, no.

> 
> Otherwise, I'd say we could turn the dev version check into a warning and just make sure that the warning shows up at the end of the configure summary (so users will see it).

Okay, may just change the `AC_MSG_ERROR` to `AC_MSG_WARN`. I'll put in a Jira ticket and pull request for it.

PR #645, opened 2026-03-13
GUACAMOLE-1998: Refactor rwlock TLS tracking to use a single global key with per-thread slot array.

sschiffli:
Alternate fix proposal that still uses native pthread key, but now only 1 global instead of 2 for every active connection within the root guacd process.

### DIFF
--- a/src/libguac/rwlock.c
+++ b/src/libguac/rwlock.c
@@ -17,9 +17,12 @@
  * under the License.
  */
 
+#include <limits.h>
 #include <pthread.h>
 #include <stdint.h>
+
 #include "guacamole/error.h"
+#include "guacamole/mem.h"
 #include "guacamole/rwlock.h"
 
 /**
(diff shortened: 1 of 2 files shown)


src/libguac/rwlock.c
```diff
@@ -38,123 +41,173 @@
...
+ * via pthread_once.
+ */
+static pthread_key_t guac_rwlock_key;
+static pthread_once_t guac_rwlock_key_init = PTHREAD_ONCE_INIT;
+
+/**
+ * Destructor called by pthreads when a thread exits. Frees the per-thread
+ * state array.
+ */
+static void guac_rwlock_free_pointer(void* pointer) {
```
mike-jumper, line 90: Please document the `pointer` parameter.

src/libguac/rwlock.c
```diff
@@ -38,123 +41,173 @@
...
- *
- * @return
- *     The difference between the number of lock and unlock requests made by
- *     the current thread.
+ * Returns the state slot for the given lock for the current thread, or NULL
+ * if the current thread does not hold that lock.
  */
-static uintptr_t get_lock_count(uintptr_t value) {
-    return value >> 4;
+static guac_rwlock_thread_state* guac_rwlock_state_get(guac_rwlock* lock) {
```
mike-jumper, line 139: Please document the parameter and return value.

src/libguac/rwlock.c
```diff
@@ -38,123 +41,173 @@
...
- *     The current count for a lock that the current thread is trying to
- *     reentrantly acquire.
- *
- * @return
- *     Zero if adding one to the current count would overflow the storage
- *     allocated to the count, or a non-zero value otherwise.
+ * Clears a state slot, marking it as empty so it can be reused.
  */
-static int would_overflow_count(uintptr_t current_count) {
+static void guac_rwlock_state_clear(guac_rwlock_thread_state* state) {
```
mike-jumper, line 191: Please document the `state` parameter.

mike-jumper (changes requested):
Looks great! Just need some docs (largely fine already, but the Doxygen annotations are missing).

mike-jumper (approved):

PR #657, opened 2026-04-09
GUACAMOLE-2259: Set initial height/width on last_frame to 0

sschiffli:
This ensures the check for layer size changes will send the proper ```size``` instruction even if the layer happens to be GUAC_DISPLAY_RESIZE_FACTOR x GUAC_DISPLAY_RESIZE_FACTOR (currently 64x64)

```
/* Commit any change in layer size */
if (current->pending_frame.width != current->last_frame.width
        || current->pending_frame.height != current->last_frame.height) {

```

Having the last_frame buffer initially NULL also more clearly conveys that no frame has been sent yet, and once a frame has been sent, the buffer reallocates on sizes changes and will function normally.

### DIFF
--- a/src/libguac/display-layer-list.c
+++ b/src/libguac/display-layer-list.c
@@ -181,13 +181,10 @@ static void XFW_guac_display_layer_buffer_resize(guac_display_layer_state* frame
 static void PFW_LFW_guac_display_layer_state_init(guac_display_layer_state* last_frame,
         guac_display_layer_state* pending_frame) {
 
-    last_frame->width = pending_frame->width = GUAC_DISPLAY_RESIZE_FACTOR;
-    last_frame->height = pending_frame->height = GUAC_DISPLAY_RESIZE_FACTOR;
+    pending_frame->width = GUAC_DISPLAY_RESIZE_FACTOR;
+    pending_frame->height = GUAC_DISPLAY_RESIZE_FACTOR;
     last_frame->opacity = pending_frame->opacity = 0xFF;
     last_frame->parent = pending_frame->parent = GUAC_DEFAULT_LAYER;
-
-    XFW_guac_display_layer_buffer_resize(last_frame,
-            last_frame->width, last_frame->height);
 
     XFW_guac_display_layer_buffer_resize(pending_frame,
             pending_frame->width, pending_frame->height);


mike-jumper (approved):

necouchman:
@sschiffli @mike-jumper I'm seeing a regression in guacd after the merge of this PR, where guacd is very inconsistent about whether it connects or not, often reporting the following:
```
guacd[3]: ERROR:	User is not responding.
guacd[3]: INFO:	User "@b85e2dfc-2dc9-452d-b29f-af610a49f1e3" disconnected (0 users remain)
guacd[3]: INFO:	Last user of connection "$b84394fe-742d-4ec0-b73f-cffc4b3e8853" disconnected
guacd[3]: DEBUG:	Requesting termination of client...
```
git bisect tracked it down to this merge:
```
0e34b9a0c1634b11e9bc6af3397ef54865619681 is the first bad commit
commit 0e34b9a0c1634b11e9bc6af3397ef54865619681 (origin/staging/1.6.1, staging/1.6.1)
Merge: 000d6198 3d69c6c3
Author: Mike Jumper <mjumper@apache.org>
Date:   Wed Apr 15 02:16:39 2026 -0700

    GUACAMOLE-2259: Merge correction to size of layer in unsent frame (should be 0x0, not 64x64).

 src/libguac/display-layer-list.c | 7 ++-----
 1 file changed, 2 insertions(+), 5 deletions(-)
```

sschiffli:
Can you confirm if the client has this fix applied or not? https://github.com/apache/guacamole-client/pull/1199

necouchman:
@sschiffli Ah, probably not. I will take a look at that and give it a shot.

This does present a bit of a challenge in maintaining backward compatibility, as this bug fix essentially represents a change that breaks compatibility at this point, and thus will likely make version 1.6.1 of guacd incompatible with older versions of the client.

sschiffli:
Yeah this is a weird case. I could implement a workaround in the server to maintain backwards compatibility.  Either setting the initial height/width to 1 here, or modifying `guac_display_dup` to not send `guac_protocol_send_cursor` if the last frame is 0x0 [here](https://github.com/apache/guacamole-server/blob/staging/1.6.1/src/libguac/display.c#L314-L320)

necouchman:
Yeah, I think _something_ should be done to make sure compatibility is maintained. If what you're suggesting maintains compatibility but also resolves the bug, that would be fine. Otherwise we could consider bumping up the Guacamole Protocol version for this release.

@mike-jumper Thoughts?

sschiffli:
Added PR to address backwards compatibility: https://github.com/apache/guacamole-server/pull/664

PR #661, opened 2026-04-14
GUACAMOLE-2263: Harden CLIPRDR channel against FreeRDP 3.x clipboard state machine.

escra:
## Summary

Six fixes for the clipboard redirection (CLIPRDR) plugin that prevent RDP
session disconnects when clipboard operations are performed with FreeRDP 3.x
clients.

FreeRDP 3.x has a significantly stricter CLIPRDR state machine than 2.x.
Several patterns that were harmless with FreeRDP 2.x now cause session
teardowns:

1. **Overlapping Format Data Requests** - FreeRDP 3.x tracks pending requests
   internally and disconnects with error 1359 if a second request arrives
   before the first is answered. Fix: add `request_pending` flag with
   `request_lock` mutex.

2. **CB_RESPONSE_FAIL not checked** - If the server responds with a failure
   flag, the old code would attempt to process the (NULL/garbage) data,
   causing a crash. Fix: check `msgFlags & CB_RESPONSE_FAIL` early.

3. **No response for unsupported formats** - When the server requests an
   unsupported clipboard format, the old code silently drops the request.
   FreeRDP 3.x times out after 10 seconds and tears down the session. Fix:
   send `CB_RESPONSE_FAIL` for unsupported formats.

4. **NULL/empty data guard** - `requestedFormatData` can be NULL or
   `dataLen` can be zero in edge cases. Fix: guard before processing.

5. **Stale pending requests** - When a new Format List arrives (server
   clipboard content changed), any pending request for old data becomes
   stale. Fix: clear `request_pending` when a new Format List is received.

6. **Non-OK return codes treated as fatal** - FreeRDP 3.x treats non-OK
   returns from CLIPRDR callbacks as fatal channel errors. Fix: all callbacks
   now return `CHANNEL_RC_OK` and log errors instead.

### Related

- [FreeRDP/FreeRDP#11847](https://github.com/FreeRDP/FreeRDP/issues/11847) -
  FreeRDP 3.x disconnects with error 1359 on clipboard operations
- [GUACAMOLE-2118](https://issues.apache.org/jira/browse/GUACAMOLE-2118) -
  Sporadic hanging connections after upgrade

### Affected versions

- `staging/1.6.1` - with FreeRDP 3.x (default in Docker image)
- `main` - same patterns exist

## Test plan

- [ ] Connect from FreeRDP 3.x client, perform rapid copy/paste operations
- [ ] Copy unsupported clipboard format (e.g., image) on server, verify no disconnect
- [ ] Verify clipboard works normally with FreeRDP 2.x (no regression)
- [ ] Stress test: rapid clipboard changes while switching windows
- [ ] Verify no error 1359 disconnects under load

### DIFF
--- a/src/protocols/rdp/channels/cliprdr.c
+++ b/src/protocols/rdp/channels/cliprdr.c
@@ -172,22 +172,23 @@ static UINT guac_rdp_cliprdr_send_capabilities(CliprdrClientContext* cliprdr) {
 static UINT guac_rdp_cliprdr_monitor_ready(CliprdrClientContext* cliprdr,
         CLIPRDR_CONST CLIPRDR_MONITOR_READY* monitor_ready) {
 
-    /* FreeRDP-specific handlers for CLIPRDR are not assigned, and thus not
-     * callable, until after the relevant guac_rdp_clipboard structure is
-     * allocated and associated with the CliprdrClientContext */
     guac_rdp_clipboard* clipboard = (guac_rdp_clipboard*) cliprdr->custom;
     assert(clipboard != NULL);
 
     guac_client_log(clipboard->client, GUAC_LOG_TRACE, "CLIPRDR: Received "
             "monitor ready.");
 
-    /* Respond with capabilities ... */
     int status = guac_rdp_cliprdr_send_capabilities(cliprdr);
     if (status != CHANNEL_RC_OK)
-        return status;
+        guac_client_log(clipboard->client, GUAC_LOG_WARNING, "CLIPRDR: "
+                "Failed to send capabilities (error %d).", status);
 
-    /* ... and supported format list */
-    return guac_rdp_cliprdr_send_format_list(cliprdr);
+    status = guac_rdp_cliprdr_send_format_list(cliprdr);
+    if (status != CHANNEL_RC_OK)
+        guac_client_log(clipboard->client, GUAC_LOG_WARNING, "CLIPRDR: "
+                "Failed to send format list (error %d).", status);
+
+    return CHANNEL_RC_OK;
 
 }
 
@@ -211,30 +212,45 @@ static UINT guac_rdp_cliprdr_monitor_ready(CliprdrClientContext* cliprdr,
 static UINT guac_rdp_cliprdr_send_format_data_request(
         CliprdrClientContext* cliprdr, UINT32 format) {
 
-    /* FreeRDP-specific handlers for CLIPRDR are not assigned, and thus not
-     * callable, until after the relevant guac_rdp_clipboard structure is
-     * allocated and associated with the CliprdrClientContext */
     guac_rdp_clipboard* clipboard = (guac_rdp_clipboard*) cliprdr->custom;
     assert(clipboard != NULL);
 
     guac_client* client = clipboard->client;
     guac_rdp_client* rdp_client = (guac_rdp_client*) client->data;
 
-    /* Create new data request */
+    /* Prevent overlapping requests — FreeRDP 3.x tracks pending requests
+     * internally and will disconnect with error 1359 if a response arrives
+     * without a matching pending request. */
+    pthread_mutex_lock(&(clipboard->request_lock));
+    if (clipboard->request_pending) {
+        guac_client_log(client, GUAC_LOG_DEBUG, "CLIPRDR: Skipping format "
+                "data request — previous request still pending.");
+        pthread_mutex_unlock(&(clipboard->request_lock));
+        return CHANNEL_RC_OK;
+    }
+
+    clipboard->requested_format = format;
+    clipboard->request_pending = 1;
+    pthread_mutex_unlock(&(clipboard->request_lock));
+
     CLIPRDR_FORMAT_DATA_REQUEST data_request = {
         .requestedFormatId = format
     };
 
-    /* Note the format we've requested for reference later when the requested
-     * data is received via a Format Data Response PDU */
-    clipboard->requested_format = format;
+    guac_client_log(client, GUAC_LOG_TRACE, "CLIPRDR: Sending format data "
+            "request for format 0x%X.", format);
 
-    guac_client_log(client, GUAC_LOG_TRACE, "CLIPRDR: Sending format data request.");
-
-    /* Send request */
     pthread_mutex_lock(&(rdp_client->message_lock));
     int retval = cliprdr->ClientFormatDataRequest(cliprdr, &data_request);
     pthread_mutex_unlock(&(rdp_client->message_lock));
+
+    if (retval != CHANNEL_RC_OK) {
+        pthread_mutex_lock(&(clipboard->request_lock));
+        clipboard->request_pending = 0;
+        pthread_mutex_unlock(&(clipboard->request_lock));
+        guac_client_log(client, GUAC_LOG_WARNING, "CLIPRDR: Failed to send "
+                "format data request (error %d).", retval);
+    }
 
     return retval;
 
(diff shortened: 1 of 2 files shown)


src/protocols/rdp/channels/cliprdr.c
```diff
@@ -172,22 +172,23 @@ static UINT guac_rdp_cliprdr_send_capabilities(CliprdrClientContext* cliprdr) {
...
 
     guac_client_log(clipboard->client, GUAC_LOG_TRACE, "CLIPRDR: Received "
             "monitor ready.");
 
-    /* Respond with capabilities ... */
     int status = guac_rdp_cliprdr_send_capabilities(cliprdr);
     if (status != CHANNEL_RC_OK)
-        return status;
+        guac_client_log(clipboard->client, GUAC_LOG_WARNING, "CLIPRDR: "
+                "Failed to send capabilities (error %d).", status);
```
necouchman, line 184: I understand the goal of this - to continue the connection even if clipboard redirection fails to sync up capabilities with the server. That said, shouldn't we at least still `return CHANNEL_RC_OK`, here, and not try to send a format list if we've failed to send capabilities?

src/protocols/rdp/channels/cliprdr.c
```diff
@@ -292,9 +308,6 @@ static int guac_rdp_cliprdr_format_supported(const CLIPRDR_FORMAT_LIST* format_l
 static UINT guac_rdp_cliprdr_format_list(CliprdrClientContext* cliprdr,
         CLIPRDR_CONST CLIPRDR_FORMAT_LIST* format_list) {
 
-    /* FreeRDP-specific handlers for CLIPRDR are not assigned, and thus not
-     * callable, until after the relevant guac_rdp_clipboard structure is
-     * allocated and associated with the CliprdrClientContext */
```
necouchman, line 296: Is this comment incorrect?

src/protocols/rdp/channels/cliprdr.c
```diff
@@ -313,23 +337,26 @@ static UINT guac_rdp_cliprdr_format_list(CliprdrClientContext* cliprdr,
         .msgFlags = CB_RESPONSE_OK
 #endif
     };
-    /* Report successful processing of format list */
```
necouchman, line 315: Is this comment not accurate?

src/protocols/rdp/channels/cliprdr.c
```diff
@@ -357,21 +384,20 @@ static UINT guac_rdp_cliprdr_format_list(CliprdrClientContext* cliprdr,
...
 
-    guac_client_log(client, GUAC_LOG_TRACE, "CLIPRDR: Received format data request.");
+    guac_client_log(client, GUAC_LOG_TRACE, "CLIPRDR: Received format data "
+            "request for format 0x%X.", format_data_request->requestedFormatId);
 
     guac_iconv_write* remote_writer;
+    int output_buf_size = clipboard->clipboard->available;
     const char* input = clipboard->clipboard->buffer;
-    char* output = guac_mem_alloc(GUAC_COMMON_CLIPBOARD_MAX_LENGTH);
+    char* output = guac_mem_alloc(output_buf_size);
```
necouchman, line 400: I'm not sure if this overlaps with #659, or is related to the discussion on that thread regarding allocating `available * 2` or `available * 4`? If this is an overlap, and is already covered by the changes in that change request, then it should probably not be, here.

src/protocols/rdp/channels/cliprdr.c
```diff
@@ -384,25 +410,42 @@ static UINT guac_rdp_cliprdr_format_data_request(CliprdrClientContext* cliprdr,
...
+
+            pthread_mutex_lock(&(rdp_client->message_lock));
+            cliprdr->ClientFormatDataResponse(cliprdr, &fail_response);
+            pthread_mutex_unlock(&(rdp_client->message_lock));
             return CHANNEL_RC_OK;
 
     }
 
-    /* Send received clipboard data to the RDP server in the format
-     * requested */
```
necouchman, line 400: Why is this comment being removed?

src/protocols/rdp/channels/cliprdr.c
```diff
@@ -418,14 +461,15 @@ static UINT guac_rdp_cliprdr_format_data_request(CliprdrClientContext* cliprdr,
...
+            "response (%d bytes).", (int)(((BYTE*) output) - start));
 
     pthread_mutex_lock(&(rdp_client->message_lock));
-    UINT result = cliprdr->ClientFormatDataResponse(cliprdr, &data_response);
+    cliprdr->ClientFormatDataResponse(cliprdr, &data_response);
     pthread_mutex_unlock(&(rdp_client->message_lock));
 
     guac_mem_free(start);
-    return result;
+    return CHANNEL_RC_OK;
```
necouchman, line 472: The documentation for this function should be updated to indicate either that it always returns `CHANNEL_RC_OK` despite failures, or the situations where failures are returned vs. `CHANNEL_RC_OK`.

src/protocols/rdp/channels/cliprdr.c
```diff
@@ -450,17 +494,35 @@ static UINT guac_rdp_cliprdr_format_data_request(CliprdrClientContext* cliprdr,
 static UINT guac_rdp_cliprdr_format_data_response(CliprdrClientContext* cliprdr,
         CLIPRDR_CONST CLIPRDR_FORMAT_DATA_RESPONSE* format_data_response) {
 
-    /* FreeRDP-specific handlers for CLIPRDR are not assigned, and thus not
-     * callable, until after the relevant guac_rdp_clipboard structure is
-     * allocated and associated with the CliprdrClientContext */
```
necouchman, line 454: Why is this comment being removed?

src/protocols/rdp/channels/cliprdr.c
```diff
@@ -659,11 +723,10 @@ void guac_rdp_clipboard_load_plugin(guac_rdp_clipboard* clipboard,
 
 void guac_rdp_clipboard_free(guac_rdp_clipboard* clipboard) {
 
-    /* Do nothing if the clipboard is not actually allocated */
```
necouchman, line 661: Why has this comment been removed?

src/protocols/rdp/channels/cliprdr.c
```diff
@@ -659,11 +723,10 @@ void guac_rdp_clipboard_load_plugin(guac_rdp_clipboard* clipboard,
...
     if (clipboard == NULL)
         return;
 
-    /* Free clipboard and underlying storage */
```
necouchman, line 665: Why has this comment been removed?

necouchman (changes requested):
The code itself looks pretty solid, but I have a few things to clear up and/or change. Overall:
* I'm concerned that several of the changes, here, overlap with #659, particularly as it relates to allocation of clipboard space. If I'm wrong about that, just let me know, but otherwise please keep those changes distinct.
* You've removed comments from the code in several places, with no clear indication as to why. I've tried to put my review comments at each of those places, but please leave comments in-line with the code, unless there's a specific reason they should be removed. If they're no longer accurate, I'd suggest updating them rather than removing them.
* Please make sure to update function documentation for any of the functions where the return value is changing - a couple of the functions always return `CHANNEL_RC_OK`, now, so that should be noted in the function documentation with an explanation as to why that is the case.

Finally, while I understand the overall goal of this pull request, and while I'm overall okay with it, I think it's worth noting that, in at least some cases, this will allow a connection to continue without clipboard support, and with no indication to the user as to why that's the case. I think it's probably okay because that's how it currently works under FreeRDP 2.x - it silently ignores clipboard errors that FreeRDP 3.x now handles more properly; however, it seems like one of those situations that could be really frustrating from a user experience point of view. Instead of the user connection getting shut down and having to go find out from the admin why it isn't working, the connection will continue but every time the user tries to copy/paste something, it will silently and inexplicably fail. I don't know if there's really anything to do about that, but figure it's worth noting.

PR #666, opened 2026-04-29
GUACAMOLE-2223: Add require-recording flag

sivagudivadaz:
JIRA: https://issues.apache.org/jira/browse/GUACAMOLE-2223

Naming convention for recording files may not always be unique leading to suffix exhaustion or file creation could fail due to various reasons. This pr adds a new `require-recording` flag to enforce recording file creation when a connection attempt is made. 

Current behavior: 
Connection is allowed even if it can't be recorded due to recording file creation failure. 

Expected behavior: 
When the session can't be recorded, reject the attempt for compliance.

### DIFF
--- a/src/protocols/ssh/settings.h
+++ b/src/protocols/ssh/settings.h
@@ -289,6 +289,13 @@ typedef struct guac_ssh_settings {
     bool recording_write_existing;
 
     /**
+     * Non-zero if the connection should be closed when a recording cannot be
+     * created. Zero by default, meaning the connection will proceed even if
+     * recording fails.
+     */
+    bool require_recording;
+
+    /**
      * The number of seconds between sending server alive messages.
      */
     int server_alive_interval;
--- a/src/protocols/ssh/settings.h
+++ b/src/protocols/ssh/settings.h
@@ -289,9 +289,9 @@ typedef struct guac_ssh_settings {
     bool recording_write_existing;
 
     /**
-     * Non-zero if the connection should be closed when a recording cannot be
-     * created. Zero by default, meaning the connection will proceed even if
-     * recording fails.
+     * Whether the connection should be closed when a recording cannot be
+     * created. Disabled by default, meaning the connection will proceed even
+     * if recording fails.
      */
     bool require_recording;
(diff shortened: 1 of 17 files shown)


src/protocols/ssh/settings.h
```diff
@@ -278,6 +278,13 @@ typedef struct guac_ssh_settings {
      */
     bool recording_write_existing;
 
+    /**
+     * Non-zero if the connection should be closed when a recording cannot be
+     * created. Zero by default, meaning the connection will proceed even if
+     * recording fails.
+     */
+    bool require_recording;
```
necouchman, line 296: You use "zero" but are using a `bool` here (instead of `int` as in the RDP code). It should probably be both consistent and the language in the comments should match the type used.

necouchman (changes requested):
Overall this looks good, and I'm pretty happy with the changes.  A couple of things that need to be addressed:
* Consistency of using `int` vs. `bool` for tracking the setting. We should probably pick one and stick with it - RDP uses `int`, the other protocols use `bool`.
* Fix the commit message - it should be descriptive and should contain the `GUACAMOLE-2223` prefix.
* Should this functionality also be added for the Kubernetes protocol?

Also, it might be worth considering what happens if the recording fails during the connection? If this is being added for a case, for example, where Guacamole is being used for audit purposes for Privileged Access Management or the like, it seems like it might also be good to find a way to disrupt the connection if the recording stops in the middle (a disk fills up, for example)?

sivagudivadaz:
> it might also be good to find a way to disrupt the connection if the recording stops in the middle

yes, will need to add the check in https://github.com/apache/guacamole-server/blob/main/src/libguac/socket-tee.c. 

I'm also thinking of adding a new error code on the frontend that reflects the connection disruption is due to recording failure.

src/protocols/ssh/settings.h
sivagudivadaz: My bad, was primarily testing rdp protocol when I first made the changes. 
Updated this & the comments to reflect it's a bool flag

PR #668, opened 2026-05-01
GUACAMOLE-2268: VNC: add UTF-8 (Unicode) support.

bbennett-ks:
See Jira issue (https://issues.apache.org/jira/browse/GUACAMOLE-2268) for details.

### DIFF
--- a/src/protocols/vnc/clipboard.c
+++ b/src/protocols/vnc/clipboard.c
@@ -29,6 +29,7 @@
 #include <guacamole/stream.h>
 #include <guacamole/user.h>
 #include <rfb/rfbclient.h>
+#include <rfb/rfbconfig.h>
 #include <rfb/rfbproto.h>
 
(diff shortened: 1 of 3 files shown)


necouchman (approved):

necouchman:
@bbennett-ks Looks like we missed the checks in the configure script for this capability - as reported on the mailing list, it breaks builds for anything prior to 0.9.15 of libvncclient, which doesn't contain the UTF8 calls. Could you open a PR to address that?

bbennett-ks:
> @bbennett-ks Looks like we missed the checks in the configure script for this capability - as reported on the mailing list, it breaks builds for anything prior to 0.9.15 of libvncclient, which doesn't contain the UTF8 calls. Could you open a PR to address that?

Apologies! [GUACAMOLE-2268 (server 671)](https://github.com/apache/guacamole-server/pull/671)

PR #670, opened 2026-05-04
GUACAMOLE-2273: Implement FreeRDP AuthenticateEx callback and handle deprecation of Authenticate callback.

necouchman:
This pull request implements the `AuthenticateEx` callback for FreeRDP, and handles versions of FreeRDP where the `Authenticate` callback has been deprecated and compiled out.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -1010,6 +1010,22 @@ then
                       [[#include <freerdp/freerdp.h>]])
 fi
 
+# Check for either AuthenticateEx or Authenticate callbacks.
+if test "x${have_freerdp}" = "xyes"
+then
+    AC_CHECK_MEMBERS([freerdp.AuthenticateEx],,
+                     [AC_CHECK_MEMBERS([freerdp.Authenticate],,
+                                       [AC_MSG_WARN([
+  ----------------------------------------------------
+   This version of FreeRDP appears to lack support
+   for both Authenticate and AuthenticateEx functions.
+   RDP connections may fail if credentials are not
+   provided with the connection.
+  ----------------------------------------------------])],
+                                       [[#include <freerdp/freerdp.h>]])],
+                      [[#include <freerdp/freerdp.h>]])
+fi
+
 if test "x${have_freerdp}" = "xyes"
 then
     AC_CHECK_DECLS([winpr_aligned_free],
--- a/configure.ac
+++ b/configure.ac
@@ -986,6 +986,21 @@ then
                       [[#include <freerdp/freerdp.h>]])
 fi
 
+# Updated authentication callback. The previous version, Authenticate, has been
+# deprecated as of 3.25.0.
+if test "x${have_freerdp}" = "xyes"
+then
+    AC_CHECK_MEMBERS([freerdp.AuthenticateEx],,,
+                      [[#include <freerdp/freerdp.h>]])
+fi
+
+# Check for old authentication callback
+if test "x${have_freerdp}" = "xyes"
+then
+    AC_CHECK_MEMBERS([freerdp.Authenticate],,,
+                      [[#include <freerdp/freerdp.h>]])
+fi
+
 if test "x${have_freerdp}" = "xyes"
 then
     AC_CHECK_DECLS([winpr_aligned_free],
--- a/src/protocols/rdp/rdp.c
+++ b/src/protocols/rdp/rdp.c
@@ -231,6 +231,7 @@ static BOOL rdp_freerdp_pre_connect(freerdp* instance) {
     return TRUE;
 }
 
+#ifdef HAVE_FREERDP_AUTHENTICATE
 /**
  * Callback invoked by FreeRDP when authentication is required but the required
  * parameters have not been provided. In the case of Guacamole clients that
@@ -324,6 +325,139 @@ static BOOL rdp_freerdp_authenticate(freerdp* instance, char** username,
     return TRUE;
 
 }
+#endif // HAVE_FREERDP_AUTHENTICATE
+
+#ifdef HAVE_FREERDP_AUTHENTICATEEX
+/**
+ * Callback invoked by FreeRDP when authentication is required but the required
+ * parameters have not been provided. In the case of Guacamole clients that
+ * support the "required" instruction, this function will send any of the three
+ * unpopulated RDP authentication parameters back to the client so that the
+ * connection owner can provide the required information.  If the values have
+ * been provided in the original connection parameters the user will not be
+ * prompted for updated parameters. If the version of Guacamole Client in use
+ * by the connection owner does not support the "required" instruction then the
+ * connection will fail. This function always returns true. This callback is for
+ * the AuthenticateEx version, which also provides the "reason" field, indicating
+ * the type of authentication that the server is requesting from the client.
+ * Guacamole currently supports PIN and traditional username, password, and domain
+ * authentication.
+ *
+ * @param instance
+ *     The FreeRDP instance associated with the RDP session requesting
+ *     credentials.
+ *
+ * @param username
+ *     Pointer to a string which will receive the user's username.
+ *
+ * @param password
+ *     Pointer to a string which will receive the user's password.
+ *
+ * @param domain
+ *     Pointer to a string which will receive the domain associated with the
+ *     user's account.
+ *
+ * @return
+ *     Always TRUE.
+ */
+static BOOL rdp_freerdp_authenticate_ex(freerdp* instance, char** username,
+        char** password, char** domain, rdp_auth_reason reason) {
+
+    rdpContext* context = GUAC_RDP_CONTEXT(instance);
+    guac_client* client = ((rdp_freerdp_context*) context)->client;
+    guac_rdp_client* rdp_client = (guac_rdp_client*) client->data;
+    guac_rdp_settings* settings = rdp_client->settings;
+    char* params[4] = {NULL};
+    int i = 0;
+    
+    /* If the client does not support the "required" instruction, warn and
+     * quit.
+     */
+    if (!guac_client_owner_supports_required(client)) {
+        guac_client_log(client, GUAC_LOG_WARNING, "Client does not support the "
+                "\"required\" instruction. No authentication parameters will "
+                "be requested.");
+        return TRUE;
+    }
+
+    switch (reason) {
+
+        /* In the case that we just need a PIN, request only the PIN/password. */
+        case AUTH_SMARTCARD_PIN:
+        case AUTH_FIDO_PIN:
+
+            /* If the password/PIN is undefined, add it to the requested parameters. */
+            if (settings->password == NULL) {
+                guac_argv_register(GUAC_RDP_ARGV_PASSWORD, guac_rdp_argv_callback, NULL, 0);
+                params[i] = GUAC_RDP_ARGV_PASSWORD;
+                i++;
+            }
+
+            break;
+
+        /* For all other cases, request anything that isn't  configured. */
+        case AUTH_RDSTLS:
+        case AUTH_TLS:
+        case AUTH_RDP:
+        case AUTH_NLA:
+        case GW_AUTH_HTTP:
+        case GW_AUTH_RDG:
+        case GW_AUTH_RPC:
+
+            /* If the username is undefined, add it to the requested parameters. */
+            if (settings->username == NULL) {
+                guac_argv_register(GUAC_RDP_ARGV_USERNAME, guac_rdp_argv_callback, NULL, 0);
+                params[i] = GUAC_RDP_ARGV_USERNAME;
+                i++;
+
+                /* If username is undefined and domain is also undefined, request domain. */
+                if (settings->domain == NULL) {
+                    guac_argv_register(GUAC_RDP_ARGV_DOMAIN, guac_rdp_argv_callback, NULL, 0);
+                    params[i] = GUAC_RDP_ARGV_DOMAIN;
+                    i++;
+                }
+
+            }
+
+            /* If the password is undefined, add it to the requested parameters. */
+            if (settings->password == NULL) {
+                guac_argv_register(GUAC_RDP_ARGV_PASSWORD, guac_rdp_argv_callback, NULL, 0);
+                params[i] = GUAC_RDP_ARGV_PASSWORD;
+                i++;
+            }
+
+            break;
+
+        /* We should never get here. */
+        default:
+            return TRUE;
+
+    }
+
+    /* NULL-terminate the array. */
+    params[i] = NULL;
+
+    if (i > 0) {
+        
+        /* Send required parameters to the owner and wait for the response. */
+        guac_client_owner_send_required(client, (const char**) params);
+        guac_argv_await((const char**) params);
+        
+        /* Free old values and get new values from settings. */
+        guac_mem_free(*username);
+        guac_mem_free(*password);
+        guac_mem_free(*domain);
+        *username = guac_strdup(settings->username);
+        *password = guac_strdup(settings->password);
+        *domain = guac_strdup(settings->domain);
+        
+    }
+    
+    /* Always return TRUE allowing connection to retry. */
+    return TRUE;
+
+}
+#endif // HAVE_FREERDP_AUTHENTICATEEX
 
 #ifdef HAVE_FREERDP_VERIFYCERTIFICATEEX
 /**
@@ -549,11 +683,20 @@ static int guac_rdp_handle_connection(guac_client* client) {
      * If the freerdp instance has a LoadChannels callback for loading plugins
      * we use that instead of the PreConnect callback to load plugins.
      */
-    #ifdef RDP_INST_HAS_LOAD_CHANNELS
+#ifdef RDP_INST_HAS_LOAD_CHANNELS
         rdp_inst->LoadChannels = rdp_freerdp_load_channels;
-    #endif
+#endif
+
     rdp_inst->PreConnect = rdp_freerdp_pre_connect;
+
+#ifdef HAVE_FREERDP_AUTHENTICATE
     rdp_inst->Authenticate = rdp_freerdp_authenticate;
+#endif
+
+#ifdef HAVE_FREERDP_AUTHENTICATEEX
+    rdp_inst->AuthenticateEx = rdp_freerdp_authenticate_ex;
+#endif
+
 
 #ifdef HAVE_FREERDP_VERIFYCERTIFICATEEX
     rdp_inst->VerifyCertificateEx = rdp_freerdp_verify_certificate;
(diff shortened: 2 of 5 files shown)


necouchman:
@mike-jumper @corentin-soriano @sschiffli Anyone do a quick review of this?

src/protocols/rdp/rdp.c
```diff
@@ -326,6 +327,137 @@ static BOOL rdp_freerdp_authenticate(freerdp* instance, char** username,
...
+ * parameters have not been provided. In the case of Guacamole clients that
+ * support the "required" instruction, this function will send any of the three
+ * unpopulated RDP authentication parameters back to the client so that the
+ * connection owner can provide the required information.  If the values have
+ * been provided in the original connection parameters the user will not be
+ * prompted for updated parameters. If the version of Guacamole Client in use
+ * by the connection owner does not support the "required" instruction then the
+ * connection will fail. This function always returns true. This callback is for
+ * the AuthenticateEx version, which also provides the "reason" field, indicating
+ * the type of authentication used to
```
corentin-soriano: I think the end of the sentence is missing.

src/protocols/rdp/rdp.c
```diff
@@ -326,6 +327,137 @@ static BOOL rdp_freerdp_authenticate(freerdp* instance, char** username,
...
+    
+    /* If the client does not support the "required" instruction, warn and
+     * quit.
+     */
+    if (!guac_client_owner_supports_required(client)) {
+        guac_client_log(client, GUAC_LOG_WARNING, "Client does not support the "
+                "\"required\" instruction. No authentication parameters will "
+                "be requested.");
+        return TRUE;
+    }
```
corentin-soriano, line 528: Part of the code in `rdp_freerdp_authenticate_ex` seems to be redundant with `rdp_freerdp_authenticate`. Perhaps we could extract the common part?

necouchman: I'll take a look and see what might make sense. It's a little tricky because the newer function is broken up by the `switch` statement trying to check for the type of authentication that is being requested, but there are still probably a couple of items that could become `static` functions and shared by both versions.

necouchman: Okay, I've done some combining of the authentication items in the RDP code.

src/protocols/rdp/rdp.c
necouchman: Oops. Should be fixed via rebase.

src/protocols/rdp/rdp.c
```diff
@@ -549,11 +718,20 @@ static int guac_rdp_handle_connection(guac_client* client) {
...
+
     rdp_inst->PreConnect = rdp_freerdp_pre_connect;
+
+#ifdef HAVE_FREERDP_AUTHENTICATE
     rdp_inst->Authenticate = rdp_freerdp_authenticate;
+#endif
+
+#ifdef HAVE_FREERDP_AUTHENTICATEEX
+    rdp_inst->AuthenticateEx = rdp_freerdp_authenticate_ex;
+#endif
```
mike-jumper, line 819: If `AuthenticateEx` is the replacement for now-deprecated `Authenticate`, should `Authenticate` not be used at all when `AuthenticateEx` is available?

necouchman: Yes, that makes sense - I've addressed this via `configure` compile-time checks and a rebase.

src/protocols/rdp/rdp.c
```diff
@@ -279,51 +386,113 @@ static BOOL rdp_freerdp_authenticate(freerdp* instance, char** username,
...
-        if (settings->domain == NULL) {
-            guac_argv_register(GUAC_RDP_ARGV_DOMAIN, guac_rdp_argv_callback, NULL, 0);
-            params[i] = GUAC_RDP_ARGV_DOMAIN;
-            i++;
-        }
-        
-    }
+    /* Add the required parameters and send them over to the client. */
+    if (guac_rdp_register_required_auth(settings, params))
+        guac_rdp_send_auth_params(client, (const char**) params, username, password, domain);
```
mike-jumper, line 464: The `password` and `domain` parameters here are swapped vs. the `guac_rdp_send_auth_params()` function definition.

Rather than just correcting the call, though, I'd suggest that the function either use the same ordering as the connection parameters, or the same ordering as the FreeRDP callback.

src/protocols/rdp/rdp.c
```diff
@@ -279,51 +386,113 @@ static BOOL rdp_freerdp_authenticate(freerdp* instance, char** username,
...
+
+        /* We should never get here. */
+        default:
+            return TRUE;
+
     }
+
+    /* If one or more params have been required, send them. */
+    if (i > 0)
+        guac_rdp_auth_send_params(client, params, username, password, domain);
```
mike-jumper: This call doesn't match the function's defined name: `guac_rdp_send_auth_params()`

configure.ac
```diff
@@ -1010,6 +1010,22 @@ then
...
+                     [AC_CHECK_MEMBERS([freerdp.Authenticate],,
+                                       [AC_MSG_WARN([
+  ----------------------------------------------------
+   This version of FreeRDP appears to lack support
+   for both Authenticate and AuthenticateEx functions.
+   RDP connections may fail if credentials are not
+   provided with the connection.
+  ----------------------------------------------------])],
+                                       [[#include <freerdp/freerdp.h>]])],
+                      [[#include <freerdp/freerdp.h>]])
```
mike-jumper, line 1026: With `-Werror`, this check will fail upon merely including `freerdp/freerdp.h` for recent FreeRDP 3.x, so this may erroneously report "no".

src/protocols/rdp/rdp.c
```diff
@@ -279,51 +386,113 @@ static BOOL rdp_freerdp_authenticate(freerdp* instance, char** username,
...
+        case AUTH_NLA:
+        case GW_AUTH_HTTP:
+        case GW_AUTH_RDG:
+        case GW_AUTH_RPC:
+            i = guac_rdp_register_required_auth(settings, params);
+            break;
+
+        /* We should never get here. */
+        default:
+            return TRUE;
```
mike-jumper: At least `AUTH_FIDO_PIN` and `AUTH_RDSTLS` were not defined until more recent versions of FreeRDP 3.x, while the `AuthenticateEx` callback has existed prior to that. If an older version of FreeRDP 3.x is used, this may fail to build.

src/protocols/rdp/rdp.c
```diff
@@ -279,51 +386,113 @@ static BOOL rdp_freerdp_authenticate(freerdp* instance, char** username,
...
+
+        /* For all other cases, request anything that isn't configured. */
+        case AUTH_RDSTLS:
+        case AUTH_TLS:
+        case AUTH_RDP:
+        case AUTH_NLA:
+        case GW_AUTH_HTTP:
+        case GW_AUTH_RDG:
+        case GW_AUTH_RPC:
+            i = guac_rdp_register_required_auth(settings, params);
```
mike-jumper: As written, `guac_rdp_register_required_auth()` only deals with the main username/password/domain parameters, not the gateway-specific variations of these. If `GW_AUTH_HTTP`, etc. are actually encountered, the user will be prompted, but the user's responses will populate the wrong settings, the gateway-specific credentials will remain empty, and auth will fail.

configure.ac
```diff
@@ -1010,6 +1010,22 @@ then
...
+if test "x${have_freerdp}" = "xyes"
+then
+    AC_CHECK_MEMBERS([freerdp.AuthenticateEx],,
+                     [AC_CHECK_MEMBERS([freerdp.Authenticate],,
+                                       [AC_MSG_WARN([
+  ----------------------------------------------------
+   This version of FreeRDP appears to lack support
+   for both Authenticate and AuthenticateEx functions.
+   RDP connections may fail if credentials are not
+   provided with the connection.
```
mike-jumper, line 1023: Things will currently fail to build with a warning in the event that neither of these callbacks are available, since the `static` utility functions that they use remain defined but no longer have any callers.

src/protocols/rdp/rdp.c
mike-jumper: We should either log a warning here (so the underlying issue is visible) or remove the `default` (so that the compiler will yell at us for any missing cases).

src/protocols/rdp/rdp.c
```diff
@@ -279,51 +386,113 @@ static BOOL rdp_freerdp_authenticate(freerdp* instance, char** username,
...
+ *
+ * @param domain
+ *     Pointer to a string which will receive the domain associated with the
+ *     user's account.
+ *
+ * @return
+ *     Always TRUE.
+ */
+static BOOL rdp_freerdp_authenticate_ex(freerdp* instance, char** username,
+        char** password, char** domain, rdp_auth_reason reason) {
```
mike-jumper, line 511: Please document the `reason` parameter.

mike-jumper (changes requested):

src/protocols/rdp/rdp.c
necouchman: I've corrected by correcting the function definition so that it matches the callback ordering.

src/protocols/rdp/rdp.c
necouchman: Corrected via rebase.

src/protocols/rdp/rdp.c
necouchman: Removed.

src/protocols/rdp/rdp.c
necouchman: Documented.

src/protocols/rdp/rdp.c
necouchman: @mike-jumper I'm trying to use the `FREERDP_VERSION=3.0.0` in the `Dockerfile` to build against a specific version of FreeRDP, but getting the following error:
```
Cloning into 'FreeRDP'...
Updating files: 100% (2962/2962), done.
Build architecture: X86
Build parallelism: 4 job(s)
Building FreeRDP @ 3.0.0(\.\d+)+ ...
error: pathspec '3.0.0(\.\d+)+' did not match any file(s) known to git
```
Am I doing something wrong?

mike-jumper: `FREERDP_VERSION` expects just the major number (2, 3, etc.) - to completely override with a specific tag or commit, you'll need to use `WITH_FREERDP`.

src/protocols/rdp/rdp.c
necouchman: I've implemented gateway authentication requests.

configure.ac
necouchman: Okay, I've added another `#if` around these functions, too.

src/protocols/rdp/rdp.c
necouchman: Aha, got it.

necouchman: I've implemented version checks to cover this based on the FreeRDP versions in which these options showed up. Couldn't figure out a good way to check `enum` members with Autoconf.

configure.ac
necouchman: Hmm...I tested up to 3.30.0, the latest tag in the FreeRDP code, and it seems to work properly?

```
...
checking for rdpSettings.AllowUnanouncedOrdersFromServer... yes
checking for freerdp.VerifyCertificateEx... yes
checking for freerdp.AuthenticateEx... yes
...
```

mike-jumper: Yeah, same - I think we're safe to assume I'm just wrong here until we encounter a real-world failure to detect, if ever.

src/protocols/rdp/rdp.c
```diff
@@ -279,51 +437,124 @@ static BOOL rdp_freerdp_authenticate(freerdp* instance, char** username,
...
+
+                /* NULL-terminate the array. */
+                params[i] = NULL;
+            }
+
+            break;
+
+        /* For all other cases, request anything that isn't configured. */
+#if (FREERDP_VERSION_MAJOR >= 3 && FREERDP_VERSION_MINOR >= 18)
+	/* RDSTLS support was added in 3.18.0 */
```
mike-jumper: Tab snuck in here.

src/protocols/rdp/rdp.c
```diff
@@ -279,51 +437,124 @@ static BOOL rdp_freerdp_authenticate(freerdp* instance, char** username,
...
+                "\"required\" instruction. No authentication parameters will "
+                "be requested.");
+        return TRUE;
     }
+
+    switch (reason) {
+
+        /* In the case that we just need a PIN, request only the PIN/password. */
+#if (FREERDP_VERSION_MAJOR >= 3 && FREERDP_VERSION_MINOR >= 25)
+	/* FIDO PIN support was added in 3.25.0 */
```
mike-jumper: Sneaky tab here, too.

src/protocols/rdp/rdp.c
```diff
@@ -231,6 +232,164 @@ static BOOL rdp_freerdp_pre_connect(freerdp* instance) {
...
+            i++;
+        }
+    }
+
+    /* If the gateway password is null, request that. */
+    if (settings->gateway_password == NULL) {
+        guac_argv_register(GUAC_RDP_ARGV_GATEWAY_PASSWORD, guac_rdp_argv_callback, NULL, 0);
+        params[i] = GUAC_RDP_ARGV_GATEWAY_PASSWORD;
+        i++;
+    }
```
mike-jumper, line 404: The `argv` callback will also need to be updated for these parameters to be handled when received.

src/protocols/rdp/rdp.c
```diff
@@ -279,51 +437,124 @@ static BOOL rdp_freerdp_authenticate(freerdp* instance, char** username,
...
+        guac_client_log(client, GUAC_LOG_WARNING, "Client does not support the "
+                "\"required\" instruction. No authentication parameters will "
+                "be requested.");
+        return TRUE;
     }
+
+    switch (reason) {
+
+        /* In the case that we just need a PIN, request only the PIN/password. */
+#if (FREERDP_VERSION_MAJOR >= 3 && FREERDP_VERSION_MINOR >= 25)
```
mike-jumper: This won't be quite right for FreeRDP 4.x (it will match >= 3.25 for 3.x, >= 4.25 for 4.x, etc.).

I think we can check enum constants using `AC_CHECK_DECL`.

src/protocols/rdp/rdp.c
```diff
@@ -231,6 +232,164 @@ static BOOL rdp_freerdp_pre_connect(freerdp* instance) {
...
+ *
+ * @param password
+ *     A pointer to the memory location where the RDP client currently
+ *     has stored the password.
+ *
+ * @param domain
+ *     A pointer to the memory location where the RDP client currently
+ *     has stored the domain.
+ */
+static void guac_rdp_send_auth_params(guac_client* client, const char** params,
```
mike-jumper, line 264: Since this never handles the gateway parameters, a gateway reason in the authenticate callback won't result in a gateway prompt.

src/protocols/rdp/rdp.c
```diff
@@ -279,51 +437,124 @@ static BOOL rdp_freerdp_authenticate(freerdp* instance, char** username,
...
+        case AUTH_RDP:
+        case AUTH_NLA:
+            i = guac_rdp_register_required_auth(settings, params);
+            break;
+
+        case GW_AUTH_HTTP:
+        case GW_AUTH_RDG:
+        case GW_AUTH_RPC:
+            i = guac_rdp_register_required_gateway_auth(settings, params);
+            break;
```
mike-jumper, line 568: Is there any potential that this callback will be invoked multiple times?

If not, then we're fine. If it _can_ be called multiple times, then these parameters will end up being registered repeatedly until `GUAC_ARGV_MAX_REGISTERED` is reached.

mike-jumper (changes requested):

src/protocols/rdp/rdp.c
necouchman: I think there are a couple of situations where it would be called multiple times:
* If Gateway authentication is requested, it could be call for both Gateway and then Server authentication. That would be twice.
* If authentication fails, the connection would retry, but I don't think this necessarily qualifies as being called multiple times, since the guacd connection would close and then restart when authentication fails.

Is there some way that multiple calls should be dealt with?

src/protocols/rdp/rdp.c
necouchman: Oops. Removed.

src/protocols/rdp/rdp.c
necouchman: Removed.

src/protocols/rdp/rdp.c
necouchman: Updated.

src/protocols/rdp/rdp.c
necouchman: Okay, I've adjusted this to be similar to the checks in `gdi.c` that break out the major versions.

src/protocols/rdp/rdp.c
necouchman: Okay, I've updated that function to handle the gateway parameters, as well.

src/protocols/rdp/rdp.c
mike-jumper: No worries, this sounds fine as-is.

src/protocols/rdp/rdp.c
```diff
@@ -231,6 +232,180 @@ static BOOL rdp_freerdp_pre_connect(freerdp* instance) {
...
+
+        if (strcmp(params[i], GUAC_RDP_ARGV_PASSWORD) == 0) {
+            guac_mem_free(*password);
+            *password = guac_strdup(settings->password);
+        }
+
+        if (strcmp(params[i], GUAC_RDP_ARGV_GATEWAY_PASSWORD) == 0) {
+            guac_mem_free(*password);
+            *password = guac_strdup(settings->gateway_password);
+        }
```
mike-jumper: These should be `else if` to avoid scanning the `params[i]` string repeatedly.

src/protocols/rdp/rdp.c
```diff
@@ -279,51 +453,126 @@ static BOOL rdp_freerdp_authenticate(freerdp* instance, char** username,
...
+                params[i] = NULL;
+            }
+
+            break;
+
+    /* For all other cases, request anything that isn't configured. */
+#if (FREERDP_VERSION_MAJOR >= 4) || \
+    (FREERDP_VERSION_MAJOR == 3 && FREERDP_VERSION_MINOR >= 18)
+    /* RDSTLS support was added in 3.18.0 */
+        case AUTH_RDSTLS:
```
mike-jumper: Beware that several of the comments here have become unindented by one level.

mike-jumper (changes requested):
Everything else: LGTM!

src/protocols/rdp/rdp.c
necouchman: Fixed via rebase.

src/protocols/rdp/rdp.c
necouchman: Oops...fixed via rebase.

mike-jumper (approved):

PR #671, opened 2026-05-07
GUACAMOLE-2268: Gate VNC UTF-8 (Extended Clipboard) support behind config check.

bbennett-ks:
SendClientCutTextUTF8 and rfbClient.GotXCutTextUTF8 are only available in libvncserver/libvncclient 0.9.15+.

This adds configure checks for the UTF-8 (Extended Clipboard) APIs and guards the affected call sites with #ifdefs. When unavailable, the VNC client falls back to the classic SendClientCutText path.

### Testing
```
libvncserver version 0.9.15:

autoreconf -I
./configure
…
checking for rfbClient.requestedResize... yes
checking for rfbClient.GotXCutTextUTF8... yes
…

libvncserver version 0.9.15:

autoreconf -I
./configure
…
checking for rfbClient.requestedResize... yes
checking for rfbClient.GotXCutTextUTF8... no
…
```

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -721,6 +721,30 @@ then
 fi
 
 #
+# libVNCServer support for the Extended Clipboard pseudo-encoding, which
+# allows clipboard text to be exchanged as UTF-8 via SendClientCutTextUTF8()
+# and the GotXCutTextUTF8 callback. Both were added in libvncserver 0.9.15.
+# If support is missing, Guacamole falls back to the classic VNC clipboard
+# with iconv-based encoding conversion.
+#
+
+if test "x${have_libvncserver}" = "xyes"
+then
+
+    have_vnc_extended_clipboard=yes
+    AC_CHECK_MEMBERS([rfbClient.GotXCutTextUTF8],
+                     [], [have_vnc_extended_clipboard=no],
+                     [[#include <rfb/rfbclient.h>]])
+
+    if test "x${have_vnc_extended_clipboard}" = "xyes"
+    then
+        AC_DEFINE([LIBVNC_CLIENT_HAS_EXTENDED_CLIPBOARD],,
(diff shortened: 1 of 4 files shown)


configure.ac
```diff
@@ -720,6 +720,30 @@ then
...
+                     [], [have_vnc_extended_clipboard=no],
+                     [[#include <rfb/rfbclient.h>]])
+
+    if test "x${have_vnc_extended_clipboard}" = "xyes"
+    then
+        AC_DEFINE([LIBVNC_CLIENT_HAS_EXTENDED_CLIPBOARD],,
+                  [Whether rfbClient supports the Extended Clipboard pseudo-encoding (UTF-8).])
+    fi
+
+fi
```
necouchman, line 745: Just FYI - this is fine, and I will merge as-is, but this can be simpler - the `AC_CHECK_MEMBERS` macro will automatically define `HAVE_RFBCLIENT_GOTXCUTTEXTUTF8` for you. For example, we use this with RDP support and the VerifyCertificateEx: https://github.com/apache/guacamole-server/blob/c802cdce5627b4e0df77987878824cb167a1f9ba/configure.ac#L985-L991

Obviously the defines end up a little less nicely-named than what you have, but the `configure.ac` ends up a little simpler.

necouchman (approved):

configure.ac
bbennett-ks: Thanks for the tip! I modeled the fix based on support for `libVNCserver support for the requestedResize`. I'll keep this in mind for next my.

FYI, I implemented similar support for `libVNCServer support FinishedFrameBufferUpdate`. Once I began to test it, I realized the the base `libVNCServer` in GitHub supported `FinishedFrameBufferUpdate` so config support was not necessary.

PR #673, opened 2026-05-14
GUACAMOLE-2123: Fix remote app render issues. Use per-window guac_dis…

jjheinon:
Fixing remote app render issues with this patch, using per-window guac_display_layer for remote apps. Also introducing a fix for stale desktop background on RDPGFX.

Using per-window guac_display_layer for remote apps and also prevent dirty-region propagation to desktop background. Added support for UpdateSurfaceArea (FreeRDP 2.x) and UpdateWindowFromSurface (FreeRDP 3.x)

When Guacamole's RDP plugin is configured with both RemoteApp and the RDPGFX pipeline enabled, remote application windows fail to render.
With GFX active, FreeRDP does not update RAIL window contents via BeginPaint/Endpaint, but via RDPGFX surfaces using windowId and callbacks.
For FreeRDP 3.x: UpdateWindowFromSurface is called per window-mapped surface with its accumulated invalid region.
For FreeRDP 2.x: UpdateSurfaceArea is called per surface with a rect list.

This patch has a dedicated guac_display_layer for each remote app and derives visibility via showState. Surface content painted with RDPGFX callbacks and scaled to correct size.
Layer is kept hidden until both window position is known and at least one surface paint request has arrived, avoiding flashes and race conditions.

Tested remote apps with FreeRDP 3.26.0 with the latest staging/1.6.1 branch.

### DIFF
--- a/src/protocols/rdp/channels/rail.c
+++ b/src/protocols/rdp/channels/rail.c
@@ -19,10 +19,21 @@
 
 #include "config.h"
 
+#include "color.h"
 #include "channels/rail.h"
+#include "common/list.h"
 #include "plugins/channels.h"
 #include "rdp.h"
 #include "settings.h"
+
+#ifdef HAVE_RDPGFX_SURFACE_AREA_UPDATE
+#include <freerdp/codec/color.h>
+#include <freerdp/codec/region.h>
+#include <freerdp/gdi/gdi.h>
+#include <freerdp/gdi/gfx.h>
+#include <guacamole/display.h>
+#include <guacamole/rect.h>
+#endif
 
 #include <freerdp/client/rail.h>
 #include <freerdp/event.h>
@@ -31,10 +42,12 @@
 #include <freerdp/window.h>
 #include <guacamole/client.h>
 #include <guacamole/mem.h>
+#include <winpr/string.h>
 #include <winpr/wtypes.h>
 #include <winpr/wtsapi.h>
 
 #include <stddef.h>
+#include <stdint.h>
 #include <string.h>
 
 /**
@@ -66,24 +79,34 @@ static UINT guac_rdp_rail_complete_handshake(RailClientContext* rail) {
     guac_client* client = (guac_client*) rail->custom;
     guac_rdp_client* rdp_client = (guac_rdp_client*) client->data;
 
+    /* Do not advertise TS_RAIL_CLIENTSTATUS_ALLOWLOCALMOVESIZE. Guacamole
+     * does not manage native local windows and does not send ClientWindowMove
+     * when a local move/resize completes. Advertising that capability can
+     * cause the server to enter a move/resize path that Guacamole cannot
+     * finish, leaving subsequent input ignored after dragging a RemoteApp
+     * window. */
     RAIL_CLIENT_STATUS_ORDER client_status = {
-        .flags =
-                TS_RAIL_CLIENTSTATUS_ALLOWLOCALMOVESIZE
-              | TS_RAIL_CLIENTSTATUS_APPBAR_REMOTING_SUPPORTED
+        .flags = TS_RAIL_CLIENTSTATUS_APPBAR_REMOTING_SUPPORTED
     };
 
     /* Send client status */
-    guac_client_log(client, GUAC_LOG_TRACE, "Sending RAIL client status.");
+    guac_client_log(client, GUAC_LOG_TRACE,
+            "Sending RAIL client status: flags=0x%08x.",
+            (unsigned int) client_status.flags);
     pthread_mutex_lock(&(rdp_client->message_lock));
     status = rail->ClientInformation(rail, &client_status);
     pthread_mutex_unlock(&(rdp_client->message_lock));
 
-    if (status != CHANNEL_RC_OK)
+    if (status != CHANNEL_RC_OK) {
+        guac_client_log(client, GUAC_LOG_DEBUG,
+                "RAIL client status send failed: status=%u.",
+                (unsigned int) status);
         return status;
+    }
 
     RAIL_SYSPARAM_ORDER sysparam = {
 
-        .dragFullWindows = FALSE,
+        .dragFullWindows = rdp_client->settings->full_window_drag_enabled,
 
         .highContrast = {
             .flags =
@@ -115,6 +138,9 @@ static UINT guac_rdp_rail_complete_handshake(RailClientContext* rail) {
             | SPI_MASK_SET_KEYBOARD_CUES
             | SPI_MASK_SET_KEYBOARD_PREF
             | SPI_MASK_SET_MOUSE_BUTTON_SWAP
+#ifdef SPI_MASK_SET_DRAG_FULL_WINDOWS
+            | SPI_MASK_SET_DRAG_FULL_WINDOWS
+#endif
             | SPI_MASK_SET_WORK_AREA
 
     };
@@ -136,10 +162,23 @@ static UINT guac_rdp_rail_complete_handshake(RailClientContext* rail) {
     };
 
     /* Execute desired RemoteApp command */
-    guac_client_log(client, GUAC_LOG_TRACE, "Executing remote application.");
+    guac_client_log(client, GUAC_LOG_TRACE,
+            "Executing remote application: program=\"%s\", args=\"%s\", "
+            "working-dir=\"%s\", flags=0x%04x.",
+            exec.RemoteApplicationProgram != NULL
+                ? exec.RemoteApplicationProgram : "",
+            exec.RemoteApplicationArguments != NULL
+                ? exec.RemoteApplicationArguments : "",
+            exec.RemoteApplicationWorkingDir != NULL
+                ? exec.RemoteApplicationWorkingDir : "",
+            (unsigned int) exec.flags);
     pthread_mutex_lock(&(rdp_client->message_lock));
     status = rail->ClientExecute(rail, &exec);
     pthread_mutex_unlock(&(rdp_client->message_lock));
+
+    guac_client_log(client, GUAC_LOG_TRACE,
+            "RAIL client execute send completed: status=%u.",
+            (unsigned int) status);
 
     return status;
 
@@ -149,7 +188,7 @@ static UINT guac_rdp_rail_complete_handshake(RailClientContext* rail) {
  * A callback function that is invoked when the RDP server sends the result
  * of the Remote App (RAIL) execution command back to the client, so that the
  * client can handle any required actions associated with the result.
- * 
+ *
  * @param context
  *     A pointer to the RAIL data structure associated with the current
  *     RDP connection.
@@ -166,6 +205,11 @@ static UINT guac_rdp_rail_execute_result(RailClientContext* context,
         const RAIL_EXEC_RESULT_ORDER* execResult) {
 
     guac_client* client = (guac_client*) context->custom;
+
+    if (execResult->execResult == RAIL_EXEC_E_NOT_IN_ALLOWLIST) {
+        guac_client_log(client, GUAC_LOG_WARNING, "RAIL command reported "
+                "application not in allowlist.");
+    }
 
     if (execResult->execResult != RAIL_EXEC_S_OK) {
         guac_client_log(client, GUAC_LOG_DEBUG, "Failed to execute RAIL command on server: %d", execResult->execResult);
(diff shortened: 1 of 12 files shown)


stephzero1:
great work

src/protocols/rdp/channels/rail.c
```diff
@@ -37,6 +47,185 @@
...
+     * The next RAIL window in the list.
+     */
+    guac_rdp_rail_window* next;
+
+};
+
+/**
+ * Returns whether RAIL window surfaces should be rendered through Guacamole
+ * layers for the given RDP client.
+ */
```
necouchman, line 491: Please complete the function documentation - it should have `@param` tags for all parameters and a `@return` tag for the return value.

src/protocols/rdp/channels/rail.c
```diff
@@ -37,6 +47,185 @@
...
+static int guac_rdp_rail_uses_gfx(guac_rdp_client* rdp_client) {
+    return rdp_client->settings->remote_app != NULL
+        && rdp_client->settings->enable_gfx;
+}
+
+/**
+ * Returns the RAIL window having the given ID, or NULL if no such window is
+ * being tracked. The rail_windows_lock must be held.
+ */
```
necouchman, line 582: As above, please complete the function documentation.

src/protocols/rdp/channels/rail.c
```diff
@@ -37,6 +47,185 @@
...
+    }
+
+    return NULL;
+
+}
+
+/**
+ * Returns the RAIL window having the given ID, creating it if necessary.
+ * The rail_windows_lock must be held.
+ */
```
necouchman, line 1119: Please complete the function documentation (see above).

src/protocols/rdp/channels/rail.c
```diff
@@ -37,6 +47,185 @@
...
+    rdp_client->rail_windows = rail_window;
+
+    return rail_window;
+
+}
+
+/**
+ * Updates the visibility of the Guacamole layer backing the given RAIL window.
+ * The rail_windows_lock must be held.
+ */
```
necouchman, line 1228: Please complete the function documentation (see above).

src/protocols/rdp/channels/rail.c
```diff
@@ -37,6 +47,185 @@
...
+            && rail_window->has_surface
+            && rail_window->has_rail_order) ? 0xFF : 0;
+    guac_display_layer_set_opacity(rail_window->layer, opacity);
+
+}
+
+/**
+ * Removes the given RAIL window from the list and frees its resources. The
+ * rail_windows_lock must be held.
+ */
```
necouchman, line 1458: Please complete the function documentation (see above).

src/protocols/rdp/channels/rail.c
```diff
@@ -279,6 +507,258 @@ static BOOL guac_rdp_rail_window_update(rdpContext* context,
 
 }
 
+/**
+ * Handler for the RAIL WindowDelete order. Removes any tracked RAIL window
+ * with the given ID.
+ */
```
necouchman, line 1827: Please complete function documentation (see above).

src/protocols/rdp/channels/rail.c
```diff
@@ -279,6 +507,258 @@ static BOOL guac_rdp_rail_window_update(rdpContext* context,
...
+ * Copies updated regions from the given RDPGFX surface to the Guacamole layer
+ * associated with that surface's RAIL window. The rail_windows_lock must NOT
+ * be held on entry; this function takes it internally.
+ *
+ * The invalidRegion of the surface is left untouched. FreeRDP's GDI pipeline
+ * owns that region and may consume or clear it independently; touching it
+ * from these callbacks led to misplaced updates in early versions of this
+ * code (menus appearing in the wrong location, etc.) because regions were
+ * being double-consumed.
+ */
```
necouchman, line 2120: Please complete function documentation (see above).

src/protocols/rdp/channels/rail.c
```diff
@@ -279,6 +507,258 @@ static BOOL guac_rdp_rail_window_update(rdpContext* context,
...
+     * creation, and only diverge when an explicit MapSurfaceToScaledWindow
+     * order rescales the surface. We honor any divergence but optimize for
+     * the common 1:1 case with a plain copy. */
+    int source_width = surface->mappedWidth;
+    int source_height = surface->mappedHeight;
+    int target_width = surface->outputTargetWidth;
+    int target_height = surface->outputTargetHeight;
+
+    if (target_width <= 0) target_width = source_width;
+    if (target_height <= 0) target_height = source_height;
```
necouchman: Please avoid these single-line `if` statements, in keeping with the established style in the code.

src/protocols/rdp/channels/rail.h
```diff
@@ -60,5 +67,35 @@
...
 
+#ifdef HAVE_RDPGFX_SURFACE_AREA_UPDATE
+/**
+ * Updates the Guacamole display layer for a RAIL window from modified areas of
+ * the given RDPGFX surface. Non-RAIL (windowId == 0) surfaces are ignored and
+ * left for FreeRDP's default GDI pipeline.
+ *
+ * Used only as a fallback when UpdateWindowFromSurface is unavailable in the
+ * installed FreeRDP. The invalidRegion of the surface is NOT cleared.
+ */
```
necouchman, line 178: As in the `rail.c` code, functions here need to be fully-documented, including `@param` tags for each parameter and `@return` tags documenting the return value.

src/protocols/rdp/channels/rail.h
```diff
@@ -60,5 +67,35 @@
...
+UINT guac_rdp_rail_update_surface_area(RdpgfxClientContext* context,
+        UINT16 surface_id, UINT32 rect_count, const RECTANGLE_16* rects);
+
+#ifdef HAVE_RDPGFX_WINDOW_SURFACE_UPDATE
+/**
+ * Updates the Guacamole display layer for a RAIL window from the given
+ * window-mapped RDPGFX surface. FreeRDP's GDI pipeline calls this once per
+ * surface per frame, with the surface's accumulated invalidRegion, and
+ * subsequently clears that region itself.
+ */
```
necouchman, line 198: Please document properly.

src/protocols/rdp/channels/rail.h
```diff
@@ -60,5 +67,35 @@
...
+UINT guac_rdp_rail_update_window_from_surface(RdpgfxClientContext* context,
+        gdiGfxSurface* surface);
+#endif
+#endif
+
+/**
+ * Frees all RAIL window state tracked for RemoteApp rendering. Safe to call
+ * even when no RAIL windows are tracked.
+ */
```
necouchman, line 210: Please document properly.

src/protocols/rdp/gdi.c
```diff
@@ -128,14 +128,24 @@ BOOL guac_rdp_gdi_end_paint(rdpContext* context) {
...
-    guac_rect_constrain(&dst_rect, &current_context->bounds);
-    guac_rect_extend(&current_context->dirty, &dst_rect);
-
-    rdp_client->gdi_modified = 1;
+    /*
+     *For RemoteApp connections the default layer is held
+     * permanently black (filled once in guac_rdp_handle_connection).
+     * Suppress dirty-region propagation so FreeRDP's ongoing painting of
+     * the server desktop into gdi->primary_buffer is never sent to the client.
+     */
```
necouchman: The comment, here, talks about what happens when RemoteApp is enabled, but it's above a section of code that deals with situations where RemoteApp is *not* in use. The language of this comment should be clarified such that it indicates what the code it is documenting _is_ actually doing.

necouchman (changes requested):
Thanks @jjheinon, this is great - I had started to work on this some time back but hit a couple of walls that I couldn't get past, and it looks like you've figured out the issues I couldn't!

The code looks pretty good to me, though I need to take time to more thoroughly review it - most of the things that need to be changed so far deal with style, particularly the comment blocks of the functions you've implemented.

src/protocols/rdp/channels/rail.c
```diff
@@ -37,6 +47,219 @@
...
+        return NULL;
+    }
+
+    rail_window->window_id = window_id;
+    rail_window->visible = 1;
+
+    /* Keep the layer hidden until we have both a RAIL order (so we know where
+     * to put it) and at least one surface paint (so we have something to
+     * show). */
+    guac_display_layer_set_opacity(rail_window->layer, 0);
```
necouchman: Why is the `visible` property being set to `1` if the layer is being hidden? Shouldn't these two items roughly match each other?

src/protocols/rdp/channels/rail.c
```diff
@@ -37,6 +47,219 @@
...
+    /**
+     * The current dimensions of the Guacamole layer.
+     */
+    int layer_width;
+    int layer_height;
+
+    /**
+     * The next RAIL window in the list.
+     */
+    guac_rdp_rail_window* next;
```
necouchman: It looks like this and several of the functions you implement are implementing a linked list. There's already a generic linked-list implementation provided within the guacd code:

https://github.com/apache/guacamole-server/blob/main/src/common/common/list.h

It might be worth at least considering re-using that rather than re-implementing all the linked list functions separately, here.

src/protocols/rdp/channels/rail.c
```diff
@@ -37,6 +47,219 @@
...
+ * @param rdp_client
+ *     The RDP client whose settings should be checked.
+ *
+ * @return
+ *     Non-zero if RAIL window surfaces should be rendered through Guacamole layers.
+ */
+static int guac_rdp_rail_uses_gfx(guac_rdp_client* rdp_client) {
+    return rdp_client->settings->remote_app != NULL
+        && rdp_client->settings->enable_gfx;
+}
```
necouchman: Does rendering RAIL windows through window surfaces really require that the graphics pipeline be enabled? Are there situations where GFX may be disabled where per-window surfaces and updates would still work, and might even be preferred?

src/protocols/rdp/channels/rail.c
```diff
@@ -37,6 +47,219 @@
...
+/**
+ * Returns the RAIL window having the given ID, or NULL if no such window is
+ * being tracked. The rail_windows_lock must be held.
+ *
+ * @param rdp_client
+ *     The RDP client instance whose tracked RAIL windows should be searched.
+ *
+ * @param window_id
+ *     The server-side ID of the RAIL window to retrieve.
+ */
```
necouchman, line 582: This comment block still needs a `@return` block.

src/protocols/rdp/channels/rail.c
```diff
@@ -37,6 +47,219 @@
...
+    }
+
+    rail_window->layer = guac_display_alloc_layer(rdp_client->display, 0);
+    if (rail_window->layer == NULL) {
+        guac_client_log(client, GUAC_LOG_WARNING, "RAIL window 0x%08x could "
+                "not be tracked. Window contents may not render.",
+                (unsigned int) window_id);
+        guac_mem_free(rail_window);
+        return NULL;
+    }
```
necouchman: It may be worth adding a `DEBUG`-level log in here to indicate more precisely what the failure is, as these existing `WARNING` messages are identical and won't help an admin track down the actual reason for the failure (memory allocation vs. display layer allocation).

src/protocols/rdp/channels/rail.c
```diff
@@ -279,6 +541,322 @@ static BOOL guac_rdp_rail_window_update(rdpContext* context,
...
+*
+* @param rect_count
+*     The number of rectangles within the rects array.
+*
+* @param rects
+*     The updated rectangular regions of the surface.
+*
+* @return
+*     CHANNEL_RC_OK if the surface was painted successfully, or an error code.
+*/
```
necouchman: Since this is defined in `rail.h`, there's no need to document it, here.

src/protocols/rdp/channels/rail.c
```diff
@@ -279,6 +541,322 @@ static BOOL guac_rdp_rail_window_update(rdpContext* context,
...
+*
+* @param context
+*     The RDPGFX client context associated with the surface update.
+*
+* @param surface
+*     The RDPGFX surface containing the updated window contents.
+*
+* @return
+*     CHANNEL_RC_OK if the surface was painted successfully, or an error code.
+*/
```
necouchman: As this function is defined in `rail.h`, the documentation can be solely there and need not be duplicated, here.

src/protocols/rdp/channels/rail.c
```diff
@@ -279,6 +541,322 @@ static BOOL guac_rdp_rail_window_update(rdpContext* context,
...
+}
+#endif
+#endif
+
+/**
+* Frees all tracked RAIL windows associated with the given RDP client.
+*
+* @param rdp_client
+*     The RDP client whose tracked RAIL windows should be freed.
+*/
```
necouchman: This function also can be documented in `rail.h` and not, here.

src/protocols/rdp/channels/rail.c
```diff
@@ -259,6 +482,45 @@ static BOOL guac_rdp_rail_window_update(rdpContext* context,
 
     UINT32 fieldFlags = orderInfo->fieldFlags;
 
+    /* Track window location and visibility for RDPGFX RemoteApp rendering. */
+    if (guac_rdp_rail_uses_gfx(rdp_client)) {
```
necouchman, line 1655: Is GFX rendering really required in order to track RAIL windows and use separate layers for updates? It seems like this is something we could implement across the board for RAIL windows, using GFX when it is supported, but falling back to non-GFX methods?

necouchman (changes requested):
@jjheinon:
A few more changes and a couple of questions. Also, one of your commits lacks the required `GUACAMOLE-2123:` tag.

Overall looking pretty good and just about ready for merge.

src/protocols/rdp/channels/rail.c
jjheinon: visible -property is a server side flag, telling the window should be visible.
We just hide it on the client side by setting layer opacity to 0, until both window coordinates and first surface data have both arrived.
If we set visibility to 0, the window never appears.

src/protocols/rdp/channels/rail.c
jjheinon: Don't know if that's feasible without major changes, as GFX rendering tracks window ownership and gives us the surface data later via windowId, but non-GFX rendering renders directly to primary desktop buffer, without ownership information which app owns the pixels. It renders only rectangles without available metadata. Figuring out to which layer those pixels should go would most likely be problematic.

necouchman: Okay, that's fair - definitely not worth holding up getting this part of it merged so that this fix can go out with 1.6.1. Maybe I'll circle back to that another time...

src/protocols/rdp/rdp.c
```diff
@@ -536,6 +536,54 @@ static int guac_rdp_handle_connection(guac_client* client) {
...
+     * buffer never reaches the client.
+     *
+     * Without this, GFX-enabled RemoteApp sessions show the server's
+     * desktop behind the application
+     * windows because gdi->primary_buffer shares memory with the default
+     * layer and FreeRDP paints all windowId==0 surfaces into it.
+     */
+    if (rdp_client->settings->remote_app != NULL) {
+
+        guac_display_layer_raw_context* bg_ctx =
```
jjheinon: Looks like this is not quite working, we set the background black too early. Desktop background (or in some cases, welcome screen) is still visible. Fixing.

jjheinon: This code was correct, the problem was elsewhere.
Basically, FreeRDPs GDI buffer points to Guacamole's default layer and kept updating it, so pixels leaked to RAIL app background.
Remote app rendering path now separates default layer from gdi->primary_buffer, so the background remains black.

src/protocols/rdp/gdi.c
```diff
@@ -169,27 +202,46 @@ BOOL guac_rdp_gdi_desktop_resize(rdpContext* context) {
...
+        /* Full desktop render path: 
+         * Update our reference to the GDI buffer, as well as any structural
+         * details, which may now all be different */
+        guac_rdp_gdi_use_primary_buffer(current_context, gdi);
+
+        /* Resize layer to match new display dimensions and underlying buffer */
+        guac_display_layer_resize(default_layer, gdi->width, gdi->height);
+        guac_display_layer_close_raw(default_layer, current_context);
+    }
+    else {
```
jjheinon, line 246: Part of the background color bugfix for resize.
Remote app rendering path keeps layers detached and doesn't call guac_rdp_gdi_use_primary_buffer

src/protocols/rdp/gdi.c
```diff
@@ -169,27 +202,46 @@ BOOL guac_rdp_gdi_desktop_resize(rdpContext* context) {
...
-    /* Update our reference to the GDI buffer, as well as any structural
-     * details, which may now all be different */
-    current_context->buffer = gdi->primary_buffer;
-    current_context->stride = gdi->stride;
-    guac_rect_init(&current_context->bounds, 0, 0, gdi->width, gdi->height);
+    if (rdp_client->settings->remote_app == NULL) {
+        /* Full desktop render path: 
+         * Update our reference to the GDI buffer, as well as any structural
+         * details, which may now all be different */
+        guac_rdp_gdi_use_primary_buffer(current_context, gdi);
```
jjheinon, line 240: Moved the code to the function guac_rdp_gdi_use_primary_buffer

src/protocols/rdp/gdi.c
```diff
@@ -35,6 +35,25 @@
...
+ * In RemoteApp sessions, this mapping would expose FreeRDP's hidden desktop
+ * buffer behind application windows, so this should only be used in full desktop mode.
+ *
+ * @param current_context
+ *     The raw context that should reference the primary GDI buffer.
+ *
+ * @param gdi
+ *     The FreeRDP GDI state whose primary buffer should be referenced.
+ */
+static void guac_rdp_gdi_use_primary_buffer(
```
jjheinon, line 51: Made this a function, as the code is called from multiple places.

src/protocols/rdp/channels/rail.c
```diff
@@ -37,6 +48,235 @@
...
+
+/**
+ * Frees the resources associated with the given RAIL window.
+ *
+ * @param data
+ *     The RAIL window to free.
+ */
+static void guac_rdp_rail_free_window_data(void* data) {
+
+    guac_rdp_rail_window* rail_window = data;
```
necouchman, line 1461: This may work as-is, but shouldn't this be explicitly type-cast?
```
guac_rdp_rail_window* rail_window = (guac_rdp_rail_window*) data;

**Review by necouchman — CHANGES_REQUESTED** [review]
Looking pretty good to me. Just one other change for an explicit type-cast at the moment.

File: src/protocols/rdp/channels/rail.c [hunk 27]
```diff
@@ -126,6 +126,77 @@ static int guac_rdp_rail_uses_gfx(guac_rdp_client* rdp_client) {
...
+ * above Guacamole's cursor layer.
+ *
+ * @param client
+ *     The Guacamole client associated with the RDP session.
+ *
+ * @return
+ *     The RemoteApp window container layer, or NULL if it could not be
+ *     allocated.
+ */
+static guac_display_layer* guac_rdp_rail_get_window_layer(
```
jjheinon, line 539: Remote App window ordering was broken and fixing that causes mouse layer to be rendered under the app windows. Attempt to fix that with negative zIndexes was not allowed/wrong approach.
So instead, we render all rail windows in the same container (in correct order), that stays under the cursor layer.

src/protocols/rdp/channels/rail.c
```diff
@@ -306,10 +403,14 @@ static UINT guac_rdp_rail_complete_handshake(RailClientContext* rail) {
...
     guac_rdp_client* rdp_client = (guac_rdp_client*) client->data;
 
+    /* Do not advertise TS_RAIL_CLIENTSTATUS_ALLOWLOCALMOVESIZE. Guacamole
+     * does not manage native local windows and does not send ClientWindowMove
+     * when a local move/resize completes. Advertising that capability can
+     * cause the server to enter a move/resize path that Guacamole cannot
+     * finish, leaving subsequent input ignored after dragging a RemoteApp
+     * window. */
     RAIL_CLIENT_STATUS_ORDER client_status = {
-        .flags =
```
jjheinon, line 68: Cannot use TS_RAIL_CLIENTSTATUS_ALLOWLOCALMOVESIZE, or we might lose mouse input. We obey positioning info from server instead, which fixes the issue.

src/protocols/rdp/channels/rail.c
```diff
@@ -416,6 +517,30 @@ static UINT guac_rdp_rail_execute_result(RailClientContext* context,
...
+ *     The RailClientContext structure used by FreeRDP to handle the RAIL
+ *     channel for the current RDP session.
+ *
+ * @param localMoveSize
+ *     The local move/size request sent by the server.
+ *
+ * @return
+ *     CHANNEL_RC_OK (zero) as the order has been ignored successfully.
+ */
+static UINT guac_rdp_rail_local_move_size(RailClientContext* rail,
```
jjheinon, line 1547: Just in case server sends the request, we answer with CHANNEL_RC_OK.

src/protocols/rdp/channels/rail.c
```diff
@@ -594,10 +718,107 @@ static BOOL guac_rdp_rail_window_delete(rdpContext* context,
...
+ * @param orderInfo
+ *     A pointer to the struct containing desktop order flags.
+ *
+ * @param monitoredDesktop
+ *     A pointer to the monitored desktop state sent by the server.
+ *
+ * @return
+ *     TRUE if the desktop order was handled successfully, otherwise FALSE.
+ */
+static BOOL guac_rdp_rail_monitored_desktop(rdpContext* context,
```
jjheinon, line 1882: The window ordering was broken earlier, now we obey server sent z-order for the windows. We also set active window according to server orders. Server z-ordering is the source of truth (if available).

src/protocols/rdp/channels/rail.c
```diff
@@ -683,6 +914,10 @@ static UINT guac_rdp_rail_paint_surface(guac_client* client,
     guac_rect source_bounds;
     guac_rect_init(&source_bounds, 0, 0, source_width, source_height);
 
+    /* RDPGFX window surfaces are copied as opaque layer content. We don't
+     * want holes on our RemoteApp windows. */
+    UINT32 dst_format = guac_rdp_get_native_pixel_format(FALSE);
```
jjheinon: GFX windows might have alpha channel, we don't want that when copying the layers to Guacamole side. (managed to get holes in paint app, where there should have been black instead). This fixes the issue.

src/protocols/rdp/channels/rail.c
jjheinon: No, void* implicitly converts to any pointer type. That would be required in C++ though, but not in C.

src/protocols/rdp/channels/rail.c
```diff
@@ -37,6 +48,332 @@
...
+ * positioned by the RAIL window-order stream and painted from the corresponding
+ * RDPGFX surface via either:
+ *
+ *   - UpdateSurfaceArea (fallback for older FreeRDP without the window
+ *     callback; called with rect lists for any surface update).
+ *
+ *   - UpdateWindowFromSurface, when available.
+ */
+
+typedef struct guac_rdp_rail_window {
```
necouchman, line 299: This `guac_rdp_rail_window` struct needs a comment describing it, it's purpose, etc.

necouchman (changes requested):
One more change, and I think it'll be ready to merge.

src/protocols/rdp/channels/rail.c
jjheinon: Added comment.

src/protocols/rdp/channels/rail.c
```diff
@@ -37,6 +48,376 @@
...
+static void guac_rdp_rail_update_window_opacity(
+        guac_rdp_rail_window* rail_window) {
+
+    int opacity = (rail_window->visible
+            && rail_window->has_surface
+            && rail_window->has_rail_order) ? 0xFF : 0;
+    guac_display_layer_set_opacity(rail_window->layer, opacity);
+}
+
+static void guac_rdp_rail_paint_background(guac_rdp_client* rdp_client) {
```
jjheinon, line 1412: Moved the blackground painting here, as earlier we rendered the black background for RAIL apps too soon before the login was done. This caused screen to remain black if login failed.

src/protocols/rdp/channels/rail.c
```diff
@@ -166,6 +551,15 @@ static UINT guac_rdp_rail_execute_result(RailClientContext* context,
         const RAIL_EXEC_RESULT_ORDER* execResult) {
 
     guac_client* client = (guac_client*) context->custom;
+    guac_rdp_client* rdp_client = (guac_rdp_client*) client->data;
+
+    if (execResult->execResult == RAIL_EXEC_E_NOT_IN_ALLOWLIST
```
jjheinon: If server tells us RAIL_EXEC_E_NOT_IN_ALLOWLIST after rendering is already working and windows being drawn, just ignore it.

src/protocols/rdp/channels/rdpgfx.c
```diff
@@ -58,16 +61,48 @@ static void guac_rdp_rdpgfx_channel_connected(rdpContext* context,
...
+ *
+ *   If only UpdateSurfaceArea is available, use the _ex initializer and
+ *   route RemoteApp window updates through our custom surface-area handler.
+ *
+ *   On older FreeRDP versions, fall back to the default pipeline. In that
+ *   case, RemoteApp with GFX will not render correctly and GFX must be
+ *   disabled.
+ */
+     
+#if defined(HAVE_RDPGFX_SURFACE_AREA_UPDATE) \
```
jjheinon: This was changed a bit, we kept on using UpdateSurfaceArea even when newer UpdateWindowFromSurface was available. Now it works better.

src/protocols/rdp/gdi.c
```diff
@@ -83,10 +102,16 @@ BOOL guac_rdp_gdi_begin_paint(rdpContext* context) {
...
-    current_context->stride = gdi->stride;
-    guac_rect_init(&current_context->bounds, 0, 0, gdi->width, gdi->height);
+    /*
+     * For full desktop sessions, the default layer shares FreeRDP's primary
+     * GDI buffer. For RemoteApp sessions, keep the default layer on its own
+     * black buffer and allow FreeRDP to paint the hidden server desktop only
+     * into its own primary buffer.
+     */
+    if (rdp_client->settings->remote_app == NULL
+            || !rdp_client->settings->enable_gfx
```
jjheinon: Use primary buffer only if RAIL is disabled, GFX is disabled or both are enabled, but we're not yet received first paint event. This fixes the issue where invalid credentials ends up to Windows login screen on non-NLA mode, but login screen is not visible. Earlier we rendered black screen in that case.

src/protocols/rdp/channels/rail.c
```diff
@@ -276,6 +733,393 @@ static BOOL guac_rdp_rail_window_update(rdpContext* context,
...
+            return ERROR_INTERNAL_ERROR;
+        }
+
+        guac_rect_extend(&layer_context->dirty, &dst_rect);
+        copied = 1;
+    }
+
+    guac_display_layer_close_raw(rail_window->layer, layer_context);
+    if (!copied) {
+        guac_common_list_unlock(rdp_client->rail_windows);
```
jjheinon, line 2416: Let's not render background before window rendering has started and we have received app data.

src/protocols/rdp/channels/rail.c
```diff
@@ -276,6 +733,393 @@ static BOOL guac_rdp_rail_window_update(rdpContext* context,
...
+        return CHANNEL_RC_OK;
+    }
+    guac_rdp_rail_paint_background(rdp_client);
```
jjheinon: Render only after data has been received. This prevents us blocking Windows login dialog on non-NLA logins, if credentials are incorrect.

src/protocols/rdp/channels/rail.c
```diff
@@ -276,10 +312,154 @@ static guac_common_list_element* guac_rdp_rail_get_window_element(
...
+ * Returns a z-order value above all currently tracked RAIL windows. The
+ * rail_windows list lock must be held.
+ *
+ * @param rdp_client
+ *     The RDP client whose tracked windows should be counted.
+ *
+ * @return
+ *     A z-order value above all currently tracked RAIL windows.
+ */
+static int guac_rdp_rail_get_top_z(guac_rdp_client* rdp_client) {
```
jjheinon, line 761: When launching multiple remote apps, the current window layering code couldn't handle popup/child windows properly. Now we handle that by tracking the window ownership and making sure the active window popups are visible and do not get hidden under other app windows.

src/protocols/rdp/channels/rail.c
```diff
@@ -687,8 +867,15 @@ static BOOL guac_rdp_rail_window_update(rdpContext* context,
 
         if (rail_window != NULL) {
 
-            int needs_opacity_update = !rail_window->has_rail_order;
+            int had_rail_order = rail_window->has_rail_order;
+            int needs_opacity_update = !had_rail_order;
             rail_window->has_rail_order = 1;
+            int needs_raise = 0;
+
+            if (fieldFlags & WINDOW_ORDER_FIELD_OWNER) {
```
jjheinon, line 1685: A window order includes the owner field. Make sure it's shown above the owner window.

src/protocols/rdp/channels/rail.c
```diff
@@ -808,6 +1019,12 @@ static BOOL guac_rdp_rail_monitored_desktop(rdpContext* context,
     if ((orderInfo->fieldFlags & WINDOW_ORDER_FIELD_DESKTOP_ZORDER)
             && monitoredDesktop->windowIds != NULL) {
 
+        for (guac_common_list_element* current = rdp_client->rail_windows->head;
+                current != NULL; current = current->next) {
+            guac_rdp_rail_window* rail_window = current->data;
+            rail_window->in_monitored_desktop_zorder = 0;
```
jjheinon: We got the official z-index for this window from the server, so we don't need to handle it separately anymore.

src/protocols/rdp/channels/rail.c
necouchman: Ah, okay - if you could just:
* Remove the empty line between the comment block and the start of the `typedef`
* Start the comment with the proper `/**`

src/protocols/rdp/channels/rail.h
```diff
@@ -60,5 +67,118 @@
...
+ *     The RDPGFX client context associated with the surface mapping.
+ *
+ * @param surface_id
+ *     The server-side ID of the mapped RDPGFX surface.
+ *
+ * @param window_id
+ *     The server-side ID of the RAIL window to which the surface was mapped.
+ *
+ * @return
+ *     CHANNEL_RC_OK if the surface mapping was handled successfully.
```
necouchman: ...or what, if the mapping was not handled correctly?

src/protocols/rdp/channels/rail.h
```diff
@@ -60,5 +67,118 @@
...
+ * RAIL window.
+ *
+ * @param context
+ *     The RDPGFX client context associated with the surface unmapping.
+ *
+ * @param window_id
+ *     The server-side ID of the RAIL window that was unmapped.
+ *
+ * @return
+ *     CHANNEL_RC_OK if the surface unmapping was handled successfully.
```
necouchman: ...or what, if the unmapping was not handled successfully?

src/protocols/rdp/channels/rail.h
```diff
@@ -60,5 +67,118 @@
...
+ *     The server-side ID of the updated RDPGFX surface.
+ *
+ * @param rect_count
+ *     The number of rectangles within the rects array.
+ *
+ * @param rects
+ *     The updated rectangular regions of the surface.
+ *
+ * @return
+ *     CHANNEL_RC_OK if the surface was painted successfully, or an error code.
```
necouchman: What type of error code? FreeRDP? Guacamole? Generic negative/positive integer?

src/protocols/rdp/channels/rail.h
```diff
@@ -60,5 +67,118 @@
...
+ * subsequently clears that region itself.
+ *
+ * @param context
+ *     The RDPGFX client context associated with the surface update.
+ *
+ * @param surface
+ *     The RDPGFX surface containing the updated window contents.
+ *
+ * @return
+ *     CHANNEL_RC_OK if the surface was painted successfully, or an error code.
```
necouchman: What type of error code?

src/protocols/rdp/channels/rdpgfx.c
```diff
@@ -58,16 +61,54 @@ static void guac_rdp_rdpgfx_channel_connected(rdpContext* context,
...
+        : gdi_graphics_pipeline_init(gdi, rdpgfx);
+#else
+    gfx_initialized = gdi_graphics_pipeline_init(gdi, rdpgfx);
+#endif
+
+    if (!gfx_initialized) {
         guac_client_log(client, GUAC_LOG_WARNING, "Rendering backend for RDPGFX "
                 "channel could not be loaded. Graphics may not render at all!");
-    else
+    } else {
```
necouchman: Please don't cuddle braces.

src/protocols/rdp/channels/rdpgfx.c
```diff
@@ -119,4 +160,3 @@ void guac_rdp_rdpgfx_load_plugin(rdpContext* context) {
     guac_freerdp_dynamic_channel_collection_add(context->settings, "rdpgfx", NULL);
 
 }
-
```
necouchman, line 122: I believe we determined recently that the empty line at the end of the file is actually required on some platforms.

src/protocols/rdp/channels/rail.c
```diff
@@ -37,6 +48,556 @@
...
+
+    /**
+     * The server-side ID of the RAIL window.
+     */
+    UINT64 window_id;
+
+    /**
+     * The server-side ID of the owning RAIL window, if any.
+     */
+    UINT64 owner_window_id;
```
necouchman, line 319: On these two comments it might be good to clarify that "server-side" means RDP server-side (versus Guacamole server/guacd), just to avoid confusion.

src/protocols/rdp/channels/rail.c
```diff
@@ -37,6 +48,556 @@
...
+     * RAIL order arrives) are held off-screen until this is set.
+     */
+    int has_rail_order;
+
+    /**
+     * Whether this window was included in the most recent server-provided
+     * monitored desktop z-order list.
+     * Used to avoid applying the raise to windows that already have authoritative z-order.
+     * We obey the server provided z-order, whenever it is available.
+     */
```
necouchman, line 381: #nitpick, but I think the `Used to avoid...` portion can be moved to the line above and everything re-spaced to make a nicer comment block.

src/protocols/rdp/channels/rail.c
```diff
@@ -37,6 +48,556 @@
...
+    int in_monitored_desktop_zorder;
+
+    /**
+     * Whether this window has already been raised due to receiving its first
+     * visible paint after its RAIL order was known.
+     * If the window does not yet have authoritative z-order, we apply this fallback raise only once.
```
necouchman: Similar to just above - the `If the window...` portion can move to the line above.

src/protocols/rdp/channels/rail.c
```diff
@@ -37,6 +48,556 @@
...
+ */
+static guac_rdp_rail_window* guac_rdp_rail_get_window(
+        guac_rdp_client* rdp_client, UINT64 window_id) {
+
+    guac_common_list_element* current = rdp_client->rail_windows->head;
+
+    while (current != NULL) {
+        guac_rdp_rail_window* rail_window = current->data;
+
+        if (rail_window->window_id == window_id)
```
necouchman: It might be worth adding a check, here for `rail_window != null`, just to protect against a null pointer de-reference.

src/protocols/rdp/channels/rail.c
```diff
@@ -37,6 +48,556 @@
...
+ */
+static guac_common_list_element* guac_rdp_rail_get_window_element(
+        guac_rdp_client* rdp_client, UINT64 window_id) {
+
+    guac_common_list_element* current = rdp_client->rail_windows->head;
+
+    while (current != NULL) {
+        guac_rdp_rail_window* rail_window = current->data;
+
+        if (rail_window->window_id == window_id)
```
necouchman: Similarly, here, might be worth checking to make sure `rail_window` is not null before trying to grab the member.

src/protocols/rdp/channels/rail.c
```diff
@@ -37,6 +48,556 @@
...
+ */
+static int guac_rdp_rail_get_top_z(guac_rdp_client* rdp_client) {
+
+    int z = 1;
+
+    for (guac_common_list_element* current = rdp_client->rail_windows->head;
+            current != NULL; current = current->next) {
+        guac_rdp_rail_window* rail_window = current->data;
+
+        if (rail_window->z_order >= z)
```
necouchman: And, here...check to make sure `rail_window` is not null.

src/protocols/rdp/channels/rail.c
```diff
@@ -37,6 +48,556 @@
...
+ */
+static int guac_rdp_rail_raise_owned_windows(guac_rdp_client* rdp_client,
+        UINT64 owner_window_id, int z) {
+
+    guac_common_list_element* current = rdp_client->rail_windows->head;
+
+    while (current != NULL) {
+        guac_rdp_rail_window* rail_window = current->data;
+
+        if (rail_window->owner_window_id == owner_window_id) {
```
necouchman: And, here...

src/protocols/rdp/channels/rail.c
necouchman: Why would we just ignore it? If the application is not in the allow list, the server shouldn't actually provide the application, and we shouldn't get anything, correct?

src/protocols/rdp/channels/rail.c
```diff
@@ -279,6 +945,486 @@ static BOOL guac_rdp_rail_window_update(rdpContext* context,
...
+ *     The RDPGFX surface containing the updated window contents.
+ *
+ * @param rect_count
+ *     The number of rectangles within the rects array.
+ *
+ * @param rects
+ *     The updated rectangular regions of the surface to copy.
+ *
+ * @return
+ *     CHANNEL_RC_OK if the surface was painted successfully, or an error code.
```
necouchman: What kind of error code?

necouchman (changes requested):
This code is looking really good - a handful of changes, mostly to comments, but overall I'm liking it.

@jjheinon Are you good with it? I see a lot of comments and pushes coming in - do we need to convert it to a draft at this point, or are you good with it?

src/protocols/rdp/channels/rail.c
jjheinon: This was an interesting edge case. Apparently it's possible for Windows RemoteApp to launch another process, which does not belong to allowed app listing for the RemoteApp session.
In this case, the requested RAIL app launches properly (and renders the app window), but server sends RAIL_EXEC_E_NOT_IN_ALLOWLIST when it prevents launching the other app process.

Not sure what should be the optimal behavior in this case, but removed the modification and reverted back to the original behavior (connection fails), but added log message about it so the Windows admin can make the requested changes on their environment.

src/protocols/rdp/channels/rail.c
```diff
@@ -167,6 +996,11 @@ static UINT guac_rdp_rail_execute_result(RailClientContext* context,
 
     guac_client* client = (guac_client*) context->custom;
 
+    if (execResult->execResult == RAIL_EXEC_E_NOT_IN_ALLOWLIST) {
```
jjheinon: Apparently it is possible to get RAIL_EXEC_E_NOT_IN_ALLOWLIST error even when RemoteApp is launched properly, if the app in question launches other processes that are not in the AllowList.
In this case you will get logged in and the app flashes on the screen but then you get kicked out.
Not sure what the optimal behavior would be in this case, but kept the old functionality and just added a warning, so the Windows admin can go and modify their remote app configuration.

necouchman:
@jjheinon I've marked this as a draft for now as it looks like you're still pretty actively working through things and making changes. When you're happy that it's ready for a final review and merge, just hit that Ready for review button.

jjheinon:
Quite happy with it now, not aware of any major issues or bugs (tested on FreeRDP 3.26.0).
For future reference, noticed that the memory usage for RDPGFX is quite massive compared to non-GFX, so something to optimize in the future. This is not just a RAIL issue but an overall optimization.

Also, the GfxProgressive codec is enabled for RAIL, but making additional codecs user configurable would be a nice improvement.
For example: FreeRDP_GfxProgressiveV2, FreeRDP_GfxH264, FreeRDP_GfxAVC444, FreeRDP_GfxAVC444v2

Will probably create a separate ticket about these in the future.

necouchman:
Thanks @jjheinon, I'll try to get it reviewed here today or tomorrow.

src/protocols/rdp/input-queue.c
```diff
@@ -59,6 +60,8 @@ static void guac_rdp_handle_mouse_event(guac_rdp_client* rdp_client,
     guac_user* user = event->user;
     int x = event->details.mouse.x;
     int y = event->details.mouse.y;
+    int rdp_x = x;
+    int rdp_y = y;
```
necouchman: I'm looking through the code, here, and I'm not seeing a reason why we need to make copies of the `x` and `y` variables in this function? Does one of the other functions where they are passed as an argument through here potentially update them? Or is there some other need to duplicate them?

src/protocols/rdp/channels/rail.h
```diff
@@ -60,5 +67,146 @@
...
+ * RAIL window.
+ *
+ * @param context
+ *     The RDPGFX client context associated with the surface unmapping.
+ *
+ * @param window_id
+ *     The server-side ID of the RAIL window that was unmapped.
+ *
+ * @return
+ *     CHANNEL_RC_OK is always returned. Surface unmapping failures are ignored as the window might not be tracked.
```
necouchman: If you could wrap this line at 80 columns that'd be nice :-).

src/protocols/rdp/gdi.c
```diff
@@ -169,27 +206,49 @@ BOOL guac_rdp_gdi_desktop_resize(rdpContext* context) {
...
-    guac_display_layer_raw_context* current_context = guac_display_layer_open_raw(default_layer);
+    guac_display_layer_raw_context* current_context = NULL;
+
+    /* Full desktop sessions share the resized GDI buffer with the default
+     * layer, and must hold an open raw context while the external buffer
+     * details change. */
+    if (rdp_client->settings->remote_app == NULL
+            || !rdp_client->settings->enable_gfx
+            || !rdp_client->rail_background_painted)
+        current_context = guac_display_layer_open_raw(default_layer);
```
necouchman, line 230: This there some reason why this particular operation - opening the raw display layer - needs to happen prior to the `gdi_resize()` call just below this? It looks like there are two identical `if` statements, here, and I'm just wondering if this `guac_display_layer_open_raw()` call could be moved inside the `if` statement down on line 224, or if it needs to happen prior to the FreeRDP GDI buffer resize?

src/protocols/rdp/channels/rail.c
```diff
@@ -66,20 +79,30 @@ static UINT guac_rdp_rail_complete_handshake(RailClientContext* rail) {
...
     RAIL_CLIENT_STATUS_ORDER client_status = {
-        .flags =
-                TS_RAIL_CLIENTSTATUS_ALLOWLOCALMOVESIZE
-              | TS_RAIL_CLIENTSTATUS_APPBAR_REMOTING_SUPPORTED
+        .flags = TS_RAIL_CLIENTSTATUS_APPBAR_REMOTING_SUPPORTED
     };
 
     /* Send client status */
-    guac_client_log(client, GUAC_LOG_TRACE, "Sending RAIL client status.");
+    guac_client_log(client, GUAC_LOG_DEBUG,
```
necouchman: Is there a reason why this should be `DEBUG` level and not remain `TRACE`? This seems like pretty low-level log messages, particularly including the actual flags, so I'm thinking `TRACE` is more appropriate?

src/protocols/rdp/channels/rail.c
```diff
@@ -136,11 +159,24 @@ static UINT guac_rdp_rail_complete_handshake(RailClientContext* rail) {
...
+            exec.RemoteApplicationArguments != NULL
+                ? exec.RemoteApplicationArguments : "",
+            exec.RemoteApplicationWorkingDir != NULL
+                ? exec.RemoteApplicationWorkingDir : "",
+            (unsigned int) exec.flags);
     pthread_mutex_lock(&(rdp_client->message_lock));
     status = rail->ClientExecute(rail, &exec);
     pthread_mutex_unlock(&(rdp_client->message_lock));
 
+    guac_client_log(client, GUAC_LOG_DEBUG,
```
necouchman: Since the log message above for the `Executing remote application...` log is `TRACE`, this should probably also be `TRACE` rather than `DEBUG`.

src/protocols/rdp/channels/rail.c
```diff
@@ -228,54 +269,1855 @@ static UINT guac_rdp_rail_handshake_ex(RailClientContext* rail,
...
+    /**
+     * The Guacamole layer that renders the contents of this window.
+     */
+    guac_display_layer* layer;
+
+    /**
+     * The position of this window within the remote desktop.
+     */
+    int x;
+    int y;
```
necouchman, line 351: This is a bit of a nit, but, in keeping with style used throughout the rest of the code, I think these variables need to be documented individually. I also think there's a technical reason behind it - I believe there are some automated processes that use the documentation to generate things like API documentation (e.g. https://guacamole.apache.org/doc/libguac/), and that documenting multiple variables with a single comment may cause some problems with that.

There are several places throughout the code, here, where this needs to be tweaked.

src/protocols/rdp/channels/rail.c
```diff
@@ -228,54 +269,1855 @@ static UINT guac_rdp_rail_handshake_ex(RailClientContext* rail,
...
+                || x >= rail_window->x + rail_window->layer_width
+                || y >= rail_window->y + rail_window->layer_height)
+            continue;
+
+        if (clicked_window == NULL || rail_window->z_order > clicked_z) {
+            clicked_window = rail_window;
+            clicked_z = rail_window->z_order;
+        }
+
+    }
```
necouchman, line 980: Might be nice to have a few comments in here to document the code flow in-line.

src/protocols/rdp/channels/rail.c
```diff
@@ -228,54 +269,1855 @@ static UINT guac_rdp_rail_handshake_ex(RailClientContext* rail,
...
+
+    if (target_window != NULL
+            && rdp_client->mouse_button_mask == 0 && mask != 0)
+        rdp_client->rail_mouse_window_id = target_window->window_id;
+
+    if (mask == 0)
+        rdp_client->rail_mouse_window_id = 0;
+
+    guac_common_list_unlock(rdp_client->rail_windows);
```
necouchman, line 1094: Similar to above, a handful of comments in here would be helpful.

necouchman (changes requested):
A few more items:
* Overall, there are some large blocks of code, particularly in `rail.c`, where some inline comments to document the code flow, particularly the "why", would be useful.
* In data structures, each variable should be documented independently - I understand the desire to document, e.g. `x` and `y` with a single comment block as they necessarily go together, but 1) it doesn't follow established style in the code, and 2) I think there are actually some reasons why we don't do things that way related to automatic generation of documentation.

necouchman:
Note that the build currently fails when using FreeRDP on Rocky9 (2.11.7-7.el9_8.3):
```
channels/rail.c: In function ‘guac_rdp_rail_repaint_surface’:
channels/rail.c:1089:47: error: ‘guac_rdp_client’ has no member named ‘rdpgfx_interface’; did you mean ‘rail_interface’?
 1089 |     RdpgfxClientContext* rdpgfx = rdp_client->rdpgfx_interface;
      |                                               ^~~~~~~~~~~~~~~~
      |                                               rail_interface
channels/rail.c: In function ‘guac_rdp_rail_window_update’:
channels/rail.c:1353:31: error: implicit declaration of function ‘ConvertWCharNToUtf8Alloc’ [-Werror=implicit-function-declaration]
 1353 |                 char* title = ConvertWCharNToUtf8Alloc(title_string,
      |                               ^~~~~~~~~~~~~~~~~~~~~~~~
channels/rail.c:1353:31: error: initialization of ‘char *’ from ‘int’ makes pointer from integer without a cast [-Werror=int-conversion]
channels/rail.c: In function ‘guac_rdp_rail_paint_surface’:
channels/rail.c:1921:23: error: implicit declaration of function ‘freerdp_image_copy_no_overlap’ [-Werror=implicit-function-declaration]
 1921 |             success = freerdp_image_copy_no_overlap(
      |                       ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~
cc1: all warnings being treated as errors
make[4]: *** [Makefile:1146: channels/libguac_client_rdp_la-rail.lo] Error 1
make[4]: Leaving directory '/devpool/guacamole/guacamole-server/src/protocols/rdp'
make[3]: *** [Makefile:1473: all-recursive] Error 1
make[3]: Leaving directory '/devpool/guacamole/guacamole-server/src/protocols/rdp'
make[2]: *** [Makefile:763: all] Error 2
make[2]: Leaving directory '/devpool/guacamole/guacamole-server/src/protocols/rdp'
make[1]: *** [Makefile:544: all-recursive] Error 1
make[1]: Leaving directory '/devpool/guacamole/guacamole-server'
make: *** [Makefile:464: all] Error 2

```

src/protocols/rdp/gdi.c
jjheinon: guac_display_layer_open_raw() needs to happen before gdi_resize() to prevent rendering thread accessing the buffer while it is being modified without write lock. guac_display_layer_open_raw() acquires the lock before we do any resizing for the buffer. Without lock, we may get artifacts or worse.
It's mentioned in the comment, changed the wording to be more precise.
Maybe thr code should be changed to have two separate if branches instead of this.

src/protocols/rdp/input-queue.c
jjheinon: This was a leftover from when I was debugging scaling issues and had to preserve the x and y coordinates. It's no longer necessary, removed it.

src/protocols/rdp/channels/rail.c
jjheinon: No reason, that was temporary, changed it back now.

src/protocols/rdp/channels/rail.c
jjheinon: Fixed this and five other places in rail.c

src/protocols/rdp/channels/rail.c
jjheinon: Added comments to make it easier to follow.

src/protocols/rdp/channels/rail.c
jjheinon: Added

jjheinon:
> Note that the build currently fails when using FreeRDP on Rocky9 (2.11.7-7.el9_8.3):
> 
> ```
> channels/rail.c: In function ‘guac_rdp_rail_repaint_surface’:
> channels/rail.c:1089:47: error: ‘guac_rdp_client’ has no member named ‘rdpgfx_interface’; did you mean ‘rail_interface’?
>  1089 |     RdpgfxClientContext* rdpgfx = rdp_client->rdpgfx_interface;
>       |                                               ^~~~~~~~~~~~~~~~
>       |                                               rail_interface
> channels/rail.c: In function ‘guac_rdp_rail_window_update’:
> channels/rail.c:1353:31: error: implicit declaration of function ‘ConvertWCharNToUtf8Alloc’ [-Werror=implicit-function-declaration]
>  1353 |                 char* title = ConvertWCharNToUtf8Alloc(title_string,
>       |                               ^~~~~~~~~~~~~~~~~~~~~~~~
> channels/rail.c:1353:31: error: initialization of ‘char *’ from ‘int’ makes pointer from integer without a cast [-Werror=int-conversion]
> channels/rail.c: In function ‘guac_rdp_rail_paint_surface’:
> channels/rail.c:1921:23: error: implicit declaration of function ‘freerdp_image_copy_no_overlap’ [-Werror=implicit-function-declaration]
>  1921 |             success = freerdp_image_copy_no_overlap(
>       |                       ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~
> cc1: all warnings being treated as errors
> make[4]: *** [Makefile:1146: channels/libguac_client_rdp_la-rail.lo] Error 1
> make[4]: Leaving directory '/devpool/guacamole/guacamole-server/src/protocols/rdp'
> make[3]: *** [Makefile:1473: all-recursive] Error 1
> make[3]: Leaving directory '/devpool/guacamole/guacamole-server/src/protocols/rdp'
> make[2]: *** [Makefile:763: all] Error 2
> make[2]: Leaving directory '/devpool/guacamole/guacamole-server/src/protocols/rdp'
> make[1]: *** [Makefile:544: all-recursive] Error 1
> make[1]: Leaving directory '/devpool/guacamole/guacamole-server'
> make: *** [Makefile:464: all] Error 2
> ```


Fixed. Tested it now compiles and works on FreeRDP 2.11.7 and FreeRDP 3.26.0. RDPGFX is automatically disabled, if using FreeRDP 2. RDPGFX can be enabled or disabled with FreeRDP 3.
Tested on Ubuntu 24.04 (ARM64).

necouchman:
@jjheinon Thanks - there are a few conflicts that need to be resolved, now. I also see @mike-jumper moved this from staging/1.6.1 to patch branch.

PR #674, opened 2026-05-29
GUACAMOLE-2283: Runtime observability: add guacd process & thread naming.

bbennett-ks:
## Overview
Add runtime process and thread naming support to `guacd`. Connection processes now expose descriptive names (e.g., `vnc user@host:5900`) and threads expose meaningful names for easier debugging, profiling, and monitoring.

This change adds process title and thread naming support to guacd through three new APIs:
-`guac_process_title_init()`: captures the process's original argv/environ region at startup and prepares it for later process title updates.- 
-`guac_process_title_set()`: updates the process title shown by tools such as ps, top, and htop, and also updates the main thread's comm name so process-oriented views reflect the active connection.- 
-`guac_thread_name_set()`: assigns a descriptive name to the calling thread using the platform's thread naming facilities.

`guac_process_title_set()` rewrites the argv/environ memory region captured at startup, replacing the original command line with a connection-specific title. Because Linux exposes this region through `/proc/<pid>/cmdline`, tools such as ps, top, and htop automatically display the updated title. The main thread's comm name is also updated to keep process-oriented views consistent.

Child processes can use `guac_process_title_set()` to expose connection-specific names, while threads created can use `guac_thread_name_set()` to publish names such as `vnc-main`, `user-input`, or `rdp-audio`. This improves observability when debugging, profiling, or monitoring running `guacd` processes.

## Thread Names
```
conn-write:     Forwards data from the connected user to the connection's child process.
conn-read:      Forwards data from the connection's child process back to the connected user.
conn-route:     Performs the protocol handshake for a new client connection and routes it to a connection process.

user-conn:      Managesa single user's connection lifecycle from handshake through disconnect.
user-input:     Readsand parses instructions from a single user's socket.
user-pending:   Periodically promotes pending users into the active connection.

client-free:    Frees a guac_client in the background, bounded by a timeout in case the free handler hangs.

display-render: Drives the display render loop, flushing completed frames to the client.
display-wrk:    Encodes and sends graphical updates for dirty layer regions.

keep-alive:     Periodically sends keep-alive NOPs on an otherwise idle socket.

k8s-in :        Reads user input and forwards it to the Kubernetes pod.
k8s-worker:     Main Kubernetes client thread; manages the websocket connection to the pod.

rdp-audio:      Flushes buffered audio input to the RDP server at the negotiated rate.
rdp-print:      Streams output from the RDP print filter process to the client.
rdp-worker:     Main RDP client thread; runs the FreeRDP connection and event loop.

ssh-in :        Reads terminal input and forwards it to the SSH server.
ssh-worker:     Main SSH client thread; runs the SSH session and drives the terminal.

telnet-in:      Reads terminal input and forwards it to the telnet server.
telnet-worker:  Main telnet client thread; runs the telnet session and event loop.

vnc-worker:     Main VNC client thread; runs the libvncclient connection and message loop.

terminal:       Renders the terminal emulator display and processes output from the remote.
```
## Username Obfuscation
Showing the full username (e.g. vnc bbennett@host:5900) in the process title is discloses information that could be accessed by a non-admin user on the guacd server and could be used as a security exploit. Even though guacd is typically deployed on locked down systems with only admin access, the username is obfuscated to mitigate this potential security issue.

For example, **vnc bbennett@host:5900**, displays as **vnc bb****tt@host:5900**; short usernames, such as root, are completely obfuscated: **telnet ****@127.0.0.1:23**.

## Example Output
```
[bbennett@rocky-9]$ ps -eo pid,ppid,comm,args --forest | sed -n '/ 272004 /,$p'
...
 272004  272002  |          gua  |                       \_ guacd -f -b 127.0.0.1 -l 4822 -L debug
 272106  272004  |          rdp  |                           \_ rdp bb****tt@192.168.238.152:3389
 272122  272004  |          ssh  |                           \_ ssh bb****tt@192.168.238.152:22
 272146  272004  |          rdp  |                           \_ rdp Br****tt@192.168.238.154:3389
 272177  272004  |          ssh  |                           \_ ssh Br****tt@192.168.238.154:22
 272188  272004  |          vnc  |                           \_ vnc bb****tt@192.168.68.50:5900
 272198  272004  |          tel  |                           \_ telnet ****@127.0.0.1:23
```

```
[bbennett@rocky-9]$  pstree -apt 272004
guacd,272004 -f -b 127.0.0.1 -l 4822 -L debug
  ├─rdp Br****,272146
  │   ├─{display-render},272392
  │   ├─{display-wrk},272157
  │   ├─{display-wrk},272158
  │   ├─{display-wrk},272159
  │   ├─{display-wrk},272160
  │   ├─{display-wrk},272162
  │   ├─{display-wrk},272163
  │   ├─{display-wrk},272165
  │   ├─{display-wrk},272166
  │   ├─{display-wrk},272167
  │   ├─{display-wrk},272168
  │   ├─{display-wrk},272169
  │   ├─{display-wrk},272170
  │   ├─{display-wrk},272171
  │   ├─{display-wrk},272172
  │   ├─{display-wrk},272173
  │   ├─{display-wrk},272174
  │   ├─{keep-alive},272151
  │   ├─{rdp-worker},272154
  │   ├─{rdp-worker},272284
  │   ├─{rdp-worker},272285
  │   ├─{rdp-worker},272286
```

### DIFF
--- a/src/libguac/proctitle.c
+++ b/src/libguac/proctitle.c
@@ -1,0 +1,372 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "config.h"
+
+#include "guacamole/proctitle.h"
+
+#include <guacamole/mem.h>
+#include <guacamole/string.h>
+
+#include <pthread.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+#include <unistd.h>
+
+#ifdef HAVE_PRCTL
+#include <sys/prctl.h>
+#endif
+
+extern char** environ;
+
+/**
+ * The size, in bytes, of the buffer used to hold a thread "comm" name. The
+ * Linux kernel limits a thread's comm to TASK_COMM_LEN bytes (16), i.e. 15
+ * characters plus a null terminator, and silently truncates anything longer,
+ * so a larger buffer would serve no purpose.
+ */
+#define GUAC_PROCTITLE_COMM_LENGTH 16
+
+/**
+ * The size, in bytes, of the buffer used to hold the /proc path addressing a
+ * thread's comm pseudo-file. This must be large enough for
+ * GUAC_PROCTITLE_COMM_PATH expanded with the widest possible PID. A 64-bit
+ * PID is at most 20 decimal digits, and the surrounding literal text is 21
+ * characters, so 64 bytes leaves comfortable headroom.
+ */
+#define GUAC_PROCTITLE_COMM_PATH_LENGTH 64
+
+/**
+ * Format string for the /proc path of a thread's comm pseudo-file. The single
+ * argument is the target thread's TID; for the thread-group leader (the
+ * process "main" thread) this equals the process PID returned by getpid().
+ */
+#define GUAC_PROCTITLE_COMM_PATH "/proc/self/task/%ld/comm"
+
+/**
+ * The number of leading and trailing characters of a username preserved by
+ * mask_username(); the characters in between are replaced with a fixed run of
+ * asterisks.
+ */
+#define GUAC_PROCTITLE_USER_REVEAL 2
+
+/**
+ * The fixed asterisk run substituted for the masked (middle) portion of a
+ * username by mask_username(). Its width is intentionally constant, rather
+ * than matching the number of characters removed, so the obfuscated form does
+ * not leak the original username's length.
+ */
+#define GUAC_PROCTITLE_USER_MASK "****"
+
+/**
+ * The minimum username length, in bytes, for which mask_username() reveals the
+ * leading and trailing characters. Names shorter than twice
+ * GUAC_PROCTITLE_USER_REVEAL characters would have all (or overlapping)
+ * characters exposed, so they are masked in their entirety instead. The "+ 1"
+ * guarantees at least one character is always masked.
+ */
+#define GUAC_PROCTITLE_USER_MIN_REVEAL (2 * GUAC_PROCTITLE_USER_REVEAL + 1)
+
+/**
+ * The size, in bytes, of a buffer guaranteed to hold any string produced by
+ * mask_username(): GUAC_PROCTITLE_USER_REVEAL characters at each end plus the
+ * GUAC_PROCTITLE_USER_MASK run (whose sizeof() includes its NUL terminator).
+ */
+#define GUAC_PROCTITLE_USER_MASKED_BUFSIZE \
+    (2 * GUAC_PROCTITLE_USER_REVEAL + sizeof(GUAC_PROCTITLE_USER_MASK))
+
+/**
+ * The inclusive bounds of the printable ASCII range (space through tilde).
+ * mask_username() treats any username byte outside this range as
+ * non-printable and masks the username in full.
+ */
+#define ASCII_PRINTABLE_MIN 0x20
+#define ASCII_PRINTABLE_MAX 0x7E
+
+/**
+ * The start of the writable argv/environ area used for process titles.
+ */
+static char* guac_proctitle_buffer = NULL;
+
+/**
+ * The number of bytes available within guac_proctitle_buffer.
+ */
+static size_t guac_proctitle_buffer_size = 0;
+
+/**
+ * Serializes access to the process-global proctitle state: the argv overlay
+ * buffer and the /proc/<pid>/task/<tgid>/comm write. Static-initialized so
+ * the first caller does not race against a missing pthread_mutex_init().
+ */
+static pthread_mutex_t guac_proctitle_lock = PTHREAD_MUTEX_INITIALIZER;
+
+/**
+ * Updates the main thread's short comm name from any thread in the same
+ * process, without changing the calling thread's own comm. Writes through
+ * /proc/<pid>/task/<tid>/comm; the main thread is identified by
+ * TID == TGID == getpid(). This relies on the Linux procfs layout and is a
+ * no-op on platforms that lack it.
+ */
+static void guac_main_thread_name_set(const char* name) {
+
+#ifdef __linux__
+    char comm_path[GUAC_PROCTITLE_COMM_PATH_LENGTH];
+    char short_name[GUAC_PROCTITLE_COMM_LENGTH];
+    FILE* comm_file;
+
+    if (name == NULL)
+        return;
+
+    memset(short_name, '\0', sizeof(short_name));
+    guac_strlcpy(short_name, name, sizeof(short_name));
+
+    snprintf(comm_path, sizeof(comm_path), GUAC_PROCTITLE_COMM_PATH,
+            (long) getpid());
+
+    /* Write the new name to the main thread's comm pseudo-file. The kernel
+     * accepts a short string (no trailing newline required) and truncates
+     * to TASK_COMM_LEN. fopen() may fail in restricted environments where
+     * /proc is unavailable or write access is denied; treat that as a
+     * silent no-op since process naming is best-effort observability. */
+    comm_file = fopen(comm_path, "w");
+    if (comm_file == NULL)
+        return;
+
+    fputs(short_name, comm_file);
+    fclose(comm_file);
+#else
+    (void) name;
+#endif
+
+}
+
+void guac_process_title_init(int argc, char** argv) {
+
+    char* buffer_end;
+    char** copied_environ;
+    int envc = 0;
+    int i;
+
+    if (argc <= 0 || argv == NULL || argv[0] == NULL)
+        return;
+
+    pthread_mutex_lock(&guac_proctitle_lock);
+
+    /* Idempotent: a second init after the buffer is claimed is a no-op. */
+    if (guac_proctitle_buffer != NULL) {
+        pthread_mutex_unlock(&guac_proctitle_lock);
+        return;
+    }
+
+    buffer_end = argv[argc - 1] + strlen(argv[argc - 1]) + 1;
+
+    for (i = 0; environ != NULL && environ[i] != NULL; i++) {
+
+        if (buffer_end == environ[i])
+            buffer_end = environ[i] + strlen(environ[i]) + 1;
+
+        envc++;
+
+    }
+
+    /* Copy environ to heap so it survives later overwrites of the argv area.
+     * If any allocation fails, leave the original environ in place and bail
+     * out without committing partial state. */
+    copied_environ = guac_mem_alloc(sizeof(char*), envc + 1);
+    if (copied_environ == NULL) {
+        pthread_mutex_unlock(&guac_proctitle_lock);
+        return;
+    }
+
+    for (i = 0; i < envc; i++) {
+
+        copied_environ[i] = guac_strdup(environ[i]);
+        if (copied_environ[i] == NULL) {
+            while (i > 0)
+                guac_mem_free(copied_environ[--i]);
+            guac_mem_free(copied_environ);
+            pthread_mutex_unlock(&guac_proctitle_lock);
+            return;
+        }
+
+    }
+    copied_environ[envc] = NULL;
+
+    environ = copied_environ;
+    guac_proctitle_buffer = argv[0];
+    guac_proctitle_buffer_size = buffer_end - argv[0];
+
+    pthread_mutex_unlock(&guac_proctitle_lock);
+
+}
+
+void guac_process_title_set(const char* title) {
+
+    size_t title_length;
+
+    if (title == NULL)
+        return;
+
+    pthread_mutex_lock(&guac_proctitle_lock);
+
+    guac_main_thread_name_set(title);
+
+    if (guac_proctitle_buffer == NULL || guac_proctitle_buffer_size == 0) {
+        pthread_mutex_unlock(&guac_proctitle_lock);
+        return;
+    }
+
+    title_length = strlen(title);
+    if (title_length >= guac_proctitle_buffer_size)
+        title_length = guac_proctitle_buffer_size - 1;
+
+    memcpy(guac_proctitle_buffer, title, title_length);
+    guac_proctitle_buffer[title_length] = '\0';
+
+    if (title_length + 1 < guac_proctitle_buffer_size) {
+        memset(guac_proctitle_buffer + title_length + 1, '\0',
+                guac_proctitle_buffer_size - title_length - 1);
+    }
+
+    pthread_mutex_unlock(&guac_proctitle_lock);
+
+}
+
+/**
+ * Writes a partially obfuscated form of the given username into the provided
+ * buffer for inclusion in a process title. The goal is to retain enough of the
+ * username for an operator to recognize a session at a glance while keeping the
+ * full target account name out of world-readable process listings
+ * (/proc/<pid>/cmdline).
+ *
+ * The algorithm is:
+ *
+ *   1. A NULL or empty username yields an empty string, so the caller omits
+ *      the "user@" portion of the title entirely.
+ *
+ *   2. A username containing any byte outside the printable ASCII range
+ *      (0x20-0x7E) is masked in full (GUAC_PROCTITLE_USER_MASK). This avoids
+ *      splitting a multi-byte UTF-8 codepoint and emitting garbage.
+ *
+ *   3. A username shorter than GUAC_PROCTITLE_USER_MIN_REVEAL bytes is masked
+ *      in full, since revealing GUAC_PROCTITLE_USER_REVEAL characters at each
+ *      end would otherwise expose every (or overlapping) character (e.g.
+ *      "root").
+ *
+ *   4. Otherwise the first and last GUAC_PROCTITLE_USER_REVEAL characters are
+ *      preserved and the middle is replaced with the fixed-width run
+ *      GUAC_PROCTITLE_USER_MASK, e.g. "bbennett" -> "bb****tt". The run width
+ *      is constant regardless of username length so the result does not leak
+ *      the original length.
+ *
+ * @param user
+ *     The username to obfuscate, which may be NULL.
+ *
+ * @param out
+ *     The buffer to receive the NUL-terminated, obfuscated username. Should be
+ *     at least GUAC_PROCTITLE_USER_MASKED_BUFSIZE bytes.
+ *
+ * @param out_size
+ *     The size of the out buffer, in bytes.
+ */
+static void mask_username(const char* user, char* out, size_t out_size) {
+
+    if (user == NULL || *user == '\0') {
+        out[0] = '\0';
+        return;
+    }
+
+    size_t length = strlen(user);
+
+    /* Reveal the prefix & suffix edges only for sufficiently long, plain
+     * printable-ASCII usernames; otherwise mask the entire value. */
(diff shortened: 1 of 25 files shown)


src/protocols/rdp/rdp.c
```diff
@@ -732,10 +733,22 @@ static int guac_rdp_handle_connection(guac_client* client) {
...
+
     guac_client* client = (guac_client*) data;
     guac_rdp_client* rdp_client = (guac_rdp_client*) client->data;
     guac_rdp_settings* settings = rdp_client->settings;
 
+    {
+        char rdp_port[GUAC_USHORT_STRING_BUFSIZE];
+        if (guac_itoa_safe(rdp_port, sizeof(rdp_port), settings->port) < 1)
+            rdp_port[0] = '\0';
+        guac_process_title_set_endpoint("rdp", settings->username,
```
necouchman: I'd suggest for all of the protocols that these protocol titles be defined somewhere as a constant.

src/libguac/proctitle.c
```diff
@@ -0,0 +1,219 @@
...
+/**
+ * Updates the main thread's short comm name from any thread in the same
+ * process, without changing the calling thread's own comm. Writes through
+ * /proc/<pid>/task/<tid>/comm; the main thread is identified by
+ * TID == TGID == getpid().
+ */
+static void guac_main_thread_name_set(const char* name) {
+
+    char comm_path[64];
+    char short_name[16];
```
necouchman: Probably worth making both of these sizes into constants and documenting the rationale for the sizes.

src/libguac/proctitle.c
```diff
@@ -0,0 +1,219 @@
...
+    FILE* comm_file;
+
+    if (name == NULL)
+        return;
+
+    memset(short_name, '\0', sizeof(short_name));
+    guac_strlcpy(short_name, name, sizeof(short_name));
+
+    snprintf(comm_path, sizeof(comm_path), "/proc/self/task/%ld/comm",
+            (long) getpid());
```
necouchman: A couple of thoughts for this path:
* I think the /proc filesystem layout is pretty standard across UNIX-like distributions, but have you checked this to make sure that it will work reliably on non-Linux UNIX platforms? I know we have some community of users that likes to run guacd on FreeBSD, and there may be others. It would be good to insure compatibility across those.
* Could this path also be handled with a `#define` macro of some sort? That would allow it to be a constant rather than a string literal, and would also allow for documenting it a little more verbosely.

necouchman (changes requested):
Overall looks pretty good to me, and I think this is a very useful addition to the code. I do have a few comments - mainly related to doing a few more `#define` constants and documenting things, but also one related to compatibility of the code across non-Linux UNIX-like platforms.

src/protocols/rdp/rdp.c
bbennett-ks: Done!

src/libguac/proctitle.c
bbennett-ks: Done!

src/libguac/proctitle.c
bbennett-ks: Done!

I created a FreeBSD to confirm the issues & test the fix.

I slipped in a fix for a FreeBSD build break in `src/libguac/tcp.c`. 

FYI, the FreeBSD compiler is a bit tighter/different than the one we use on Linux. I had to disable some errors:
```
./configure CPPFLAGS="... -Wno-error=strict-prototypes -Wno-error=newline-eof -Wno-error=unused-but-set-variable"
```
I now understand why most files have an empty line at the end of the file... :)

src/protocols/ssh/ssh.c
```diff
@@ -201,6 +202,10 @@ static char* guac_ssh_get_credential(guac_client *client, char* cred_name) {
 
 void* ssh_input_thread(void* data) {
 
+    /* Thread name ssh-in: reads terminal STDIN and forwards it to the
```
eugen-keeper: The thread name comment says `ssh-in`, but the code sets `ssh-stdin`. I am not sure if this is intentional. Not critical anyway.

src/protocols/telnet/telnet.c
```diff
@@ -358,6 +359,10 @@ static void __guac_telnet_event_handler(telnet_t* telnet, telnet_event_t* event,
  */
 static void* __guac_telnet_input_thread(void* data) {
 
+    /* Thread name telnet-in: reads terminal STDIN and forwards it to the
```
eugen-keeper: The thread name comment says `telnet-in`, but the code sets `telnet-stdin`. Not critical.

src/protocols/kubernetes/kubernetes.c
```diff
@@ -176,6 +178,10 @@ struct lws_protocols guac_kubernetes_lws_protocols[] = {
  */
 static void* guac_kubernetes_input_thread(void* data) {
 
+    /* Thread name k8s-in: reads user input and forwards it to the
```
eugen-keeper: The thread name comment says `k8s-in`, but the code sets `k8s-input`. Not critical.

src/libguac/tcp.c
```diff
@@ -30,6 +30,11 @@
 #include <sys/socket.h>
 #include <unistd.h>
 
+/* Fallback for platforms that do not define EBADFD. */
+#ifndef EBADFD
+#define EBADFD EBADF
```
eugen-keeper, line 35: `tcp.c` already used `EBADFD` before this commit. Is this required for your changes, or a portability fix you noticed while building on another platform?

src/protocols/kubernetes/kubernetes.c
```diff
@@ -207,6 +217,13 @@ void* guac_kubernetes_client_thread(void* data) {
     pthread_t input_thread;
     char endpoint_path[GUAC_KUBERNETES_MAX_ENDPOINT_LENGTH];
 
+    char kubernetes_port[GUAC_USHORT_STRING_BUFSIZE];
+    if (guac_itoa_safe(kubernetes_port, sizeof(kubernetes_port),
+            settings->port) < 1)
+        kubernetes_port[0] = '\0';
+    guac_process_title_set_endpoint(GUAC_KUBERNETES_PROCESS_TITLE_NAME, NULL,
```
eugen-keeper: Would it make sense to include `settings->kubernetes_namespace` and `settings->kubernetes_pod` in the process title instead of (or in addition to) the API server hostname? Both fields are already available at this point, and something like `k8s default/mypod` would likely be very useful.

src/protocols/vnc/vnc.c
```diff
@@ -452,10 +453,20 @@ static rfbBool guac_vnc_handle_messages(guac_client* client) {
...
+    /* Thread name vnc-worker: main VNC client thread; runs the libvncclient
+     * connection and message loop. */
+    guac_thread_name_set("vnc-worker");
+
     guac_client* client = (guac_client*) data;
     guac_vnc_client* vnc_client = (guac_vnc_client*) client->data;
     guac_vnc_settings* settings = vnc_client->settings;
 
+    char vnc_port[GUAC_USHORT_STRING_BUFSIZE];
+    if (guac_itoa_safe(vnc_port, sizeof(vnc_port), settings->port) < 1)
```
eugen-keeper: Do I understand correctly that if the port is not specified, we will have something like `vnc bbennett@192.168.68.50:0` here? See `src\protocols\vnc\settings.c` lines 457 - 459.

eugen-keeper (changes requested):
The formatted process titles expose connection metadata (username, host, port) in `ps` and `/proc`, where other local users may potentially see it. Do we consider that acceptable for typical **guacd** deployments?

bbennett-ks:
> The formatted process titles expose connection metadata (username, host, port) in `ps` and `/proc`, where other local users may potentially see it. Do we consider that acceptable for typical **guacd** deployments?

I had concerns as well, especially for the username, as the other information is readily available to non-root users.. My thought is that because of the Guacamole deployment model, `guacd` is only accessible by admins, this is acceptable. But I could see an argument for removing username.

I added a comment to `proctitle.h` to emphasize the security aspect.

src/protocols/ssh/ssh.c
bbennett-ks: Thanks! Stale comment.

src/libguac/tcp.c
bbennett-ks: Yep, testing on FreeBSD I hit a build break because of  `EBADFD` being defined as `EBADF`. I didn't think worthwhile creating a separate issue for this.

src/protocols/kubernetes/kubernetes.c
bbennett-ks: Yep, absolutely.

src/protocols/vnc/vnc.c
bbennett-ks: Yep, should show ` ` is not specified, not `:0`.

src/protocols/kubernetes/kubernetes.c
bbennett-ks: Done!

src/protocols/kubernetes/kubernetes.c
bbennett-ks: Done!

src/protocols/vnc/vnc.c
bbennett-ks: Done!

eugen-keeper (approved):
LGTM

necouchman:
> > The formatted process titles expose connection metadata (username, host, port) in `ps` and `/proc`, where other local users may potentially see it. Do we consider that acceptable for typical **guacd** deployments?
> 
> I had concerns as well, especially for the username, as the other information is readily available to non-root users.. My thought is that because of the Guacamole deployment model, `guacd` is only accessible by admins, this is acceptable. But I could see an argument for removing username.
> 
> I added a comment to `proctitle.h` to emphasize the security aspect.

This is _probably_ okay - I would tend to agree that typical deployments are not going to grant just anyone access to the system running guacd (or a system running a set of containers that includes guacd).

Curious what @mike-jumper @corentin-soriano @sschiffli think about this, particularly if there are any objections?

bbennett-ks:
> > > The formatted process titles expose connection metadata (username, host, port) in `ps` and `/proc`, where other local users may potentially see it. Do we consider that acceptable for typical **guacd** deployments?
> > 
> > 
> > I had concerns as well, especially for the username, as the other information is readily available to non-root users.. My thought is that because of the Guacamole deployment model, `guacd` is only accessible by admins, this is acceptable. But I could see an argument for removing username.
> > I added a comment to `proctitle.h` to emphasize the security aspect.
> 
> This is _probably_ okay - I would tend to agree that typical deployments are not going to grant just anyone access to the system running guacd (or a system running a set of containers that includes guacd).
> 
> Curious what @mike-jumper @corentin-soriano @sschiffli think about this, particularly if there are any objections?

If there are objections, I could show the UID instead of the username. Another possibility is showing the connection name.

necouchman:
> > > > The formatted process titles expose connection metadata (username, host, port) in `ps` and `/proc`, where other local users may potentially see it. Do we consider that acceptable for typical **guacd** deployments?
> > > 
> > > 
> > > I had concerns as well, especially for the username, as the other information is readily available to non-root users.. My thought is that because of the Guacamole deployment model, `guacd` is only accessible by admins, this is acceptable. But I could see an argument for removing username.
> > > I added a comment to `proctitle.h` to emphasize the security aspect.
> > 
> > 
> > This is _probably_ okay - I would tend to agree that typical deployments are not going to grant just anyone access to the system running guacd (or a system running a set of containers that includes guacd).
> > Curious what @mike-jumper @corentin-soriano @sschiffli think about this, particularly if there are any objections?
> 
> If there are objections, I could show the UID instead of the username. Another possibility is showing the connection name.

Yeah - or, here's another idea that occurred to me, if the goal is to aid in "runtime observability", what a socket opened by guacd that could be read from that contains information on the guacd process(es)? This could be one of the following variants:
* A single socket opened by the main guacd process that records information about each of the forked child processes that contains useful/required information - some of the stuff, here, and maybe some additional items (counters, maybe?). This has the advantage of limiting the number of sockets that get opened, particularly on high-volume systems. This could be some sort of uniformly formatted output (CSV, JSON, YAML, etc.) that would contain the PIDs of the forked processes and the information about each one (protocol, user, hostname, connection UUID, total bytes transferred, etc., etc.).
* A socket per forked guacd process that contains this information for that specific process, in a uniformly-formatted output. This has the advantage of being able to more easily parse information for a specific process, but would result in a lot of sockets being opened.

I'm kind of channeling the way HAProxy provides a statistics socket that can be read from and will output the information in various formats. This seems like it might achieve some of the observability requirements, but has the advantage of being able to use filesystem permissions to limit who can read the data and interact with the socket.

I'm throwing that out as an idea - I'm not saying it necessarily should completely replace what you're doing, here, I still think this PR, as-is, is useful, and provided there aren't any objections to this information being accessible to anyone logged into the guacd system, I'm still okay with it. I'm just throwing it out as an idea, perhaps for a future change or to address some of the potential concerns, here.

bbennett-ks:
> > > The formatted process titles expose connection metadata (username, host, port) in `ps` and `/proc`, where other local users may potentially see it. Do we consider that acceptable for typical **guacd** deployments?
> > 
> > 
> > I had concerns as well, especially for the username, as the other information is readily available to non-root users.. My thought is that because of the Guacamole deployment model, `guacd` is only accessible by admins, this is acceptable. But I could see an argument for removing username.
> > I added a comment to `proctitle.h` to emphasize the security aspect.
> 
> This is _probably_ okay - I would tend to agree that typical deployments are not going to grant just anyone access to the system running guacd (or a system running a set of containers that includes guacd).
> 
> Curious what @mike-jumper @corentin-soriano @sschiffli think about this, particularly if there are any objections?

To  mitigate this, I've obfuscated the username, Details in the overview.

corentin-soriano:
> ```
> [bbennett@rocky-9]$ ps -eo pid,ppid,comm,args --forest | sed -n '/ 272004 /,$p'
> ...
>  272004  272002  |          gua  |                       \_ guacd -f -b 127.0.0.1 -l 4822 -L debug
>  272106  272004  |          rdp  |                           \_ rdp bbennett@192.168.238.152:3389
>  272122  272004  |          ssh  |                           \_ ssh bbennett@192.168.238.152:22
>  272146  272004  |          rdp  |                           \_ rdp BradleyBennett@192.168.238.154:3389
>  272177  272004  |          ssh  |                           \_ ssh BradleyBennett@192.168.238.154:22
>  272188  272004  |          vnc  |                           \_ vnc bbennett@192.168.68.50:5900
>  272198  272004  |          tel  |                           \_ telnet bbennett@127.0.0.1:23
> ```



> > > The formatted process titles expose connection metadata (username, host, port) in `ps` and `/proc`, where other local users may potentially see it. Do we consider that acceptable for typical **guacd** deployments?
> > 
> > 
> > I had concerns as well, especially for the username, as the other information is readily available to non-root users.. My thought is that because of the Guacamole deployment model, `guacd` is only accessible by admins, this is acceptable. But I could see an argument for removing username.
> > I added a comment to `proctitle.h` to emphasize the security aspect.
> 
> This is _probably_ okay - I would tend to agree that typical deployments are not going to grant just anyone access to the system running guacd (or a system running a set of containers that includes guacd).
> 
> Curious what @mike-jumper @corentin-soriano @sschiffli think about this, particularly if there are any objections?

I think it's not good practice, but that might not justify blocking this pull request.

The target's IP address and port are already visible via netstat with a non-privileged account so I think the biggest risk is in the leakage of the username and the username naming format.

Server access should be restricted to administrators, but an unauthorized user who has gained access to a non-privileged shell by exploiting a vulnerability in another process will be able to access this data.

bbennett-ks:
> Yeah - or, here's another idea that occurred to me, if the goal is to aid in "runtime observability", what a socket opened by guacd that could be read from that contains information on the guacd process(es)? This could be one of the following variants:
> 
> * A single socket opened by the main guacd process that records information about each of the forked child processes that contains useful/required information - some of the stuff, here, and maybe some additional items (counters, maybe?). This has the advantage of limiting the number of sockets that get opened, particularly on high-volume systems. This could be some sort of uniformly formatted output (CSV, JSON, YAML, etc.) that would contain the PIDs of the forked processes and the information about each one (protocol, user, hostname, connection UUID, total bytes transferred, etc., etc.).
> * A socket per forked guacd process that contains this information for that specific process, in a uniformly-formatted output. This has the advantage of being able to more easily parse information for a specific process, but would result in a lot of sockets being opened.
> 
> I'm kind of channeling the way HAProxy provides a statistics socket that can be read from and will output the information in various formats. This seems like it might achieve some of the observability requirements, but has the advantage of being able to use filesystem permissions to limit who can read the data and interact with the socket.
> 
> I'm throwing that out as an idea - I'm not saying it necessarily should completely replace what you're doing, here, I still think this PR, as-is, is useful, and provided there aren't any objections to this information being accessible to anyone logged into the guacd system, I'm still okay with it. I'm just throwing it out as an idea, perhaps for a future change or to address some of the potential concerns, here.

I think this would be a great addition, though I believe this would be a separate issue: having a subset of the information a HAProxy-like socket interface is useful in debugging and analyze system performance and scalability issues in the context of the entire system.

I like the single socket approach, though it provides limited information (connection ID, protocol, PID (of connection process), and start/up time.  There is a bi-directional socker between the guacd and the connection process that could be extended to provide the rest of the information (username, host, port...).

I'll file a Jira for this.

bbennett-ks:
> > ```
> > > > The formatted process titles expose connection metadata (username, host, port) in `ps` and `/proc`, where other local users may potentially see it. Do we consider that acceptable for typical **guacd** deployments?
> > > 
> > > 
> > > I had concerns as well, especially for the username, as the other information is readily available to non-root users.. My thought is that because of the Guacamole deployment model, `guacd` is only accessible by admins, this is acceptable. But I could see an argument for removing username.
> > > I added a comment to `proctitle.h` to emphasize the security aspect.
> > 
> > 
> > This is _probably_ okay - I would tend to agree that typical deployments are not going to grant just anyone access to the system running guacd (or a system running a set of containers that includes guacd).
> > Curious what @mike-jumper @corentin-soriano @sschiffli think about this, particularly if there are any objections?
> 
> I think it's not good practice, but that might not justify blocking this pull request.
> 
> The target's IP address and port are already visible via netstat with a non-privileged account so I think the biggest risk is in the leakage of the username and the username naming format.
> 
> Server access should be restricted to administrators, but an unauthorized user who has gained access to a non-privileged shell by exploiting a vulnerability in another process will be able to access this data.

Does the username obfuscation mitigation mitigate the security concernse? I just pushed this a few minutes ago & updated the overview.

bbennett-ks:
> I think this would be a great addition, though I believe this would be a separate issue: having a subset of the information a HAProxy-like socket interface is useful in debugging and analyze system performance and scalability issues in the context of the entire system.
> 
> I like the single socket approach, though it provides limited information (connection ID, protocol, PID (of connection process), and start/up time. There is a bi-directional socker between the guacd and the connection process that could be extended to provide the rest of the information (username, host, port...).
> 
> I'll file a Jira for this.

https://issues.apache.org/jira/browse/GUACAMOLE-2288

src/protocols/kubernetes/kubernetes.c
```diff
@@ -214,6 +223,32 @@ void* guac_kubernetes_client_thread(void* data) {
...
+                    "%s %s/%s/%s", GUAC_KUBERNETES_PROCESS_TITLE_NAME,
+                    kubernetes_namespace, settings->kubernetes_pod,
+                    settings->kubernetes_container);
+        else
+            snprintf(kubernetes_title, sizeof(kubernetes_title),
+                    "%s %s/%s", GUAC_KUBERNETES_PROCESS_TITLE_NAME,
+                    kubernetes_namespace, settings->kubernetes_pod);
+
+        guac_process_title_set(kubernetes_title);
+    }
```
necouchman: Is there some reason that this code is in braces, but without any sort of controlling statement (if/else, switch, etc.)? I think it's perfectly legal - it'll compile and run - but is there some reason why it's useful or necessary?

bbennett-ks: Apologies, a change I evidently un-reverted from a very earlier version where I was using a lock. The embedded curly braces is intended to limit the scope of a lock to minimize the time it's held. I'll remove it.

necouchman: No worries - I did some searching before asking the question to see if there was a particular use for that in C, and I couldn't find a good answer (even among the AI Chatbots!) amongst all of the standard articles about using braces for functions, if/else statements, switch statement, loops, etc.

bbennett-ks: Done!

necouchman (approved):

PR #675, opened 2026-06-02
GUACAMOLE-2281: Add auth-challenge and auth-response instructions to libguac.

aleitner:
Adds two new protocol instructions for authentication exchanges:

auth-challenge: opens a stream carrying the body of a challenge for a pending authentication exchange. Wire form is auth-challenge,<stream_idx>,<mimetype>,<challenge_id>. The mimetype identifies the auth flavor; the challenge_id is an opaque identifier the peer references in its matching auth-response.

auth-response: opens a stream carrying the response body for a previously-issued auth-challenge identified by the same challenge_id. Same wire shape as auth-challenge.

Each instruction's body rides as blobs on the announced stream, terminated by end.

Also adds a destructor (free_handler) on guac_stream, called from guac_user_free_stream / guac_client_free_stream and from guac_user_free / guac_client_free for streams still open at disconnect, so consumers can attach per-stream state that needs cleanup if the peer drops mid-stream.

The first consumer is WebAuthn passthrough (companion PR https://github.com/apache/guacamole-client/pull/1218) using application/x-webauthn-create+json and application/x-webauthn-get+json mimetypes.

[Guacamole client PR](https://github.com/apache/guacamole-client/pull/1218)

### DIFF
--- a/src/libguac/guacamole/protocol.h
+++ b/src/libguac/guacamole/protocol.h
@@ -1120,6 +1120,38 @@ int guac_protocol_send_clipboard(guac_socket* socket, const guac_stream* stream,
 int guac_protocol_send_name(guac_socket* socket, const char* name);
 
 /**
+ * Sends an auth-challenge instruction over the given guac_socket
+ * connection, announcing a stream carrying the challenge body for a
+ * pending authentication exchange. The peer responds via an
+ * auth-response instruction carrying the same challenge_id.
+ *
+ * The body is shipped on the given stream as blobs, terminated by an
+ * end, after this function returns. Callers are responsible for
+ * allocating and freeing the stream.
+ *
+ * @param socket
+ *     The guac_socket connection to use.
+ *
+ * @param stream
+ *     The stream along which the challenge body will be sent.
+ *
+ * @param mimetype
+ *     The mimetype of the data that will be sent along the given stream.
+ *     The mimetype identifies the auth flavor (e.g. WebAuthn create vs.
+ *     get).
+ *
+ * @param challenge_id
+ *     Opaque identifier carried by the auth-challenge. The peer's
+ *     matching auth-response will reference this identifier.
+ *
+ * @return
+ *     Zero on success, non-zero on error.
+ */
+int guac_protocol_send_auth_challenge(guac_socket* socket,
+        const guac_stream* stream, const char* mimetype,
+        const char* challenge_id);
+
+/**
  * Decodes the given base64-encoded string in-place. The base64 string must
  * be NULL-terminated.
  *
--- a/src/libguac/guacamole/stream.h
+++ b/src/libguac/guacamole/stream.h
@@ -105,6 +105,34 @@ struct guac_stream {
      */
     guac_user_end_handler* end_handler;
 
+    /**
+     * Optional cleanup callback for whatever the stream's data pointer
+     * holds. Called from guac_user_free_stream when the stream is freed,
+     * and from guac_user_free for any streams still open when the user
+     * disconnects. NULL if no cleanup is needed.
+     *
+     * Example:
+     * @code
+     *     static void my_data_free(guac_stream* stream) {
+     *         my_data* d = (my_data*) stream->data;
+     *         free(d->buffer);
+     *         free(d);
+     *     }
+     *
+     *     int my_pipe_handler(guac_user* user, guac_stream* stream,
+     *             char* mimetype, char* name) {
+     *         my_data* d = malloc(sizeof(*d));
+     *         d->buffer = NULL;
+     *         stream->data = d;
+     *         stream->blob_handler = my_blob_handler;
+     *         stream->end_handler = my_end_handler;
+     *         stream->free_handler = my_data_free;
+     *         return 0;
+     *     }
+     * @endcode
+     */
+    guac_stream_free_handler* free_handler;
+
 };
 
 #endif
--- a/src/libguac/guacamole/user-fntypes.h
+++ b/src/libguac/guacamole/user-fntypes.h
@@ -381,6 +381,19 @@ typedef int guac_user_ack_handler(guac_user* user, guac_stream* stream,
 typedef int guac_user_end_handler(guac_user* user, guac_stream* stream);
 
 /**
+ * Optional cleanup callback installed on a guac_stream for whatever the
+ * stream's data pointer holds. Called from guac_user_free_stream and
+ * guac_client_free_stream when the stream is freed, and from
+ * guac_user_free and guac_client_free for any streams still open when
+ * the user or client is disposed.
+ *
+ * @param stream
+ *     The stream being freed. The handler is responsible for releasing
+ *     whatever the stream's data pointer holds.
+ */
+typedef void guac_stream_free_handler(guac_stream* stream);
+
+/**
  * Handler for Guacamole join events. A join event is fired by the
  * guac_client whenever a guac_user joins the connection. There is no
  * instruction associated with a join event.
@@ -500,7 +513,7 @@ typedef int guac_user_put_handler(guac_user* user, guac_object* object,
         guac_stream* stream, char* mimetype, char* name);
 
 /**
- * Handler for Guacamole USB connect events, invoked when a "usbconnect" 
+ * Handler for Guacamole USB connect events, invoked when a "usbconnect"
  * instruction has been received from a user. This indicates that the user
  * has connected a USB device via WebUSB and it is available for redirection.
  *
@@ -540,9 +553,35 @@ typedef int guac_user_put_handler(guac_user* user, guac_object* object,
  *     an error occurred.
  */
 typedef int guac_user_usbconnect_handler(guac_user* user, const char* device_id,
-    int vendor_id, int product_id, const char* device_name, 
+    int vendor_id, int product_id, const char* device_name,
     const char* serial_number, int device_class, int device_subclass,
     int device_protocol, const char* interface_data);
+
+/**
+ * Handler for Guacamole "auth-response" events, invoked when an
+ * "auth-response" instruction has been received from a user. An
+ * auth-response announces a stream carrying the response body for a
+ * previously-issued auth-challenge identified by the same challenge_id.
+ *
+ * @param user
+ *     The user that sent the auth-response.
+ *
+ * @param stream
+ *     The stream along which the response body will be received.
+ *
+ * @param mimetype
+ *     The mimetype of the data that will be received along the given
+ *     stream.
+ *
+ * @param challenge_id
+ *     The challenge identifier of the originating auth-challenge being
+ *     responded to.
+ *
+ * @return
+ *     Zero on success, non-zero on error.
+ */
+typedef int guac_user_auth_response_handler(guac_user* user,
+        guac_stream* stream, char* mimetype, char* challenge_id);
 
 /**
 * Handler for Guacamole USB data events, invoked when a "usbdata" instruction
--- a/src/libguac/protocol.c
+++ b/src/libguac/protocol.c
@@ -1299,13 +1299,34 @@ int guac_protocol_send_video(guac_socket* socket, const guac_stream* stream,
     int ret_val;
 
     guac_socket_instruction_begin(socket);
-    ret_val = 
+    ret_val =
            guac_socket_write_string(socket, "5.video,")
         || __guac_socket_write_length_int(socket, stream->index)
         || guac_socket_write_string(socket, ",")
         || __guac_socket_write_length_int(socket, layer->index)
         || guac_socket_write_string(socket, ",")
         || __guac_socket_write_length_string(socket, mimetype)
+        || guac_socket_write_string(socket, ";");
+    guac_socket_instruction_end(socket);
+
+    return ret_val;
+
+}
+
+int guac_protocol_send_auth_challenge(guac_socket* socket,
+        const guac_stream* stream, const char* mimetype,
+        const char* challenge_id) {
+
+    int ret_val;
+
+    guac_socket_instruction_begin(socket);
+    ret_val =
+           guac_socket_write_string(socket, "14.auth-challenge,")
+        || __guac_socket_write_length_int(socket, stream->index)
+        || guac_socket_write_string(socket, ",")
+        || __guac_socket_write_length_string(socket, mimetype)
+        || guac_socket_write_string(socket, ",")
+        || __guac_socket_write_length_string(socket, challenge_id)
         || guac_socket_write_string(socket, ";");
     guac_socket_instruction_end(socket);
 
--- a/src/libguac/protocol.c
+++ b/src/libguac/protocol.c
@@ -1301,13 +1301,34 @@ int guac_protocol_send_video(guac_socket* socket, const guac_stream* stream,
     int ret_val;
 
     guac_socket_instruction_begin(socket);
-    ret_val = 
+    ret_val =
            guac_socket_write_string(socket, "5.video,")
         || __guac_socket_write_length_int(socket, stream->index)
         || guac_socket_write_string(socket, ",")
         || __guac_socket_write_length_int(socket, layer->index)
         || guac_socket_write_string(socket, ",")
         || __guac_socket_write_length_string(socket, mimetype)
+        || guac_socket_write_string(socket, ";");
+    guac_socket_instruction_end(socket);
+
+    return ret_val;
+
+}
+
+int guac_protocol_send_auth_challenge(guac_socket* socket,
+        const guac_stream* stream, const char* mimetype,
+        const char* challenge_id) {
+
+    int ret_val;
+
+    guac_socket_instruction_begin(socket);
+    ret_val =
+           guac_socket_write_string(socket, "14.auth-challenge,")
+        || __guac_socket_write_length_int(socket, stream->index)
+        || guac_socket_write_string(socket, ",")
+        || __guac_socket_write_length_string(socket, mimetype)
+        || guac_socket_write_string(socket, ",")
+        || __guac_socket_write_length_string(socket, challenge_id)
         || guac_socket_write_string(socket, ";");
     guac_socket_instruction_end(socket);
 
--- a/src/libguac/tests/protocol/auth_send.c
+++ b/src/libguac/tests/protocol/auth_send.c
@@ -1,0 +1,159 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+#include <CUnit/CUnit.h>
+#include <guacamole/protocol.h>
+#include <guacamole/socket.h>
+#include <guacamole/stream.h>
+
+#include <stdlib.h>
+#include <string.h>
+#include <sys/wait.h>
+#include <unistd.h>
+
+/**
+ * Reads everything available from the given file descriptor into a
+ * heap-allocated buffer. The buffer grows geometrically as needed.
+ * Closes the file descriptor before returning.
+ *
+ * @param fd
+ *     The file descriptor to read from.
+ *
+ * @param out_len
+ *     Out parameter that receives the number of bytes read.
+ *
+ * @return
+ *     A heap-allocated buffer containing the bytes read. The caller
+ *     takes ownership and must free the buffer.
+ */
+static char* read_all(int fd, int* out_len) {
+
+    int capacity = 4096;
+    char* buffer = malloc(capacity);
+    int offset = 0;
+    int numread;
+
+    while ((numread = read(fd, buffer + offset, capacity - offset)) > 0) {
+        offset += numread;
+        if (offset == capacity) {
+            capacity *= 2;
+            buffer = realloc(buffer, capacity);
+        }
+    }
+
+    close(fd);
+    *out_len = offset;
+    return buffer;
+
+}
+
+/**
+ * Runs the given sender inside a forked child process, writing to
+ * write_fd, then reads everything from read_fd in the parent and asserts
+ * equality against the expected bytes.
+ *
+ * @param sender
+ *     The function the child process invokes to write its bytes to
+ *     write_fd.
+ *
+ * @param write_fd
+ *     The file descriptor the sender should write to. Closed in the
+ *     parent before reading.
+ *
+ * @param read_fd
+ *     The file descriptor the parent reads from. Closed in the child
+ *     before sending.
+ *
+ * @param expected
+ *     The byte sequence expected to be received on read_fd.
+ *
+ * @param expected_len
+ *     The length of the expected byte sequence.
+ */
+static void run_sender_check(void (*sender)(int), int write_fd, int read_fd,
+        const char* expected, int expected_len) {
+
+    int childpid;
+    CU_ASSERT_NOT_EQUAL_FATAL((childpid = fork()), -1);
+
+    if (childpid == 0) {
+        close(read_fd);
+        sender(write_fd);
+        _exit(0);
+    }
+
+    close(write_fd);
+
+    int len;
+    char* buffer = read_all(read_fd, &len);
+
+    int status;
+    waitpid(childpid, &status, 0);
+
+    CU_ASSERT_EQUAL(len, expected_len);
+    if (len == expected_len)
+        CU_ASSERT_EQUAL(memcmp(buffer, expected, expected_len), 0);
+
+    free(buffer);
+
+}
+
+/**
+ * Sender for the auth-challenge test. Emits a single
+ * guac_protocol_send_auth_challenge with a synthetic UUID-shaped
+ * challenge id and a WebAuthn create mimetype, then flushes the socket
+ * so the parent process can read the bytes.
+ *
+ * @param fd
+ *     The file descriptor to write the wire bytes to.
+ */
+static void write_auth_challenge(int fd) {
+
+    guac_socket* socket = guac_socket_open(fd);
+    if (socket == NULL)
+        return;
+
+    guac_stream stream = { .index = 7 };
+    guac_protocol_send_auth_challenge(socket, &stream,
+            "application/x-webauthn-create+json",
+            "aaaaaaaa-aaaa-4aaa-8aaa-aaaaaaaaaaaa");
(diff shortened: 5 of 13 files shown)


src/libguac/protocol.c
```diff
@@ -1315,6 +1315,61 @@ int guac_protocol_send_video(guac_socket* socket, const guac_stream* stream,
...
+    int ret_val;
+
+    guac_socket_instruction_begin(socket);
+    ret_val =
+           guac_socket_write_string(socket, "4.done,")
+        || __guac_socket_write_length_string(socket, id)
+        || guac_socket_write_string(socket, ",")
+        || __guac_socket_write_length_string(socket, status)
+        || guac_socket_write_string(socket, ",")
+        || __guac_socket_write_length_string(socket, payload)
```
eugen-keeper: I'm not entirely sure about this  but I wonder whether carrying the full RPC body in a single `payload` argument here might run into the `GUAC_INSTRUCTION_MAX_LENGTH` 8192 limit. Do you expect **WebAuthn** payloads to always fit it?

aleitner: WebAuthn payloads can exceed 8192. A YubiKey assertion is under 1 KiB which is what I have been testing with, but TPM attestation objects (Windows Hello platform authenticator) run 8 to 11 KiB once the attestation chain is included.

Switched to a stream rather than a single message. The body of every req or done rides a paired pipe with the correlation id as its name and a reserved mimetype (`GUAC_USER_RPC_PAYLOAD_MIMETYPE`). libguac assembles the body internally and hands it to the handler via `body`/`body_len`, so req and done themselves stay simple at two args each. Generic across any req/done consumer rather than WebAuthn specific

src/libguac/guacamole/protocol.h
```diff
@@ -1119,6 +1120,88 @@ int guac_protocol_send_clipboard(guac_socket* socket, const guac_stream* stream,
...
+ * @param socket
+ *     The guac_socket connection to use.
+ *
+ * @param id
+ *     The correlation identifier of the request to cancel.
+ *
+ * @return
+ *     Zero on success, non-zero on error.
+ */
+int guac_protocol_send_cancel(guac_socket* socket, const char* id);
```
bbennett-ks: `guac_protocol_send_cancel(guac_client* ..`?

bbennett-ks:
Maybe add a UT for the new framework? Similar to `./src/libguac/tests/socket/fd_send_instruction.c`? May be worthwhile, since new framework code paths aren't exercise by the the Guacamole product.

src/libguac/guacamole/protocol.h
bbennett-ks, line 1116: Out of date comment: `param socket The `guac_socket`.

bbennett-ks:
Thinking through shared sessions: since **send_req()** broadcasts to all connected users, we would receive multiple "done" responses for a single request. Could we follow the existing `guac_client_owner_send_required()` pattern and target only the session owner instead of broadcasting?

This follows the existing RDP & VNC pattern of prompting only the session owner for authentication-related input.

src/libguac/user-handlers.c
```diff
@@ -403,6 +406,119 @@ int __guac_handle_file(guac_user* user, int argc, char** argv) {
...
+        if (new_buf == NULL) {
+            guac_protocol_send_ack(user->socket, stream,
+                    "Out of memory buffering RPC payload",
+                    GUAC_PROTOCOL_STATUS_SERVER_ERROR);
+            free(a->id);
+            free(a->buffer);
+            free(a);
+            stream->data = NULL;
+            stream->blob_handler = NULL;
+            stream->end_handler = NULL;
```
bbennett-ks: Missing `stream->ack_handler = NULL`.

It maybe be worth defining the following, as the `free()`'s and `stream->... = NULL` as used several times.:
```
static void __guac_rpc_body_assembly_free(
        __guac_rpc_body_assembly* a);

static void __guac_stream_clear_handlers(
        guac_stream* stream);
```

src/libguac/user-handlers.c
```diff
@@ -857,3 +999,95 @@ int __guac_user_call_opcode_handler(__guac_instruction_handler_mapping* map,
...
+    if (argc != 2) {
+        guac_user_log(user, GUAC_LOG_WARNING, "Received \"req\" "
+                "instruction with wrong number of arguments (expected 2, got %d).",
+                argc);
+        return 0;
+    }
+
+    char* body = NULL;
+    int body_len = 0;
+    __guac_rpc_pending_take(user, argv[0], &body, &body_len);
```
bbennett-ks: Add warning if `__guac_rpc_pending_take() < 0`?

src/libguac/user-handlers.c
```diff
@@ -857,3 +999,95 @@ int __guac_user_call_opcode_handler(__guac_instruction_handler_mapping* map,
...
+    if (argc != 2) {
+        guac_user_log(user, GUAC_LOG_WARNING, "Received \"done\" "
+                "instruction with wrong number of arguments (expected 2, got %d).",
+                argc);
+        return 0;
+    }
+
+    char* body = NULL;
+    int body_len = 0;
+    __guac_rpc_pending_take(user, argv[0], &body, &body_len);
```
bbennett-ks: Add warning if `__guac_rpc_pending_take() < 0`?

src/libguac/user-handlers.c
```diff
@@ -403,6 +406,119 @@ int __guac_handle_file(guac_user* user, int argc, char** argv) {
...
+ * Pulls the buffered body matching the given correlation id. The caller
+ * takes ownership of *body and must free it. Returns 0 if a body
+ * matched (in which case *body and *body_len are set), or -1 if nothing
+ * is buffered or the id does not match.
+ */
+static int __guac_rpc_pending_take(guac_user* user, const char* id,
+        char** body, int* body_len) {
+
+    if (user->__rpc_pending_id == NULL
+            || strcmp(user->__rpc_pending_id, id) != 0)
```
bbennett-ks: If `id` doesn't match, should we free anyway? To prevent a memory leak?

src/libguac/protocol.c
```diff
@@ -1315,6 +1317,90 @@ int guac_protocol_send_video(guac_socket* socket, const guac_stream* stream,
 
 }
 
+/**
+ * Sends a "req" or "done" body over a paired pipe named for the
```
bbennett-ks: This design seems to assume only one in-flight RPC body per socket at a time. That appears to be true for the current WebAuthn usage, but is it guaranteed for future consumers of this mechanism? It may be worth documenting that callers must serialize RPC sends on a given socket.

src/libguac/user-handlers.c
```diff
@@ -411,14 +527,40 @@ int __guac_handle_pipe(guac_user* user, int argc, char** argv) {
...
+            free(a);
+            guac_protocol_send_ack(user->socket, stream, "Out of memory",
+                    GUAC_PROTOCOL_STATUS_SERVER_ERROR);
+            return 0;
+        }
+        a->buffer = NULL;
+        a->buffer_len = 0;
+        a->buffer_capacity = 0;
+
+        stream->data = a;
```
bbennett-ks: I was curious why other handlers don't allocate a structure for `stream->data`. Looking around, it seems other users of `stream->data` (e.g. RDP upload handling) allocate and free that state outside the connection-handling code. With RPC, the connection layer itself allocates `stream->data` (and `->id` / `->buffer`) and is for freeing it.

\Are there edge cases where this state isn't freed, e.g. if `__guac_rpc_body_end_handler()` is never called?

Would it make sense to move stream-specific cleanup into `guac_user_free_stream()`? Rather than adding RPC specific cleanup logic, perhaps `guac_stream` could be extended to have a "destructor" callback? `guac_user_free_stream()` would call it (if !NULL). That would (hopefully!) handle edge cases & simplify the RPC handler code by removing much of the free logic...

bbennett-ks (changes requested):

src/libguac/guacamole/protocol.h
necouchman: Not necessarily, no - `guac_protocol_send_name` also uses `guac_socket`, and the implementation in `protocol.c` uses `guac_socket` and not `guac_client`.

src/libguac/user-handlers.c
```diff
@@ -403,6 +406,119 @@ int __guac_handle_file(guac_user* user, int argc, char** argv) {
...
+ * "done" payload pipe.
+ */
+typedef struct __guac_rpc_body_assembly {
+    char* id;
+    char* buffer;
+    int buffer_len;
+    int buffer_capacity;
+} __guac_rpc_body_assembly;
+
+static int __guac_rpc_body_blob_handler(guac_user* user, guac_stream* stream,
```
necouchman: This function needs some documentation.

src/libguac/user-handlers.c
```diff
@@ -403,6 +406,119 @@ int __guac_handle_file(guac_user* user, int argc, char** argv) {
...
+        a->buffer_capacity = new_cap;
+    }
+
+    memcpy(a->buffer + a->buffer_len, data, length);
+    a->buffer_len += length;
+    return 0;
+
+}
+
+static int __guac_rpc_body_end_handler(guac_user* user, guac_stream* stream) {
```
necouchman: This function also needs some documentation.

src/libguac/user-handlers.c
```diff
@@ -403,6 +406,119 @@ int __guac_handle_file(guac_user* user, int argc, char** argv) {
...
+    return 0;
+
+}
+
+/**
+ * Pulls the buffered body matching the given correlation id. The caller
+ * takes ownership of *body and must free it. Returns 0 if a body
+ * matched (in which case *body and *body_len are set), or -1 if nothing
+ * is buffered or the id does not match.
+ */
```
necouchman: This should include `@param` tags and `@return` documentation.

necouchman (changes requested):

bbennett-ks:
I think this (and it's companion: https://github.com/apache/guacamole-client/pull/1218) are actually drafts & not intended to be merged at this point. Maybe we should change the PR status?

necouchman:
> I think this (and it's companion: [apache/guacamole-client#1218](https://github.com/apache/guacamole-client/pull/1218)) are actually drafts & not intended to be merged at this point. Maybe we should change the PR status?

Okay, I've marked them as Draft for now.

src/libguac/guacamole/protocol.h
aleitner: No body to ship, so `send_cancel` doesn't need to allocate a stream. Keeping it on `guac_socket` matches `send_name` and the other instruction-only senders

src/libguac/guacamole/protocol.h
aleitner: `send_name` is unchanged by this PR, so the `@param socket` line still matches its signature. Were you flagging a different function

src/libguac/user-handlers.c
aleitner: Added `stream->ack_handler = NULL` everywhere and extracted `__guac_rpc_body_assembly_free` for the duplicated cleanup. Left the stream-handler clears inline for now (three sites), happy to fold into a helper if you'd prefer

src/libguac/user-handlers.c
aleitner: Added a warning. The body pipe always precedes the req in this protocol, so `take` returning -1 means the sender violated the convention

src/libguac/user-handlers.c
aleitner: Same warning added on the done path

src/libguac/user-handlers.c
aleitner: Looked at this more carefully and match-only is the right behavior. `end_handler` already drop-and-replaces pending when a new pipe ends, and `guac_user_free` cleans up anything left at disconnect. Adding the drop to `take` would also fire from `cancel`, which can land while another exchange is mid-flight, killing a legitimate buffered body. Bounded by the 64 KiB body cap and the connection lifetime

src/libguac/protocol.c
aleitner: Documented in the `send_req` and `send_done` doc comments. Callers on the same `guac_client` need to serialize RPC sends so the four wire instructions (pipe, blobs, end, req or done) do not interleave with another RPC's

src/libguac/user-handlers.c
aleitner: Thanks for the suggestion. Added a `destructor` callback to `guac_stream`, called from `guac_user_free_stream` / `guac_client_free_stream`, and from `guac_user_free` / `guac_client_free` for streams still open at disconnect. Pipe handler installs `__guac_rpc_body_assembly_free` as the destructor when it allocates the assembly

src/libguac/user-handlers.c
aleitner: Added a function-level docstring

src/libguac/user-handlers.c
aleitner: Same on the end handler

src/libguac/user-handlers.c
aleitner: Converted the prose to `@param` and `@return` tags

src/libguac/guacamole/protocol.h
bbennett-ks: Apologies, I meant to delete that. I believe instead I deleted a comment for:
```
 * @param socket
 *     The guac_socket connection to use.
 *
 * @param id
 *     The correlation identifier of the request to cancel.
 *
 * @return
 *     Zero on success, non-zero on error.
 */
int guac_protocol_send_cancel(guac_socket* socket, const char* id);
```

I was wondering if it made sense to change `guac_socket* socket` to `guac_client* client`, like the 2 other RPC APIs.

aleitner: Refactored all three to take guac_user instead. Cancel now has the same shape as send_req and send_done, and it also fixes your other comment about send_req broadcasting in shared sessions, since each call now goes to one user via user->socket rather than the client broadcast socket. Callers that want owner-targeting walk guac_client_for_owner first

aleitner:
Added unit tests in the latest push

bbennett-ks (approved):
I looked over the existing primitives (msg, pipe, argv) to understand the need for a new RPC set, and this is a good extension: none of them offer a correlated request/response, whereas this provides:
  - Bidirectional request/response channel (either side can initiate).
  - Client-generated UUID correlation that stays unique across the multiple processes in the round-trip workflow (where a per-connection stream/object index couldn't).
  - Payload breakup & reassembly: the body streams over a paired pipe as blobs and is reassembled by libguac, allowing payloads larger than a single instruction can carry.
  - Cancellation support.

mike-jumper (commented):
I'm mainly not sure about the set of `req`/`done`/`cancel` instructions and use of `pipe`. The structure resembles the established "object" pattern that we already use for `filesystem`, so in this case I'd lean more toward an object-style `service` that can be given a name.

For example:

1. `service` declares a service object named `webauthn`.
2. Client retrieves the object's JSON index of available streams, which here would be `create` and `get`.
3. When the client wishes to issue a request, it creates a `put` stream to the relevant named stream of the object, including an opaque request ID in the path (ie: `create/SOME_ID`).
4. When the server needs to send the response for a request, it issues a `body` for the same path.

Would that align with what you're seeing with the structure of WebAuthN (and possibly plans for USB)?

aleitner:
> I'm mainly not sure about the set of `req`/`done`/`cancel` instructions and use of `pipe`. The structure resembles the established "object" pattern that we already use for `filesystem`, so in this case I'd lean more toward an object-style `service` that can be given a name.
> 
> For example:
> 
>     1. `service` declares a service object named `webauthn`.
> 
>     2. Client retrieves the object's JSON index of available streams, which here would be `create` and `get`.
> 
>     3. When the client wishes to issue a request, it creates a `put` stream to the relevant named stream of the object, including an opaque request ID in the path (ie: `create/SOME_ID`).
> 
>     4. When the server needs to send the response for a request, it issues a `body` for the same path.
> 
> 
> Would that align with what you're seeing with the structure of WebAuthN (and possibly plans for USB)?

A couple clarifications before reshaping the PR.

WebAuthn passthrough is server initiated. The server forwards a challenge from upstream and the browser is the responder. The existing object pattern has get and put flowing client to server with body as the response back, and I couldn't find an example of a server-pushed body without a preceding client get. Would the shape you have in mind involve a server-issued put toward the client carrying the request, or is there an existing primitive for server-initiated push I'm missing

mike-jumper:
A challenge/response flow could just be:

1. An `auth-challenge` stream, with the challenge having an arbitrary challenge ID distinct from the stream ID. The fact that the challenge is for consumption via WebAuthN could be embedded into the mimetype.
2. An `auth-response` stream, with the response pointing at the challenge ID.

I add the `auth-` prefixes here just because having a naked `response` opcode feels like we're commandeering a broad term for too specific a purpose.

aleitner:
> A challenge/response flow could just be:
> 
>     1. An `auth-challenge` stream, with the challenge having an arbitrary challenge ID distinct from the stream ID. The fact that the challenge is for consumption via WebAuthN could be embedded into the mimetype.
> 
>     2. An `auth-response` stream, with the response pointing at the challenge ID.
> 
> 
> I add the `auth-` prefixes here just because having a naked `response` opcode feels like we're commandeering a broad term for too specific a purpose.

Okay so I am working on a redesign. Two new stream-opening instructions, `auth-challenge` and `auth-response`, each announcing a stream the body rides on directly. This way we have no paired pipe, reserved mimetype, or single-slot per-user state.

```
auth-challenge,<stream_idx>,<mimetype>,<challenge_id>
auth-response,<stream_idx>,<mimetype>,<challenge_id>
```
The mimetype identifies the ceremony kind. I went with `application/x-webauthn-create+json` and `application/x-webauthn-get+json`

src/libguac/guacamole/stream.h
```diff
@@ -105,6 +105,34 @@ struct guac_stream {
...
+     *         d->buffer = NULL;
+     *         stream->data = d;
+     *         stream->blob_handler = my_blob_handler;
+     *         stream->end_handler = my_end_handler;
+     *         stream->destructor = my_data_free;
+     *         return 0;
+     *     }
+     * @endcode
+     */
+    void (*destructor)(void* data);
```
mike-jumper: The established pattern here is a `free_handler`, which would need its own stream-specific typedef (just like the other structs).

src/libguac/user.c
```diff
@@ -122,13 +137,21 @@ guac_stream* guac_user_alloc_stream(guac_user* user) {
...
 }
 
 void guac_user_free_stream(guac_user* user, guac_stream* stream) {
 
+    /* Run the registered destructor for the stream's data, if any */
+    if (stream->destructor != NULL) {
+        stream->destructor(stream->data);
+        stream->destructor = NULL;
+        stream->data = NULL;
+    }
```
mike-jumper: Good idea to NULL out `stream->data` here, but that should probably be done more broadly and regardless of whether a destructor is provided. For example:

```c
if (stream->destructor != NULL)
    stream->destructor(stream->data);

/* Clean up any remaining dangling pointers before permitting reuse of the
 * stream */
stream->data = NULL;
stream->destructor = NULL;
stream->another_handler = NULL;
stream->yet_another_handler = NULL;
```

src/libguac/tests/user/auth_dispatch.c
```diff
@@ -0,0 +1,181 @@
...
+/**
+ * State captured by the test auth handlers so the test can verify what
+ * they were invoked with.
+ */
+typedef struct auth_capture {
+    int call_count;
+    int stream_index;
+    char mimetype[64];
+    char challenge_id[64];
+} auth_capture;
```
mike-jumper, line 57: Please document.

src/libguac/tests/user/auth_dispatch.c
```diff
@@ -0,0 +1,181 @@
...
+
+/**
+ * Handler that copies its arguments into the user's data pointer
+ * (assumed to be an auth_capture) so the test can assert them. Used for
+ * both the auth-challenge and auth-response paths since they share the
+ * same handler signature.
+ */
+static int capture_auth(guac_user* user, guac_stream* stream,
+        char* mimetype, char* challenge_id) {
```
mike-jumper, line 80: We need to always document all parameters and return values.

src/libguac/tests/user/auth_dispatch.c
```diff
@@ -0,0 +1,181 @@
...
+    int devnull = open("/dev/null", O_WRONLY);
+    CU_ASSERT_TRUE_FATAL(devnull >= 0);
+    user->socket = guac_socket_open(devnull);
+    CU_ASSERT_PTR_NOT_NULL_FATAL(user->socket);
+
+    return user;
+
+}
+
+static void teardown_auth_user(guac_user* user) {
```
mike-jumper, line 130: Please document this function.

src/libguac/tests/protocol/auth_send.c
```diff
@@ -0,0 +1,163 @@
...
+    run_sender_check(write_auth_challenge, fd[1], fd[0],
+            expected, sizeof(expected) - 1);
+
+}
+
+/**
+ * Sender for the auth-response test: emits send_auth_response referencing
+ * an earlier challenge id.
+ */
+static void write_auth_response(int fd) {
```
mike-jumper: Same here - we need to document all parameters.

src/libguac/user-handlers.c
```diff
@@ -58,6 +58,8 @@ __guac_instruction_handler_mapping __guac_instruction_handler_map[] = {
    {"usbconnect",    __guac_handle_usbconnect},
    {"usbdata",       __guac_handle_usbdata},
    {"usbdisconnect", __guac_handle_usbdisconnect},
+   {"auth-challenge", __guac_handle_auth_challenge},
```
mike-jumper: What are the semantics of receiving an auth challenge from the user?

src/libguac/tests/protocol/auth_send.c
```diff
@@ -0,0 +1,163 @@
...
+
+}
+
+/**
+ * Runs the given sender inside a forked child process, writing to write_fd,
+ * then reads everything from read_fd in the parent and asserts equality
+ * against expected (a string of the given length).
+ */
+static void run_sender_check(void (*sender)(int), int write_fd, int read_fd,
+        const char* expected, int expected_len) {
```
mike-jumper, line 90: Here, too, please - all parameters need to be documented.

src/libguac/tests/protocol/auth_send.c
```diff
@@ -0,0 +1,163 @@
...
+#include <string.h>
+#include <sys/wait.h>
+#include <unistd.h>
+
+/**
+ * Reads everything available from the given file descriptor into a
+ * heap-allocated buffer, returning the buffer and writing the length to
+ * *out_len. Caller takes ownership of the buffer. Closes the file descriptor.
+ */
+static char* read_all(int fd, int* out_len) {
```
mike-jumper, line 45: And here - documentation needed for parameters and return values.

mike-jumper (changes requested):

src/libguac/tests/user/auth_dispatch.c
aleitner: Documented the params and return

src/libguac/tests/user/auth_dispatch.c
aleitner: Documented setup_auth_user and teardown_auth_user

src/libguac/tests/protocol/auth_send.c
aleitner: Documented the params and return

src/libguac/tests/protocol/auth_send.c
aleitner: Documented all params

src/libguac/tests/protocol/auth_send.c
aleitner: Removed along with send_auth_response, nothing emits auth-response from the server. See the map entry for the removed auth-challenge

src/libguac/user-handlers.c
aleitner: Removed auth-challenge since nothing registers a handler. Also dropped send_auth_response, the typedef, and the field since the server only emits auth-challenge and only receives auth-response. Same on the JS side, dropped the auth-response handler, onauthresponse, and sendAuthChallenge. Can add the reverse direction back if anything else needs it

src/libguac/guacamole/stream.h
aleitner: Renamed to free_handler and added a guac_stream_free_handler typedef in user-fntypes.h alongside the other stream handler typedefs

src/libguac/user.c
aleitner: Moved the cleanup out of the if-block. data, ack_handler, blob_handler, end_handler, and free_handler are now all NULL'd unconditionally before the slot is released. Same fix in guac_client_free_stream

src/libguac/tests/user/auth_dispatch.c
aleitner: Added docs to each field

src/libguac/guacamole/user-fntypes.h
```diff
@@ -380,6 +380,18 @@ typedef int guac_user_ack_handler(guac_user* user, guac_stream* stream,
...
+ * Optional cleanup callback installed on a guac_stream for whatever the
+ * stream's data pointer holds. Called from guac_user_free_stream and
+ * guac_client_free_stream when the stream is freed, and from
+ * guac_user_free and guac_client_free for any streams still open when
+ * the user or client is disposed.
+ *
+ * @param data
+ *     The stream's data pointer at the time the stream is being freed.
+ */
+typedef void guac_stream_free_handler(void* data);
```
mike-jumper: This should receive the `guac_stream` being freed.

src/libguac/tests/user/auth_dispatch.c
```diff
@@ -0,0 +1,186 @@
...
+    int call_count;
+
+    /** Stream index of the most recent invocation's stream. */
+    int stream_index;
+
+    /** Mimetype argument of the most recent invocation. */
+    char mimetype[64];
+
+    /** Challenge id argument of the most recent invocation. */
+    char challenge_id[64];
```
mike-jumper: These should match the:

```c
/**
 * Doxygen block comments used elsewhere.
 */
```

rather than:

```c
/** Single-line Doxygen comment (which has no precedent here). */
```

src/libguac/user-handlers.c
```diff
@@ -857,3 +859,28 @@ int __guac_user_call_opcode_handler(__guac_instruction_handler_mapping* map,
...
+        return 0;
+
+    /* If supported, call handler */
+    if (user->auth_response_handler)
+        return user->auth_response_handler(
+            user,
+            stream,
+            argv[1], /* mimetype */
+            argv[2]  /* challenge_id */
+        );
```
mike-jumper, line 882: `argc` needs to be verified before dereferencing `argv`.

mike-jumper (changes requested):

src/libguac/guacamole/user-fntypes.h
aleitner: Switched the typedef to take the stream and updated the call sites and the test handler

src/libguac/tests/user/auth_dispatch.c
aleitner: Converted to block style

src/libguac/user-handlers.c
aleitner: Added an argc guard ahead of the argv reads

PR #677, opened 2026-06-10
GUACAMOLE-2265: Fix worker process leak after client disconnect.

escra:
This is the **main**-targeted forward-port of #662 (GUACAMOLE-2265). The original PR targets the frozen `staging/1.6.1` branch; this applies the same fix on top of current `main` (HEAD e775052a).

### Problem
A guacd worker process could leak after client disconnect: the blocking `guacd_recv_fd()` loop in `guacd_exec_proc()` never returns when the connection drops without a clean FD close, leaving the child alive in the process map.

### Fix
- Replace the blocking receive loop with a `poll()`-based loop with idle timeout, plus reconnect-grace and max-idle handling (`GUACD_PROC_IDLE_TIMEOUT_MS`, `GUACD_PROC_RECONNECT_GRACE_MS`, `GUACD_PROC_MAX_IDLE_MS`).
- Handle `POLLERR`/`POLLHUP`/`POLLNVAL`.
- Add `pthread_create()` failure handling in `guacd_proc_add_user()` (close fd, free params, stop proc).

### Verification
Builds cleanly on `main`; `guacd` links successfully. Substance is identical to #662 (verified diff parity).

### DIFF
--- a/src/guacd/proc.c
+++ b/src/guacd/proc.c
@@ -34,6 +34,7 @@
 #include <guacamole/user.h>
 
 #include <errno.h>
+#include <poll.h>
 #include <pthread.h>
 #include <signal.h>
 #include <stdlib.h>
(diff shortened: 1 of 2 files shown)


mike-jumper:
Closing as GUACAMOLE-2265 is marked as a duplicate of GUACAMOLE-2118.

PR #679, opened 2026-06-11
GUACAMOLE-2002: Warn on clipboard truncation instead of silently dropping data

escra:
Building on the configurable clipboard buffer work in GUACAMOLE-2002, this addresses a remaining silent-failure case: **charset expansion** during clipboard conversion can make the converted data outgrow the configured `clipboard-buffer-size`.

`clipboard->available` only bounds the *input*; it does not account for expansion such as UTF-8 ↔ UTF-16, CRLF translation, or multi-byte characters (e.g. emoji). `guac_iconv()` signals this by returning `0` (output buffer filled before all input was consumed), but the RDP (`cliprdr`) and VNC clipboard paths ignored that return value, so oversized clipboard transfers were **silently truncated or dropped** with no operator-visible signal.

## Changes

- **RDP `cliprdr.c`** (both the clipboard request and the `format_data_response` directions) and **VNC `clipboard.c`** (`end_handler` and `cut_text` directions) now check the `guac_iconv()` return value.
- On truncation we emit a `GUAC_LOG_WARNING` naming the byte limit and pointing the operator at the `clipboard-buffer-size` setting.
- For inbound data the converted (possibly partial) text is still **delivered/recorded** rather than dropped entirely — truncated clipboard text is still useful — and the warning is logged afterwards.
- This preserves `main`'s heap-allocated clipboard buffers (`guac_mem_alloc(clipboard->available)`) and the clipboard recording added in GUACAMOLE-1969.

## Relation to #659 and reviewer feedback

This follows up on #659 (now closed). In that thread @bbennett-ks and @eugen-keeper raised that simply over-allocating the buffer (e.g. `available * 4` to absorb worst-case expansion) is unsafe — with a 50 MiB `clipboard-buffer-size` that would mean pre-allocating up to **200 MiB** per clipboard transfer.

This PR deliberately chooses **detection + warning over pre-allocation**: the allocation stays bound to the operator-configured `clipboard-buffer-size`, and overflow is reported cleanly (with partial delivery where possible) instead of either silently dropping data or ballooning memory. cc @bbennett-ks @eugen-keeper

## Validation

- **Unit:** extended the `guac_iconv` CUnit suite with a truncation case (`test_iconv__truncation`) asserting a `0` return and no overflow for ASCII→UTF-16, CRLF, and emoji expansion.
- **Container/CI:** `test_guacamole_2002_clipboard.sh` and `test_cliprdr_hardening.sh` (static + functional checks), wired into our pipeline.
- **E2E against a real guacd build:** a 1 MiB clipboard transfer (the GUACAMOLE-2002 case) completes **without crash/disconnect** (session-clipboard suite: 21 passed); RDP + SSH sessions green.
- Honest status note: an XRDP large-buffer variant is still **deferred** due to a defective XRDP test target (not a guacd issue); further verification is ongoing. The PR can be treated as a draft until that completes if maintainers prefer.

Relates to / follows up on #659.

### DIFF
--- a/src/protocols/rdp/channels/cliprdr.c
+++ b/src/protocols/rdp/channels/cliprdr.c
@@ -403,8 +403,19 @@ static UINT guac_rdp_cliprdr_format_data_request(CliprdrClientContext* cliprdr,
      * requested */
     BYTE* start = (BYTE*) output;
     guac_iconv_read* local_reader = settings->normalize_clipboard ? GUAC_READ_UTF8_NORMALIZED : GUAC_READ_UTF8;
-    guac_iconv(local_reader, &input, clipboard->clipboard->length,
-            remote_writer, &output, output_buf_size);
+
+    /* A zero return indicates the output buffer filled before all input was
+     * consumed (truncation). Charset expansion (UTF-8 -> UTF-16, CRLF
+     * translation, or multi-byte characters such as emoji) can outgrow the
+     * configured clipboard buffer, so warn the operator rather than silently
+     * sending truncated data. */
+    if (!guac_iconv(local_reader, &input, clipboard->clipboard->length,
+            remote_writer, &output, output_buf_size)) {
+        guac_client_log(client, GUAC_LOG_WARNING, "CLIPRDR: Outbound clipboard "
+                "data was truncated to fit the %d-byte clipboard buffer. "
+                "Increase \"clipboard-buffer-size\" to transfer larger "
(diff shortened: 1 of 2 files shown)


necouchman (commented):
Overall I'm fine with the changes - I do have one question: Is this worth trying to warn the _user_ about rather than the _admin_? The logs will only go to the admin, but I'm wondering if sending a message to the user, similar to what we do for things like admins joining a session or file uploads would be useful?

PR #682, opened 2026-06-26
GUACAMOLE-1852: Send a desktop scale factor to the RDP server.

stephzero1:
### Add `desktop-scale-factor` RDP parameter

Add a new `desktop-scale-factor` RDP parameter to control the UI scaling of the remote Windows session.

* **Range:** `100–500%`
* **`0`:** disables the feature
* **Out-of-range values:** automatically clamped to the supported range
* **Parameter omitted:** previous behavior is unchanged

When configured, the value is sent to the server as the `[MS-RDPBCGR]` `desktopScaleFactor`. It is paired with a protocol-valid `deviceScaleFactor` (`100`, `140`, or `180`), as required by the specification.

This allows the client to request a **device-pixel, high-resolution framebuffer** while keeping the remote Windows UI and text at a comfortable size, resulting in a sharper display without making the UI excessively small.

The same scale factor is also included in the `[MS-RDPEDISP]` `DISPLAYCONTROL_MONITOR_LAYOUT` message, ensuring that the scaling is preserved across dynamic display resizes.

Overall, this makes it possible to combine **high-resolution rendering** with **comfortable UI scaling** in remote Windows sessions.

### Related issue

This addresses the use case described in **[[GUACAMOLE-1852](https://issues.apache.org/jira/browse/GUACAMOLE-1852)](https://issues.apache.org/jira/browse/GUACAMOLE-1852)**, which requests DPI/remote desktop scaling support to achieve Retina/high-resolution display quality without excessively small fonts and icons.

### Related client PR

The corresponding Guacamole Client UI support is available in **[[apache/guacamole-client#1222](https://github.com/apache/guacamole-client/pull/1222)](https://github.com/apache/guacamole-client/pull/1222)**, which exposes the `desktop-scale-factor` field in the RDP connection's **Display** settings.

### DIFF
--- a/src/protocols/rdp/channels/disp.c
+++ b/src/protocols/rdp/channels/disp.c
@@ -224,6 +224,23 @@ void guac_rdp_disp_update_size(guac_rdp_disp* disp,
     }
 
     else if (settings->resize_method == GUAC_RESIZE_DISPLAY_UPDATE) {
+
+        /*
+         * Forward any requested desktop scale factor to the server so the
+         * remote session keeps scaling its UI across dynamic resizes. Both
+         * fields default to 0 (no scaling requested), which the server ignores
+         * per [MS-RDPEDISP]. When set, the desktop scale factor MUST be paired
+         * with a protocol-valid device scale factor (100, 140, or 180) or both
+         * are ignored.
+         */
+        UINT32 desktop_scale_factor = 0;
+        UINT32 device_scale_factor = 0;
+        if (settings->desktop_scale_factor != 0) {
+            desktop_scale_factor = settings->desktop_scale_factor;
+            device_scale_factor =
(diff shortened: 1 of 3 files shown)


stephzero1:
The difference between before:
image
and after:
image

on my setup: 42" 4k display with 125% scale factor

m0pus:
Just used this patch on my guac server to test, and it is such a great improvement.

Hope this gets merged in soon.

stephzero1:
> Just used this patch on my guac server to test, and it is such a great improvement.
> 
> Hope this gets merged in soon.

Thanks for the "bump";
the patch was very straightforward (thanks to AI, of course).

The improvement is also pretty painless and essentially just an enhancement (it’s optional—you can use it if you want, or otherwise it defaults to the original behavior).

To me, integrating it is a no-brainer.

PR #685, opened 2026-07-05
GUACAMOLE-2294: Address multiple minor resource leaks.

necouchman:
This addresses a few minor resource leaks identified by Coverity/Blackduck. Nothing major, just some cleanup.

It's worth noting that our macro for freeing memory, `PRIV_guac_mem_free`, doesn't seem to handle freeing `const *` members/pointers.

### DIFF
--- a/src/terminal/display.c
+++ b/src/terminal/display.c
@@ -269,6 +269,10 @@ guac_terminal_display* guac_terminal_display_alloc(guac_client* client,
     if (guac_terminal_display_set_font(display, font_name, font_size, dpi)) {
         guac_client_abort(display->client, GUAC_PROTOCOL_STATUS_SERVER_ERROR,
                 "Unable to set initial font \"%s\"", font_name);
+        guac_common_surface_free(display->display_surface);
+        guac_client_free_layer(client, display->display_layer);
+        guac_client_free_layer(client, display->select_layer);
+        guac_mem_free(display->default_palette);
         guac_mem_free(display);
         return NULL;
     }
--- a/src/terminal/terminal.c
+++ b/src/terminal/terminal.c
@@ -557,6 +557,10 @@ guac_terminal* guac_terminal_create(guac_client* client,
     /* Fail if display init failed */
     if (term->display == NULL) {
         guac_client_log(client, GUAC_LOG_DEBUG, "Display initialization failed");
+        guac_terminal_buffer_free(term->alternate_buffer);
+        free((char *)(term->color_scheme));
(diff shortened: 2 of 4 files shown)


src/terminal/terminal.c
```diff
@@ -557,6 +557,10 @@ guac_terminal* guac_terminal_create(guac_client* client,
     /* Fail if display init failed */
     if (term->display == NULL) {
         guac_client_log(client, GUAC_LOG_DEBUG, "Display initialization failed");
+        guac_mem_free(term->alternate_buffer);
```
eugen-keeper: Should we use `guac_terminal_buffer_free()` here?

src/terminal/display.c
```diff
@@ -269,6 +269,9 @@ guac_terminal_display* guac_terminal_display_alloc(guac_client* client,
     if (guac_terminal_display_set_font(display, font_name, font_size, dpi)) {
         guac_client_abort(display->client, GUAC_PROTOCOL_STATUS_SERVER_ERROR,
                 "Unable to set initial font \"%s\"", font_name);
+        guac_mem_free(display->display_layer);
```
eugen-keeper: Font-failure cleanup still looks incomplete: layers should use `guac_client_free_layer()`, the surface should use `guac_common_surface_free()`, and `display->default_palette` should be freed here too probably.

eugen-keeper (changes requested):

src/terminal/terminal.c
necouchman: Probably so, yes.

src/terminal/display.c
necouchman: Okay, I'll take another look.

src/terminal/terminal.c
necouchman: Okay, I've switched over to that for the buffers.

src/terminal/display.c
necouchman: I've switched to the proper `free`ing functions. It's worth noting that there is a `guac_palette_free()` function, but it isn't part of the public API and is limited to callers in the `libguac` source tree.

mike-jumper:
> It's worth noting that our macro for freeing memory, `PRIV_guac_mem_free`, doesn't seem to handle freeing `const *` members/pointers.

We provide `guac_mem_free_const()` as a convenience for this, but yeah - the automatic `NULL` assignment of `guac_mem_free()` just can't (or shouldn't) be done in the `const` case.

mike-jumper (approved):

PR #686, opened 2026-07-05
GUACAMOLE-2295: Check setsockopt and lseek return values.

necouchman:


src/protocols/rdp/fs.c
```diff
@@ -414,6 +414,7 @@ int guac_rdp_fs_open(guac_rdp_fs* fs, const char* path,
 int guac_rdp_fs_read(guac_rdp_fs* fs, int file_id, uint64_t offset,
         void* buffer, int length) {
 
+    int lseek_offset;
```
eugen-keeper: Not critical. **lseek** returns `off_t`, but the result is stored in int and checked with `< 0`. Successful seeks to offsets where `(offset mod 4 GiB) >= 2 GiB` may truncate to a negative int and are misreported as errors.

src/protocols/rdp/fs.c
```diff
@@ -424,7 +425,9 @@ int guac_rdp_fs_read(guac_rdp_fs* fs, int file_id, uint64_t offset,
     }
 
     /* Attempt read */
-    lseek(file->fd, offset, SEEK_SET);
+    lseek_offset = lseek(file->fd, offset, SEEK_SET);
+    if (lseek_offset < 0)
```
eugen-keeper, line 429: Again if `lseek_offset` is not **off_t**, this can be mistakenly true.

mike-jumper (changes requested):
Worth correcting the `off_t` truncation issue noted by @eugen-keeper - otherwise LGTM.

src/protocols/rdp/fs.c
necouchman: Yep, makes sense. Fixed via force push.

src/protocols/rdp/fs.c
necouchman: Fixed via force push.

mike-jumper (approved):

PR #687, opened 2026-07-06
GUACAMOLE-1370: guacenc output to stdout.

grintor:
Added a -o option to guacenc allowing the output path to be specified explicitly. It defaults to FILE.m4v as before when unspecified for backwards compatibility. Specifying "-" as the output path streams the encoded video to stdout. Normal MP4 (which is what's being used for file output still) doesn't support streaming, so we use fragmented MP4 here when outputting to stdout. This paves the way to streaming that stdout to a video player or even the browser in real time as it's being encoded.


src/guacenc/guacenc.c
```diff
@@ -106,15 +120,27 @@ int main(int argc, char* argv[]) {
...
+            int len = snprintf(out_buffer, sizeof(out_buffer), "%s.m4v", path);
+
+            /* Do not write if filename exceeds maximum length */
+            if (len >= sizeof(out_buffer)) {
+                guacenc_log(GUAC_LOG_ERROR, "Cannot write output file for "
+                        "\"%s\": Name too long", path);
+                continue;
+            }
+
+            out_path = out_buffer;
```
necouchman, line 142: What if the filename that the user specifies is too long? It looks like this logic only catches generated filenames that may be too long, and doesn't account for something being given by the user that is too long?

src/guacenc/video.c
```diff
@@ -54,8 +56,16 @@ guacenc_video* guacenc_video_alloc(const char* path, const char* codec_name,
     int ret;
     int failed_header = 0;
 
+    /* The container format cannot be guessed from a pipe URL and must be
+     * specified explicitly. The "ipod" container is used, matching the
+     * container that would be guessed from the ".m4v" extension of the output
+     * files normally produced */
+    bool is_pipe = (strncmp(path, "pipe:", 5) == 0);
+    const char* format_name = is_pipe ? "ipod" : NULL;
+
```
necouchman, line 65: Is this an appropriate assumption to make universally, or should this be configurable, as well? Are there situations where users may want to stream in a format different than iPod?

grintor: "ipod" was already an assumption being made universally before, just not explicitly. When writing to a file, the third argument to avformat_alloc_output_context2() is NULL so libavformat guesses the container from the extension of the output path. Since that was hardcoded to "FILE.m4v" it was always making the same guess based on ".m4v" -- 'ipod'. This is a pretty universally supported format, even in web browsers, so I think it's a good default. Maybe someone will want it to be configurable in the future, but we can just wait for that feature request.

src/guacenc/guacenc.c
grintor: Yea, this isn't handling the user input, but the user input isn't going into a buffer anywhere, it's being passed directly in to guacenc_encode(). There, if the filename is too long, the actual file-open call fails and guacenc logs it as a failure "Error occurred while opening output file".

This check is solely for the rare edge case where the default "FILE.m4v" ends up more than 4096 characters deep in a directory tree. But I didn't add that logic, it was in the original code [here](https://github.com/apache/guacamole-server/pull/687/changes/6f40cb5b1b42bf5394773142514ca5f0001b076e#diff-05daedc99f9bd28adae51e5f02797ec783ae34b53b9e758cf83473e7a0c9f55dL113-L117)

src/guacenc/video.c
necouchman: I agree that it was assumed by virtue of the fact that all files were necessarily `m4v` files. Now we're giving folks an option on what to name the files, which implicitly allows them to also adjust the format - and we're giving them an option to pipe everything to `stdout`, but forcing them to use the `ipod` format when going to `stdout`.

I'm not going to block this PR from going through if you don't want to do this, it's just something that seems to be in the same spirit as the ability to send it to `stdout`, and change the filename.

[It's also worth noting that simply having the ability to send to `stdout` means that you could also pipe it back into ffmpeg and reformat it that way, but you may start to lose the fidelity of the video that way due to lossy encodings, etc.]

necouchman:
@grintor You may need to rebase this on top of the current `main` branch - there were some issues with the main branch there for a while, and the build check is currently failing.

grintor:
Ok this latest commit closes GUACAMOLE-974 in addition to the GUACAMOLE-1370 the original pull request by adding your requested feature. I also noticed that my original pull request closes GUACAMOLE-973 as well.

grintor:
Ok the commits are on top of the current main branch now.

vlouvet:
I built b248d24b on Ubuntu 24.04 (FFmpeg 6.1.1, libvpx 1.14.0) and ran it against a small synthetic recording and a 1280x720 one. Everything below was checked with ffprobe:

| Command | Result |
|---|---|
| `guacenc -s 320x240 in.guac` | mpeg4 in `in.guac.m4v`, same as before |
| `-c libvpx-vp9 -F webm` | VP9 in `in.guac.webm` |
| `-c libvpx -F webm` | VP8 in `in.guac.webm` |
| `-c libaom-av1 -F webm` | AV1 in `in.guac.webm` |
| `-c libx264 -F mp4` | H.264 in `in.guac.mp4` |
| `-c libvpx-vp9 -o out.webm` | WebM, guessed from the extension |
| `-c libvpx-vp9 -F webm -o -` | valid WebM on stdout |
| `-o -` (defaults) | fragmented MP4 on stdout |
| `-F notaformat` | rejected before encoding, exit 1 |
| `-c not_a_codec` | "Failed to locate codec", no file written |
| second run into an existing file | refused, original left untouched |

One thing I ran into: `guacenc -c libvpx-vp9 in.guac` with no `-F` still names the output `in.guac.m4v`, so the ipod muxer is picked and the header write fails:

```
[ipod @ ...] Could not find tag for codec vp9 in stream #0, codec not currently supported in container
guacenc: ERROR: Error occurred while writing output file header.
```

The man page does say the codec has to fit the container, so I don't think this needs to hold up the PR. Once it's merged I'd be glad to send a small follow-up under GUACAMOLE-2298 that picks the default extension from the codec when neither `-F` nor `-o` is given (VP8/VP9/AV1 to `.webm`), with unit tests.

Separately, and the same in 1.6.0: guacenc exits 0 even when every file fails (bad codec, codec/container mismatch, refusing to overwrite). That matters a bit more now that output can go to a pipe. I can open an issue for it if it's worth changing.

Thanks for adding the format option. It covers most of what I filed GUACAMOLE-2298 for.

grintor:
Good call on the exit code @vlouvet
I added a fix here for it.

Sorry to duplicate your work on GUACAMOLE-2298 I hadn't seen that.

PR #690, opened 2026-07-13
GUACAMOLE-2136: Correct constness of results per GCC warnings.

mike-jumper:
Recent GCC fails with warnings regarding constness based on known behavior of `memchr()`, `bsearch()`, etc. regardless of the declared non-const return type.


necouchman:
@mike-jumper Did you intend this to go into the `patch` branch and not `staging/1.6.1`?

Nevermind...I see the original pull for this issue also went into `patch`.

necouchman (approved):

PR #692, opened 2026-07-13
GUACAMOLE-2300: Add IPMI Serial-over-LAN protocol support

ciroiriarte:
This adds a new `ipmi` protocol module to guacamole-server providing **Serial-over-LAN (SOL)** console access and **chassis power management** for BMCs (baseboard management controllers), built on FreeIPMI (`libipmiconsole` for SOL, `libfreeipmi` for chassis control).

JIRA: https://issues.apache.org/jira/browse/GUACAMOLE-2300

### What's included
- New `libguac-client-ipmi` protocol module — SOL console rendered through the shared `guac_terminal`.
- Chassis control over a client-agnostic `ipmi-control` pipe channel (newline-delimited JSON): power on/off/cycle/reset, soft shutdown, diagnostic interrupt (NMI), chassis identify, System Event Log (SEL) viewer, and live power/SOL status.
- In-terminal control menu (`Ctrl+]`) with an alternate-screen UI, run asynchronously so BMC round-trips never block the console.
- Multivendor support: configurable cipher suite / privilege level / workaround flags, hex `K_g`, encryption policy. Verified end-to-end against Supermicro (ATEN), Lenovo XCC, and Dell iDRAC 9.
- Connection parameters including persistent boot-device selection and keepalive interval.
- CUnit tests for the control-channel JSON parser, wired into `make check`.

### Testing
- 12/12 unit tests pass.
- AddressSanitizer clean on live SOL sessions. Two memory bugs were found and fixed during review: a NULL-deref on teardown, and a use-after-free that freed the terminal before joining the SOL worker.
- Full-stack live regression (browser → guacamole-client → guacd → BMC): SOL console I/O, live status push, chassis identify, SEL read, and Send Break all verified **non-destructively** against a Supermicro BMC.

### Related PRs
- Client: https://github.com/apache/guacamole-client/pull/1228
- Manual: https://github.com/apache/guacamole-manual/pull/298

### DIFF
--- a/src/protocols/ipmi/argv.c
+++ b/src/protocols/ipmi/argv.c
@@ -88,7 +88,8 @@ void guac_ipmi_send_current_argv_batch(
 
     /* Send current font size */
     char font_size[64];
-    sprintf(font_size, "%i", guac_terminal_get_font_size(terminal));
+    snprintf(font_size, sizeof(font_size), "%i",
+            guac_terminal_get_font_size(terminal));
     guac_client_stream_argv(client, socket, "text/plain",
             GUAC_IPMI_ARGV_FONT_SIZE, font_size);
 
--- a/src/protocols/ipmi/argv.c
+++ b/src/protocols/ipmi/argv.c
@@ -1,0 +1,95 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+#include "config.h"
+#include "argv.h"
+#include "ipmi.h"
+#include "terminal/terminal.h"
+
+#include <guacamole/protocol.h>
+#include <guacamole/socket.h>
+#include <guacamole/user.h>
+
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+
+int guac_ipmi_argv_callback(guac_user* user, const char* mimetype,
+        const char* name, const char* value, void* data) {
+
+    guac_client* client = user->client;
+    guac_ipmi_client* ipmi_client = (guac_ipmi_client*) client->data;
+    guac_terminal* terminal = ipmi_client->term;
+
+    /* Skip if terminal not yet ready */
+    if (terminal == NULL)
+        return 0;
+
+    /* Update color scheme */
+    if (strcmp(name, GUAC_IPMI_ARGV_COLOR_SCHEME) == 0)
+        guac_terminal_apply_color_scheme(terminal, value);
+
+    /* Update font name */
+    else if (strcmp(name, GUAC_IPMI_ARGV_FONT_NAME) == 0)
+        guac_terminal_apply_font(terminal, value, -1, 0);
+
+    /* Update only if font size is sane */
+    else if (strcmp(name, GUAC_IPMI_ARGV_FONT_SIZE) == 0) {
+        int size = atoi(value);
+        if (size > 0)
+            guac_terminal_apply_font(terminal, NULL, size,
+                    ipmi_client->settings->resolution);
+    }
+
+    return 0;
+
+}
+
+void* guac_ipmi_send_current_argv(guac_user* user, void* data) {
+
+    /* Defer to the batch handler, using the user's socket to send the data */
+    guac_ipmi_send_current_argv_batch(user->client, user->socket);
+
+    return NULL;
+
+}
+
+void guac_ipmi_send_current_argv_batch(
+        guac_client* client, guac_socket* socket) {
+
+    guac_ipmi_client* ipmi_client = (guac_ipmi_client*) client->data;
+    guac_terminal* terminal = ipmi_client->term;
+
+    /* Send current color scheme */
+    guac_client_stream_argv(client, socket, "text/plain",
+            GUAC_IPMI_ARGV_COLOR_SCHEME,
+            guac_terminal_get_color_scheme(terminal));
+
+    /* Send current font name */
+    guac_client_stream_argv(client, socket, "text/plain",
+            GUAC_IPMI_ARGV_FONT_NAME,
+            guac_terminal_get_font_name(terminal));
+
+    /* Send current font size */
+    char font_size[64];
+    sprintf(font_size, "%i", guac_terminal_get_font_size(terminal));
+    guac_client_stream_argv(client, socket, "text/plain",
+            GUAC_IPMI_ARGV_FONT_SIZE, font_size);
+
+}
--- a/src/protocols/ipmi/argv.c
+++ b/src/protocols/ipmi/argv.c
@@ -17,7 +17,6 @@
  * under the License.
  */
 
-#include "config.h"
 #include "argv.h"
 #include "ipmi.h"
 #include "terminal/terminal.h"
--- a/src/protocols/ipmi/chassis.c
+++ b/src/protocols/ipmi/chassis.c
@@ -368,3 +368,126 @@ int guac_ipmi_chassis_identify(guac_client* client, int interval, bool force) {
     return result;
 
 }
+
+/**
+ * The maximum number of System Event Log entries retained by the viewer (the
+ * most recent entries).
+ */
+#define GUAC_IPMI_SEL_MAX_ENTRIES 20
+
+/**
+ * The maximum length, in bytes, of a single formatted SEL entry line.
+ */
+#define GUAC_IPMI_SEL_LINE_LENGTH 160
+
+/**
+ * State accumulated while parsing the System Event Log. Formatted entries are
+ * stored in a ring buffer such that, after parsing, the buffer holds the most
+ * recent GUAC_IPMI_SEL_MAX_ENTRIES entries.
+ */
+typedef struct guac_ipmi_sel_state {
+
+    /**
+     * Ring buffer of formatted SEL entry lines.
+     */
+    char lines[GUAC_IPMI_SEL_MAX_ENTRIES][GUAC_IPMI_SEL_LINE_LENGTH];
+
+    /**
+     * The total number of SEL records parsed so far.
+     */
+    int total;
+
+} guac_ipmi_sel_state;
+
+/**
+ * libfreeipmi SEL parse callback. Formats the current SEL record into the ring
+ * buffer maintained by the given guac_ipmi_sel_state.
+ *
+ * @param sel_ctx
+ *     The SEL parse context, positioned at the record to format.
+ *
+ * @param data
+ *     A pointer to the guac_ipmi_sel_state accumulating formatted entries.
+ *
+ * @return
+ *     Zero to continue parsing.
+ */
+static int guac_ipmi_sel_callback(ipmi_sel_ctx_t sel_ctx, void* data) {
+
+    guac_ipmi_sel_state* state = (guac_ipmi_sel_state*) data;
+    char* slot = state->lines[state->total % GUAC_IPMI_SEL_MAX_ENTRIES];
+
+    /* Format "<id>  <date> <time>  <sensor>  <event>" using the current record.
+     * A NULL record with zero length refers to the record the callback is
+     * positioned on. */
+    if (ipmi_sel_parse_read_record_string(sel_ctx, "%i  %d %t  %s  %e",
+                NULL, 0, slot, GUAC_IPMI_SEL_LINE_LENGTH,
+                IPMI_SEL_STRING_FLAGS_DEFAULT) < 0)
+        slot[0] = '\0';
+
+    state->total++;
+    return 0;
+
+}
+
+int guac_ipmi_chassis_sel(guac_client* client, char* buffer, int size) {
+
+    ipmi_ctx_t ctx = guac_ipmi_chassis_connect(client);
+    if (ctx == NULL)
+        return 1;
+
+    int result = 1;
+
+    /* A NULL SDR context yields slightly less descriptive sensor names but
+     * avoids the cost and complexity of caching the SDR repository. */
+    ipmi_sel_ctx_t sel_ctx = ipmi_sel_ctx_create(ctx, NULL);
+    if (sel_ctx != NULL) {
+
+        guac_ipmi_sel_state state;
+        state.total = 0;
+
+        if (ipmi_sel_parse(sel_ctx, IPMI_SEL_RECORD_ID_FIRST,
+                    IPMI_SEL_RECORD_ID_LAST, guac_ipmi_sel_callback,
+                    &state) >= 0) {
+
+            int offset = 0;
+
+            if (state.total == 0)
+                snprintf(buffer, size, "System Event Log is empty.\r\n");
+
+            else {
+
+                /* Emit only the most recent entries retained in the ring */
+                int shown = state.total < GUAC_IPMI_SEL_MAX_ENTRIES
+                        ? state.total : GUAC_IPMI_SEL_MAX_ENTRIES;
+                int start = state.total - shown;
+
+                offset += snprintf(buffer + offset, size - offset,
+                        "Last %i of %i System Event Log entries:\r\n",
+                        shown, state.total);
+
+                for (int i = start; i < state.total && offset < size - 1; i++)
+                    offset += snprintf(buffer + offset, size - offset, "%s\r\n",
+                            state.lines[i % GUAC_IPMI_SEL_MAX_ENTRIES]);
+
+            }
+
+            result = 0;
+        }
+
+        else
+            guac_client_log(client, GUAC_LOG_ERROR,
+                    "Unable to read System Event Log: %s",
+                    ipmi_sel_ctx_errormsg(sel_ctx));
+
+        ipmi_sel_ctx_destroy(sel_ctx);
+    }
+
+    else
+        guac_client_log(client, GUAC_LOG_ERROR,
+                "Unable to create System Event Log context.");
+
+    guac_ipmi_chassis_disconnect(ctx);
+    return result;
+
+}
--- a/src/protocols/ipmi/chassis.c
+++ b/src/protocols/ipmi/chassis.c
@@ -281,7 +281,11 @@ int guac_ipmi_chassis_set_boot_device(guac_client* client,
 
 }
 
-int guac_ipmi_chassis_status(guac_client* client, char* buffer, int size) {
+int guac_ipmi_chassis_status(guac_client* client, char* buffer, int size,
+        guac_ipmi_power_state* power) {
+
+    if (power != NULL)
+        *power = GUAC_IPMI_POWER_STATE_UNKNOWN;
 
     ipmi_ctx_t ctx = guac_ipmi_chassis_connect(client);
     if (ctx == NULL)
@@ -323,6 +327,11 @@ int guac_ipmi_chassis_status(guac_client* client, char* buffer, int size) {
                         power_fault ? " (fault)" : "",
                         power_overload ? " (overload)" : "");
 
+                if (power != NULL)
+                    *power = power_is_on
+                            ? GUAC_IPMI_POWER_STATE_ON
+                            : GUAC_IPMI_POWER_STATE_OFF;
+
                 result = 0;
             }
         }
@@ -335,7 +344,7 @@ int guac_ipmi_chassis_status(guac_client* client, char* buffer, int size) {
 
 }
 
-int guac_ipmi_chassis_identify(guac_client* client, int interval, bool force) {
+int guac_ipmi_chassis_identify(guac_client* client, int interval) {
 
     ipmi_ctx_t ctx = guac_ipmi_chassis_connect(client);
     if (ctx == NULL)
@@ -346,14 +355,9 @@ int guac_ipmi_chassis_identify(guac_client* client, int interval, bool force) {
     if (obj_cmd_rs != NULL) {
 
         uint8_t interval_value = (uint8_t) interval;
-        uint8_t force_value = force
-            ? IPMI_CHASSIS_FORCE_IDENTIFY_ON
-            : IPMI_CHASSIS_FORCE_IDENTIFY_OFF;
+        uint8_t force_value = IPMI_CHASSIS_FORCE_IDENTIFY_OFF;
 
-        /* When forcing the LED on indefinitely, the interval is not meaningful
-         * and is passed as NULL. */
-        if (ipmi_cmd_chassis_identify(ctx,
-                    force ? NULL : &interval_value,
+        if (ipmi_cmd_chassis_identify(ctx, &interval_value,
                     &force_value, obj_cmd_rs) < 0)
             guac_client_log(client, GUAC_LOG_ERROR,
                     "Chassis identify command failed: %s",
--- a/src/protocols/ipmi/chassis.c
+++ b/src/protocols/ipmi/chassis.c
@@ -17,8 +17,6 @@
  * under the License.
  */
 
-#include "config.h"
-
 #include "chassis.h"
 #include "ipmi.h"
 #include "settings.h"
@@ -47,6 +45,36 @@
  * command sessions.
  */
 #define GUAC_IPMI_CHASSIS_RETRANSMISSION_TIMEOUT 1000
+
+/**
+ * The maximum number of System Event Log entries retained by the viewer (the
+ * most recent entries).
+ */
+#define GUAC_IPMI_SEL_MAX_ENTRIES 20
+
+/**
+ * The maximum length, in bytes, of a single formatted SEL entry line.
+ */
+#define GUAC_IPMI_SEL_LINE_LENGTH 160
+
+/**
+ * State accumulated while parsing the System Event Log. Formatted entries are
+ * stored in a ring buffer such that, after parsing, the buffer holds the most
+ * recent GUAC_IPMI_SEL_MAX_ENTRIES entries.
+ */
+typedef struct guac_ipmi_sel_state {
+
+    /**
+     * Ring buffer of formatted SEL entry lines.
+     */
+    char lines[GUAC_IPMI_SEL_MAX_ENTRIES][GUAC_IPMI_SEL_LINE_LENGTH];
+
+    /**
+     * The total number of SEL records parsed so far.
+     */
+    int total;
+
+} guac_ipmi_sel_state;
 
 /**
  * Maps a guac_ipmi_privilege_level to the corresponding libfreeipmi
@@ -281,17 +309,14 @@ int guac_ipmi_chassis_set_boot_device(guac_client* client,
 
 }
 
-int guac_ipmi_chassis_status(guac_client* client, char* buffer, int size,
-        guac_ipmi_power_state* power) {
-
-    if (power != NULL)
-        *power = GUAC_IPMI_POWER_STATE_UNKNOWN;
+guac_ipmi_power_state guac_ipmi_chassis_status(guac_client* client,
+        char* buffer, int size) {
 
     ipmi_ctx_t ctx = guac_ipmi_chassis_connect(client);
     if (ctx == NULL)
-        return 1;
+        return GUAC_IPMI_POWER_STATE_UNKNOWN;
 
-    int result = 1;
+    guac_ipmi_power_state power = GUAC_IPMI_POWER_STATE_UNKNOWN;
     fiid_obj_t obj_cmd_rs = fiid_obj_create(tmpl_cmd_get_chassis_status_rs);
     if (obj_cmd_rs != NULL) {
 
@@ -327,12 +352,9 @@ int guac_ipmi_chassis_status(guac_client* client, char* buffer, int size,
                         power_fault ? " (fault)" : "",
                         power_overload ? " (overload)" : "");
 
-                if (power != NULL)
-                    *power = power_is_on
-                            ? GUAC_IPMI_POWER_STATE_ON
-                            : GUAC_IPMI_POWER_STATE_OFF;
-
-                result = 0;
+                power = power_is_on
+                        ? GUAC_IPMI_POWER_STATE_ON
+                        : GUAC_IPMI_POWER_STATE_OFF;
             }
         }
 
@@ -340,7 +362,7 @@ int guac_ipmi_chassis_status(guac_client* client, char* buffer, int size,
     }
 
     guac_ipmi_chassis_disconnect(ctx);
-    return result;
+    return power;
 
 }
 
@@ -372,36 +394,6 @@ int guac_ipmi_chassis_identify(guac_client* client, int interval) {
     return result;
 
 }
-
-/**
- * The maximum number of System Event Log entries retained by the viewer (the
- * most recent entries).
- */
-#define GUAC_IPMI_SEL_MAX_ENTRIES 20
-
-/**
- * The maximum length, in bytes, of a single formatted SEL entry line.
- */
-#define GUAC_IPMI_SEL_LINE_LENGTH 160
-
-/**
- * State accumulated while parsing the System Event Log. Formatted entries are
- * stored in a ring buffer such that, after parsing, the buffer holds the most
- * recent GUAC_IPMI_SEL_MAX_ENTRIES entries.
- */
-typedef struct guac_ipmi_sel_state {
-
-    /**
-     * Ring buffer of formatted SEL entry lines.
-     */
-    char lines[GUAC_IPMI_SEL_MAX_ENTRIES][GUAC_IPMI_SEL_LINE_LENGTH];
-
-    /**
-     * The total number of SEL records parsed so far.
-     */
-    int total;
-
-} guac_ipmi_sel_state;
 
 /**
  * libfreeipmi SEL parse callback. Formats the current SEL record into the ring
--- a/src/protocols/ipmi/chassis.c
+++ b/src/protocols/ipmi/chassis.c
@@ -1,0 +1,370 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+#include "config.h"
+
+#include "chassis.h"
+#include "ipmi.h"
+#include "settings.h"
+
+#include <guacamole/client.h>
+
+#include <freeipmi/freeipmi.h>
+
+#include <stdbool.h>
+#include <stdio.h>
+#include <string.h>
+
+/**
+ * The IPMI session timeout, in milliseconds, used for the short-lived
+ * out-of-band sessions opened to issue chassis commands. Chassis commands are
+ * simple request/response operations, so a bounded timeout (rather than the
+ * library default) is used to limit how long a chassis command can stall when
+ * the BMC is slow or unreachable. This matters because chassis commands
+ * triggered from the in-terminal control menu run synchronously within the
+ * user input handler.
+ */
+#define GUAC_IPMI_CHASSIS_SESSION_TIMEOUT 8000
+
+/**
+ * The IPMI packet retransmission timeout, in milliseconds, used for chassis
+ * command sessions.
+ */
+#define GUAC_IPMI_CHASSIS_RETRANSMISSION_TIMEOUT 1000
+
+/**
+ * Maps a guac_ipmi_privilege_level to the corresponding libfreeipmi
+ * IPMI_PRIVILEGE_LEVEL_* constant.
+ *
+ * @param privilege_level
+ *     The privilege level to map.
+ *
+ * @return
+ *     The equivalent IPMI_PRIVILEGE_LEVEL_* constant.
+ */
+static uint8_t guac_ipmi_map_privilege_level(
+        guac_ipmi_privilege_level privilege_level) {
+
+    switch (privilege_level) {
+        case GUAC_IPMI_PRIVILEGE_USER:
+            return IPMI_PRIVILEGE_LEVEL_USER;
+        case GUAC_IPMI_PRIVILEGE_OPERATOR:
+            return IPMI_PRIVILEGE_LEVEL_OPERATOR;
+        case GUAC_IPMI_PRIVILEGE_ADMIN:
+        default:
+            return IPMI_PRIVILEGE_LEVEL_ADMIN;
+    }
+
+}
+
+/**
+ * Maps a guac_ipmi_power_action to the corresponding libfreeipmi
+ * IPMI_CHASSIS_CONTROL_* constant.
+ *
+ * @param action
+ *     The power action to map. Must not be GUAC_IPMI_POWER_NONE.
+ *
+ * @return
+ *     The equivalent IPMI_CHASSIS_CONTROL_* constant.
+ */
+static uint8_t guac_ipmi_map_power_action(guac_ipmi_power_action action) {
+
+    switch (action) {
+        case GUAC_IPMI_POWER_ON:
+            return IPMI_CHASSIS_CONTROL_POWER_UP;
+        case GUAC_IPMI_POWER_OFF:
+            return IPMI_CHASSIS_CONTROL_POWER_DOWN;
+        case GUAC_IPMI_POWER_CYCLE:
+            return IPMI_CHASSIS_CONTROL_POWER_CYCLE;
+        case GUAC_IPMI_POWER_RESET:
+            return IPMI_CHASSIS_CONTROL_HARD_RESET;
+        case GUAC_IPMI_POWER_SOFT_SHUTDOWN:
+            return IPMI_CHASSIS_CONTROL_INITIATE_SOFT_SHUTDOWN;
+        case GUAC_IPMI_POWER_DIAGNOSTIC_INTERRUPT:
+            return IPMI_CHASSIS_CONTROL_PULSE_DIAGNOSTIC_INTERRUPT;
+        default:
+            return IPMI_CHASSIS_CONTROL_POWER_DOWN;
+    }
+
+}
+
+/**
+ * Maps a guac_ipmi_boot_device to the corresponding libfreeipmi
+ * IPMI_SYSTEM_BOOT_OPTION_BOOT_FLAG_BOOT_DEVICE_FORCE_* constant.
+ *
+ * @param device
+ *     The boot device to map. Must not be GUAC_IPMI_BOOT_NONE.
+ *
+ * @return
+ *     The equivalent boot device constant.
+ */
+static uint8_t guac_ipmi_map_boot_device(guac_ipmi_boot_device device) {
+
+    switch (device) {
+        case GUAC_IPMI_BOOT_PXE:
+            return IPMI_SYSTEM_BOOT_OPTION_BOOT_FLAG_BOOT_DEVICE_FORCE_PXE;
+        case GUAC_IPMI_BOOT_DISK:
+            return IPMI_SYSTEM_BOOT_OPTION_BOOT_FLAG_BOOT_DEVICE_FORCE_HARD_DRIVE;
+        case GUAC_IPMI_BOOT_CDROM:
+            return IPMI_SYSTEM_BOOT_OPTION_BOOT_FLAG_BOOT_DEVICE_FORCE_CD_DVD;
+        case GUAC_IPMI_BOOT_BIOS:
+            return IPMI_SYSTEM_BOOT_OPTION_BOOT_FLAG_BOOT_DEVICE_FORCE_BIOS_SETUP;
+        default:
+            return IPMI_SYSTEM_BOOT_OPTION_BOOT_FLAG_BOOT_DEVICE_NO_OVERRIDE;
+    }
+
+}
+
+/**
+ * Opens a short-lived IPMI 2.0 out-of-band session against the BMC described
+ * by the given client's settings. The returned context must be closed with
+ * ipmi_ctx_close() and freed with ipmi_ctx_destroy() when no longer needed.
+ *
+ * @param client
+ *     The guac_client whose settings describe the BMC to connect to.
+ *
+ * @return
+ *     A connected ipmi_ctx_t on success, or NULL on failure (with an error
+ *     logged on behalf of the client).
+ */
+static ipmi_ctx_t guac_ipmi_chassis_connect(guac_client* client) {
+
+    guac_ipmi_client* ipmi_client = (guac_ipmi_client*) client->data;
+    guac_ipmi_settings* settings = ipmi_client->settings;
+
+    ipmi_ctx_t ctx = ipmi_ctx_create();
+    if (ctx == NULL) {
+        guac_client_log(client, GUAC_LOG_ERROR,
+                "Unable to allocate IPMI context for chassis command.");
+        return NULL;
+    }
+
+    /* The BMC key (K_g) is optional; its length is required separately as it
+     * may legitimately contain null bytes. */
+    const unsigned char* k_g = (const unsigned char*) settings->k_g;
+    unsigned int k_g_len = settings->k_g != NULL ? strlen(settings->k_g) : 0;
+
+    /* Open an IPMI 2.0 (RMCP+) out-of-band session. Session and retransmission
+     * timeouts of 0 request the library defaults. */
+    if (ipmi_ctx_open_outofband_2_0(ctx,
+                settings->hostname,
+                settings->username,
+                settings->password,
+                k_g, k_g_len,
+                guac_ipmi_map_privilege_level(settings->privilege_level),
+                (uint8_t) settings->cipher_suite,
+                GUAC_IPMI_CHASSIS_SESSION_TIMEOUT,
+                GUAC_IPMI_CHASSIS_RETRANSMISSION_TIMEOUT,
+                0 /* workaround flags */,
+                0 /* flags */) < 0) {
+        guac_client_log(client, GUAC_LOG_ERROR,
+                "Unable to open IPMI session for chassis command: %s",
+                ipmi_ctx_errormsg(ctx));
+        ipmi_ctx_destroy(ctx);
+        return NULL;
+    }
+
+    return ctx;
+
+}
+
+/**
+ * Closes and destroys the given IPMI context.
+ *
+ * @param ctx
+ *     The context to close and destroy.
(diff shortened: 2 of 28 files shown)


src/protocols/ipmi/tests/control/json_get.c
```diff
@@ -0,0 +1,96 @@
...
+
+/**
+ * Verifies that a non-string value (e.g. a number) is not matched, as only
+ * quoted string values are extracted.
+ */
+void test_json_get__non_string_value(void) {
+    char out[64];
+    CU_ASSERT_FALSE(guac_ipmi_control_json_get(
+            "{\"n\":123}", "n", out, sizeof(out)));
+}
```
necouchman, line 143: Why are only string values allowed? My experience in general with JSON is that string values are expected for strings, but other values - boolean, numeric, etc. - are perfectly valid JSON?

src/protocols/ipmi/argv.c
```diff
@@ -0,0 +1,96 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+#include "config.h"
```
necouchman: Mike recently submitted a PR that makes including `config.h` unnecessary - see #691.

src/protocols/ipmi/chassis.c
```diff
@@ -0,0 +1,497 @@
...
+        case GUAC_IPMI_PRIVILEGE_USER:
+            return IPMI_PRIVILEGE_LEVEL_USER;
+        case GUAC_IPMI_PRIVILEGE_OPERATOR:
+            return IPMI_PRIVILEGE_LEVEL_OPERATOR;
+        case GUAC_IPMI_PRIVILEGE_ADMIN:
+        default:
+            return IPMI_PRIVILEGE_LEVEL_ADMIN;
+    }
+
+}
```
necouchman, line 102: Is there a particular reason we need to define new values for these privilege levels as `GUAC_` constants and shouldn't just re-use the `IPMI_PRIVILEGE_LEVEL_` values throughout the code?

src/protocols/ipmi/chassis.c
```diff
@@ -0,0 +1,497 @@
...
+            return IPMI_CHASSIS_CONTROL_HARD_RESET;
+        case GUAC_IPMI_POWER_SOFT_SHUTDOWN:
+            return IPMI_CHASSIS_CONTROL_INITIATE_SOFT_SHUTDOWN;
+        case GUAC_IPMI_POWER_DIAGNOSTIC_INTERRUPT:
+            return IPMI_CHASSIS_CONTROL_PULSE_DIAGNOSTIC_INTERRUPT;
+        default:
+            return IPMI_CHASSIS_CONTROL_POWER_DOWN;
+    }
+
+}
```
necouchman, line 133: Similar to the privilege levels - is there a particular reason that we need to create and map them rather than just using the ones provided by FreeIPMI?

src/protocols/ipmi/chassis.c
```diff
@@ -0,0 +1,497 @@
...
+            return IPMI_SYSTEM_BOOT_OPTION_BOOT_FLAG_BOOT_DEVICE_FORCE_HARD_DRIVE;
+        case GUAC_IPMI_BOOT_CDROM:
+            return IPMI_SYSTEM_BOOT_OPTION_BOOT_FLAG_BOOT_DEVICE_FORCE_CD_DVD;
+        case GUAC_IPMI_BOOT_BIOS:
+            return IPMI_SYSTEM_BOOT_OPTION_BOOT_FLAG_BOOT_DEVICE_FORCE_BIOS_SETUP;
+        default:
+            return IPMI_SYSTEM_BOOT_OPTION_BOOT_FLAG_BOOT_DEVICE_NO_OVERRIDE;
+    }
+
+}
```
necouchman, line 160: And, similar to above two - why do we need to redefine these?

src/protocols/ipmi/chassis.c
```diff
@@ -0,0 +1,497 @@
...
+    guac_ipmi_chassis_disconnect(ctx);
+    return result;
+
+}
+
+int guac_ipmi_chassis_status(guac_client* client, char* buffer, int size,
+        guac_ipmi_power_state* power) {
+
+    if (power != NULL)
+        *power = GUAC_IPMI_POWER_STATE_UNKNOWN;
```
necouchman: Based on the rest of the code in this function, I'm not sure I understand why this is being checked to be not null and forced to the `UNKNOWN` state?
* It looks like the only other time that the `power` variable is referenced is when it is set below in lines 330-334?
* If it is `NULL`, it will not be set to `UNKNOWN` here, nor will the actual detected value be set in lines 330-334?
* If it is not `NULL`, this `UNKNOWN` state is set, but that value is never used in this function?

Wouldn't it be better to check to see if it _is_ `NULL` and allocate it to `UNKNOWN` if it is?

src/protocols/ipmi/chassis.c
```diff
@@ -0,0 +1,497 @@
...
+        fiid_obj_destroy(obj_cmd_rs);
+    }
+
+    guac_ipmi_chassis_disconnect(ctx);
+    return result;
+
+}
+
+int guac_ipmi_chassis_status(guac_client* client, char* buffer, int size,
+        guac_ipmi_power_state* power) {
```
necouchman: Why not have this function return the `guac_ipmi_power_state` value rather than returning a result value?

src/protocols/ipmi/chassis.c
```diff
@@ -0,0 +1,497 @@
...
+
+                snprintf(buffer, size, "Power: %s%s%s",
+                        power_is_on ? "ON" : "OFF",
+                        power_fault ? " (fault)" : "",
+                        power_overload ? " (overload)" : "");
+
+                if (power != NULL)
+                    *power = power_is_on
+                            ? GUAC_IPMI_POWER_STATE_ON
+                            : GUAC_IPMI_POWER_STATE_OFF;
```
necouchman: Are `ON` and `OFF` really the only valid IPMI power states?

src/protocols/ipmi/chassis.c
```diff
@@ -0,0 +1,497 @@
...
+     * Ring buffer of formatted SEL entry lines.
+     */
+    char lines[GUAC_IPMI_SEL_MAX_ENTRIES][GUAC_IPMI_SEL_LINE_LENGTH];
+
+    /**
+     * The total number of SEL records parsed so far.
+     */
+    int total;
+
+} guac_ipmi_sel_state;
```
necouchman: I believe the established style in the code is to have all of the `#define` and `typedef` statements/definitions at the top of the file.

src/protocols/ipmi/control.c
```diff
@@ -0,0 +1,428 @@
...
+    int length;
+
+} guac_ipmi_control_stream;
+
+/**
+ * Escapes the given string into the body of a JSON string (without the
+ * surrounding quotes), writing at most dstlen-1 bytes plus a NUL terminator.
+ */
+void guac_ipmi_control_json_escape(char* dst, int dstlen,
+        const char* src) {
```
necouchman: It looks like this function is documented in the `control.h` file, so no need to document it, here.

src/protocols/ipmi/control.c
```diff
@@ -0,0 +1,428 @@
...
+    dst[j] = '\0';
+
+}
+
+/**
+ * Extracts the string value of the given key from a flat JSON object into out.
+ * Returns non-zero if the key was found.
+ */
+int guac_ipmi_control_json_get(const char* json, const char* key,
+        char* out, int outlen) {
```
necouchman: I believe this function is documented in `control.h`, so need to document it here.

src/protocols/ipmi/control.c
```diff
@@ -0,0 +1,428 @@
...
+
+    return 0;
+
+}
+
+/**
+ * Sends one newline-delimited JSON message to the given user on a fresh
+ * outbound "ipmi-control" pipe stream.
+ */
+static void guac_ipmi_control_send(guac_user* user, const char* json) {
```
necouchman, line 167: The parameters for this function need to be documented.

src/protocols/ipmi/control.c
```diff
@@ -0,0 +1,428 @@
...
+/**
+ * Pushes a state message to the given user. If query is true, the current
+ * power state is actively read from the BMC (which may take several seconds)
+ * and included in the message; otherwise only the SOL health is sent, leaving
+ * the client's last-known power value untouched. This lets health-only updates
+ * (initial state, and the broadcast when SOL connects/disconnects) refresh the
+ * connection badge without regressing the displayed power state to "unknown".
+ */
+static void guac_ipmi_control_send_state(guac_user* user, guac_client* client,
+        bool query) {
```
necouchman, line 198: Please document the function parameters.

src/protocols/ipmi/control.c
```diff
@@ -0,0 +1,428 @@
...
+ */
+static void guac_ipmi_control_send(guac_user* user, const char* json) {
+
+    guac_stream* stream = guac_user_alloc_stream(user);
+    guac_protocol_send_pipe(user->socket, stream, "application/json",
+            GUAC_IPMI_CONTROL_PIPE_NAME);
+    guac_protocol_send_blob(user->socket, stream, json, strlen(json));
+    guac_protocol_send_end(user->socket, stream);
+    guac_socket_flush(user->socket);
+    guac_user_free_stream(user, stream);
```
necouchman, line 175: Almost all of these functions return some sort of value to indicate success or failure. Maybe the return values should be checked?

src/protocols/ipmi/control.c
```diff
@@ -0,0 +1,428 @@
...
+                power, health);
+
+    }
+    else
+        snprintf(json, sizeof(json),
+                "{\"type\":\"state\",\"health\":\"%s\"}", health);
+
+    guac_ipmi_control_send(user, json);
+
+}
```
necouchman, line 238: It would be good to add some comments to document the flow of this function.

src/protocols/ipmi/control.c
```diff
@@ -0,0 +1,428 @@
...
+
+static void* guac_ipmi_control_broadcast_state_callback(guac_user* user,
+        void* data) {
```
necouchman, line 254: This function needs to be documented.

src/protocols/ipmi/control.c
```diff
@@ -0,0 +1,428 @@
...
+    guac_client_foreach_user(client,
+            guac_ipmi_control_broadcast_state_callback, client);
+
+}
+
+/**
+ * Pushes a command result message to the given user.
+ */
+static void guac_ipmi_control_send_result(guac_user* user, const char* id,
+        bool ok, const char* message) {
```
necouchman, line 284: Please document function parameters.

src/protocols/ipmi/control.c
```diff
@@ -0,0 +1,428 @@
...
+    guac_ipmi_control_send(user, json);
+
+}
+
+/**
+ * Reads the System Event Log and pushes it to the given user. The entries are
+ * currently delivered as a preformatted text field; structured per-entry
+ * delivery is a planned refinement.
+ */
+static void guac_ipmi_control_send_sel(guac_user* user, guac_client* client) {
```
necouchman, line 308: Please complete function parameter documentation.

src/protocols/ipmi/control.c
```diff
@@ -0,0 +1,428 @@
...
+    }
+    else
+        snprintf(json, GUAC_IPMI_CONTROL_SEL_LENGTH * 2,
+                "{\"type\":\"sel\",\"error\":\"Unable to read System Event Log.\"}");
+
+    guac_ipmi_control_send(user, json);
+    guac_mem_free(sel);
+    guac_mem_free(json);
+
+}
```
necouchman, line 332: A couple of inline comments might be good, here, to document code flow.

src/protocols/ipmi/control.c
```diff
@@ -0,0 +1,428 @@
...
+
+/**
+ * Maps a command string to a chassis power action, or GUAC_IPMI_POWER_NONE if
+ * the command is not a power action.
+ */
+static guac_ipmi_power_action guac_ipmi_control_power_action(const char* cmd) {
```
necouchman, line 345: Please complete function documentation - parameters and return value.

src/protocols/ipmi/control.c
```diff
@@ -0,0 +1,428 @@
...
+    if (strcmp(cmd, "diagnostic-interrupt") == 0) return GUAC_IPMI_POWER_DIAGNOSTIC_INTERRUPT;
+    return GUAC_IPMI_POWER_NONE;
+}
+
+/**
+ * Attempts to acquire the shared chassis operation lock, so control-channel
+ * operations and the in-terminal menu never run overlapping BMC sessions.
+ * Returns non-zero if the lock was acquired.
+ */
+static int guac_ipmi_control_try_acquire(guac_ipmi_client* ipmi_client) {
```
necouchman, line 366: Please document function parameters and return value.

src/protocols/ipmi/control.c
```diff
@@ -0,0 +1,428 @@
...
+    pthread_mutex_lock(&ipmi_client->menu_lock);
+    if (!ipmi_client->chassis_busy) {
+        ipmi_client->chassis_busy = true;
+        acquired = 1;
+    }
+    pthread_mutex_unlock(&ipmi_client->menu_lock);
+    return acquired;
+}
+
+static void guac_ipmi_control_release(guac_ipmi_client* ipmi_client) {
```
necouchman, line 384: Please document the function.

src/protocols/ipmi/control.c
```diff
@@ -0,0 +1,428 @@
...
+    pthread_mutex_lock(&ipmi_client->menu_lock);
+    ipmi_client->chassis_busy = false;
+    pthread_mutex_unlock(&ipmi_client->menu_lock);
+}
+
+/**
+ * Parses and executes a single inbound control command line.
+ */
+static void guac_ipmi_control_dispatch(guac_user* user, const char* line) {
```
necouchman, line 399: Please document function parameters.

src/protocols/ipmi/control.c
```diff
@@ -0,0 +1,428 @@
...
+    else {
+        guac_ipmi_power_action action = guac_ipmi_control_power_action(command);
+        if (action == GUAC_IPMI_POWER_NONE)
+            guac_ipmi_control_send_result(user, id, false, "Unknown command.");
+        else {
+            int ok = guac_ipmi_chassis_power(client, action) == 0;
+            guac_ipmi_control_send_result(user, id, ok,
+                    ok ? "Command sent successfully." : "Command failed.");
+        }
+    }
```
necouchman: I'm not sure that assuming that, if we've passed all other command checks, the user must want to do a power action, is a good assumption, and that automatically tring that command out with the power action function is a good idea.

I think it'd be better to have another `else if` to check to see if the command is a power command, and then a final `else` block that handles situations where the user has not provided a valid command. Keep in mind that, while this project develops both Client and Server side components, there are many out there who use guacd with their own client implementations, so we shouldn't assume that it will be used correctly.

src/protocols/ipmi/control.c
```diff
@@ -0,0 +1,428 @@
...
+    guac_ipmi_control_release(ipmi_client);
+
+}
+
+/**
+ * Blob handler for the inbound control pipe: reassembles newline-delimited JSON
+ * and dispatches each complete line.
+ */
+static int guac_ipmi_control_blob_handler(guac_user* user, guac_stream* stream,
+        void* data, int length) {
```
necouchman, line 488: Please document parameters and return value.

src/protocols/ipmi/control.c
```diff
@@ -0,0 +1,428 @@
...
+            /* Oversized line; drop it to resynchronize on the next newline */
+            cs->length = 0;
+    }
+
+    guac_protocol_send_ack(user->socket, stream, "OK",
+            GUAC_PROTOCOL_STATUS_SUCCESS);
+    guac_socket_flush(user->socket);
+    return 0;
+
+}
```
necouchman, line 515: A little more inline commentary would be helpful, here.

src/protocols/ipmi/control.c
```diff
@@ -0,0 +1,428 @@
...
+
+static int guac_ipmi_control_end_handler(guac_user* user, guac_stream* stream) {
```
necouchman, line 530: Please document the function.

src/protocols/ipmi/control.c
```diff
@@ -0,0 +1,428 @@
...
+    guac_ipmi_control_stream* cs =
+        guac_mem_zalloc(sizeof(guac_ipmi_control_stream));
+
+    stream->data = cs;
+    stream->blob_handler = guac_ipmi_control_blob_handler;
+    stream->end_handler = guac_ipmi_control_end_handler;
+
+    guac_protocol_send_ack(user->socket, stream, "OK",
+            GUAC_PROTOCOL_STATUS_SUCCESS);
+    guac_socket_flush(user->socket);
```
necouchman, line 547: I believe both of these functions have return values - should they really not be checked?

src/protocols/ipmi/settings.c
```diff
@@ -0,0 +1,910 @@
...
+
+    /**
+     * The hostname or IP address of the BMC to connect to. Required.
+     */
+    IDX_HOSTNAME,
+
+    /**
+     * The RMCP port. Informational only; libipmiconsole uses 623/UDP.
+     */
+    IDX_PORT,
```
necouchman: What does `Informational only` mean? Does libipmiconsole always use 623/UDP? If this is not configurable and hard-coded into the library, is there any point in passing the setting back and forth?

src/protocols/ipmi/settings.c
```diff
@@ -0,0 +1,910 @@
...
+ *     The parsed privilege level, or GUAC_IPMI_PRIVILEGE_ADMIN if the value is
+ *     blank or unrecognized.
+ */
+static guac_ipmi_privilege_level guac_ipmi_parse_privilege_level(
+        guac_user* user, const char* value) {
+
+    if (value == NULL || strcmp(value, "") == 0
+            || strcasecmp(value, "admin") == 0
+            || strcasecmp(value, "administrator") == 0)
+        return GUAC_IPMI_PRIVILEGE_ADMIN;
```
necouchman: Two things, here:
* This parsing function should not return `GUAC_IPMI_PRIVILEGE_ADMIN` for blank or `NULL` values. That should be handled as a default value further down when parsing the settings received by the client. This should deal strictly in converting the `value` to the privilege level.
* I also do not think there is any reason to have multiple values for Admin-level privileges - if this is coming from the Guacamole Client, we should be able to just use a single value. Pick `admin` or `administrator` and go with that.

src/protocols/ipmi/settings.c
```diff
@@ -0,0 +1,910 @@
...
+
+    if (strcasecmp(value, "operator") == 0)
+        return GUAC_IPMI_PRIVILEGE_OPERATOR;
+
+    if (strcasecmp(value, "user") == 0)
+        return GUAC_IPMI_PRIVILEGE_USER;
+
+    guac_user_log(user, GUAC_LOG_WARNING, "Unrecognized privilege level "
+            "\"%s\"; defaulting to \"admin\".", value);
```
necouchman: I think you can eliminate the logging, here, and have this function strictly responsible for parsing. Deal with the cases where an invalid privilege level is provided when actually using this function to parse the setting.

src/protocols/ipmi/settings.c
```diff
@@ -0,0 +1,910 @@
...
+ *     The power action string.
+ *
+ * @return
+ *     The parsed power action, or GUAC_IPMI_POWER_NONE if the value is blank
+ *     or unrecognized.
+ */
+static guac_ipmi_power_action guac_ipmi_parse_power_action(
+        guac_user* user, const char* value) {
+
+    if (value == NULL || strcmp(value, "") == 0
```
necouchman: Similar to the privilege parsing, I believe that this function should strictly parse provided string values into the corresponding `guac_ipmi_power_action` and should not concern itself with handling default/missing values.

src/protocols/ipmi/settings.c
```diff
@@ -0,0 +1,910 @@
...
+        return GUAC_IPMI_POWER_ON;
+
+    if (strcasecmp(value, "reset") == 0)
+        return GUAC_IPMI_POWER_RESET;
+
+    if (strcasecmp(value, "cycle") == 0)
+        return GUAC_IPMI_POWER_CYCLE;
+
+    guac_user_log(user, GUAC_LOG_WARNING, "Unrecognized power-on-connect "
+            "action \"%s\"; no power action will be taken.", value);
```
necouchman: Logging can also be eliminated, here.

src/protocols/ipmi/settings.c
```diff
@@ -0,0 +1,910 @@
...
+ *     The boot device string ("none", "pxe", "disk", "cdrom", or "bios").
+ *
+ * @return
+ *     The parsed boot device, or GUAC_IPMI_BOOT_NONE if the value is blank or
+ *     unrecognized.
+ */
+static guac_ipmi_boot_device guac_ipmi_parse_boot_device(
+        guac_user* user, const char* value) {
+
+    if (value == NULL || strcmp(value, "") == 0
```
necouchman: Same as above - parsing function need not handle default/unknown values.

src/protocols/ipmi/settings.c
```diff
@@ -0,0 +1,910 @@
...
+        return GUAC_IPMI_BOOT_DISK;
+
+    if (strcasecmp(value, "cdrom") == 0 || strcasecmp(value, "cd") == 0)
+        return GUAC_IPMI_BOOT_CDROM;
+
+    if (strcasecmp(value, "bios") == 0 || strcasecmp(value, "setup") == 0)
+        return GUAC_IPMI_BOOT_BIOS;
+
+    guac_user_log(user, GUAC_LOG_WARNING, "Unrecognized boot device "
+            "\"%s\"; boot order will be left unchanged.", value);
```
necouchman: And as above, no need to log, here.

src/protocols/ipmi/settings.c
```diff
@@ -0,0 +1,910 @@
...
+}
+
+/**
+ * Parses the given string into a guac_ipmi_encryption_policy. Unrecognized
+ * values fall back to the (secure) default of "required".
+ */
+static guac_ipmi_encryption_policy guac_ipmi_parse_encryption_policy(
+        guac_user* user, const char* value) {
+
+    if (value == NULL || strcmp(value, "") == 0
```
necouchman: Also, here, do not handle default/blank values in the parser.

src/protocols/ipmi/settings.c
```diff
@@ -0,0 +1,910 @@
...
+        return GUAC_IPMI_ENCRYPTION_REQUIRED;
+
+    if (strcasecmp(value, "preferred") == 0)
+        return GUAC_IPMI_ENCRYPTION_PREFERRED;
+
+    if (strcasecmp(value, "none") == 0)
+        return GUAC_IPMI_ENCRYPTION_NONE;
+
+    guac_user_log(user, GUAC_LOG_WARNING, "Unrecognized encryption policy "
+            "\"%s\"; requiring encryption.", value);
```
necouchman: And no need to log anything, here.

src/protocols/ipmi/settings.c
```diff
@@ -0,0 +1,910 @@
...
+int guac_ipmi_cipher_provides_confidentiality(int cipher_suite) {
+
+    /* IPMI 2.0 RMCP+ cipher suites whose confidentiality algorithm is NOT
+     * "none" (i.e. the payload is encrypted). Suites 0, 1, 2, 6, 7, 11, 15,
+     * and 16 provide no confidentiality and are therefore excluded. */
+    switch (cipher_suite) {
+        case 3:  case 4:  case 5:
+        case 8:  case 9:  case 10:
+        case 12: case 13: case 14:
+        case 17: case 18: case 19:
```
necouchman: Following established code style, these should be one per line.

src/protocols/ipmi/settings.c
```diff
@@ -0,0 +1,910 @@
...
+ * A single FreeIPMI workaround flag, mapping a user-facing token to the
+ * corresponding libipmiconsole (SOL) and libfreeipmi (chassis) flag bits. A
+ * chassis bit of 0 indicates the flag has no equivalent in the chassis
+ * (out-of-band 2.0) namespace and applies to the SOL session only.
+ */
+typedef struct guac_ipmi_workaround_token {
+    const char* name;
+    unsigned int sol_flag;
+    unsigned int chassis_flag;
+} guac_ipmi_workaround_token;
```
necouchman, line 478: Please document the structure members.

src/protocols/ipmi/settings.c
```diff
@@ -0,0 +1,910 @@
...
+};
+
+/**
+ * A vendor preset expanding to a set of individual workaround tokens known to
+ * be commonly required for that vendor's BMCs.
+ */
+typedef struct guac_ipmi_workaround_preset {
+    const char* name;
+    const char* tokens;
+} guac_ipmi_workaround_preset;
```
necouchman, line 513: Please document the structure members.

src/protocols/ipmi/settings.c
```diff
@@ -0,0 +1,910 @@
...
+    { "hpe",        "" },
+    { "lenovo",     "" }
+};
+
+/**
+ * Applies a single individual workaround flag token (not a preset) to the
+ * given SOL and chassis masks. Returns non-zero if the token was recognized.
+ */
+static int guac_ipmi_apply_workaround_token(const char* token,
+        unsigned int* sol, unsigned int* chassis) {
```
necouchman, line 548: Please document parameters and return value.

src/protocols/ipmi/settings.c
```diff
@@ -0,0 +1,910 @@
...
+
+}
+
+/**
+ * Parses a comma-separated list of workaround flag tokens and/or vendor preset
+ * names, accumulating the resulting flags into the given SOL (libipmiconsole)
+ * and chassis (libfreeipmi) masks. Both masks are cleared to 0 first.
+ */
+static void guac_ipmi_parse_workaround_flags(guac_user* user,
+        const char* value, unsigned int* sol, unsigned int* chassis) {
```
necouchman, line 584: Please document parameters.

src/protocols/ipmi/settings.c
```diff
@@ -0,0 +1,910 @@
...
+
+    *sol = 0;
+    *chassis = 0;
+
+    if (value == NULL || strcmp(value, "") == 0)
+        return;
```
necouchman: Might be worth spitting out a log message, at `WARN` or `DEBUG` level, for this case.

src/protocols/ipmi/settings.c
```diff
@@ -0,0 +1,910 @@
...
+
+/**
+ * Parses the raw K_g argument into a newly-allocated byte buffer, returning
+ * the buffer and storing its length in *length. If the value begins with "0x",
+ * it is decoded as a hexadecimal byte string (allowing binary keys that may
+ * contain NUL bytes); otherwise it is treated as a raw passphrase. Returns NULL
+ * (with *length set to 0) if the value is NULL or invalid.
+ */
+static char* guac_ipmi_parse_k_g(guac_user* user, const char* raw,
+        int* length) {
```
necouchman, line 678: Please document parameters and return value.

src/protocols/ipmi/settings.c
```diff
@@ -0,0 +1,910 @@
...
+
+    *length = 0;
+
+    if (raw == NULL || *raw == '\0')
+        return NULL;
```
necouchman: Might be good to log this at `WARN` or `DEBUG` level?

src/protocols/ipmi/settings.h
```diff
@@ -0,0 +1,476 @@
...
+/**
+ * The privilege level to authenticate with, expressed using the
+ * IPMICONSOLE_PRIVILEGE_* constants from libipmiconsole. The values here
+ * intentionally match those constants.
+ */
+typedef enum guac_ipmi_privilege_level {
+    GUAC_IPMI_PRIVILEGE_USER     = 0,
+    GUAC_IPMI_PRIVILEGE_OPERATOR = 1,
+    GUAC_IPMI_PRIVILEGE_ADMIN    = 2
+} guac_ipmi_privilege_level;
```
necouchman, line 97: Please document each of the values. Also, as mentioned in one of the other files, is there any reason not to just use the `IPMICONSOLE_PRIVILEGE_*` values?? Why redefine them?

src/protocols/ipmi/ipmi.c
```diff
@@ -0,0 +1,494 @@
...
+#include <string.h>
+#include <unistd.h>
+
+/**
+ * Lock guarding the process-global libipmiconsole engine reference count. The
+ * libipmiconsole engine is a single process-wide resource; initializing or
+ * tearing it down per connection would race with, and disrupt, any other IPMI
+ * connections active within the same guacd process.
+ */
+static pthread_mutex_t guac_ipmi_engine_lock = PTHREAD_MUTEX_INITIALIZER;
```
necouchman: When you say "any other IPMI connections active within the same guacd process" - does this mean:
* Within the same guacd parent process?
* Within the same guacd child/connection process?

guacd forks a new process for each connection, and, since this particular lock would only be initialized in the context of a single connection, and that would be *after* the child connection has been forked by guacd, there actually should not be any issue with initializing it or tearing it down...

...*UNLESS*, you also mean that the IPMI chassis connections would interfere with the SOL connection, in which case, this would still make sense.

src/protocols/ipmi/menu.c
```diff
@@ -0,0 +1,457 @@
...
+#define GUAC_IPMI_MENU_ALT_EXIT "\x1B[?1049l"
+
+/**
+ * Displays a "press any key to return" prompt and marks the current control
+ * menu operation complete. Forward-declared for use by the chassis worker.
+ *
+ * @param client
+ *     The guac_client associated with the IPMI connection.
+ */
+static void guac_ipmi_menu_await_dismiss(guac_client* client);
```
necouchman: Our practice throughout the code for `static` functions is generally just to implement them where they need to be in the code and not prototype them.

necouchman (changes requested):
@ciroiriarte This looks really great, I'm excited to have this in the Guacamole code! I do have some initial comments - don't despair, most of these are style and documentation clean-up, although some deal with implementation details. Generally:
* In several places you've defined Guacamole-specific data structures or enums that mirror ones provided by the IPMI libraries. Is there any reason to do this? It seems like it would simplify things and cut down the amount of code to just reuse what is already there, unless there's a technical or implementation-specific reason why not?
* Make sure you've fully-documented all of the functions and/or function prototypes. There are several places where thinks like `@param` and `@return` tags are missing, and a handful where there was no documentation.
* There are a few places where I think some more verbose comments within the code would help people more easily understand what's being done and why.
* I see what you're doing with the IPMI menu, but I wonder if there's a way to use Guacamole's built-in menu to handle these functions rather than implementing the text-based menu you have and then having to implement the JSON-related code around it? I haven't looked at this in any detail, it just seems to me that that's what our existing hidden Guacamole menu is meant for, and there ought to be a way to handle things like chassis commands and the like similar to how we handle file transfers, terminal parameters, etc.

src/protocols/ipmi/chassis.c
ciroiriarte: Makes sense — I'll return `guac_ipmi_power_state` directly with `UNKNOWN` as the failure signal, and drop the separate result code.

That also resolves the out-parameter question below it. For the record on what it does today: `power` is caller-owned, and the `UNKNOWN` at entry is so the failure paths leave the caller with a defined value rather than an uninitialised one — it's read by the caller after a non-zero return, not inside the function. But returning the enum makes that mechanism unnecessary rather than merely explained, so I'd rather just make the change.

src/protocols/ipmi/chassis.c
ciroiriarte: Scoped to the `power` value specifically, those are exhaustive: `current_power_state.power_is_on` in the Get Chassis Status response is a single bit, and `UNKNOWN` covers a failed read. The richer state isn't lost — power fault and overload are separate bits and already appear in the human-readable status buffer.

src/protocols/ipmi/control.c
ciroiriarte: I'll restructure this into the explicit `else if` (power command) / `else` (invalid) shape you describe — it should be obvious on sight, and it isn't.

One clarification just so the current behaviour isn't on the record as worse than it is: `guac_ipmi_control_power_action()` is a pure string parser that returns `GUAC_IPMI_POWER_NONE` for anything unrecognised, and the `else` block replies "Unknown command." — nothing unvalidated reaches the BMC. But the flow reads as though it could, which is reason enough to change it.

src/protocols/ipmi/control.c
ciroiriarte: I'll add the checks. I'd followed the convention in the surrounding protocol drivers, where these returns generally aren't checked, but that's not an argument for new code repeating it. I'll check where a failure can affect control flow or leave the client with a misleading status, and document anything that stays best-effort.

src/protocols/ipmi/tests/control/json_get.c
ciroiriarte: `guac_ipmi_control_json_get()` bails if the value isn't quoted, so `{"id":123}` from a third-party client parses as absent. Given your point about independent client implementations, I'll extend it to accept numeric and boolean scalars.

src/protocols/ipmi/ipmi.c
ciroiriarte: The comment is misleading, not just unclear. This lock has nothing to do with chassis-vs-SOL interference; that's handled separately by `menu_lock`/`chassis_busy` in `control.c`. All this guards is the refcount around `ipmiconsole_engine_init()`/`ipmiconsole_engine_teardown()`, and your reading of the fork model is the right one: guacd forks per connection, the refcount lives in that child, so it can only ever go 0→1 and the lock never contends. As written it's scaffolding for a sharing scenario that can't occur.

Let me simplify it rather than justify it. If it's worth keeping any of it, the argument would be that libipmiconsole's engine is process-global and the refcount is cheap insurance if guacd ever ran connections in threads rather than forks — but that's speculative, and I'd rather not carry speculative concurrency machinery. I'll cut it back to a plain init/teardown and fix the comment.

src/protocols/ipmi/chassis.c
ciroiriarte: Addressed together with the "return the power state" suggestion just above — see my reply there (https://github.com/apache/guacamole-server/pull/692#discussion_r3604202841). Returning `guac_ipmi_power_state` directly removes this out-parameter entirely, so the up-front `UNKNOWN` init goes away with it.

src/protocols/ipmi/control.c
ciroiriarte: Same as the return-value comment earlier in this file — Ill add the checks; details in my reply there (https://github.com/apache/guacamole-server/pull/692#discussion_r3604203451).

src/protocols/ipmi/settings.h
ciroiriarte: Two things here — the "why redefine" question and the request to document the values.

On reuse: this is the one enum I'd argue for keeping, because the module talks to two libraries whose privilege constants don't agree. SOL goes through libipmiconsole (`IPMICONSOLE_PRIVILEGE_USER/OPERATOR/ADMIN` = 0/1/2); chassis goes through libfreeipmi (`IPMI_PRIVILEGE_LEVEL_USER/OPERATOR/ADMIN` = 0x02/0x03/0x04). So "just reuse the library's values" is ambiguous, and a value picked for one library is silently wrong on the other path — an `IPMICONSOLE_PRIVILEGE_ADMIN` (2) reaching the chassis code reads as `IPMI_PRIVILEGE_LEVEL_USER` (0x02), a privilege downgrade with nothing to flag it.

The current code does invite the question fairly, though: the enum is defined as `USER=0, OPERATOR=1, ADMIN=2` — hard-coded to the libipmiconsole numbering — and `ipmi.c` assigns it straight into `ipmiconsole_ipmi_config.privilege_level` with no mapping, while only the chassis side maps (via `guac_ipmi_map_privilege_level()`). So on the SOL path the enum really is just a copy of libipmiconsole's constants, which is why it reads as redundant. I'll give the values arbitrary numbering (not 0/1/2) and add a companion `guac_ipmi_map_privilege_level_ipmiconsole()` mirroring the chassis switch, so neither back-end ever receives the raw enum value.

To be straight about it: two conflicting constant sets don't prove a third enum is the only option — parsing directly to each library's constant at its boundary would work too. The internal enum just keeps the parsed setting semantic between the settings layer and the two back-ends instead of committing to one library's numbering at parse time. If you'd rather drop it and map at the boundary, I'll do that instead.

On the documentation: I'll add a doc comment to each value as part of that change.

src/protocols/ipmi/chassis.c
ciroiriarte: Different rationale from the privilege levels here, and a weaker one — I'd rather be upfront about that than stretch the same argument to cover it. There's no conflicting-constants problem for power actions or boot devices; these enums are less a wrapper around the FreeIPMI constants than internal command state. Each carries a `NONE` member that has no IPMI equivalent (`IPMI_CHASSIS_CONTROL_*` can't express "no action requested"), and they're used well away from this mapping — parsing the client's command strings, the menu's display labels, and the pending-confirmation state in `guac_ipmi_client`. The translation to the FreeIPMI constant happens only here, at the boundary.

`NONE` could be modelled as optional/nullable state instead, which would let these collapse into the raw constants. I lean toward keeping them as internal state enums for the parsing and menu use, but if you'd prefer the nullable approach I'll switch.

src/protocols/ipmi/chassis.c
ciroiriarte: Same privilege question — answered in full under the enum definition in `settings.h` (https://github.com/apache/guacamole-server/pull/692#discussion_r3604260265). Short version: libipmiconsole and libfreeipmi disagree on the privilege values (0/1/2 vs 0x02/0x03/0x04), so reusing either is silently wrong on the other path. I'll keep an internal enum but give it arbitrary values and map explicitly on both paths.

src/protocols/ipmi/chassis.c
ciroiriarte: Same rationale as the power-action enum just above (https://github.com/apache/guacamole-server/pull/692#discussion_r3604260498): these are internal command state with a `NONE` sentinel that has no `IPMI_CHASSIS_CONTROL_*` / boot-device equivalent, mapped to the FreeIPMI constant only here at the boundary.

src/protocols/ipmi/chassis.c
necouchman: I guess I was more wondering if there were states like "Sleep" or "Hibernate" that represent status that are 1) not "OFF", but also not "ON", but might be states where you'd want to do something power-related to the system, like wake it up?

src/protocols/ipmi/argv.c
ciroiriarte: Removed in 86f66774 — config.h is auto-included in every .c file since GUACAMOLE-2182, so the explicit include is redundant.

src/protocols/ipmi/chassis.c
ciroiriarte: Moved these to the top of the file alongside the other file-level declarations in 86f66774.

src/protocols/ipmi/control.c
ciroiriarte: Removed the duplicate in 86f66774; the canonical documentation stays in control.h.

src/protocols/ipmi/control.c
ciroiriarte: Removed the duplicate in 86f66774; the canonical documentation stays in control.h.

src/protocols/ipmi/control.c
ciroiriarte: Documented in 86f66774.

src/protocols/ipmi/control.c
ciroiriarte: Documented in 86f66774.

src/protocols/ipmi/control.c
ciroiriarte: Added inline comments describing the flow in 86f66774.

src/protocols/ipmi/control.c
ciroiriarte: Documented in 86f66774.

src/protocols/ipmi/control.c
ciroiriarte: Documented in 86f66774.

src/protocols/ipmi/control.c
ciroiriarte: Documented in 86f66774.

src/protocols/ipmi/control.c
ciroiriarte: Added inline comments describing the flow in 86f66774.

src/protocols/ipmi/control.c
ciroiriarte: Documented in 86f66774.

src/protocols/ipmi/control.c
ciroiriarte: Documented in 86f66774.

src/protocols/ipmi/control.c
ciroiriarte: Documented in 86f66774.

src/protocols/ipmi/control.c
ciroiriarte: Documented in 86f66774.

src/protocols/ipmi/control.c
ciroiriarte: Documented in 86f66774.

src/protocols/ipmi/control.c
ciroiriarte: Added inline comments describing the flow in 86f66774.

src/protocols/ipmi/control.c
ciroiriarte: Documented in 86f66774.

src/protocols/ipmi/settings.c
ciroiriarte: Removed the port setting entirely in 86f66774 — both libipmiconsole and libfreeipmi hardwire 623/UDP, so it was read but never honored.

src/protocols/ipmi/settings.c
ciroiriarte: Done in 86f66774 — the parser is now a pure string-to-enum converter returning an INVALID sentinel; the default and warning moved to the caller, and the "administrator" alias was dropped in favour of "admin".

src/protocols/ipmi/settings.c
ciroiriarte: Done in 86f66774 — logging removed; the parser only converts now.

src/protocols/ipmi/settings.c
ciroiriarte: Done in 86f66774 — pure converter; the default is applied by the caller.

src/protocols/ipmi/settings.c
ciroiriarte: Done in 86f66774 — logging removed from the parser.

src/protocols/ipmi/settings.c
ciroiriarte: Done in 86f66774 — pure converter; the default is applied by the caller.

src/protocols/ipmi/settings.c
ciroiriarte: Done in 86f66774 — logging removed from the parser.

src/protocols/ipmi/settings.c
ciroiriarte: Done in 86f66774 — pure converter; the default is applied by the caller.

src/protocols/ipmi/settings.c
ciroiriarte: Done in 86f66774 — logging removed from the parser.

src/protocols/ipmi/settings.c
ciroiriarte: Done in 86f66774 — one case label per line now.

src/protocols/ipmi/settings.c
ciroiriarte: Documented in 86f66774.

src/protocols/ipmi/settings.c
ciroiriarte: Documented in 86f66774.

src/protocols/ipmi/settings.c
ciroiriarte: Documented in 86f66774.

src/protocols/ipmi/settings.c
ciroiriarte: Documented in 86f66774.

src/protocols/ipmi/settings.c
ciroiriarte: Added a DEBUG log for this case in 86f66774.

src/protocols/ipmi/settings.c
ciroiriarte: Documented in 86f66774.

src/protocols/ipmi/settings.c
ciroiriarte: Added a DEBUG log for this case in 86f66774.

src/protocols/ipmi/menu.c
ciroiriarte: Done in 86f66774 — dropped the prototype and moved the definition ahead of its first use.

src/protocols/ipmi/chassis.c
ciroiriarte: Good question — worth separating two things.

The value this function reports comes from Get Chassis Status, whose `current_power_state` is a single "power is on" bit (alongside the overload/interlock/fault bits). It has no field for ACPI sleep states, so from this command a running system and one suspended to RAM are indistinguishable — ON/OFF (plus UNKNOWN on a failed read) really are exhaustive for what it can tell us.

Sleep (S3) / hibernate (S4) live in a separate IPMI command, *Get ACPI Power State*, which reports the system's ACPI S-state and which we don't currently query. If surfacing "suspended" distinctly — and offering a targeted wake — is worthwhile, that's a sensible follow-up: read the ACPI power state alongside the chassis status and extend `guac_ipmi_power_state` with the S-states. It would be additive to this function rather than a change to the on/off bit it reports today.

On waking specifically: IPMI chassis control has no dedicated wake-from-sleep action — the existing power-up/reset actions operate at the chassis level, and waking from S3 is usually an OS or Wake-on-LAN concern. So there is no distinct state we would act on differently through chassis control right now.

ciroiriarte:
Thanks again for the thorough review, @necouchman — I've pushed 86f66774 addressing the feedback, and replied inline on each thread. A summary of what changed:

**Documentation**
* Completed `@param`/`@return` across control.c and settings.c, documented the workaround struct members and each enum value, and removed the doc blocks that duplicated control.h.

**Style**
* Dropped the now-redundant `config.h` includes (auto-included since GUACAMOLE-2182), one `case` label per line, moved file-level `#define`/`typedef` declarations to the top of chassis.c, and inlined the menu.c static function instead of forward-declaring it.

**Settings parsers**
* The privilege / power-action / boot-device / encryption-policy parsers are now pure string-to-enum converters — no logging, no defaults. Each returns a new `*_INVALID` sentinel and the default plus warning is applied by the caller (this also fixed a small leak where the parsed argument strings were never freed). Collapsed the `administrator` alias to `admin`.
* Removed the unused `port` setting entirely — both libipmiconsole and libfreeipmi hardwire 623/UDP, so it was read but never honored.

**Behavior** (discussed inline)
* Accept non-string JSON scalars in the control parser, return the power state directly from `guac_ipmi_chassis_status()`, reject unknown control commands explicitly, and drop the vestigial engine reference count/lock.

Still to come, from the enum discussion above: giving the privilege enum library-independent values and adding the symmetric `guac_ipmi_map_privilege_level_ipmiconsole()` mapper so neither back-end receives the raw enum value. I'll follow up with that shortly. Happy to iterate on anything here.

src/protocols/ipmi/settings.h
ciroiriarte: Follow-up from above: this is now done in c29f725d. The enum values are library-independent (INVALID = 0, then USER/OPERATOR/ADMIN), and the SOL path maps through a new `guac_ipmi_map_privilege_level_ipmiconsole()` mirroring the chassis-side `guac_ipmi_map_privilege_level()` — so neither back-end receives the raw enum value anymore.

src/protocols/ipmi/control.c
necouchman: Fair enough - I do think changing it is the right thing to do, as it helps avoid bugs that could creep up in the future if someone isn't paying attention, and also helps keep the code flow clear.

src/protocols/ipmi/chassis.c
ciroiriarte: Following up on this one so it isn't left hanging. Unlike the privilege enum, the power-action and boot-device enums have no conflicting-constants problem — they're internal command state, each with a `NONE` member that has no `IPMI_CHASSIS_CONTROL_*` equivalent, used in command-string parsing, the menu labels, and the pending-confirmation state, and translated to the FreeIPMI constant only at the boundary. I'm happy either way: keep them as internal state enums, or collapse them into the raw FreeIPMI constants with "no action" modeled as a nullable/optional value. Do you have a preference? I'll make the change if you'd rather drop them.

src/protocols/ipmi/control.c
ciroiriarte: These two calls — `guac_protocol_send_ack()` and `guac_socket_flush()` — are best-effort writes to a single user's socket. I followed the existing convention here: `guac_socket_flush()` is called unchecked in the same pipe/broadcast paths in ssh (`ssh/pipe.c`), telnet (`telnet/pipe.c`), and vnc (`vnc/client.c`), and `guac_protocol_send_ack()`'s return isn't checked anywhere in the tree. A failed write means that user's socket is already broken, which guacd surfaces and tears down through the normal user lifecycle, so there's no extra recovery to do here. Checking them would make the IPMI code inconsistent with the other protocols for no behavioral gain — but if you'd prefer them checked regardless, I'll add them.

ciroiriarte:
@necouchman — following up on your suggestion to handle the chassis functions through Guacamole's built-in menu rather than the in-terminal text menu plus the JSON around it. I dug into how far that can go, and I'd like to agree the target before doing a large rework.

**Where I landed:** the strongest version of the point is really about the JSON, not just the text menu — guacd bundles no JSON parser, and the idiomatic control channel is argv name=value streams with `GUAC_ARGV_OPTION_ECHO` (which is what already backs the built-in menu for e.g. the ssh color-scheme/font params). So the hand-rolled `json_get`/`json_escape` is the non-idiomatic surface, and a good chunk of the control channel can move onto argv.

**Fits the built-in argv/menu model cleanly:**
- `boot-device`, `boot-persistent` — genuine settings → argv params; surface in the built-in menu with no custom client code.
- `power-state`, SOL `health`, `identify` state — push via `OPTION_ECHO` for display.

**Doesn't fit argv:**
- The momentary chassis *actions* (power on/off/cycle/reset/soft-shutdown/diagnostic-interrupt, identify, send-break) and their per-command results ("sent" / "failed" / "operation already in progress"). argv has only `ONCE`/`ECHO` — no action or request→response semantics — and nothing in guacd exposes action-buttons via argv today (terminal params are settings; file-transfer uses file streams).
- SEL is a multi-KB log; it fits a standard outbound pipe/download stream better than an embedded JSON blob.

**Proposed plan:**
1. Move settings + state (`boot-device`, `boot-persistent`, `power-state`, `health`, `identify`) to argv — this deletes `json_escape` and most of the pushed-state JSON.
2. Deliver SEL as a pipe/download stream rather than JSON.
3. Keep a small, non-JSON channel only for the momentary actions + their results.

Two calls I'd like your read on before I start:
- **The momentary-action channel:** OK to keep a minimal dedicated channel for the actions/results (argv can't express them), or would you rather I model actions as an argv trigger and drop the per-command result text, reflecting outcome only via the pushed `power-state`?
- **The in-terminal text menu (~446 lines):** it's currently the only client-agnostic path — it works with any client or a plain terminal, no custom UI, whereas the built-in menu covers only custom-UI clients. Happy to drop it, but wanted to confirm you're comfortable requiring a custom client for chassis control rather than assume it.

Once you pick on those two, I'll rework accordingly.

PR #696, opened 2026-07-22
GUACAMOLE-2306: Add support for the RDPSND Wave2 (SNDC_WAVE2) PDU.

mkhawam:
Modern Windows RDP servers may transmit audio using the single-PDU Wave2 (SNDC_WAVE2) form rather than the legacy WaveInfo/Wave pair. Per MS-RDPEA, a server uses Wave2 only when both the client and server advertise RDPSND protocol version 8 or higher. Guacamole previously advertised version 6 and had no SNDC_WAVE2 handler, so Wave2 could not be negotiated, and any Wave2 PDU that did arrive was silently dropped (no audio decoded and no Wave Confirmation returned, which stalls the server's audio stream).

This advertises client version 8 and adds a handler that decodes the self-contained Wave2 audio block, plays it, and acknowledges it via a Wave Confirmation PDU. The legacy WaveInfo/Wave path is retained for servers negotiating a lower version.

### DIFF
--- a/src/protocols/rdp/channels/rdpsnd/rdpsnd-messages.c
+++ b/src/protocols/rdp/channels/rdpsnd/rdpsnd-messages.c
@@ -89,7 +89,7 @@ void guac_rdpsnd_formats_handler(guac_rdp_common_svc* svc,
 
     /* Version and padding */
     Stream_Write_UINT8(output_stream,  0);
-    Stream_Write_UINT16(output_stream, 6);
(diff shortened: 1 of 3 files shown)


clhedrick:
It's not obvous from this description, but this patch is needed to make audio work with Gnome Remote Desktop, least with Guacamole on our Redhat 10.2 system. Without the patch there is no sound.

PR #698, opened 2026-08-04
GUACAMOLE-1625: Close sockets inherited by connection processes.

Heyzi:
Connection processes inherit every descriptor open in the parent, as guacd_create_proc() closes only its own end of the socketpair it shares with the new child. For a busy parent, this means the parent's end of every other connection's user socketpair, the parent's end of every other connection process' socketpair, and the socket guacd listens on. FD_CLOEXEC cannot help, as connection processes are forked but never exec'd.

Those inherited descriptors keep unrelated sockets referenced after the parent has closed them, so their peers never observe EOF. A write which should fail with EPIPE instead blocks forever, and that blocked write is performed while holding a read lock on the client's list of users, so guac_client_remove_user() can never acquire the corresponding write lock. The count of connected users therefore never reaches zero, guacd_proc_stop() is never called, and the process never exits.

The effect compounds, as the more connections are open, the more descriptors each new process inherits. A parent with 469 such processes was observed holding tens of gigabytes, with descriptor counts rising from 98 in the oldest process to 756 in the youngest, and a single socket held by 40 of 40 sampled processes.

Close inherited sockets within the newly-forked process. Only sockets are closed, as descriptors for anything else may be held by libraries unaware that they are being closed, and reusing such a number for a later socket would silently redirect their reads and writes.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -150,7 +150,7 @@ AC_SUBST(UUID_LIBS)
 AC_SUBST(CUNIT_LIBS)
 
 # Library functions
-AC_CHECK_FUNCS([clock_gettime gettimeofday memmove memset select strdup nanosleep prctl])
+AC_CHECK_FUNCS([clock_gettime gettimeofday memmove memset select strdup nanosleep prctl close_range])
 
 AC_CHECK_DECL([png_get_io_ptr],
     [AC_DEFINE([HAVE_PNG_GET_IO_PTR],,
--- a/src/guacd/proc.c
+++ b/src/guacd/proc.c
@@ -32,12 +32,15 @@
 #include <guacamole/socket.h>
 #include <guacamole/user.h>
 
+#include <dirent.h>
 #include <errno.h>
 #include <pthread.h>
 #include <signal.h>
 #include <stdlib.h>
 #include <string.h>
+#include <syslog.h>
 #include <unistd.h>
+#include <sys/stat.h>
 #include <sys/time.h>
 #include <sys/types.h>
 #include <sys/socket.h>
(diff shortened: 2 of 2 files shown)


necouchman:
@Heyzi As with #699, could you rebase against `staging/1.6.1`?

mike-jumper:
I'm concerned about the potential performance impact of looping through all file descriptors at connection start. Would `close_range()` be an option here?

Heyzi:
> I'm concerned about the potential performance impact of looping through all file descriptors at connection start. Would `close_range()` be an option here?

Yeah, fair point. I'll push an update soon.

Heyzi:
> I'm concerned about the potential performance impact of looping through all file descriptors at connection start. Would `close_range()` be an option here?

Switched to `close_range()`, with two fallbacks - both are load-bearing:

- musl has no wrapper (the guacd image is Alpine), so the syscall is invoked directly
- RHEL/Rocky 8 has neither — glibc 2.28, Linux 4.18 — so `/proc/self/fd` is used there

Benchmarked against the previous per-descriptor approach, 25 runs per point:

| open fds | scan + fstat | close_range |
|---|---|---|
| 100 | 0.363 ms | 0.077 ms |
| 500 | 1.582 ms | 0.381 ms |
| 1000 | 3.101 ms | 0.819 ms |
| 5000 | 18.115 ms | 7.172 ms |

The RHEL 8 fallback matters too: with the default `ulimit -n` of 1048576 in a Rocky 8 container, walking the full range costs 229ms per connection vs 0.05ms via `/proc/self/fd`.

We've deployed a build of this to our own production, since the issue had become critical for us.
v1
close_range
Graphs show ~200 sessions per hour, so nearly all sessions turn over hourly.

mike-jumper (approved):

PR #699, opened 2026-08-04
GUACAMOLE-2118: Signal end of user input to connection processes.

Heyzi:
Once no further data can be read from a user, the thread forwarding that user's input simply returns, leaving its end of the socketpair shared with the connection process open. A user which vanishes without sending "disconnect" therefore leaves that process blocked awaiting input indefinitely: nothing further informs it that its last user has left, so it never stops itself and never exits.

The thread reading from that process cannot detect this either, as it remains blocked on a socket which an idle process will never write to again, and so never reaches the point at which it would close its end.

Shut down the write side of the socketpair when no further user input is possible, such that the connection process observes end of input.

### DIFF
--- a/src/guacd/connection.c
+++ b/src/guacd/connection.c
@@ -129,6 +129,16 @@ static void* guacd_connection_write_thread(void* data) {
             break;
     }
 
+    /* Signal end of input to the connection process. Without this, a user
+     * which vanishes without sending "disconnect" leaves that process blocked
+     * awaiting input indefinitely, as nothing further will inform it that its
+     * last user has left. */
+    if (shutdown(params->fd, SHUT_WR))
+        guacd_log(GUAC_LOG_ERROR, "Unable to signal end of user input to "
+                "connection process: %s. That process may remain running but "
+                "inactive, retaining the memory of its connection until guacd "
+                "is restarted.", strerror(errno));
+
     return NULL;
 
 }


necouchman:
Thanks @Heyzi - one request, since this Jira issue is slated for 1.6.1, could you rebase against the `staging/1.6.1` branch?

Heyzi:
> Thanks @Heyzi - one request, since this Jira issue is slated for 1.6.1, could you rebase against the `staging/1.6.1` branch?

The Docker build is broken for everyone right now.

libwebsockets v5.0.0 turns on HTTP/3 by default, and HTTP/3 force-enables `LWS_WITH_GNUTLS`, which the builder stage has no packages for. Since `WITH_LIBWEBSOCKETS` is an unpinned pattern, autobuild.sh picks up v5.0.0 automatically.

guacd only uses libwebsockets for Kubernetes and doesn't need HTTP/3, so adding this to `LIBWEBSOCKETS_OPTS` fixes it without pinning:
`-DLWS_WITH_HTTP3=OFF`


Alternatively, pinning to the 4.x series on line 57 works too, and still picks up the newest 4.x release:
`ARG WITH_LIBWEBSOCKETS='v4(.\d+)+'`

Want me to open a PR, or would you rather take it?

necouchman:
> > Thanks @Heyzi - one request, since this Jira issue is slated for 1.6.1, could you rebase against the `staging/1.6.1` branch?
> 
> The Docker build is broken for everyone right now.
> 
> libwebsockets v5.0.0 turns on HTTP/3 by default, and HTTP/3 force-enables `LWS_WITH_GNUTLS`, which the builder stage has no packages for. Since `WITH_LIBWEBSOCKETS` is an unpinned pattern, autobuild.sh picks up v5.0.0 automatically.
> 
> guacd only uses libwebsockets for Kubernetes and doesn't need HTTP/3, so adding this to `LIBWEBSOCKETS_OPTS` fixes it without pinning: `-DLWS_WITH_HTTP3=OFF`
> 
> Alternatively, pinning to the 4.x series on line 57 works too, and still picks up the newest 4.x release: `ARG WITH_LIBWEBSOCKETS='v4(.\d+)+'`
> 
> Want me to open a PR, or would you rather take it?

Thanks for the info on that - I'll put in a separate Jira issue for it, at least, and you can open a PR for it if you want. I don't have a terribly strong preference on direction at this point - disabling HTTP3 seems fine to me, but good either way.

necouchman:
> > Want me to open a PR, or would you rather take it?
> 
> Thanks for the info on that - I'll put in a separate Jira issue for it, at least, and you can open a PR for it if you want. I don't have a terribly strong preference on direction at this point - disabling HTTP3 seems fine to me, but good either way.

Opened GUACAMOLE-2315 for this build issue.

src/guacd/connection.c
```diff
@@ -129,6 +129,14 @@ static void* guacd_connection_write_thread(void* data) {
             break;
     }
 
+    /* Signal end of input to the connection process. Without this, a user
+     * which vanishes without sending "disconnect" leaves that process blocked
+     * awaiting input indefinitely, as nothing further will inform it that its
+     * last user has left. */
+    if (shutdown(params->fd, SHUT_WR))
+        guacd_log(GUAC_LOG_DEBUG, "Unable to signal end of user input to "
+                "connection process: %s", strerror(errno));
```
mike-jumper: If the consequence of this failing is a stuck process that does not get cleaned up, this should probably be at least a warning (if not an error), and that warning should include enough information that the administrator can understand the consequences of what happened.

For example:

https://github.com/apache/guacamole-server/blob/de97609007c088b5e6afd827eff5e9076013a247/src/guacd/proc.c#L555-L559

mike-jumper (changes requested):

src/guacd/connection.c
Heyzi: Left the connection ID out for now - the params struct doesn't carry it, and a pointer wouldn't be safe: the routing thread frees the client right after `waitpid()`, while both I/O threads are detached. If the child exits on its own, `conn-write` can still be sitting in `guac_socket_read()` well after that.

Adding it means copying the string into the params and freeing it alongside. Let me know if you want that...

mike-jumper (approved):

PR #700, opened 2026-08-04
GUACAMOLE-2313: Disambiguate handling of lock modifiers and Mac-specific keys.

mike-jumper:
This change is the server half of the changes proposed by https://github.com/apache/guacamole-client/pull/1238, adding handling for the Command/Option keysyms and ensuring keypresses of modifiers and locks do not cause the terminal to suddenly scroll down.


necouchman (approved):
Good with the changes, but feel like we need to address the build issue before merging (GUACAMOLE-2315).

mike-jumper:
> ... feel like we need to address the build issue before merging ([GUACAMOLE-2315](https://issues.apache.org/jira/browse/GUACAMOLE-2315)).

Yep, I agree.

PR #705, opened 2026-08-18
GUACAMOLE-2319: Fix SFTP symlink dirs shown as files when root is not system root.

eugen-keeper:
When SFTP is enabled and the configured root directory is the system root, symlinked directories are listed and navigated correctly. The issue only occurs when a custom SFTP root directory is configured that is not the system root. In that case, symlinked directories appear as files in the side panel instead of folders.


necouchman:
@eugen-keeper Seems simple enough that this could slip into 1.6.1 - can you rebase for `staging/1.6.1`?

necouchman (approved):

PR #709, opened 2026-08-24
GUACAMOLE-2323: Abort the connection when FreeRDP reports a connection error

xelan:
Fixes https://issues.apache.org/jira/browse/GUACAMOLE-2323

### DIFF
--- a/src/protocols/rdp/rdp.c
+++ b/src/protocols/rdp/rdp.c
@@ -884,14 +884,22 @@ static int guac_rdp_handle_connection(guac_client* client) {
 
         int wait_result = rdp_guac_client_wait_for_events(client, GUAC_RDP_MESSAGE_CHECK_INTERVAL);
 
-        /* Abort if the wait failed or FreeRDP recorded a connection error, as
-         * is the case when an RDP server tears down the connection without
-         * first sending a Set Error Info PDU. Simply leaving the loop would
-         * return from this function with the guac_client still running, which
-         * guac_rdp_client_thread() takes as a request to reconnect. */
         if (wait_result < 0) {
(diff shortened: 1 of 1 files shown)


src/protocols/rdp/rdp.c
```diff
@@ -883,8 +883,16 @@ static int guac_rdp_handle_connection(guac_client* client) {
...
+
+        /* Abort if the wait failed or FreeRDP recorded a connection error, as
+         * is the case when an RDP server tears down the connection without
+         * first sending a Set Error Info PDU. Simply leaving the loop would
+         * return from this function with the guac_client still running, which
+         * guac_rdp_client_thread() takes as a request to reconnect. */
+        if (wait_result < 0) {
+            guac_rdp_client_abort(client, rdp_inst);
             break;
+        }
```
necouchman: How is guacd to distinguish events which _should_ trigger a reconnection from the logout events which _should not_?

necouchman (commented):
See my comment below - I think we've discussed this in the past, and I'm pretty sure the behavior we have currently is the desired behavior - generally speaking, failures in the RDP client should trigger a reconnect, and changing that behavior for the sake of broken versions of Windows that don't send the correct disconnect/logout event to the client I'm not sure is the way to go - at least, not without some other checks to distinguish between the two. Otherwise we might end up simply aborting connections in situations where we should actually retry them?

src/protocols/rdp/rdp.c
xelan: Thanks for your feedback @necouchman! I've added another commit which improves the detection logic. Now the guac_rdp_client_abort is only done if FreeRDP reports an appropriate info code. Tested with Windows XP SP3. What do you think about this approach?

